PR #16, opened 2016-04-06
Small cleanups

akshayjshah:
1. Add links to the competitive benchmarking suite and the lock file to the README.
2. Add an output assertion to the package-level example.
3. Make the package-level GoDoc match the GitHub repo description.
4. Add a test for runtime log level changes.
5. Fix the Makefile and Glide manifest for Travis.
6. Only attempt to use the Coveralls API after tests pass, and don't fail the build if the API is flaky.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,5 +12,7 @@ cache:
 install:
   - make dependencies
 script:
-  - make coveralls
+  - make test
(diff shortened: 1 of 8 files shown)


prashantv:
lgtm

PR #20, opened 2016-04-06
Add a field factory for stacktraces

### LINKED ISSUE #8: Add a field factory for stacktraces
Add a field factory for stacktraces. We should probably add a convenient `ErrStack(error)` factory, too, since this will be a common combination.

akshayjshah:
Add a field factory that wraps `runtime.Stack`, so users don't need
to pick a byte size for their stacktraces. The wrapper attempts to
re-use a buffer from the pool, but allocates if it's necessary to
capture the whole stacktrace.

Note that the `Stack` function is *really* expensive - casting the trace
to a string costs an allocation, but just taking a stacktrace is ~10
microseconds (roughly the same amount of time as logging 100
non-stacktrace fields).

Fixes #8.

### DIFF
--- a/field.go
+++ b/field.go
@@ -100,7 +100,11 @@ func Stack() Field {
 	// Try to avoid allocating a buffer.
 	enc := newJSONEncoder()
 	bs := enc.bytes[:cap(enc.bytes)]
-	field := String("stacktrace", string(takeStacktrace(bs, false)))
+	// Returning the stacktrace as a string costs an allocation, but saves us
+	// from expanding the Field union struct to include a byte slice. Since
+	// taking a stacktrace is already so expensive (~10us), the extra allocation
+	// is okay.
+	field := String("stacktrace", takeStacktrace(bs, false))
 	enc.Free()
 	return field
 }
--- a/field.go
+++ b/field.go
@@ -92,6 +92,19 @@ func Err(err error) Field {
 	return String("error", err.Error())
 }
 
+// Stack constructs a Field that stores a stacktrace of the current goroutine
+// under the key "stacktrace". Keep in mind that taking a stacktrace is
+// extremely expensive (relatively speaking); this function both makes an
+// allocation and takes ~10 microseconds.
+func Stack() Field {
+	// Try to avoid allocating a buffer.
+	enc := newJSONEncoder()
+	bs := enc.bytes[:cap(enc.bytes)]
+	field := String("stacktrace", string(takeStacktrace(bs, false)))
+	enc.Free()
+	return field
+}
+
 // Duration constructs a Field with the given key and value. It represents
 // durations as an integer number of nanoseconds.
 func Duration(key string, val time.Duration) Field {
--- a/field_test.go
+++ b/field_test.go
@@ -22,11 +22,13 @@ package zap
 
 import (
 	"errors"
+	"strings"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type fakeUser struct{ name string }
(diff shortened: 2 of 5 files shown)


field_test.go
```diff
@@ -129,3 +129,14 @@ func TestNestField(t *testing.T) {
...
+
+func TestStackField(t *testing.T) {
+	enc := newJSONEncoder()
+	defer enc.Free()
+
+	Stack().addTo(enc)
+	output := string(enc.bytes)
+
+	assert.Equal(t, `"stacktrace":`, output[:13], "Stacktrace added under an unexpected key.")
+	assert.Contains(t, output[13:], "zap.takeStacktrace", "Expected stacktrace to contain our stacktrace capture function.")
```
prashantv: maybe check for TestStackField instead so we know the caller is in there

field_test.go
```diff
@@ -129,3 +129,14 @@ func TestNestField(t *testing.T) {
...
 }
+
+func TestStackField(t *testing.T) {
+	enc := newJSONEncoder()
+	defer enc.Free()
+
+	Stack().addTo(enc)
+	output := string(enc.bytes)
+
+	assert.Equal(t, `"stacktrace":`, output[:13], "Stacktrace added under an unexpected key.")
```
prashantv: might want to do `strings.HasPrefix` so the test doesn't panic if `output` is smaller than expected, and make it a `require` so that it doesn't run the below check if the string doesn't start with `"stacktrace":`

stacktrace.go
```diff
@@ -0,0 +1,45 @@
...
+// takeStacktrace attempts to use the provided byte slice to take a stacktrace.
+// If the provided slice isn't large enough, takeStacktrace will allocate
+// successively larger slices until it can capture the whole stack.
+func takeStacktrace(buf []byte, includeAllGoroutines bool) string {
+	n := runtime.Stack(buf, includeAllGoroutines)
+	for n >= len(buf) {
+		// Buffer wasn't large enough, allocate a larger one. No need to copy
+		// previous buffer's contents.
+		size := 2 * n
+		if size == 0 {
```
prashantv: maybe this should be the first thing we check,
```go
if len(buf) == 0 {
  buf := make(...)
}
```

field.go
```diff
@@ -92,6 +92,19 @@ func Err(err error) Field {
...
 
+// Stack constructs a Field that stores a stacktrace of the current goroutine
+// under the key "stacktrace". Keep in mind that taking a stacktrace is
+// extremely expensive (relatively speaking); this function both makes an
+// allocation and takes ~10 microseconds.
+func Stack() Field {
+	// Try to avoid allocating a buffer.
+	enc := newJSONEncoder()
+	bs := enc.bytes[:cap(enc.bytes)]
+	field := String("stacktrace", string(takeStacktrace(bs, false)))
```
prashantv: `takeStacktrace` already returns a string, no need to cast again.

stacktrace.go
```diff
@@ -0,0 +1,45 @@
...
+			// We may have been passed a nil byte slice.
+			size = 1024
+		}
+		buf = make([]byte, size)
+		n = runtime.Stack(buf, includeAllGoroutines)
+	}
+	// Casting this to a string costs an allocation, but saves us from expanding
```
prashantv: strings make more sense for stacks anyway in terms of encoding. this comment seems like it shouldn't be in this function (this function doesn't really care about fields etc) but in `Stack()` instead?

field_test.go
akshayjshah: I was so convinced that the function was strings.StartsWith, and I couldn't find it. :facepunch:

stacktrace.go
akshayjshah: I'll shuffle things around so that it makes sense. Ideally, the cast and the comment would live together in `Stack`.

prashantv:
lgtm

PR #22, opened 2016-04-06
Specify an error sink when creating the logger

prashantv:
We had a `errW` that was not being used, wired up a new parameter to `NewJSON` that is used as the error sink. If it's nil, we use `_defaultErrSink`.

I'm not a fan of the new parameter, maybe we need to go the functional options route (with `NewLogger` not needing any parameters, but using some defaults like "log everything to stdout". Can do that as a follow up if you agree with that approach.

### DIFF
--- a/logger_test.go
+++ b/logger_test.go
@@ -31,12 +31,14 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func withJSONLogger(f func(*jsonLogger, func() []string), fields ...Field) {
-	sink := bytes.NewBuffer(nil)
-	jl := NewJSON(All, sink, fields...)
+func withJSONLogger(t testing.TB, f func(*jsonLogger, func() []string), fields ...Field) {
+	sink := &bytes.Buffer{}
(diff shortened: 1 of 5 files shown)


logger_test.go
```diff
@@ -31,12 +31,14 @@ import (
...
-	sink := bytes.NewBuffer(nil)
-	jl := NewJSON(All, sink, fields...)
+func withJSONLogger(t testing.TB, f func(*jsonLogger, func() []string), fields ...Field) {
+	sink := &bytes.Buffer{}
+	errSink := &bytes.Buffer{}
+	jl := NewJSON(All, sink, errSink, fields...)
 	jl.StubTime()
 
 	f(jl.(*jsonLogger), func() []string { return strings.Split(sink.String(), "\n") })
+	assert.Empty(t, errSink.String(), "Expected error sink to be empty")
```
akshayjshah: :+1: very thorough

akshayjshah:
:shipit:

PR #23, opened 2016-04-07
Update zap.NewJSON to use functional options

prashantv:
Default options are:
- Info
- Write logs to stdout
- Write logging errors to stderr
- No fields

Log levels implement the option interface so you can pass them directly.

Added examples for NewJSON that show default vs passing options.


options.go
```diff
@@ -0,0 +1,64 @@
...
+type Option interface {
+	apply(*jsonLogger) error
+}
+
+// optionFunc wraps a func so it satisfies the Option interface.
+type optionFunc func(*jsonLogger) error
+
+func (f optionFunc) apply(jl *jsonLogger) error {
+	return f(jl)
+}
```
akshayjshah: :+1:

options.go
```diff
@@ -0,0 +1,64 @@
...
+
+// This allows any Level to be used as an option.
+func (l Level) apply(jl *jsonLogger) error {
+	*jl.level = int32(l)
+	return nil
+}
```
akshayjshah: Level is atomic (though we clearly can't trigger a race from the current options application site). Just for consistency, can we use `jl.SetLevel` here instead?

prashantv: Good idea, cleans up the code. Done

akshayjshah:
lgtm

PR #24, opened 2016-04-07
Test unknown field types

akshayjshah:
I noticed that this portion of `Field.addTo` wasn't tested. Since it's
easy enough to exercise, let's make sure we panic on unknown field
types.


prashantv:
lgtm

PR #25, opened 2016-04-07
Add hooks to logger

akshayjshah:
Extend the logger to support a variable number of pre-log hooks. The
hooks can alter the in-flight message and/or add to the logging context.

For this first revision, I've included hooks to (a) annotate each log
message with the logging call site, and (b) include stack traces for all
log messages above a given level.

### DIFF
--- a/hook_test.go
+++ b/hook_test.go
@@ -34,7 +34,7 @@ func TestHookAddCaller(t *testing.T) {
 	logger := NewJSON(All, Output(buf), AddCaller())
 	logger.Info("Callers.")
 
-	re := regexp.MustCompile(`hook_test.go:[\d]+: Callers\.`)
+	re := regexp.MustCompile(`"msg":"hook_test.go:[\d]+: Callers\."`)
 	assert.Regexp(t, re, buf.String(), "Expected to find package name and file name in output.")
 }
 
@@ -55,9 +55,17 @@ func TestHookAddCallerFail(t *testing.T) {
 func TestHookAddStacks(t *testing.T) {
 	buf := &bytes.Buffer{}
 	logger := NewJSON(All, Output(buf), AddStacks(Info))
-	logger.Info("Stacks.")
 
+	logger.Info("Stacks.")
 	output := buf.String()
 	require.Contains(t, output, "zap.TestHookAddStacks", "Expected to find test function in stacktrace.")
 	assert.Contains(t, output, `"stacktrace":`, "Stacktrace added under an unexpected key.")
+
+	buf.Reset()
+	logger.Warn("Stacks.")
+	assert.Contains(t, buf.String(), `"stacktrace":`, "Expected to include stacktrace at Warn level.")
+
+	buf.Reset()
+	logger.Debug("No stacks.")
+	assert.NotContains(t, buf.String(), "Unexpected stacktrace at Debug level.")
 }
--- a/hook_test.go
+++ b/hook_test.go
@@ -34,7 +34,7 @@ func TestHookAddCaller(t *testing.T) {
 	logger := NewJSON(All, Output(buf), AddCaller())
 	logger.Info("Callers.")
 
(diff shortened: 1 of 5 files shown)


hook_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"regexp"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHookAddCaller(t *testing.T) {
+	buf := &bytes.Buffer{}
+	logger := NewJSON(All, Output(buf), ErrorOutput(ioutil.Discard), Hooks(AddCaller()))
```
prashantv: here and below, suggest skipping the `ErrorOutput` option since the tests shouldn't produce errors. If they do, they'll go to stderr and we'll be more likely to notice them.

prashantv: Could our default hooks just implement `Option` so you can pass them directly without having to rewrap them?

hook_test.go
```diff
@@ -0,0 +1,64 @@
...
+	logger.Info("Callers.")
+
+	re := regexp.MustCompile(`github\.com\\/uber-common\\/zap\\/hook_test.go:[\d]+: Callers\.`)
```
prashantv: hmm should we do `filepath.Base` or something to only get the leaf directory? I think that's what our internal option does, it seems a little too heavy.

Or we can make that an option to `AddCaller`, but I'd rather keep it simple for now.

logger_bench_test.go
```diff
@@ -133,6 +133,20 @@ func BenchmarkObjectField(b *testing.B) {
 	})
 }
 
+func BenchmarkAddCallerHook(b *testing.B) {
+	logger := zap.NewJSON(
+		zap.Output(ioutil.Discard),
+		zap.ErrorOutput(ioutil.Discard),
```
prashantv: same as earlier, suggest removing `ErrorOutput` option

hook_test.go
akshayjshah: Great idea. Done!

hook_test.go
akshayjshah: Yep - I wasn't sure about this either. Given zap's focus on performance, I wasn't excited about doing more string processing in this hook.

hook_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHookAddCaller(t *testing.T) {
+	buf := &bytes.Buffer{}
+	logger := NewJSON(All, Output(buf), AddCaller())
+	logger.Info("Callers.")
+
+	re := regexp.MustCompile(`hook_test.go:[\d]+: Callers\.`)
```
prashantv: nit: can we ensure that only the base path is included, by making this look for `"msg":"hook_test.go:[...]`

hook_test.go
```diff
@@ -0,0 +1,63 @@
...
+	_callerSkip = 1e3
+	defer func() { _callerSkip = originalSkip }()
+
+	logger := NewJSON(All, Output(buf), ErrorOutput(errBuf), AddCaller())
+	logger.Info("Failure.")
+	assert.Equal(t, "failed to get caller\n", errBuf.String(), "Didn't find expected failure message.")
+	assert.Contains(t, buf.String(), `"msg":"Failure."`, "Expected original message to survive failures in runtime.Caller.")
+}
+
+func TestHookAddStacks(t *testing.T) {
```
prashantv: i think we want to test both branches in `AddStacks`, where the level is higher than or equal to the minimum, and one for when it's lower. Maybe a simple table test that takes a logger configured for `Warning` and ensures `Warning` & `Error` add the stack, while `Info` doesn't?

akshayjshah: Done.

prashantv:
lgtm

PR #27, opened 2016-04-09
Replace io.Writer with zap.WriteSyncer

### LINKED ISSUE #17: Flush when writing at panic level or higher
Require callers to construct loggers with something a bit richer than a plain `io.Writer` and also require a `Sync` method. Use `Sync` to flush on panics and higher.

akshayjshah:
We want to flush our output files whenever we log at the panic or fatal
levels, otherwise there's a chance the program will crash before we
actually write our logs. This is particularly important if we use
a buffered output writer.

To accommodate this, replace our use of `io.Writer` with `zap.WriteSyncer`.
All `*os.File`s already satisfy our interface, and this diff adds
a wrapper for other `io.Writer`s.

Fixes #17.


logger_test.go
```diff
@@ -169,14 +170,11 @@ func TestJSONLoggerNoOpsDisabledLevels(t *testing.T) {
 	})
 }
 
-func TestJSONLoggerInternalErrorHandlingNoSink(t *testing.T) {
-	errBuf := bytes.NewBuffer(nil)
-	_defaultErrSink = errBuf
-	defer func() { _defaultErrSink = os.Stderr }()
-
-	buf := &bytes.Buffer{}
+func TestJSONLoggerInternalErrorHandlingWithSink(t *testing.T) {
```
prashantv: nit: can remove `WithSink` since there's no `NoSink` anymore

writer.go
```diff
@@ -0,0 +1,76 @@
...
+// A WriteSyncer is an io.Writer that can also flush any buffered data. Note
+// that *os.File (and thus, os.Stderr and os.Stdout) implement WriteSyncer.
+type WriteSyncer interface {
+	io.Writer
+	Sync() error
+}
+
+// AddSync converts an io.Writer to a WriteSyncer. It attempts to be
+// intelligent: if the concrete type of the io.Writer implements WriteSyncer or
+// WriteFlusher, we'll use the existing Sync or Flush methods. If it doesn't,
```
prashantv: Seems like something is missing after "If it doesn't,"

writer.go
```diff
@@ -0,0 +1,76 @@
...
+func AddSync(w io.Writer) WriteSyncer {
+	switch w.(type) {
+	case WriteSyncer:
+		// The concrete type is already a WriteSyncer (e.g., an *os.File).
+		return w.(WriteSyncer)
+	case WriteFlusher:
+		// The concrete type implements a suitable Flush method, which we'll
+		// just use instead of Sync.
+		return flusherWrapper{w.(WriteFlusher)}
```
prashantv: with `switch`+`case`, you can avoid the type cast in the case by doing:
```go
switch w := w.(type) {
case WriteSyncer:
  return w
case WriteFlusher:
  return flusherWrapper{w}
default:
  return writeWrapper{w}
}
```

(the comments in the code are just duplicating the doc comments, so not really worth having).

writer_test.go
```diff
@@ -0,0 +1,84 @@
...
+	ws := AddSync(concrete)
+	requireWriteWorks(t, ws)
+
+	require.NoError(t, ws.Sync(), "Unexpected error syncing a WriteSyncer.")
+	require.True(t, concrete.FlushCalled, "Expected to dispatch to concrete type's Flush method.")
+
+	concrete.Err = errors.New("fail")
+	assert.Error(t, ws.Sync(), "Expected to propagate errors from concrete type's Flush method.")
+}
+
```
prashantv: let's put a test for `AddSync` where there's a `Sync` and a `Flush` method -- only the `Sync` should be called, but just want to have it tested.

writer.go
akshayjshah: good catch - fixed.

prashantv:
lgtm

PR #28, opened 2016-04-09
Add a stdlib- and bark-compatible wrappers

### LINKED ISSUE #18: Provide a wrapper compatible with the standard library

### LINKED ISSUE #19: Provide a sampling wrapper
Provide a sampling wrapper that allows callers to log the first n messages and every k'th message thereafter.

akshayjshah:
Add wrappers to make zap compatible with the stdlib and bark.

Fixes #18 and #34.

### DIFF
--- a/field.go
+++ b/field.go
@@ -41,7 +41,6 @@ const (
 	int64Type
 	stringType
 	marshalerType
-	unsafeBytesType
 )
 
 // A Field is a deferred marshaling operation used to add a key-value pair to
@@ -135,16 +134,6 @@ func Nest(key string, fields ...Field) Field {
 	return Field{key: key, fieldType: marshalerType, obj: multiFields(fields)}
 }
 
-// UnsafeBytes constructs a Field which will add arbitrary bytes to the logging
-// context. While the key is escaped, the passed bytes aren't escaped or sanity
-// checked in any way - passing improperly encoded bytes can easily make the
-// logger's final output unparsable.
-//
-// This is unsafe - use with caution!
-func UnsafeBytes(key string, val []byte) Field {
-	return Field{key: key, fieldType: unsafeBytesType, str: string(val)}
-}
-
 func (f Field) addTo(kv KeyValue) error {
 	switch f.fieldType {
 	case boolType:
@@ -162,8 +151,6 @@ func (f Field) addTo(kv KeyValue) error {
 		err := f.obj.MarshalLog(kv)
 		closer.CloseField()
 		return err
-	case unsafeBytesType:
-		kv.UnsafeAddBytes(f.key, []byte(f.str))
 	default:
 		panic(fmt.Sprintf("unknown field type found: %v", f))
 	}
--- a/field.go
+++ b/field.go
@@ -41,6 +41,7 @@ const (
 	int64Type
 	stringType
 	marshalerType
+	unsafeBytesType
 )
 
 // A Field is a deferred marshaling operation used to add a key-value pair to
@@ -134,6 +135,16 @@ func Nest(key string, fields ...Field) Field {
 	return Field{key: key, fieldType: marshalerType, obj: multiFields(fields)}
 }
 
+// UnsafeBytes constructs a Field which will add arbitrary bytes to the logging
+// context. While the key is escaped, the passed bytes aren't escaped or sanity
+// checked in any way - passing improperly encoded bytes can easily make the
+// logger's final output unparsable.
+//
+// This is unsafe - use with caution!
+func UnsafeBytes(key string, val []byte) Field {
+	return Field{key: key, fieldType: unsafeBytesType, str: string(val)}
+}
+
 func (f Field) addTo(kv KeyValue) error {
 	switch f.fieldType {
 	case boolType:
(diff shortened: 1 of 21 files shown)


akshayjshah:
Benchmarks aren't quite as bad as I'd feared, but they're not awesome:

```
BenchmarkStandardPrint-4         1000000              1071 ns/op             168 B/op          7 allocs/op
BenchmarkStandardPrintf-4        1000000              1303 ns/op             184 B/op          7 allocs/op
```

prashantv:
Should we separate wrapping a longer to match the standard logging library interface vs wrapping to match the logrus interface? Logrus is a superset of the standard logging interface, so if we wrote one wrapper for logrus, that should cover both cases.

I'd like to avoid adding more APIs and instead having a single API to wrap (and that could also live outside of the core `zap` package).

akshayjshah:
Fair enough. I went back and forth about separating these into their own sub-package - I'll do that, and I'll make the wrapper full-featured enough to support the full bark API. That'll require some additional API support in zap:

* unmarshaling the fields already serialized into a map[string]interface{}
* a zap.UnsafeBytes() field.

rap/doc.go
```diff
@@ -0,0 +1,23 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package rap provides wrappers to make zap compatible with other logging
+// libraries.
+package rap
```
akshayjshah: `rap` sounds like "wrap" and looks like `zap`, which I found amusing. Feel free to object.

breerly: i don't think maintainers will get the joke in a couple years - suggest reserving fun names for just the project name, and then dull but clear names within the lib.

akshayjshah: Sigh. I suppose you're right.

akshayjshah:
As gross as it is, the bark wrapper is still way faster than logrus.

Adding fields at each log site:
```
BenchmarkLogrusAddingFields-4                     100000             17295 ns/op            5275 B/op         78 allocs/op
BenchmarkZapBarkifyAddingFields-4                 300000              6221 ns/op            3601 B/op         23 allocs/op
```

Logging a static string after we've already accumulated some context:
```
BenchmarkLogrusWithAccumulatedContext-4           100000             13003 ns/op            3438 B/op         61 allocs/op
BenchmarkZapBarkifyWithAccumulatedContext-4      3000000               593 ns/op              48 B/op          3 allocs/op
```

field.go
```diff
@@ -151,6 +162,8 @@ func (f Field) addTo(kv KeyValue) error {
 		err := f.obj.MarshalLog(kv)
 		closer.CloseField()
 		return err
+	case unsafeBytesType:
+		kv.UnsafeAddBytes(f.key, []byte(f.str))
```
prashantv: We're taking in `[]byte`, copying to make a `string` to store in the Field, then copying again to make a `[]byte` that is appended. Ideally we:
 - only use a single type all the way. since we want to use fields in the string, maybe it should just be string?
 - or we have a second `UnsafeString` version to avoid all the copying.

zbark/bark_bench_test.go
```diff
@@ -0,0 +1 @@
+package zbark
```
prashantv: remove file?

akshayjshah: Yep, missed that one :)

field.go
akshayjshah: Fixed, opted to not export the UnsafeBytes stuff.

prashantv:
lgtm. what are the new benchmarks with reduced copying?

akshayjshah:
Faster, but same number of allocations.

Before:

```
$ go test -bench="Barkify" -run="^$" -benchmem -cpuprofile=cpu.pprof -memprofile=mem.pprof ./benchmarks
testing: warning: no tests to run
PASS
BenchmarkZapBarkifyAddingFields-4                 200000              5572 ns/op            3600 B/op         23 allocs/op
BenchmarkZapBarkifyWithAccumulatedContext-4      3000000               518 ns/op              48 B/op          3 allocs/op
```

After:
```
$ go test -bench="Barkify" -run="^$" -benchmem -cpuprofile=cpu.pprof -memprofile=mem.pprof ./benchmarks
testing: warning: no tests to run
PASS
BenchmarkZapBarkifyAddingFields-4                 300000              4072 ns/op            3601 B/op         23 allocs/op
BenchmarkZapBarkifyWithAccumulatedContext-4      3000000               431 ns/op              48 B/op          3 allocs/op
```

prashantv:
Cool, perf win is always good. 🚢

PR #29, opened 2016-04-09
Improve documentation for Marshaler

akshayjshah:
Clarify the GoDoc comments and add an example for Marshaler.

### DIFF
--- a/marshaler_example_test.go
+++ b/marshaler_example_test.go
@@ -1,0 +1,78 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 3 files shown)


marshaler_example_test.go
```diff
@@ -0,0 +1,78 @@
...
+	Name string `json:"name"`
+	Age  int    `json:"age"`
+	Auth Auth   `auth:"auth"`
+}
+
+func (u User) MarshalLog(kv zap.KeyValue) error {
+	kv.AddString("name", u.Name)
+	kv.AddInt("age", u.Age)
+
+	// Put the authentication information in a nested object.
```
prashantv: After seeing this example, I feel like `KeyValue` maybe should just have `AddObject`. We should chat about the benefits of `Nest`

marshaler_example_test.go
```diff
@@ -0,0 +1,78 @@
...
+	defer kv.Nest("auth").CloseField()
+	if err := u.Auth.MarshalLog(kv); err != nil {
```
prashantv: optionally: `return u.Auth.MarshalLog(kv)`

prashantv:
lgtm, the discussion of `Nest` vs `AddObject` can be tackled separately.

marshaler_example_test.go
akshayjshah: We can definitely add an `AddObject` method, but I think we'll still need `Nest` - not all embedded structs are defined by the user, so I think we still need some way to create a nested namespace for non-`Marshaler`s.

marshaler_example_test.go
akshayjshah: good point.

PR #30, opened 2016-04-11
Support specifying level through a flag (uses standard flag package)

prashantv:
This is similar to how inbuilt types work,
https://golang.org/src/flag/flag.go?s=21248:21321#L144


akshayjshah:
lgtm

akshayjshah:
Way to lower our test coverage, though. That 0.6% is killing us ;)

prashantv:
Deleted the `Get()` method which we isn't used, coverage should be better now :)

PR #31, opened 2016-04-11
Move String method on Level to value receiver

prashantv:
We want pretty-printing of levels when we pass the level by value


akshayjshah:
:shipit:

PR #32, opened 2016-04-12
Add Stringer field which will call String() method

prashantv:
This avoids some line noise in log call sites when passing things like `net.IP`, since code like:
```go
zap.Info("Started.", zap.String("ip", ipAddr.String()))
```
becomes:
```go
zap.Info("Started.", zap.Stringer("ip", ipAddr))
```


blampe:
Type safety is so cool 👍

PR #35, opened 2016-04-30
Miscellaneous ergonomic improvements

akshayjshah:
This PR introduces a few small ergonomic improvements to zap's APIs (might as well get them in while we're in beta!):

* `Nest` now takes a function to populate the nested namespace, instead of returning a `FieldCloser` and leaving state management up to the user.
* The `KeyValue` interface now has an `AddObject` method which adds a `Marshaler` to the logging context.
* Loggers now have a `DFatal` method and a `Development` option.
* I also updated the benchmarks in the README. In a [previous PR](https://github.com/uber-common/zap/commit/078a0b69f4dce79002c4387d2f9cc99f2d1c3f63), I fixed one of the logrus benchmarks - it's actually even slower than the first round of benchmarks suggested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # :zap: zap [![GoDoc][doc-img]][doc] [![Build Status][ci-img]][ci] [![Coverage Status][cov-img]][cov]
 
-Fast, structured, leveled logging in Go.
+Blazing fast, structured, leveled logging in Go.
 
 ## Structure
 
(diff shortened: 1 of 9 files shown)


akshayjshah:
Tests are failing locally and on Travis with the cryptic `Usage of /tmp/go-build839778272/github.com/uber-common/zap/_test/zap.test:`, which seems to suggest that our internal tools are somehow polluting the build. The temp directory mentioned doesn't even exist on my local filesystem, and `go clean`, reinstalling packages, and clearing out `$GOPATH/pkg` don't solve the problem.

Will investigate more today.

prashantv:
lgtm

PR #36, opened 2016-05-02
Add a sampling wrapper

### LINKED ISSUE #19: Provide a sampling wrapper

akshayjshah:
Add a sampling wrapper that logs the first N messages and every Mth
message thereafter. Results of the included sampler benchmark:

```
BenchmarkSampledZap-4   50000000    33.9 ns/op  0 B/op  0 allocs/op
```

Fixes #19, which @nomis52 wanted.

### DIFF
--- a/spy/logger.go
+++ b/spy/logger.go
@@ -117,9 +117,10 @@ func (l *Logger) SetLevel(new zap.Level) {
 // With creates a new Logger with additional fields added to the logging context.
 func (l *Logger) With(fields ...zap.Field) zap.Logger {
 	return &Logger{
-		level:   l.level,
-		sink:    l.sink,
-		context: append(l.context, fields...),
+		level:       l.level,
+		sink:        l.sink,
+		context:     append(l.context, fields...),
+		development: l.development,
 	}
 }
 
--- a/spy/logger.go
+++ b/spy/logger.go
@@ -1,0 +1,177 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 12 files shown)


akshayjshah:
Rebased on master, pulled in `DFatal`.

spy/logger.go
```diff
@@ -0,0 +1,177 @@
...
+	sync.Mutex
+
+	level       zap.Level
+	sink        *Sink
+	context     []zap.Field
+	development bool
+}
+
+// New returns a new spy logger at the default level.
+func New(sink *Sink) *Logger {
```
prashantv: nit: we'll most likely always want a new `Sink` when returning a new `Logger`, so we could make this:
```go
func New() (*Sink, *Logger)
```

zwrap/sample.go
```diff
@@ -0,0 +1,123 @@
...
+// concurrent use.
+//
+// The children of sampling loggers (i.e., the results of calls to the With method)
+// inherit their parent's sampling logic but maintain independent counters.
+func Sample(zl zap.Logger, tick time.Duration, first, thereafter int) zap.Logger {
+	return &sampler{
+		Logger:     zl,
+		tick:       tick,
+		first:      uint64(first),
+		thereafter: uint64(thereafter),
```
prashantv: If I want to configure the Sampler to only print the first 100 messages, and then nothing else for that time period, what do you thik is cleaner? the user passing a large number for `thereafter` or making `thereafter == 0` mean that it should never log anything after `first`?

spy/logger.go
akshayjshah: Good point, that would be more convenient.

akshayjshah:
Sampling by call site is about 10x slower than the previous version. Will investigate more some other time...

```
BenchmarkZapSampleWithoutFields-4        5000000               321 ns/op               0 B/op          0 allocs/op
```

akshayjshah:
Sampling by message (and eating the string hashing cost) turns out to be quite a bit faster than using `runtime.Callers`.

```
BenchmarkZapSampleWithoutFields-4       20000000                63.1 ns/op             0 B/op          0 allocs/op
```

Here's a CPU profile of the same benchmark: http://imgh.us/pprof001_3.svg

zwrap/sample.go
akshayjshah: I'd prefer to have users pass a large number. I'd really rather not encourage users to stop caring completely after the first N messages; even if `thereafter` is very large, users should get some feedback when their applications are way off the rails.

zwrap/sample.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (c *counters) Inc(key string) uint64 {
+	c.RLock()
+	count, ok := c.counts[key]
+	c.RUnlock()
+
+	if !ok {
+		one := uint64(1)
+		c.Lock()
```
prashantv: double check if the key exists under the lock, before setting the value. It could be that someone else added the key right after the `RUnlock`

zwrap/example_test.go
```diff
@@ -45,3 +47,20 @@ func Example() {
 	// Output:
 	// {"msg":"Encountered 0 errors.","level":"warn","ts":0,"fields":{}}
 }
+
+func Example_sample() {
+	sampledLogger := zwrap.Sample(zap.NewJSON(), time.Second, 1, 100)
+	// Stub the current time in tests.
+	sampledLogger.StubTime()
+
+	sampledLogger.Error("Unusual failure.")
```
prashantv: nit: think we should move this to after the With, so we know that it's printed even if the common failure is being sampled

prashantv:
lgtm

zwrap/example_test.go
akshayjshah: Good point, done.

zwrap/sample.go
akshayjshah: Yep, completely forgot. Fixed.

PR #37, opened 2016-05-02
Renaming in preparation for an arbitrary-object field

akshayjshah:
In preparation for a real arbitrary-object field, rename `Object` -> `Marshaler` and `Marshaler` -> `LogMarshaler`.


prashantv:
lgtm

PR #38, opened 2016-05-02
Add zap.Object to log arbitrary objects

akshayjshah:
Introduce zap.Object, which uses encoding-appropriate reflection to log
truly arbitrary objects. This introduces an API that's not fast by
default, but it makes the zbark wrapper much cleaner.


prashantv:
lgtm

PR #39, opened 2016-05-03
Add a benchmark for disabled log levels

akshayjshah:
Realized that we weren't benchmarking the overhead of disabled logging. It's not free, but it's pretty close.

```
BenchmarkZapDisabledLevels-4    200000000                6.07 ns/op            0 B/op          0 allocs/op
```

### DIFF
--- a/benchmarks/zap_bench_test.go
+++ b/benchmarks/zap_bench_test.go
@@ -49,39 +49,8 @@ var _jane = user{
 	CreatedAt: time.Date(1980, 1, 1, 12, 0, 0, 0, time.UTC),
 }
 
-func BenchmarkZapDisabledLevels(b *testing.B) {
-	logger := zap.NewJSON(zap.Error, zap.Output(zap.Discard))
-	b.ResetTimer()
-	b.RunParallel(func(pb *testing.PB) {
-		for pb.Next() {
-			logger.Info("Should be discarded.")
-		}
-	})
-}
-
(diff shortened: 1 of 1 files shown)


blampe:
🚢

prashantv:
Can you also add benchmarks with some fields? (I think unfortunately we'll still do allocs even with a level disabled)

akshayjshah:
Done. Results are as expected - we allocate for the slice to hold the variadic args and when casting concrete types to interfaces.

```
BenchmarkZapDisabledLevelsWithoutFields-4           200000000                7.78 ns/op            0 B/op          0 allocs/op
BenchmarkZapDisabledLevelsAccumulatedContext-4  200000000                6.48 ns/op            0 B/op          0 allocs/op
BenchmarkZapDisabledLevelsAddingFields-4         3000000               353 ns/op             704 B/op          2 allocs/op
BenchmarkZapWithoutFields-4                     10000000               220 ns/op               0 B/op          0 allocs/op
BenchmarkZapWithAccumulatedContext-4             5000000               233 ns/op               0 B/op          0 allocs/op
BenchmarkZapAddingFields-4                       1000000              1129 ns/op             705 B/op          2 allocs/op
```

prashantv:
Thanks, yep that's what I expected. We'll probably want to do a "is level enabled" check before logging in the critical path (especially for debugging logging) similar to TChannel-Go.

lgtm

PR #41, opened 2016-05-06
Stop escaping '/' which JSON does not require

prashantv:
Escaping '/' makes URLs longer and harder to read, and isn't required by
the JSON spec.


breerly:
👏

akshayjshah:
lgtm

PR #42, opened 2016-05-07
Handle NaN and +/- Inf

akshayjshah:
The type system won't protect us from this, so we should handle these cases explicitly.

### DIFF
--- a/json_encoder.go
+++ b/json_encoder.go
@@ -105,19 +105,16 @@ func (enc *jsonEncoder) AddInt64(key string, val int64) {
 // grade-school notation otherwise).
 func (enc *jsonEncoder) AddFloat64(key string, val float64) {
 	enc.addKey(key)
-	if math.IsNaN(val) {
+	switch {
+	case math.IsNaN(val):
 		enc.bytes = append(enc.bytes, `"NaN"`...)
-		return
-	}
-	if math.IsInf(val, 1) {
+	case math.IsInf(val, 1):
 		enc.bytes = append(enc.bytes, `"+Inf"`...)
-		return
-	}
-	if math.IsInf(val, -1) {
+	case math.IsInf(val, -1):
 		enc.bytes = append(enc.bytes, `"-Inf"`...)
-		return
+	default:
+		enc.bytes = strconv.AppendFloat(enc.bytes, val, 'g', -1, 64)
 	}
-	enc.bytes = strconv.AppendFloat(enc.bytes, val, 'g', -1, 64)
 }
 
 // Nest allows the caller to populate a nested object under the provided key.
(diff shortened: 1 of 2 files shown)


json_encoder.go
```diff
@@ -104,6 +105,18 @@ func (enc *jsonEncoder) AddInt64(key string, val int64) {
 // grade-school notation otherwise).
 func (enc *jsonEncoder) AddFloat64(key string, val float64) {
 	enc.addKey(key)
+	if math.IsNaN(val) {
+		enc.bytes = append(enc.bytes, `"NaN"`...)
```
prashantv: Hmm should we be encoding the same key with a different type? What do you think of the idea of using something like "key-error" or something?

It's a pain JSON doesn't allow `NaN` or infinity

json_encoder.go
```diff
@@ -104,6 +105,18 @@ func (enc *jsonEncoder) AddInt64(key string, val int64) {
...
+		return
+	}
+	if math.IsInf(val, 1) {
+		enc.bytes = append(enc.bytes, `"+Inf"`...)
+		return
+	}
+	if math.IsInf(val, -1) {
```
prashantv: you can replace the 2 infinity checks with just one,
```go
if math.IsInf(val, 0) {
  // infinity of either sign
}
```

akshayjshah: Right, but we'll still need to distinguish +Inf from -Inf. Is

```go
if math.IsInf(val, 0) {
   if val > 0 {
       // +Inf
    } else {
        // -Inf
    }
}
```
better than two `math.IsInf` checks?

json_encoder.go
akshayjshah: I'm fine with `key-error`, but let's defer that to another change - there are a few places where we handle errors like this. Created #43.

json_encoder.go
prashantv: Good point, I totally missed that we're doing different values for them. What you have is fine.

json_encoder.go
```diff
@@ -104,6 +105,18 @@ func (enc *jsonEncoder) AddInt64(key string, val int64) {
 // grade-school notation otherwise).
 func (enc *jsonEncoder) AddFloat64(key string, val float64) {
 	enc.addKey(key)
+	if math.IsNaN(val) {
```
prashantv: optional: `switch` might make make this easier to read
```go
switch {
case math.IsNaN(val) {
  enc.bytes = append(enc.bytes, `"NaN"`...)
case math.IsInf(val, 1) {
  // ...
case math.IsInf(val, -1) {
  // ...
default:
  // ...
}
```

prashantv:
lgtm

json_encoder.go
akshayjshah: Argh, I keep forgetting that `switch` is an option. Fixed.

PR #44, opened 2016-05-10
Add a Log method to Logger interface

akshayjshah:
Add a method to all the loggers that allows users to pass the level as a parameter (instead of calling a named method). This is a precursor to introducing a faster (though less convenient) sampling API.

### DIFF
--- a/json_encoder_test.go
+++ b/json_encoder_test.go
@@ -22,6 +22,7 @@ package zap
 
 import (
 	"bytes"
+	"errors"
 	"io"
 	"math"
 	"strings"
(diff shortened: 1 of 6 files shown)


akshayjshah:
Crazytown. On tip, apparently `encoding/json` now serializes `map[int]string` to JSON by casting the ints to strings.

https://travis-ci.org/uber-common/zap/jobs/129036303

Fixed the test for JSON serialization errors to use a `json.Marshaler` that always fails.

prashantv:
lgtm

oh wow, I'm kind of excited for serialization of `map[int]*`, will make `yab` a little cleaner.

PR #45, opened 2016-05-17
Reduce allocations for disabled/sampled logs

### LINKED ISSUE #56: Add a way to sample without allocating a slice of fields
As-is, sampling is okay...but calls to the leveled logging methods still requires allocating a slice to hold the variadic `...Field` argument. Add a token-like API so that we allocate only if we're actually going to log a message.

akshayjshah:
Even when a log level is disabled or heavily sampled, calling `logger.Error(msg, someField, someOtherField)` will make at least one allocation to hold the fields. Since we're going to discard the message anyways, this is wasted work. This PR introduces the `logger.Check` API, which allows particularly performance-sensitive applications to avoid these allocations by using a slightly wordier call pattern.

Using `Check` makes it virtually free to pass fields to a sampled logger or collect lots of debug logs:

```
BenchmarkZapDisabledLevelsAddingFields-4         3000000        439 ns/op             704 B/op          2 allocs/op
BenchmarkZapDisabledLevelsCheckAddingFields-4   200000000         5.74 ns/op            0 B/op          0 allocs/op

BenchmarkZapSampleWithoutFields-4       20000000                60.9 ns/op             0 B/op          0 allocs/op
BenchmarkZapSampleCheckWithoutFields-4  20000000                61.4 ns/op             0 B/op          0 allocs/op

BenchmarkZapSampleAddingFields-4         2000000               575 ns/op             704 B/op          2 allocs/op
BenchmarkZapSampleCheckAddingFields-4   20000000                65.2 ns/op             0 B/op          0 allocs/op
```

This fixes #56.

### DIFF
--- a/benchmarks/zap_bench_test.go
+++ b/benchmarks/zap_bench_test.go
@@ -110,7 +110,7 @@ func BenchmarkZapDisabledLevelsCheckAddingFields(b *testing.B) {
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
-			if m := logger.Check(zap.Info, "Should be discarded."); m != nil {
+			if m := logger.Check(zap.Info, "Should be discarded."); m.OK() {
 				m.Write(fakeFields()...)
 			}
 		}
@@ -185,7 +185,7 @@ func BenchmarkZapSampleCheckWithoutFields(b *testing.B) {
 		i := 0
 		for pb.Next() {
 			i++
-			if cm := logger.Check(zap.Info, messages[i%1000]); cm != nil {
+			if cm := logger.Check(zap.Info, messages[i%1000]); cm.OK() {
 				cm.Write()
 			}
 		}
@@ -201,7 +201,7 @@ func BenchmarkZapSampleCheckAddingFields(b *testing.B) {
 		i := 0
 		for pb.Next() {
 			i++
-			if m := logger.Check(zap.Info, messages[i%1000]); m != nil {
+			if m := logger.Check(zap.Info, messages[i%1000]); m.OK() {
 				m.Write(fakeFields()...)
 			}
 		}
--- a/benchmarks/zap_bench_test.go
+++ b/benchmarks/zap_bench_test.go
@@ -66,6 +66,14 @@ func fakeFields() []zap.Field {
 	}
 }
 
+func fakeMessages(n int) []string {
+	messages := make([]string, n)
+	for i := range messages {
+		messages[i] = fmt.Sprintf("Test logging, but use a somewhat realistic message length. (#%v)", i)
+	}
+	return messages
+}
+
 func BenchmarkZapDisabledLevelsWithoutFields(b *testing.B) {
 	logger := zap.NewJSON(zap.Error, zap.Output(zap.Discard))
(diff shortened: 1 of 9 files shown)


benchmarks/zap_bench_test.go
```diff
@@ -96,6 +104,19 @@ func BenchmarkZapDisabledLevelsAddingFields(b *testing.B) {
 		}
 	})
 }
+
+func BenchmarkZapDisabledLevelsCheckAddingFields(b *testing.B) {
+	logger := zap.NewJSON(zap.Error, zap.Output(zap.Discard))
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			if m := logger.Check(zap.Info, "Should be discarded."); m != nil {
```
prashantv: optional: you could make the `CheckedMessage` type have a method (`OK() bool`, `Log() bool`, not sure), and that method could also support `nil` as well so you don't have to return a new object.

instead of `m != nil`, we could have `m.OK()`

logger.go
```diff
@@ -27,6 +27,37 @@ import (
...
+	}
+}
+
+// Write logs the pre-checked message with the supplied fields. It panics if
+// called more than once.
+func (m *CheckedMessage) Write(fields ...Field) {
+	if n := atomic.AddUint32(&m.used, 1); n > 1 {
+		// We could call logger.Log at a (possibly large) performance cost, but it's
+		// better to make this an obvious user error.
+		panic("Can't use a CheckedMessage more than once.")
```
prashantv: only panic in `Development` -- in prod, we should silently ignore

benchmarks/zap_bench_test.go
akshayjshah: done.

logger.go
akshayjshah: Compromised on logging an error with `DFatal`.

abhinav:
👍

checked_message.go
```diff
@@ -0,0 +1,58 @@
...
+		msg:    msg,
+	}
+}
+
+// Write logs the pre-checked message with the supplied fields. It should only
+// be used once; if a CheckedMessage is re-used, it also logs an error message
+// with the underlying logger's DFatal method.
+func (m *CheckedMessage) Write(fields ...Field) {
+	if n := atomic.AddUint32(&m.used, 1); n > 1 {
+		m.logger.DFatal("Shouldn't re-use a CheckedMessage.")
```
prashantv: you probably want to return after this to avoid logging again in prod

akshayjshah: yep, done.

checked_message.go
```diff
@@ -0,0 +1,63 @@
...
+		msg:    msg,
+	}
+}
+
+// Write logs the pre-checked message with the supplied fields. It should only
+// be used once; if a CheckedMessage is re-used, it also logs an error message
+// with the underlying logger's DFatal method.
+func (m *CheckedMessage) Write(fields ...Field) {
+	if n := atomic.AddUint32(&m.used, 1); n > 1 {
+		if n == 2 {
```
prashantv: I really like this approach

prashantv:
lgtm

PR #46, opened 2016-05-17
Add a test for Development option

### LINKED ISSUE #55: Add tests for development option and Fatal methods
Add unit tests for the `Development` option, which requires stubbing calls to `os.Exit`.

akshayjshah:
Fixes #55.

### DIFF
--- a/logger_test.go
+++ b/logger_test.go
@@ -24,6 +24,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"strings"
 	"testing"
 
@@ -33,6 +34,20 @@ import (
 
 func opts(opts ...Option) []Option {
 	return opts
+}
+
+type stubbedExit struct {
+	Status int
+}
+
+func (se *stubbedExit) Unstub() {
+	_exit = os.Exit
+}
+
+func stubExit() *stubbedExit {
+	stub := &stubbedExit{}
+	_exit = func(s int) { stub.Status = s }
+	return stub
 }
 
 func withJSONLogger(t testing.TB, opts []Option, f func(*jsonLogger, func() []string)) {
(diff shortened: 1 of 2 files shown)


prashantv:
Rather than this, I'd rather take the `var` stubbing approach to test this. That would allow us to test the `Fatal` calls in general, as well as `DFatal`

There is another option, which is kind of hacky, but could work, which is to:
 - Start a separate goroutine for the test
 - use a custom sink that will call `runtime.Goexit` (probably in the `Flush` call)
 - verify that the goroutine ends and check the contents

I think the first option is probably cleaner though

akshayjshah:
Sure, works for me.

options_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestOptionDevelopment(t *testing.T) {
+	prod := NewJSON().(*jsonLogger)
+	require.False(t, prod.development, "Expected default logger to be in production mode.")
+
+	dev := NewJSON(Development()).(*jsonLogger)
```
breerly: where does `Development()` come from? testify?

prashantv: It's an option you can pass when creating a logger:
https://godoc.org/github.com/uber-common/zap#Development

We're using it to support the idea of `DFatal` which is an error log in production mode, but a fatal in test mode.

logger_test.go
```diff
@@ -35,6 +36,20 @@ func opts(opts ...Option) []Option {
 	return opts
 }
 
+type stubbedExit struct {
+	Status int
```
prashantv: to distinguish vs being called with `Exit(0)` vs not being called, can we either use a separate bool to track whether it was called, or make `Status` a pointer.

akshayjshah: Good point, I completely overlooked that.

akshayjshah: Done.

logger_test.go
```diff
@@ -136,10 +151,19 @@ func TestJSONLoggerLog(t *testing.T) {
...
 		assert.Panics(t, func() { jl.Log(Panic, "foo") }, "Expected logging at Panic level to panic.")
 		assertMessage(t, "panic", "foo", output()[0])
 	})
+
+	stub := stubExit()
+	defer stub.Unstub()
+	withJSONLogger(t, nil, func(jl *jsonLogger, output func() []string) {
+		jl.Log(Fatal, "foo")
+		assertMessage(t, "fatal", "foo", output()[0])
+		assert.Equal(t, 1, *stub.Status, "Expected to call os.Exit with status 1.")
```
prashantv: optional nit: if `exit` isn't called, rather than failing with a nice error, it'll cause a panic. we could do
```go
if assert.NotNil(t, stub.Status, "Expected a call to os.Exit") {
  assert.Equal(t, 1, *stub.Status, "os.Exit got unexpected exit code")
}
```

(i also try to not mention the specific value we expect in the message, since `Equal` will print the value it expected vs what it got already, and if we change it, it means less places to change / accidentally miss)

prashantv:
lgtm with minor nit

PR #48, opened 2016-05-19
Update references from uber-common to uber-go

prashantv:


blampe:
🚢

akshayjshah:
A++ would review again.

PR #53, opened 2016-05-23
Fix the 'Bool' method output by checking the field value

### LINKED ISSUE #52: zap.Bool() is not expected
Hi,

I type & run: 
```
package main

import (
	"os"

	"github.com/uber-go/zap"
)

func main() {
	writeSyncer := zap.AddSync(os.Stderr)
	logger := zap.NewJSON(
		zap.Output(writeSyncer),
		zap.ErrorOutput(writeSyncer),
	)
	logger.With(zap.Bool("wrong_bool", false)).Info("Hi!")
}
```
then, I got
```
{"msg":"Hi!","level":"info","ts":1463991525501189122,"fields":{"wrong_bool":true}}
```
The value of field named `wrong_bool` should be false, but is not expected.

pravj:
According to the issue #52, the logger was always reporting a boolean key-value pair as 'true'.

It was happening because of the hardcoded field value as 1, in the file ['field.go'](https://github.com/uber-go/zap/blob/master/field.go#L55).

Test cases were unable to detect it probably because the test case is giving input [only as](https://github.com/uber-go/zap/blob/master/field_test.go#L75-L78) 'true' type.

### DIFF
--- a/field.go
+++ b/field.go
@@ -52,7 +52,7 @@ type Field struct {
 
 // Bool constructs a Field with the given key and value.
 func Bool(key string, val bool) Field {
-	var ival int64 = 0
+	var ival int64
 	if val {
 		ival = 1
 	}
--- a/field.go
+++ b/field.go
@@ -52,7 +52,12 @@ type Field struct {
 
 // Bool constructs a Field with the given key and value.
 func Bool(key string, val bool) Field {
-	return Field{key: key, fieldType: boolType, ival: 1}
+	var ival int64 = 0
+	if val {
+		ival = 1
+	}
+
+	return Field{key: key, fieldType: boolType, ival: ival}
 }
 
 // Float64 constructs a Field with the given key and value. The floating-point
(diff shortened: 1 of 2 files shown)


prashantv:
@pravj Thanks for the fix!

Couple of things:
- Can you please sign our [Uber CLA](https://docs.google.com/a/uber.com/forms/d/1pAwS_-dA1KhPlfxzYLBqK6rsSWwRwH95OCCZrcsY5rk/viewform)
- Can you add unit tests to field_test.go, similar to https://github.com/uber-go/zap/blob/master/field_test.go#L75

pravj:
Hi @prashantv,

I have signed the CLA.

I have added unit tests also, as you suggested.

Methods 'TestTrueBoolField' and 'TestFalseBoolField' will check for equality of expected and returned boolean values (true and false respectively).

Method 'TestUnlikeBoolField' will check for inequality of different boolean value pairs.

prashantv:
lgtm, thanks for the fixes @pravj

PR #54, opened 2016-05-24
Allow CallerSkip to be set dynamically.

justyntemme:
_Caller skip can now bet set everytime addcaller is called. I'm not sure how you want to proceed from here.

### DIFF
--- a/hook.go
+++ b/hook.go
@@ -32,10 +32,8 @@ import (
 var _callerSkip = 3
 
 //Function to set Skip caller variable.
-func IncCallerSkip(i int) int {
+func IncCallerSkip(i int) {
 	_callerSkip = i
-	return 0
-
 }
 
 // A hook is executed each time the Logger writes a message. It receives the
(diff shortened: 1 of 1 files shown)


prashantv:
Thanks for the PR @justyntemme 

However, I'm not sure if we want this to be statically configured at log creation. There's a few issues with the current approach:
 - An option is changing a global variable, which is unexpected and means that you can't have multiple loggers that require different levels of skipping
 - You need to know the wrappers at creation time, rather than when adding the wrappers

Ideally, the way I'd want this to work is:
 - You can create a logger, and it uses the standard skip amount
 - When wrapping the logger, only calls from the wrapped logger have an incremented skip amount
 - Both the unwrapped and wrapped loggers should work correctly at the same time (e.g. no globals)

I think the user shouldn't have to configure skip levels, these should be handled by the wrappers. I don't think we've come up with the right design for this yet though. cc @akshayjshah

justyntemme:
I do not think i have the time to re-work the PR for that situation. I agree it is a better solution. Will close this PR to allow for others to help. Thanks
-justyn

PR #68, opened 2016-05-27
Defer serializing Stringers

akshayjshah:
Since we've already got an `interface{}` in the field struct, we might as well defer calling `String()`.


abhinav:
LGTM

prashantv:
lgtm

Issue #58, opened 2016-05-25
Add an http.Handler to make runtime logging changes


akshayjshah:
Add a constructor that takes a zap logger and returns an HTTP handler to manage common runtime changes (e.g., changing the log level). Applications can mount that handler wherever they like.

pravj:
Hi @akshayjshah, I have worked on a prototype for this. Am I allowed to send a PR?

akshayjshah:
Of course! I'd love to take a look at what you have.

Please make sure to [sign Uber's Contributor License Agreement](https://docs.google.com/a/uber.com/forms/d/1pAwS_-dA1KhPlfxzYLBqK6rsSWwRwH95OCCZrcsY5rk/viewform) when you open the PR.

akshayjshah:
Added in #71 and #72.

PR #71, opened 2016-05-29
Add http.Handler to enable runtime changes (#58)

pravj:
- [x] [signed Uber's Contributor License Agreement](https://docs.google.com/a/uber.com/forms/d/1pAwS_-dA1KhPlfxzYLBqK6rsSWwRwH95OCCZrcsY5rk/viewform)
- [x] added tests to cover your changes.
- [x] ran the test suite locally (`make test`).
- [x] ran the linters locally (`make lint`).

I have already signed the CLA for my previous pull request #53.

---

The pull request implements a `HTTPHandler` struct that takes a `Logger` instance as a field.
It provides a `ChangeLogLevel` method that accepts a `Level` argument and returns a `http.Handler`.

Right now the handler will set the new log level and returns the response after checking the level via `Enabled` method. The response string is "true" if the level is in effect and "false" otherwise.

* What are your comments on this response workflow?
* I feel that a good enhancement would be passing the log level through the request argument, what do you think?


pravj:
This is a sample code [(gist)](https://gist.github.com/pravj/1a043a498f5e350a47ad252031a74276) I have written that demonstrates its working.

It can be viewed as an application which is trying to load its configurations initially.
So the log level is `Debug` in starting. The application is emitting both `Info` and `Debug` messages.

Once the config files are distributed perfectly (simulated by time.Sleep). It makes a request using the provided `Handler` and the log level is changed to `Info`. After that, the `Debug` messages are ignored.

```go
package main

import (
    "os"
    "fmt"
    "time"
    "net/http"
    "github.com/uber-go/zap"
)

func main() {
    writeSyncer := zap.AddSync(os.Stderr)
    logger := zap.NewJSON(
        zap.Debug,
        zap.Output(writeSyncer),
        zap.ErrorOutput(writeSyncer),
    )

    go func() {
        var i int
        for i <= 4 {
            logger.Debug(fmt.Sprintf("Debug - loading config - time left: %v", 5 - i))
            logger.Info(fmt.Sprintf("Info - loading config - time left: %v", 5 - i))

            time.Sleep(time.Millisecond * 1000)
            i = i + 1
        }

        _, err := http.Get("http://localhost:3000/change")
        if err != nil {
            logger.Fatal("Error in GET request")
        }
        
        // intentionally leaving the check for response data

        i = 1
        for i <= 4 {
            logger.Debug(fmt.Sprintf("Debug - loaded config - since: %v", i))
            logger.Info(fmt.Sprintf("Info - loaded config - since: %v", i))

            time.Sleep(time.Millisecond * 1000)
            i = i + 1
        }

        os.Exit(0)
    }()

    hh := zap.NewHTTPHandler(logger)

    mux := http.NewServeMux()
    mux.Handle("/change", hh.ChangeLogLevel(zap.Info))

    http.ListenAndServe(":3000", mux)
}
```

logger.go
```diff
@@ -38,7 +38,7 @@ type Logger interface {
 	// Check the minimum enabled log level.
 	Level() Level
 	// Change the level of this logger, as well as all its ancestors and
-	// descendants. This makes makes it easy to change the log level at runtime
+	// descendants. This makes it easy to change the log level at runtime
```
akshayjshah: Thanks for catching this 👍

http_handler.go
```diff
@@ -0,0 +1,52 @@
...
+
+		var res = "false"
+		if h.logger.Enabled(lvl) {
+			res = "true"
+		}
+		w.Write([]byte(res))
+	}
+
+	return http.HandlerFunc(fn)
+}
```
akshayjshah: This is an interesting approach. I'd imagined something more flexible, where `NewHTTPHandler` returns an `http.Handler` directly. That handler accepts GET and PUT requests; GETs return the current logging level, and PUTs accept a form-encoded level.

I'd really like something like this to work:

```bash
$ curl http://localhost:1234/  # shows the current log level
$ curl --data "level=info" http://localhost:1234/  # sets log level to info & shows log level
```

http_handler_test.go
```diff
@@ -0,0 +1,62 @@
...
+		h := NewHTTPHandler(jl)
+
+		// test server to handle the http.Handler
+		ts := httptest.NewServer(h.ChangeLogLevel(Debug))
+		defer ts.Close()
+
+		res, err := http.Get(ts.URL)
+		if err != nil {
+			jl.Fatal("Error in GET request")
+		}
```
akshayjshah: Using `jl.Fatal` in the tests doesn't exit cleanly, so the `go test` command doesn't show which test failed. Instead, use `github.com/stretchr/testify/require`, which cleanly stops the test as soon as an assertion fails.

In this case, you'd replace the `if err != nil` block with `require.NoError(t, err, "Error making GET request.")`.

http_handler.go
pravj: Ohh, that sounds good.

So we will be exporting just a function `NewHTTPHandler` returning the `http.Handler`, and it will act according to the HTTP method.

* GET - Retrieve log level
* PUT - Update log level based on the form-encoded data

Thanks for clearing both my doubts. I will implement it soon.

prashantv: If the handler is only going to be used to get/set levels, do we need form-encoded data? Should it just be the level as data? That way to modify the level, it would just be
```
// assume the handler is registered as /debug/loglevel

$ curl -X PUT -d "info" http://localhost:1234/debug/loglevel
```

http_handler.go
```diff
@@ -0,0 +1,52 @@
...
+
+package zap
+
+import (
+	"net/http"
+)
+
+// A HTTPHandler takes a logger instance and provides methods to enable
+// runtime changes to the logger via http.handler interface.
+type HTTPHandler struct {
```
prashantv: This shouldn't need to be exported

http_handler.go
akshayjshah: I'd rather get form-encoded data (1) we can extend the handler's functionality later if necessary, and (2) we can also support JSON in the same handler (assuming that the caller sets the content type correctly).

Even if we don't want to build fancy debug pages, accepting JSON and form-encoded input will make it much easier for other folks to build nice config pages.

prashantv: I think we should be consistent in the `GET` and `PUT` methods.

If we want the ability to extend the handlers in future, then `GET` should also return a key-value list, probably as JSON initially. Ideally, the response from `GET` is a valid input for `PUT`.

Then it makes sense to also support form encoded key-value pairs as input to `PUT` as well (ideally based on content-type).

http_handler.go
pravj: Now the `NewHTTPHandler` function is returning a `http.Handler` directly (as suggested by @akshayjshah), it's no more in use.

http_handler_test.go
pravj: I have replaced these blocks using the `require` package. Thanks for the hint.

http_handler.go
pravj: I have implemented it to return JSON string response according to the discussion.

```sh
# assume that the handler is registered as /debug/loglevel
$ curl http://localhost:3000/debug/loglevel
{"level":"info"}

$ curl -X PUT -d '{"level":"warn"}' http://localhost:3000/debug/loglevel
{"level":"warn"}

$ curl http://localhost:3000/debug/loglevel
{"level":"warn"}
```

akshayjshah:
Great - thanks for being so flexible, @pravj! There are a few small style nits remaining, but I can fix those up in a follow-up change - I'll tag you in the PR so you can see the changes.

pravj:
@akshayjshah  @prashantv 
Thanks for your guidance, I believe I have been disturbing you with my questions.

I have learned so much just by reading the source code of zap, it's awesome.

* Should we also log events happening in the handler like 'receiving a request', 'bad request' etc.?

akshayjshah:
Nope, this is great as-is. At least for now, let's not log.

PR #72, opened 2016-05-31
Refactor HTTP handler

akshayjshah:
Refactor @pravj's HTTP handler a little - the core of the code is the same, but I rearranged things a bit.

- Create a struct for the handler, which makes it easier to separate the get and put logic.
- Return JSON even for non-200 responses.
- Leverage the `Level` type's `UnmarshalText` method.
- Use the spy logger instead of a JSON logger in these tests.

### DIFF
--- a/http_handler.go
+++ b/http_handler.go
@@ -25,49 +25,60 @@ import (
 	"net/http"
 )
 
-// payload struct defines the format of the request payload received
-type payload struct {
-	Level string `json:"level"`
+type httpPayload struct {
+	Level *Level `json:"level"`
 }
 
-// NewHTTPHandler takes a logger instance and provides methods to enable
-// runtime changes to the logger level via http.handler interface.
+type errorResponse struct {
+	Error string `json:"error"`
+}
+
+type levelHandler struct {
+	logger Logger
+}
+
+// NewHTTPHandler returns an HTTP handler that can change the logging level at
+// runtime.
+//
+// GET requests return a JSON description of the current logging level. PUT
+// requests change the logging level and expect a payload like
+//   {"level":"info"}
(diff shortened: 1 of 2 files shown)


http_handler.go
```diff
@@ -25,49 +25,60 @@ import (
...
+	json.NewEncoder(w).Encode(httpPayload{Level: &current})
+}
 
-		res := payload{currentLevel}
-		json.NewEncoder(w).Encode(res)
+func (h *levelHandler) putLevel(w http.ResponseWriter, r *http.Request) {
+	var p httpPayload
+	decoder := json.NewDecoder(r.Body)
+	if err := decoder.Decode(&p); err != nil {
+		h.error(w, "Request body must be well-formed JSON.", http.StatusBadRequest)
```
prashantv: optional nit: should we put `err` into the error message being returned?

http_handler_test.go
```diff
@@ -18,94 +18,114 @@
...
 
-		statusCode, responseStr := sampleHTTPClient(t, "GET", handler, nil)
+	req, err := http.NewRequest(method, ts.URL, reader)
+	require.NoError(t, err, "Error constructing %s request.", method)
 
-		assert.Equal(t, http.StatusOK, statusCode, "Unexpected response status code.")
-		assert.Equal(t, fmt.Sprintf("%s", responseStr), fmt.Sprintf("{\"level\":\"%v\"}\n", jl.Level().String()), "Unexpected logger level.")
-	})
-}
+	client := &http.Client{}
```
prashantv: nit: can just use `http.DefaultClient` instead

prashantv:
lgtm

PR #75, opened 2016-06-03
Use uber-go/atomic for atomic operations

### LINKED ISSUE #63: Convert to uber-go/atomic
Once we flip `uber-go/atomic` to be public, use it here.

prashantv:
Fixes #63


akshayjshah:
awesome.

PR #76, opened 2016-06-03
Add explanatory comment for zap.Err shortcut

### LINKED ISSUE #74: Why does zap.Err() have a hardcoded key name?
I noticed that of all the functions that return a zap.Field,  Err is the only one that doesn't let you specify the name.

I don't have any use case where this is a problem, but it just popped out at me as inconsistent and I didn't see any comments or docs around it explaining why it is different.

akshayjshah:
Fixes #74.

### DIFF
--- a/field.go
+++ b/field.go
@@ -95,7 +95,9 @@ func Time(key string, val time.Time) Field {
 	return Int64(key, val.UnixNano())
 }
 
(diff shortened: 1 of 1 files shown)


prashantv:
lgtm

PR #77, opened 2016-06-05
Rename levels and Err field constructor

### LINKED ISSUE #47: zap.Err breaks convention
All the other Field wrappers use the full type name, only zap.Err uses a shortened name.

akshayjshah:
Rename all the `Level` constants from `Foo` to `FooLevel`, which allows
us to rename the `Err` field constructor to `Error`. This is a large bag
of changes, but they were all created with `gorename`.

Fixes #47.


abhinav:
LGTM

prashantv:
code lgtm, might want to add a note about String/MarshalText and UnmarshalText, since we serialize/deserialize without the level suffix.

Maybe we should have an example which deserializes from YAML which shows how it's specified

Issue #57, opened 2016-05-25
Add a Debarkify wrapper


akshayjshah:
`zbark.Barkify` wraps a zap logger in the bark interface; `Debarkify` should do the opposite. This makes migrating from bark to zap (or vice versa) easier, since libraries and applications don't have to move in lock-step.

PR #82, opened 2016-06-20
Export field.AddTo to support wrappers

billf:
Export `Field.addTo` so it can be used by external wrapper libraries.

(This is step 1 towards addressing #57.)


akshayjshah:
:shipit: Looks good to me.

PR #85, opened 2016-06-23
Add debarkify to wrap bark.Logger in zap.Logger

billf:


billf:
i'll edit the commits/descriptions if this is ok. open questions around unit tests and behavior of a few of the receivers.

billf:
updated to use KeyValueMap from #89

zbark/debark_test.go
```diff
@@ -0,0 +1,140 @@
...
+	buf := &bytes.Buffer{}
+	logger := logrus.New()
+	logger.Out = buf
+	logger.Level = logrus.DebugLevel
+	return Debarkify(bark.NewLoggerFromLogrus(logger)), buf
+}
+
+func TestDebark_CastNoop(t *testing.T) {
+	orig := zap.NewJSON(zap.DebugLevel)
+	assert.Equal(t, orig, Debarkify(Barkify(orig)))
```
prashantv: This will do a deep equals, I think you want to do a pointer check here, `assert.True(t, orig == Debarkify(...), ...)`

zbark/debark.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (zbf zapperBarkFields) Fields() map[string]interface{} {
+	return zbf.KeyValueMap
+}
+
+func zapToBark(zfs []zap.Field) bark.LogFields {
+	zbf := zapperBarkFields{make(zwrap.KeyValueMap, len(zfs))}
+	for _, zf := range zfs {
+		_ = zf.AddTo(zbf)
```
prashantv: why the `_ =` here? just `zf.AddTo(zbf)` should be fine

zbark/debark.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (z *zapper) Check(l zap.Level, msg string) *zap.CheckedMessage {
+	if l < z.lvl {
+		return nil
+	}
+	return zap.NewCheckedMessage(z, l, msg)
+}
+
+type zapperBarkFields struct {
+	zwrap.KeyValueMap
```
prashantv: any reason why this is embedded?

zbark/debark.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (z *zapper) Error(msg string, fields ...zap.Field) {
+	z.bl.WithFields(zapToBark(fields)).Error(msg)
+}
+
+func (z *zapper) Panic(msg string, fields ...zap.Field) {
+	z.bl.WithFields(zapToBark(fields)).Panic(msg)
+}
+
+func (z *zapper) Fatal(msg string, fields ...zap.Field) {
```
prashantv: the `SetLevel` doesn't impact any of these methods. I'd expect that calling `SetLevel(zap.ErrorLevel)` would mean that any `zap.Debug` logs I do would be ignored

zbark/debark.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/uber-common/bark"
+)
+
+// Debarkify wraps bark.Logger to make it compatible with zap's JSON logger
+func Debarkify(b bark.Logger) zap.Logger {
+	if wrapper, ok := b.(*barker); ok {
+		return wrapper.zl
+	}
+	return &zapper{
+		lvl: zap.DebugLevel,
```
prashantv: rather than default to debug, should we take level as an argument to Debarkify? I'm not 100% sure how we want zap's levels to interact with the logrus loggers

zbark/debark.go
billf: it was a note to myself to consider checking the error value and do something if possible.

zbark/debark_test.go
billf: good catch

zbark/debark.go
billf: yeah, i picked `zap.DebugLevel` so that `Log()` would default open and `SetLevel()` could be used to clamp down if desired. i'm not sure what callers expect.

zbark/debark.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (z *zapper) Level() zap.Level {
+	return z.lvl
+}
+
+// Change the level of this logger, as well as all its ancestors and
+// descendants. This makes it easy to change the log level at runtime
+// without restarting your application.
+func (z *zapper) SetLevel(l zap.Level) {
+	z.lvl = l
```
billf: panic if unknown level?

zbark/debark.go
prashantv: Yeah I think that makes sense

zbark/debark.go
prashantv: I'm fine with not checking for now, we don't check in the underlying `jsonLogger` implementation either

zbark/bark.go
```diff
@@ -29,89 +29,89 @@ import (
 	"github.com/uber-common/bark"
 )
 
-// Barkify wraps zap's JSON logger to make it compatible with the bark.Logger
+// Barkify wraps zap's JSON barker to make it compatible with the bark.Logger
```
prashantv: find and replace typo I assume: s/barker/logger

zbark/debark.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+}
+
+type zapper struct {
+	lvl zap.Level
+	bl  bark.Logger
+}
+
+func (z *zapper) DFatal(msg string, fields ...zap.Field) {
+	// Implement development/DFatal?
```
prashantv: nit: can you put `TODO: ` prefix on this comment

zbark/debark.go
```diff
@@ -0,0 +1,143 @@
...
+	case zap.WarnLevel:
+		bl.Warn(msg)
+	case zap.ErrorLevel:
+		bl.Error(msg)
+	case zap.PanicLevel:
+		bl.Panic(msg)
+	case zap.FatalLevel:
+		bl.Fatal(msg)
+	default:
+		panic("passed an unknown zap.Level")
```
prashantv: should we mention the unrecognized level in the output?
```go
panic(fmt.Errorf("passed an unknown zap.Level: %v", l))
```

zbark/debark.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (z *zapper) Check(l zap.Level, msg string) *zap.CheckedMessage {
+	if l < z.lvl {
+		return nil
+	}
+	return zap.NewCheckedMessage(z, l, msg)
+}
+
+type zapperBarkFields zwrap.KeyValueMap
```
prashantv: can we move this type to the top, and the implementation to the bottom. Ideally the implementation for `zapperBarkFields` and `zapToBark` are not in between the `zapper` methods

zbark/debark_test.go
```diff
@@ -0,0 +1,153 @@
...
+		assert.Equal(t, 0, buf.Len(), "buffer not zero")
+		logger.Log(l, "ohai")
+		assert.NotEqual(t, 0, buf.Len(), "%q did not log", l)
+		buf.Reset()
+	}
+
+	logger.SetLevel(zap.FatalLevel)
+	assert.Equal(t, 0, buf.Len(), "buffer not zero")
+	for _, l := range append(levels, zap.PanicLevel) {
+		logger.Log(l, "ohai")
```
prashantv: if you log using `Panic`, shouldn't it cause this test to panic and fail?

zbark/debark_test.go
```diff
@@ -0,0 +1,153 @@
...
+	logger, buf := newDebark()
+	logger.SetLevel(zap.DebugLevel)
+	for _, l := range levels {
+		buf.Reset()
+		assert.Equal(t, 0, buf.Len(), "buffer not zero")
+		logger.Log(l, "ohai")
+		assert.NotEqual(t, 0, buf.Len(), "%q did not log", l)
+		buf.Reset()
+	}
+
```
prashantv: if you set the level to something like `Error`, can you make sure that nothing logged with a lower level logs anything

zbark/debark_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/uber-go/zap"
+
+	"github.com/uber-common/bark"
+)
+
```
prashantv: apart from ensuring there's output, we're not really testing the output specifically. I don't think our tests should be tied to the logrus output format, but I also think we should make sure that the output is what we expect. One way to do this is:
 - Debarkify a logger, and log something with a few fields
 - Log the same message using logrus with the same fields
 - ensure both the output buffers are the same

(we should set logrus' `DisableTimestamp` so that the tests don't rely on time)

prashantv:
The code looks good, just a few comments on the tests. Let me know if you'd like help with the tests, I can help make some test changes too.

Can we also improve the title + description for this change?

Thanks @billf

zbark/debark_test.go
billf: this is the "nothing logged with a lower level logs anything" test. should `logger.Log(zap.PanicLevel, ...)` panic unconditionally, even in the level is lower? shoud `logger.Log(zap.FatalLevel, ...)` always fatal regardless of level?

prashantv: That's a good question. I think the answer is yes -- even if the logger is set to ignore panic logs, we should still panic, but skip logging the message.

I verified behaviours fatal/panic behaviours in a branch:
https://github.com/uber-go/zap/compare/pv_panic_always?expand=1

This actually found a minor issue with how the `Check` works for panics and fatals, but otherwise, the logger works as it should by default, it will skip the log but still panic/fatal.

billf:
recent update: port #126 changes here as well

zbark/debark_test.go
```diff
@@ -0,0 +1,198 @@
...
+func TestDebark_LeveledLogging(t *testing.T) {
+	logger, buf := newDebark()
+	logger.SetLevel(zap.DebugLevel)
+	for _, l := range levels {
+		buf.Reset()
+		assert.Equal(t, 0, buf.Len(), "buffer not zero")
+		logger.Log(l, "ohai")
+		assert.NotEqual(t, 0, buf.Len(), "%q did not log", l)
+		buf.Reset()
```
prashantv: do you need both the `buf.Reset()` calls? I think the one at the beginning can be removed?

zbark/example_test.go
```diff
@@ -39,3 +44,18 @@ func Example() {
 	// Output:
 	// {"level":"info","msg":"Bark accepts arbitrary types.","errors":0}
 }
+
+func ExampleDebarkify() {
+	logrusLogger := logrus.New()
+	logrusLogger.Out = os.Stdout
+	logrusLogger.Formatter = new(logrus.JSONFormatter)
+	logrusLogger.Formatter.(*logrus.JSONFormatter).TimestampFormat = "lies"
```
prashantv: can the above 2 lines be collpased into:
```go
logrusLogger.Formatter = &logrus.JSONFormatter{TimestampFormat: "..."}
```

which should also avoid the typecast

zbark/example_test.go
```diff
@@ -39,3 +44,18 @@ func Example() {
...
+	barkLogger := bark.NewLoggerFromLogrus(logrusLogger).WithField("errors", 0)
+	logger := zbark.Debarkify(barkLogger, zap.DebugLevel)
+
+	// The wrapped logger has zap's actually fluent API.
+	logger.Info("Zap accepts", zap.String("typed", "fields"))
+
+	// Output:
+	// {"errors":0,"level":"info","msg":"Zap accepts","time":"lies","typed":"fields"}
```
prashantv: does logrus output an `errors` field for every log?

prashantv:
lgtm, just a couple of nits, and it's ready to merge. Thanks for being so patient @billf

zbark/example_test.go
billf: yeah, sorry, this is a remnant from when i was modifying the default `logrus.TextFormatter` instead of creating one.

zbark/example_test.go
```diff
@@ -39,3 +44,18 @@ func Example() {
 	// Output:
 	// {"level":"info","msg":"Bark accepts arbitrary types.","errors":0}
 }
+
+func ExampleDebarkify() {
+	logrusLogger := logrus.New()
+	logrusLogger.Out = os.Stdout
+	logrusLogger.Formatter = new(logrus.JSONFormatter)
+	logrusLogger.Formatter.(*logrus.JSONFormatter).TimestampFormat = "lies"
+	barkLogger := bark.NewLoggerFromLogrus(logrusLogger).WithField("errors", 0)
```
billf: `errors` comes from here and i was just matching the field use above in `ExampleBarkify()`

prashantv: Ah I see, OK makes sense.

billf:
vet failure seems spurious:

```
zbark/example_test.go:33: ExampleBarkify refers to unknown identifier: Barkify
zbark/example_test.go:48: ExampleDebarkify refers to unknown identifier: Debarkify
```

prashantv:
vet can give flaky errors if the package is not installed. We might need to do a `go test -i $(PKGS)` before we run the tests.

billf:
vet issue is addressed in #129

prashantv:
lgtm, thanks @billf

Issue #87, opened 2016-06-27
Consolidate Nest and AddMarshaler in KeyValue interface


prashantv:
They're basically the same thing -- one takes an interface, one takes a function directly. We should be able to reduce the API by removing `Nest`, and instead creating a type, `LogMarshalerFunc` that will take a `func(KeyValue) error` and wrapping it to make it a `LogMarshaler`.

PR #88, opened 2016-06-27
Add parameter names to the KeyValue interface

prashantv:
Parameter serve as documentation, especially for AddString which accepts
two strings `(string, string)` vs `(key, value string)`

### DIFF
--- a/keyvalue.go
+++ b/keyvalue.go
@@ -26,14 +26,14 @@ package zap
 //
 // See Marshaler for an example.
 type KeyValue interface {
-	AddBool(string, bool)
-	AddFloat64(string, float64)
-	AddInt(string, int)
-	AddInt64(string, int64)
-	AddMarshaler(string, LogMarshaler) error
+	AddBool(key string, value bool)
+	AddFloat64(key string, value float64)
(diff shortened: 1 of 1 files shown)


keyvalue.go
```diff
@@ -26,14 +26,14 @@ package zap
...
+	AddInt64(key string, value int64)
+	AddMarshaler(key string, marshaler LogMarshaler) error
 	// AddObject uses reflection to serialize arbitrary objects, so it's slow and
 	// allocation-heavy. Consider implementing the LogMarshaler interface instead.
-	AddObject(string, interface{})
-	AddString(string, string)
-	Nest(string, func(KeyValue) error) error
+	AddObject(key string, value interface{})
+	AddString(key, value string)
+	Nest(key string, f func(KeyValue) error) error
```
prashantv: I don't think `f` is a great name here, but I want to remove it very soon anyway (https://github.com/uber-go/zap/issues/87), so didn't spend much time coming up with a better name for it.

abhinav:
👍

akshayjshah:
:shipit:

PR #89, opened 2016-06-27
Add KeyValueMap, a KeyValue implemented backed by a map

prashantv:
This is a useful utility when testing custom MarhsalLog implementations

cc @billf: Can you make sure that this works for you to use in #85

### DIFF
--- a/zwrap/kvmap.go
+++ b/zwrap/kvmap.go
@@ -1,0 +1,56 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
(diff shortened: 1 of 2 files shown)


akshayjshah:
What do you think about putting this in its own package? Once we clean up and export the `encoder` type, we'll want this to implement the encoder interface, and we'll probably want to use it in some tests, in `zwrap`, and in `zbark`.

prashantv:
Yep, I thought it belonged to another package, but wasn't sure what that package should be. 

I considered `zencoder` but don't really think that's right. Any suggestions?

akshayjshah:
How about `mapencoder`? `mapencoder.New` sounds like a reasonable constructor.

prashantv:
Hmm I'm not sure about encoder in the name since it's an implementation of the `KeyValue` interface. The interface abstracts the JSON encoder and encoding in general, but the `KeyValueMap` doesn't really fulfill a full encoder, it only handles the fields.

However, I don't have strong opinions on this so if we can't come up with anything better, then `mapencoder` is OK.

akshayjshah:
lgtm. let's drop it wherever makes sense, maybe in the core zap package.

prashantv:
Leaving it in `zwrap` for now, we can move it in to core later as we make changes for 1.0

PR #90, opened 2016-06-27
Add benchmark for Nest, remove extra alloc in BenchmarkMarshalerField

prashantv:
Some pre-work for #87

### DIFF
--- a/json_encoder_bench_test.go
+++ b/json_encoder_bench_test.go
@@ -34,14 +34,15 @@ type logRecord struct {
 	Fields  map[string]interface{} `json:"fields"`
 }
 
+var s string
+
 func BenchmarkJSONEncoderNest(b *testing.B) {
-	f := func(kv KeyValue) error {
-		kv.AddString("s", "string")
-		return nil
-	}
 	for i := 0; i < b.N; i++ {
 		enc := newJSONEncoder()
-		enc.Nest("nested", f)
+		enc.Nest("nested", func(kv KeyValue) error {
+			kv.AddInt("i", i)
+			return nil
+		})
 		enc.Free()
 	}
 }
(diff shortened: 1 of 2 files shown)


json_encoder_bench_test.go
```diff
@@ -34,6 +34,18 @@ type logRecord struct {
...
+	f := func(kv KeyValue) error {
+		kv.AddString("s", "string")
+		return nil
+	}
+	for i := 0; i < b.N; i++ {
+		enc := newJSONEncoder()
+		enc.Nest("nested", f)
+		enc.Free()
+	}
+}
```
akshayjshah: Does this benchmark reflect how we expect this method to be used? I'd expected most users to pass an anonymous function, like this:

```go
enc := newJSONEncoder()
enc.Nest("nested", func(kv KeyValue) error {
  kv.AddString("s", "string")
  return nil
})
enc.Free()
```

logger_bench_test.go
```diff
@@ -41,7 +41,7 @@ func (u user) MarshalLog(kv zap.KeyValue) error {
 	return nil
 }
 
-var _jane = user{
+var _jane = &user{
```
akshayjshah: Thanks for explaining this more to me offline. Can we also change the user struct's `MarshalLog` method to have a pointer receiver?

json_encoder_bench_test.go
prashantv: Good point, replaced with something that is more realistic by having an inline closure + use of a variable as the parameter (which actually makes it ~10ns slower).

logger_bench_test.go
prashantv: 👍

prashantv:
Updated

akshayjshah:
lgtm. Merge away.

PR #91, opened 2016-06-28
Replace Nest in KeyValue interface with LogMarshalerFunc type

### LINKED ISSUE #87: Consolidate Nest and AddMarshaler in KeyValue interface

prashantv:
This reduces the KeyValue interface while still allowing a function
to be used to build an object using LogMarshalerFunc

Fixes #87

### DIFF
--- a/field_test.go
+++ b/field_test.go
@@ -163,6 +163,11 @@ func TestNestField(t *testing.T) {
 	assertCanBeReused(t, nest)
 }
 
+func TestLogMarshalerFunc(t *testing.T) {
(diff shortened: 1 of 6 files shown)


akshayjshah:
lgtm

PR #93, opened 2016-06-29
Encode errors under different keys

### LINKED ISSUE #43: Encode errors under different keys
If we run into an error encoding a value, we're currently just putting the error message under the original key. (I.e., we output `"foo":"failed to serialize User"`.) Instead, we should put the error message under a different key (I.e., `"foo-error":"failed to serialize User"`). This will help users emit data with a more consistent schema.

akshayjshah:

Currently, we don't handle encoding errors in a consistent way - `AddObject` serializes the error into the log message, but `AddMarshaler` logs the message to standard out. Even though very few portions of the code can fail at runtime, we end up checking and managing errors all over the place.

This PR allows the encoder to return an error for `AddObject` and `AddMarshaler`, which are the only two methods that can fail. However, the field handles that error by serializing it under a different key
(errors for the key `foo` are serialized under `fooError`).

This should make our APIs less cumbersome and our error handling more consistent. It fixes #43.


akshayjshah:
Force-pushed after resolving merge conflict.

prashantv:
lgtm

akshayjshah:
Resolved more merge conflicts, force-pushed. Will merge once tests pass.

Issue #92, opened 2016-06-28
zap.Error() nil pointer dereference


knarz:
Hello,

`zap.Error()` is a convenience method so I feel like making it handle `nil` errors gracefully seems like a reasonable move.

Example use case: I have an http middleware that logs all requests. The wrapped http handler returns an `error` and then the next line might be something like this:
```go
logger.Info("time to process request", zap.Error(err), zap.Duration("delta", time.Since(t0)), zap.String("path", path))
```

With the current `zap.Error()` this would panic if `err = nil` and require a check for nil to prevent just that.

I'd be happy to submit a PR if this seems like a sane addition.

akshayjshah:
Great point - I'd love a PR and a unit test to cover it!

prashantv:
Fixed by #95.

PR #95, opened 2016-06-29
allows Error() to handle nil as an argument

knarz:
This is the PR for #92 

Is the empty string appropriate for a nil error or should it be `null` instead?

### DIFF
--- a/field.go
+++ b/field.go
@@ -38,6 +38,7 @@ const (
 	marshalerType
 	objectType
 	stringerType
+	skipType
 )
 
 // A Field is a deferred marshaling operation used to add a key-value pair to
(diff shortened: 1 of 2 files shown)


prashantv:
Another option: skip the field completely if the error is `nil`

akshayjshah:
Yep. I think the best option here is to add a new field type and make `AddTo` a no-op for that type.

knarz:
How about `skipType`?

akshayjshah:
Sure, that name works.

knarz:
Updated

prashantv:
lgtm, Thanks for the fix @knarz

PR #102, opened 2016-07-25
Clean up hooks and logger config

akshayjshah:
This PR is a mish-mash of interwoven changes I made on a plane. Oops.

It accomplishes a few goals:

1. It adds an `Entry` type, which packages together the data passed between loggers, hooks, and encoders. At the end of this series of diffs, we'll be able to move most of `Logger.WriteEntry`'s logic onto the entry itself, which will remove the ugly type-cast inside `jsonEncoder`.
2. It pulls implementation-agnostic logger configuration into a separate struct, which loggers can embed to reduce boilerplate. This also makes it easy for other logger implementations to work with zap's built-in options.
3. It switches to representing times as floating-point seconds since epoch. Most aggregation and indexing systems can work with second timestamps, but very few of them can work with nanosecond timestamps, so this is a more sensible default.

Remaining tasks for this round of changes:

1. Produce a similar implementation-agnostic configuration for encoders. This'll let us customize timestamp, level, and message handling, and it'll remove the type-cast from our encoders.
2. Flatten the serialized logs.
3. Lock file handles.
3. Clean up docs and examples.


hook.go
```diff
@@ -27,27 +27,33 @@ import (
...
+	// runtime.Caller.
+	_callerSkip = 3
+)
 
-// A hook is executed each time the Logger writes a message. It receives the
-// message's priority, the message itself, and the logging context, and it
-// returns a modified message and an error.
-type hook func(Level, string, KeyValue) (string, error)
+// A hook is executed each time the logger writes a message. It can modify the
+// message, but must not retain references to the message or any of its
```
prashantv: s/message/entry (since the whole entry is mutable)

hook.go
```diff
@@ -27,27 +27,33 @@ import (
...
 func AddCaller() Option {
-	return hook(func(_ Level, msg string, _ KeyValue) (string, error) {
+	return hook(func(e *Entry) error {
+		if e == nil {
+			return errHookNilEntry
+		}
 		_, filename, line, ok := runtime.Caller(_callerSkip)
 		if !ok {
-			return msg, errors.New("failed to get caller")
+			return errors.New("failed to get caller")
```
prashantv: should we also have a top-level error for this similar to `errHookNilEntry`?

hook_test.go
```diff
@@ -68,3 +68,18 @@ func TestHookAddStacks(t *testing.T) {
 	logger.Debug("No stacks.")
 	assert.NotContains(t, buf.String(), "Unexpected stacktrace at Debug level.")
 }
+
+func TestHooksNilEntry(t *testing.T) {
+	tests := []struct {
+		hook hook
+		name string
+	}{
+		{hook: AddStacks(InfoLevel).(hook), name: "AddStacks"},
```
prashantv: nit: suggest putting name first, and then hook (in the struct, and inline here).

since it's also pretty simple, I'd be OK with dropping the field names in the inline struct, this seems pretty readable:
```go
{"AddStacks", AddStacks(InfoLevel).(hook)}
{"AddCaller", AddCaller().(hook)}
```

json_encoder.go
```diff
@@ -38,14 +38,20 @@ const (
...
-	New: func() interface{} {
-		return &jsonEncoder{
-			// Pre-allocate a reasonably-sized buffer for each encoder.
-			bytes: make([]byte, 0, _initialBufSize),
-		}
-	},
-}
+var (
+	// ErrNilEntry signals that Encoder.WriteEntry was called with a nil *Entry.
+	ErrNilEntry = errors.New("can't encode a nil *Entry")
```
prashantv: nit: prefer not exporting these errors. I don't think there's a reasonable way to use these errors in code since it would be a bad code bug.

json_encoder.go
```diff
@@ -38,14 +38,20 @@ const (
...
+	// ErrNilSink signals that Encoder.WriteEntry was called with a nil WriteSyncer.
+	ErrNilSink = errors.New("can't write encoded message a nil WriteSyncer")
+	jsonPool   = sync.Pool{
```
prashantv: nit: blank line to group the errors and the pool separately (i like to related vars together, even using separate blocks for unrelated variables)

json_encoder.go
```diff
@@ -151,24 +157,37 @@ func (enc *jsonEncoder) AddFields(fields []Field) {
 	addFields(enc, fields)
 }
 
-// WriteMessage writes a complete log message to the supplied writer, including
+// Bytes returns a mutable reference to the encoder's underlying bytes.
+func (enc jsonEncoder) Bytes() []byte {
```
prashantv: Can we remove this method, and just do type casts in `WriteEntry`

config.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/uber-go/atomic"
+)
+
+// Config is implementation-agnostic configuration for Loggers. Most Logger
+// implementations can reduce the required boilerplate by embedding a *Config.
+//
+// Note that while Level and SetLevel are safe for concurrent use, direct field
+// access and modification is not.
+type Config struct {
```
prashantv: As discussed offline, let's name this something other than `Config`. `Meta` is not great but good enough for now.

akshayjshah: Done.

json_encoder.go
akshayjshah: Done.

json_encoder.go
akshayjshah: Done.

json_encoder.go
akshayjshah: Done.

hook_test.go
akshayjshah: Fair enough - done.

hook.go
akshayjshah: Sure.

hook.go
akshayjshah: Done.

json_encoder.go
```diff
@@ -151,24 +156,32 @@ func (enc *jsonEncoder) AddFields(fields []Field) {
...
 
 	final.bytes = append(final.bytes, `{"msg":"`...)
-	final.safeAddString(msg)
+	final.safeAddString(e.Message)
 	final.bytes = append(final.bytes, `","level":"`...)
-	final.bytes = append(final.bytes, lvl...)
+	final.bytes = append(final.bytes, e.Level.String()...)
 	final.bytes = append(final.bytes, `","ts":`...)
-	final.bytes = strconv.AppendInt(final.bytes, ts.UnixNano(), 10)
+	final.bytes = strconv.AppendFloat(final.bytes, timeToSeconds(e.Time), 'f', -1, 64)
```
prashantv: Let's add a test for writing out a future time (maybe +100 years) and ensure that it's written without exponents

meta.go
```diff
@@ -0,0 +1,80 @@
...
+	"os"
+
+	"github.com/uber-go/atomic"
+)
+
+// Meta is implementation-agnostic state management for Loggers. Most Logger
+// implementations can reduce the required boilerplate by embedding a *Meta.
+//
+// Note that while Level and SetLevel are safe for concurrent use, direct field
+// access and modification are not.
```
prashantv: call out the specific field `lvl`? Otherwise it seems like no fields are safe for direct access

prashantv:
lgtm, just add a small test for the time formatting

entry.go
```diff
@@ -0,0 +1,62 @@
...
+package zap
+
+import (
+	"sync"
+	"time"
+)
+
+var (
+	_timeNow   = time.Now // for tests
+	_entryPool = sync.Pool{New: func() interface{} { return &Entry{} }}
```
tamird: this is pretty odd. ISTM that `Entry` is safe to copy, why bother using pointers at all? Just pass it around by value.

meta.go
```diff
@@ -0,0 +1,80 @@
...
+	Output      WriteSyncer
+	ErrorOutput WriteSyncer
+
+	lvl *atomic.Int32
+}
+
+// NewMeta returns a new meta struct with sensible defaults: logging at
+// InfoLevel, a JSON encoder, development mode off, and writing to standard error
+// and standard out.
+func NewMeta() *Meta {
```
tamird: another example of something which appears to be relatively safe to pass around by value. the one exception is `(hook).apply()` which has to take it as a pointer, but there's no reason for the embedding contract of this structure to be by-pointer.

entry.go
akshayjshah: It is safe to copy, but it's a little large - at some point pre-1.0, we should test to see which option is faster.

tamird:
```go
type Entry struct {
	Level   Level // int32
	Time    time.Time // int64, int32, pointer
	Message string // represented as reflect.StringHeader: uintptr, int
	enc     encoder // interface, two pointers
}
```

Assuming 64-bit, that gives us:

4 + 8 + 4 + 8 + 8 + 8 + 8 + 8 = 56 bytes

Hard to imagine that using a sync.Pool is faster?

meta.go
akshayjshah: All the options need to take this as a pointer, since they're modifying it.

tamird: It's still better to embed it by value since e.g. jsonLogger is already always boxed. In other words, embedding by value into a struct that is already allocated on the heap saves an allocation over embedding by pointer.

entry.go
tamird: Regarding your amendment: some benchmarking is needed, since creating the entry on the stack and passing a reference to the hooks may still cause the Go compiler to promote the Entry to the heap. Such is the price of fast compilation :(

meta.go
akshayjshah: Ah, I see what you're saying. Good point, will change in a future PR.

PR #103, opened 2016-07-26
Add Base64 field for encoding byte slices

### LINKED ISSUE #97: Add support for []byte
Are there any plans to add support for []byte type? I'm not gonna lie, I did a TL;DR on the source code.

prashantv:
The `Base64` field encodes the byte slice into a base64 string (with padding).

The data is encoded immediately for a couple of reasons:
 - all other primitive types (ints, strings, etc.) are either immutable or the value is copied. Since byte slices are mutable, if we kept a reference around, it may change between creating the field and encoding the value.
 - we don't want to add yet another union field to the `Field` type. We could store the contents as a `string` but we'd either need to do a full copy of the contents, or use unsafe to create a string out of the byte slice.

Fixes #97

### DIFF
--- a/field.go
+++ b/field.go
@@ -21,6 +21,7 @@
 package zap
 
 import (
+	"encoding/base64"
 	"fmt"
 	"math"
 	"time"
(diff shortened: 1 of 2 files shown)


field.go
```diff
@@ -52,6 +53,12 @@ type Field struct {
 	obj       interface{}
 }
 
+// Base64 constructs a field that encodes the given value as a
+// padded base64 string.
```
akshayjshah: Can we amend this comment to clarify that the encoding is eager?

akshayjshah:
lgtm (with comment change)

field.go
prashantv: Done

PR #104, opened 2016-07-26
Always use grade-school notation for floats

akshayjshah:
In JSON output, always use grade-school notation for floats. Exponential
notation is allowed by the JSON spec, but this is more obviously
correct (and it matches the desired behavior for timestamps, which will
soon use float fields).


prashantv:
lgtm

PR #105, opened 2016-07-26
Add options for JSON encoders

akshayjshah:
Let users specify how they'd like the message, level, and timestamp formatted for each log entry. 

In order to keep the proposed change small, this PR introduces a small performance regression in `jsonEncoder.WriteEntry` - it copies more bytes than necessary. In the next PR, I'm going to flatten the message structure, which will remove this extra copy and require many changes to the test code.

(I'm holding off on marking the related issues resolved until we export the `newJSONEncoder` function.)

### DIFF
--- a/encoder.go
+++ b/encoder.go
@@ -20,7 +20,10 @@
 
 package zap
 
-import "io"
+import (
+	"io"
+	"time"
+)
 
 // encoder is a format-agnostic interface for all log field encoders. It's not
 // safe for concurrent use.
@@ -30,5 +33,5 @@ type encoder interface {
 	AddFields([]Field)
 	Clone() encoder
 	Free()
-	WriteEntry(io.Writer, *Entry) error
+	WriteEntry(io.Writer, string, Level, time.Time) error
 }
--- a/json_encoder.go
+++ b/json_encoder.go
@@ -74,7 +74,7 @@ func newJSONEncoder(options ...JSONOption) *jsonEncoder {
 	enc.timeF = defaultTimeF
 	enc.levelF = defaultLevelF
 	for _, opt := range options {
-		opt.Apply(enc)
+		opt.apply(enc)
 	}
 
 	return enc
@@ -189,25 +189,22 @@ func (enc *jsonEncoder) WriteEntry(sink io.Writer, msg string, lvl Level, t time
 	// will require updating many tests.
 	final.truncate()
 	final.bytes = append(final.bytes, '{')
-	final.AddFields([]Field{
-		final.messageF(msg),
-		final.levelF(lvl),
-		final.timeF(t),
-	})
+	final.messageF(msg).AddTo(final)
+	final.levelF(lvl).AddTo(final)
+	final.timeF(t).AddTo(final)
 	final.bytes = append(final.bytes, `,"fields":{`...)
 	final.bytes = append(final.bytes, enc.bytes...)
 	final.bytes = append(final.bytes, "}}\n"...)
 
+	expectedBytes := len(final.bytes)
 	n, err := sink.Write(final.bytes)
+	final.Free()
 	if err != nil {
-		final.Free()
 		return err
 	}
-	if n != len(final.bytes) {
-		final.Free()
-		return fmt.Errorf("incomplete write: only wrote %v of %v bytes", n, len(final.bytes))
+	if n != expectedBytes {
+		return fmt.Errorf("incomplete write: only wrote %v of %v bytes", n, expectedBytes)
 	}
-	final.Free()
 	return nil
 }
 
--- a/json_encoder.go
+++ b/json_encoder.go
@@ -28,6 +28,7 @@ import (
 	"math"
 	"strconv"
 	"sync"
+	"time"
 	"unicode/utf8"
 )
 
@@ -39,10 +40,13 @@ const (
 )
 
 var (
-	// errNilEntry signals that Encoder.WriteEntry was called with a nil *Entry.
-	errNilEntry = errors.New("can't encode a nil *Entry")
 	// errNilSink signals that Encoder.WriteEntry was called with a nil WriteSyncer.
 	errNilSink = errors.New("can't write encoded message a nil WriteSyncer")
+
+	// Default formatters for JSON encoders.
+	defaultMessageF = MessageKey("msg")
+	defaultTimeF    = EpochFormatter("ts")
+	defaultLevelF   = LevelString("level")
 
 	jsonPool = sync.Pool{New: func() interface{} {
 		return &jsonEncoder{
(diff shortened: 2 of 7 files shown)


encoder.go
```diff
@@ -30,5 +33,5 @@ type encoder interface {
 	AddFields([]Field)
 	Clone() encoder
 	Free()
-	WriteEntry(io.Writer, *Entry) error
+	WriteEntry(io.Writer, string, Level, time.Time) error
```
akshayjshah: This undoes a change made in the previous PR - after some more thought, I think it's better to keep encoders and entries separate. This removes the concern about type mismatches between the receiver of `WriteEntry` and `Entry.encoder`.

I do still think that the `WriteEntry` name is appropriate, though, since we're writing out a single log line.

json_options.go
```diff
@@ -0,0 +1,88 @@
+package zap
+
+import "time"
+
+// JSONOption is used to set options for a JSON encoder.
+type JSONOption interface {
```
prashantv: Does this need to be exported?

prashantv: I'd also unexport the `Apply` method so that the custom formatter types won't display `Apply` in the documentation

json_options.go
```diff
@@ -0,0 +1,88 @@
...
+// MessageKey encodes log messages under the provided key.
+func MessageKey(key string) MessageFormatter {
+	return MessageFormatter(func(msg string) Field {
+		return String(key, msg)
+	})
+}
+
+// NoMessage drops the message field altogether. It's sometimes useful for
+// testing.
+func NoMessage() MessageFormatter {
```
prashantv: We're not using this in any of our tests now, I don't think we should add this to the exported API. It's trivial to implement outside of the library too

json_options.go
```diff
@@ -0,0 +1,88 @@
...
+// level's String method to serialize it.
+func LevelString(key string) LevelFormatter {
+	return LevelFormatter(func(l Level) Field {
+		return String(key, l.String())
+	})
+}
+
+// NoLevel drops the entry's log level altogether. It's occasionally useful in
+// testing.
+func NoLevel() LevelFormatter {
```
prashantv: Similar to `NoMessage`, suggest dropping this. `NoTime` is useful to provide since you can't test log output which contains time. This isn't true for message + levels, so don't need to make it as easy -- and if someone really wants to, they can easily write their own formatters to do this.

encoder.go
prashantv: SGTM

json_encoder.go
```diff
@@ -144,8 +159,12 @@ func (enc *jsonEncoder) AddObject(key string, obj interface{}) error {
 
 // Clone copies the current encoder, including any data already encoded.
 func (enc *jsonEncoder) Clone() encoder {
-	clone := newJSONEncoder()
+	clone := jsonPool.Get().(*jsonEncoder)
```
prashantv: why this change to no longer use `newJSONEncoder` which does the `jsonPool.Get()` + `clone.truncate`?

json_encoder.go
```diff
@@ -158,37 +177,37 @@ func (enc *jsonEncoder) AddFields(fields []Field) {
...
-	final.bytes = strconv.AppendFloat(final.bytes, timeToSeconds(e.Time), 'f', -1, 64)
+
+	final := enc.Clone().(*jsonEncoder)
+
+	// TODO: When we flatten the serialized messages, remove the extra truncate
+	// and copy below. This will change the shape of the serialized message, which
+	// will require updating many tests.
+	final.truncate()
+	final.bytes = append(final.bytes, '{')
+	final.AddFields([]Field{
```
prashantv: Rather than build up a slice, and then iterate over the slice etc, it will probably be faster to do:
```go
final.messageF(msg).AddTo(final)
final.levelF(lvl).AddTo(final)
final.timeF(t).AddTo(final)
```

Avoids a bunch of indirection.

json_encoder.go
```diff
@@ -158,37 +177,37 @@ func (enc *jsonEncoder) AddFields(fields []Field) {
...
 	final.bytes = append(final.bytes, `,"fields":{`...)
-	if e.enc != nil {
-		final.bytes = append(final.bytes, e.enc.(*jsonEncoder).bytes...)
-	}
+	final.bytes = append(final.bytes, enc.bytes...)
 	final.bytes = append(final.bytes, "}}\n"...)
 
 	n, err := sink.Write(final.bytes)
 	if err != nil {
+		final.Free()
```
prashantv: consider breaking up this function into 2, one that writes out the entry given a `*jsonEncoder`, and the caller just gets the encoder + calls free afterwards. I'm not a fan of multiple return paths having to remember to `Free`

json_encoder_test.go
```diff
@@ -266,3 +268,19 @@ func TestJSONJSONEscaping(t *testing.T) {
 		assertJSON(t, output, enc)
 	}
 }
+
+func TestJSONOptions(t *testing.T) {
+	enc := newJSONEncoder(
```
prashantv: can you also verify that if you call `Clone` and use that `Clone` to log, you get the same output. (I want to make sure we test the clone's options are set correctly too)

prashantv:
Completed the first pass, the formatters look good. I'd like to reduce the exported API a fair bit more though (unexporting `JSONOption` and related functions, and removing some of the skip formatters).

json_options.go
akshayjshah: Unexporting `Apply` definitely makes sense...not sure why I exported it to begin with. We should keep `JSONOption` exported, though, since it'll be part of the public `NewJSONEncoder` signature.

json_options.go
akshayjshah: Sure, can do.

json_encoder.go
akshayjshah: The constructor also sets the message/time/level formatters back to the default, which we don't want here. I could pass the current formatters in as options, but that'll allocate a slice for the variadic args.

json_encoder.go
akshayjshah: Good point.

json_encoder.go
akshayjshah: Turns out that I didn't think about this very carefully. We can just free the encoder after writing to the sink - no need to free immediately before each return.

prashantv:
lgtm

akshayjshah:
## Benchmarks

Pre:
```
BenchmarkZapAddingFields-4                       1000000              1390 ns/op             705 B/op          2 allocs/op
BenchmarkZapWithAccumulatedContext-4             5000000               355 ns/op               0 B/op          0 allocs/op
```

Post:
```
BenchmarkZapAddingFields-4                       1000000              1537 ns/op             705 B/op          2 allocs/op
BenchmarkZapWithAccumulatedContext-4             3000000               457 ns/op               0 B/op          0 allocs/op
```

PR #106, opened 2016-07-27
Support concurrent logging to a single WriteSyncer

### LINKED ISSUE #99: Logging With Concurrency
I tried using this logger in a concurrent application, but the messages will run overtop of each other in the log. I am wondering if I am using it wrong, or the library isn't concurrent safe. I have a simple example: 
```go
package main

import "github.com/uber-go/zap"

var logger zap.Logger

func main() {
	logger = zap.NewJSON()

	for i := 0; i < 800; i++ {
		go printer(i)
	}
}

func printer(i int) {
	logger.Info("Counting",
		zap.Int("i", i),
	)
}
```

Here is a section of output from the above program: 
```
{"msg":"Counting","level":"info","ts":1469196244924860064,"fields":{"i":8}}
{"msg":"Counting","level":"info","ts":1469196244924867134,"fields":{"i":42}}
{"msg":"Counting","level":"info","ts":1469196244924866589,"fields":{"i":20}}
{"msg":"Counti{"msg":"Counting","level":"info","ts":1469196244924879429,"fields":{"i":21}}
ng","level":"info","ts":1469196244924870959,"fields":{"i":9}}
{"msg":"Counting","level":"info","ts":1469196244924879050,"fields":{"i":36}}
{"msg":"Counting","level":"info","ts":1469196244924873937,"fields":{"i":43}}
{"msg":"Counting","level":"info","ts":1469196244924904543,"fields":{"i":22}}
{"msg":"Counting","level":"info","ts":1469196244924910632,"fields":{"i":10}}
{"msg":"Counting","level":"info","ts":1469196244924916792,"fields":{"i":23}}
```

prashantv:
Currently, the WriteSyncer assumes the underlying Writer will support
concurrent Writes correctly. However, this is not true for common
io.Writer implementations:
 - *os.File may call the write syscall multiple times
 - *bytes.Buffer does not support concurrent writes at all

Instead, we will wrap all passed in WriteSyncers in a lockedWriteSyncer
which will use a mutex to serialize calls to the underlying WriteSyncer.

In future, we may want to expose a way of setting a WriteSyncer that
already supports concurrent calls, but since most common implementations
cannot handle concurrent writes correctly, locking the writer seems like
a simpler default.

Fixes #99


logger_test.go
```diff
@@ -274,3 +275,41 @@ func TestJSONLoggerSyncsOutput(t *testing.T) {
...
+
+func TestLoggerConcurrent(t *testing.T) {
+	buf := &bytes.Buffer{}
+	withJSONLogger(t, []Option{Output(AddSync(buf))}, func(jl *jsonLogger, output func() []string) {
+		jl.StubTime()
+		jl2 := jl.With(String("foo", "bar"))
+
+		wg := &sync.WaitGroup{}
+		runNTimes(5, wg, func() {
+			for i := 0; i < 10; i++ {
```
akshayjshah: Why nest another loop here instead of setting `n` to 50 in `runNTimes`?

logger_test.go
```diff
@@ -274,3 +275,41 @@ func TestJSONLoggerSyncsOutput(t *testing.T) {
...
+
+		wg.Wait()
+
+		// Make sure the output doesn't contain interspersed entries.
+		expected := `{"msg":"info","level":"info","ts":0,"fields":{"foo":"bar"}}` + "\n"
+		assert.Equal(t, strings.Repeat(expected, 200), buf.String())
+	})
+}
+
+func runNTimes(n int, wg *sync.WaitGroup, f func()) {
```
akshayjshah: Is it worth capturing the full pattern you're using? This helper could easily be `runNTimes(n numGoroutines, m iterationsPerGoroutine, wg *sync.WaitGroup, f func())`.

prashantv: sure, can do that. It's a little less clear to have 2 ints as args, so I added comments for the arguments to make it clearer.

prashantv:
Updated with test fix

akshayjshah:
:shipit:

PR #107, opened 2016-07-28
Flatten serialized entries

akshayjshah:
Log aggregation and processing systems typically require some metadata in the top-level JSON. In zap, we included the message, level, and timestamp, but that's clearly not enough to support many use cases. We could support adding information to this envelope via a separate API, but the simplest option is to flatten the entry.

This is a small change to the production code, but it requires large changes to the tests. To keep review semi-sane, the production and test changes are in separate commits.

### DIFF
--- a/checked_message_test.go
+++ b/checked_message_test.go
@@ -28,16 +28,20 @@ import (
 )
 
 func TestJSONLoggerCheck(t *testing.T) {
-	withJSONLogger(t, opts(InfoLevel), func(jl *jsonLogger, output func() []string) {
-		assert.False(t, jl.Check(DebugLevel, "Debug.").OK(), "Expected CheckedMessage to be not OK at disabled levels.")
+	withJSONLogger(t, opts(InfoLevel), func(logger Logger, buf *testBuffer) {
+		assert.False(
+			t,
+			logger.Check(DebugLevel, "Debug.").OK(),
+			"Expected CheckedMessage to be not OK at disabled levels.",
+		)
 
-		cm := jl.Check(InfoLevel, "Info.")
+		cm := logger.Check(InfoLevel, "Info.")
 		require.True(t, cm.OK(), "Expected CheckedMessage to be OK at enabled levels.")
 		cm.Write(Int("magic", 42))
 		assert.Equal(
 			t,
-			`{"msg":"Info.","level":"info","ts":0,"fields":{"magic":42}}`,
-			output()[0],
+			`{"level":"info","msg":"Info.","magic":42}`,
+			buf.Stripped(),
 			"Unexpected output after writing a CheckedMessage.",
 		)
 	})
@@ -45,14 +49,14 @@ func TestJSONLoggerCheck(t *testing.T) {
 
 func TestCheckedMessageIsSingleUse(t *testing.T) {
 	expected := []string{
-		`{"msg":"Single-use.","level":"info","ts":0,"fields":{}}`,
-		`{"msg":"Shouldn't re-use a CheckedMessage.","level":"error","ts":0,"fields":{"original":"Single-use."}}`,
+		`{"level":"info","msg":"single-use"}`,
+		`{"level":"error","msg":"Shouldn't re-use a CheckedMessage.","original":"single-use"}`,
 	}
-	withJSONLogger(t, nil, func(jl *jsonLogger, output func() []string) {
-		cm := jl.Check(InfoLevel, "Single-use.")
+	withJSONLogger(t, nil, func(logger Logger, buf *testBuffer) {
+		cm := logger.Check(InfoLevel, "single-use")
 		cm.Write() // ok
 		cm.Write() // first re-use logs error
 		cm.Write() // second re-use is silently ignored
-		assert.Equal(t, expected, output(), "Expected re-using a CheckedMessage to log an error.")
+		assert.Equal(t, expected, buf.Lines(), "Expected re-using a CheckedMessage to log an error.")
 	})
 }
--- a/json_encoder.go
+++ b/json_encoder.go
@@ -182,19 +182,17 @@ func (enc *jsonEncoder) WriteEntry(sink io.Writer, msg string, lvl Level, t time
 		return errNilSink
 	}
 
-	final := enc.Clone().(*jsonEncoder)
-
-	// TODO: When we flatten the serialized messages, remove the extra truncate
-	// and copy below. This will change the shape of the serialized message, which
-	// will require updating many tests.
+	final := jsonPool.Get().(*jsonEncoder)
 	final.truncate()
 	final.bytes = append(final.bytes, '{')
-	final.messageF(msg).AddTo(final)
-	final.levelF(lvl).AddTo(final)
-	final.timeF(t).AddTo(final)
-	final.bytes = append(final.bytes, `,"fields":{`...)
-	final.bytes = append(final.bytes, enc.bytes...)
-	final.bytes = append(final.bytes, "}}\n"...)
+	enc.levelF(lvl).AddTo(final)
+	enc.timeF(t).AddTo(final)
+	enc.messageF(msg).AddTo(final)
+	if len(enc.bytes) > 0 {
+		final.bytes = append(final.bytes, ',')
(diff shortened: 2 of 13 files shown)


checked_message_test.go
```diff
@@ -28,31 +28,35 @@ import (
...
 
-		cm := jl.Check(InfoLevel, "Info.")
+		cm := logger.Check(InfoLevel, "Info.")
 		require.True(t, cm.OK(), "Expected CheckedMessage to be OK at enabled levels.")
 		cm.Write(Int("magic", 42))
 		assert.Equal(
 			t,
-			`{"msg":"Info.","level":"info","ts":0,"fields":{"magic":42}}`,
-			output()[0],
+			`{"magic":42,"msg":"Info.","level":"info"}`,
```
prashantv: While the JSON output isn't intended for users, it will often be consumed by humans. The current ordering of keys isn't very human-friendly, I'd prefer: time, level, msg, fields. (specifically, the time, level and message should be before the fields).

Since we're copying the bytes in clone anyway, I don't think it'll be any less efficient to create a new buffer, append the time/level/message fields before copying over the fields buffer.

json_encoder_test.go
```diff
@@ -36,16 +36,37 @@ import (
...
-	enc.AddString("foo", "bar")
 	f(enc)
 	enc.Free()
 }
 
+type testBuffer struct {
+	*bytes.Buffer
+}
+
+func newTestBuffer() *testBuffer {
```
prashantv: if you're returning a pointer here, you can embed the `bytes.Buffer` without a pointer. the other option is to use a value for `testBuffer` and leave `*bytes.Buffer` as a pointer. I personally prefer the first, since you could just do `&testBuffer{}` and use it without needing `newTestBuffer`

json_encoder_test.go
```diff
@@ -66,6 +88,7 @@ func TestJSONAddString(t *testing.T) {
 
 func TestJSONAddBool(t *testing.T) {
 	withJSONEncoder(func(enc *jsonEncoder) {
+		enc.AddString("foo", "bar")
```
prashantv: should we just delete the `"foo": "bar"` rather than calling `AddString` in these tests? Not sure what the purpose of the `AddString` is

json_encoder_test.go
```diff
@@ -126,49 +152,51 @@ func TestJSONAddFloat64(t *testing.T) {
...
+		entry.Time,
+	), "Expected an error writing to a nil sink.")
+
+	// Messages should be escaped.
+	err := enc.WriteEntry(sink, entry.Message, entry.Level, entry.Time)
+	assert.NoError(t, err, "WriteEntry returned an unexpected error.")
+	assert.Equal(
+		t,
+		`{"foo":"bar","msg":"hello\\","level":"info","ts":0}`,
+		sink.FirstLine(),
```
prashantv: Instead of `FirstLine`, can we have a method that returns the contents with the new line stripped? We want the tests to fail if there is an unexpected extra line, the easiest way seems to be to keep the previous logic, but have it as a method on `testBuffer`

logger_test.go
```diff
@@ -62,43 +61,43 @@ func stubExit() *stubbedExit {
...
-	jl.StubTime()
+	logger := newLogger(newJSONEncoder(NoTime()), allOpts...)
 
-	f(jl.(*jsonLogger), func() []string {
-		output := strings.Split(sink.String(), "\n")
-		return output[:len(output)-1]
-	})
-	assert.Empty(t, errSink.String(), "Expected error sink to be empty")
+	f(logger, sink)
+	assert.Empty(t, errSink.Lines(), "Expected error sink to be empty")
```
prashantv: when checking if it's empty, you can probably just use `errSink.String()`, no point trying to convert to lines right?

logger_test.go
```diff
@@ -62,43 +61,43 @@ func stubExit() *stubbedExit {
...
 }
 
-func assertMessage(t testing.TB, level, expectedMsg, actualLog string) {
-	expectedLog := fmt.Sprintf(`{"msg":"%s","level":"%s","ts":0,"fields":{}}`, expectedMsg, level)
-	assert.Equal(t, expectedLog, actualLog, "Unexpected log output.")
+func TestJSONLoggerSetLevel(t *testing.T) {
+	withJSONLogger(t, nil, func(logger Logger, _ *testBuffer) {
+		assert.Equal(t, DebugLevel, logger.Level(), "Unexpected initial level.")
+		logger.SetLevel(DebugLevel)
```
prashantv: Not sure how this test slipped in the way it is, but seems like we should be changing the level here instead of using the same value as the initial level?

json_encoder_test.go
akshayjshah: Validates that we're correctly separating fields with commas.

prashantv: Ah OK, by testing that there's a comma separating them, we're ignoring the case where there's no comma separator right now.

Can we have a test helper that's intended for testing adding a field that will run multiple cases:
 - Don't add a field before, make sure there's no comma
 - Add a field before (and after), make sure there's a comma

akshayjshah: No, that case is checked immediately after this assertion; the same is true for all the other AddFoo methods (any lacunae are bugs).

Nevertheless, I'll see about writing a good helper for this. It's a bit challenging to get the types right, since this code doesn't deal in Fields at all.

logger_test.go
akshayjshah: Yep, this is an oversight; the default used to be `AllLevel`.

json_encoder_test.go
akshayjshah: Yep, makes sense.

logger_test.go
akshayjshah: Sure.

akshayjshah:
Updated.

json_encoder.go
```diff
@@ -182,19 +182,20 @@ func (enc *jsonEncoder) WriteEntry(sink io.Writer, msg string, lvl Level, t time
...
+	enc.timeF(t).AddTo(final)
+	enc.messageF(msg).AddTo(final)
+	if len(enc.bytes) > 0 {
+		if len(final.bytes) > 1 {
+			// All the formatters may have been no-ops.
+			final.bytes = append(final.bytes, ',')
+		}
+		final.bytes = append(final.bytes, enc.bytes...)
+	}
+	final.bytes = append(final.bytes, "}\n"...)
```
prashantv: Let's change this to:
```go
final.bytes = append(final.bytes, '}', '\n')
```

It looks like this is slightly faster,
https://github.com/prashantv/go-bench/blob/master/append_bytes_test.go
```
BenchmarkAppendString-8 500000000                3.72 ns/op            0 B/op          0 allocs/op
BenchmarkAppendBytes-8  2000000000               1.63 ns/op            0 B/op          0 allocs/op
```

json_encoder_test.go
```diff
@@ -41,162 +41,154 @@ func assertJSON(t *testing.T, expected string, enc *jsonEncoder) {
...
+	return nil
 }
 
-func TestJSONAddString(t *testing.T) {
-	withJSONEncoder(func(enc *jsonEncoder) {
-		enc.AddString("baz", "bing")
-		assertJSON(t, `"foo":"bar","baz":"bing"`, enc)
+func (b *testBuffer) Lines() []string {
+	output := strings.Split(b.String(), "\n")
+	return output[:len(output)-1]
```
prashantv: can simplify this to:
```go
return strings.Split(b.Stripped(), "\n")
```

prashantv:
lgtm with minor comments

json_encoder.go
blampe: You guys are ridiculous :P

prashantv: Every nanosecond counts!

akshayjshah: 🎶 Eeeeevery nanosecond is precious... :notes:

json_encoder_test.go
akshayjshah: Unfortunately, that doesn't behave the same way when the buffer is empty - splitting the empty string returns `[]string{""}`, which isn't what we want.

PR #113, opened 2016-08-01
[v1, 2/8] Export the Encoder type

akshayjshah:
Export the Encoder type, since we'd like to allow external implementations. (Exporting a constructor for the JSON encoder comes in the next PR.)

This is the second of eight PRs to finish up the large pre-v1.0.0 refactoring:
- [ ] Export Encoder type
- [ ] Export the JSON encoder constructor
- [ ] Add a `NewLogger` constructor and remove `NewJSON`
- [ ] Remove `StubTime`
- [ ] Export the Hook type
- [ ] Improve GoDoc.

### DIFF
--- a/encoder.go
+++ b/encoder.go
@@ -25,13 +25,13 @@ import (
 	"time"
 )
 
-// encoder is a format-agnostic interface for all log field encoders. It's not
+// Encoder is a format-agnostic interface for all log field encoders. It's not
 // safe for concurrent use.
-type encoder interface {
+type Encoder interface {
 	KeyValue
 
 	AddFields([]Field)
-	Clone() encoder
+	Clone() Encoder
 	Free()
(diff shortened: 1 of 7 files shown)


encoder.go
```diff
@@ -25,13 +25,13 @@ import (
 	"time"
 )
 
-// encoder is a format-agnostic interface for all log field encoders. It's not
+// Encoder is a format-agnostic interface for all log field encoders. It's not
 // safe for concurrent use.
-type encoder interface {
+type Encoder interface {
```
prashantv: If we are exporting this, let's add comments on the below methods, especially `Clone`/`Free` since we're expecting the implementors to do the pooling.

akshayjshah: That's covered in PR 8/8, if that's okay.

prashantv: 👍

prashantv:
lgtm

PR #115, opened 2016-08-01
[v1, 4/8] Replace NewJSON constructor with NewLogger

akshayjshah:
Make the logger type name more generic, since it's not actually JSON-specific, and make the logger constructor take an encoder implementation. Since we're injecting encoders, we can stop using the ugly `StubTime` hack (which we'll remove altogether in the next PR).

This is the fourth of eight PRs to finish up the large pre-v1.0.0 refactoring:
- [x] Export Encoder type
- [ ] Add a `NewLogger` constructor and remove `NewJSON`
- [ ] Remove `StubTime`
- [ ] Export the Hook type
- [ ] Improve GoDoc.

### DIFF
--- a/benchmarks/zap_bench_test.go
+++ b/benchmarks/zap_bench_test.go
@@ -75,7 +75,7 @@ func fakeMessages(n int) []string {
 }
 
 func BenchmarkZapDisabledLevelsWithoutFields(b *testing.B) {
-	logger := zap.NewJSON(zap.ErrorLevel, zap.Output(zap.Discard))
+	logger := zap.NewLogger(zap.NewJSONEncoder(), zap.ErrorLevel, zap.Output(zap.Discard))
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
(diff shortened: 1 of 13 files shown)


prashantv:
Did you consider this alternate interface:
- `New` that creates a logger with all the default options + encoder
- a `zap.Encoder` which lets you pass a custom encoder as an option

This makes creating a new default logger a fair bit more verbose:
```go
zap.NewLogger(zap.NewJSONEncoder(), [..])
```
vs 
```go
zap.New([...])
```

And in the case that someone wants to customize the encoder, it's not very complicated either:
```go
zap.New(zap.Encoder(zap.NewJSONEncoder()), [...])
```

akshayjshah:
I considered that (and roughed out some code), but ended up rejecting it for two reasons:

1. It introduces an order dependency between the `Fields` and `Encoder` option, since changing the encoder discards any accumulated context. This is solvable, but definitely introduces more complexity.
2. Even if we solve the first problem, I suspect that all production users will want to customize the encoder options. Timestamp formats, level enums, and envelope keys are just too deployment-specific for one-size-fits-all defaults. If users want to customize these options, using a functional option just introduces more noise.

benchmarks/zap_bench_test.go
```diff
@@ -86,7 +86,12 @@ func BenchmarkZapDisabledLevelsWithoutFields(b *testing.B) {
 
 func BenchmarkZapDisabledLevelsAccumulatedContext(b *testing.B) {
 	context := fakeFields()
-	logger := zap.NewJSON(zap.ErrorLevel, zap.Output(zap.Discard), zap.Fields(context...))
+	logger := zap.NewLogger(
```
prashantv: Optional: might be worth adding a helper to create a new JSON logger with output discarded, that takes a list of additional options on top.

prashantv:
Makes sense. lgtm

akshayjshah:
Will change `NewLogger` to `New`, wait for tests, and then land.

PR #117, opened 2016-08-01
[v1, 5/8] Remove StubTime in favor of Encoder options

akshayjshah:
Our ugly `StubTime` hack is no longer necessary now that we have the `NoTime` option for the JSON encoder. Remove it!

This is the fifth of eight PRs to finish up the large pre-v1.0.0 refactoring:
- [x] Export Encoder type
- [ ] Remove `StubTime`
- [ ] Export the Hook type
- [ ] Improve GoDoc.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -23,7 +23,6 @@ package zap
 import (
 	"fmt"
 	"os"
-	"time"
 )
 
 // For tests.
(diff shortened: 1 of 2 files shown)


prashantv:
yay! 🚢

PR #118, opened 2016-08-01
[v1, 6/8] Export the Hook type

akshayjshah:
Export the `Hook` type so users can define their own hooks.

This is the sixth of eight PRs to finish up the large pre-v1.0.0 refactoring:
- [x] Export Encoder type
- [x] Remove `StubTime`
- [ ] Export the Hook type
- [ ] Improve GoDoc.

### DIFF
--- a/hook.go
+++ b/hook.go
@@ -35,20 +35,22 @@ var (
 	_callerSkip = 3
 )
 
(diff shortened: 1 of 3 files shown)


prashantv:
lgtm, :shipit:

PR #119, opened 2016-08-01
[v1, 7/8] Remove Encoder.AddFields

akshayjshah:
Remove `AddFields` from the `Encoder` interface: it increases the number of methods each encoder must implement, and it's not actually adding much value. Users can rely on `Field.AddTo` and a for loop, and within zap we're already using a small helper function.

This is the seventh of eight PRs to finish up the large pre-v1.0.0 refactoring:
- [x] Export Encoder type
- [x] Remove `StubTime`
- [x] Export the Hook type
- [ ] Improve GoDoc.

### DIFF
--- a/encoder.go
+++ b/encoder.go
@@ -30,7 +30,6 @@ import (
 type Encoder interface {
 	KeyValue
 
-	AddFields([]Field)
 	Clone() Encoder
 	Free()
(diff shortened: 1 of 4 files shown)


prashantv:
lgtm

PR #120, opened 2016-08-02
[v1, 8/8] Improve documentation

akshayjshah:
This PR improves and updates documentation throughout the project. As part of CR feedback, it also changes error fields to be lazy.

This is the last of eight PRs to finish up the large pre-v1.0.0 refactoring:
- [x] Export Encoder type
- [x] Remove `StubTime`
- [x] Export the Hook type
- [x] Remove `Encoder.AddFields` (since we already have `Field.AddTo`)
- [ ] Improve GoDoc.


encoder.go
```diff
@@ -25,12 +25,33 @@ import (
...
 )
 
-// Encoder is a format-agnostic interface for all log field encoders. It's not
-// safe for concurrent use.
+// Encoder is a format-agnostic interface for all log entry marshalers. Since
+// log encoders don't need to support the same wide range of use cases as
+// general-purpose marshalers, it's possible to make them much faster and
+// lower-allocation.
+//
+// Encoders need not be fully safe for concurrent use, but it should be safe to
```
prashantv: Now that this is documented, it seems very strange to require some methods to work concurrently, but not others. Should we actually have 2 interfaces here that separate a mutable state compared to the immutable state that is safe for concurrent use? E.g., breaking out the `WriteEntry` into a separate interface that is safe for concurrent use, while adding fields and `Clone` are not safe for concurrent use.

field.go
```diff
@@ -96,8 +95,8 @@ func String(key string, val string) Field {
 	return Field{key: key, fieldType: stringType, str: val}
 }
 
-// Stringer constructs a Field with the given key and value. The value
-// is the result of the String method.
+// Stringer constructs a Field with the given key and the output of the value's
+// String method.
```
prashantv: Document that this is lazy and the `String` method is only called if/when the entry is written.

prashantv: Recommend the same documentation for the other lazy methods

field.go
```diff
@@ -108,9 +107,11 @@ func Time(key string, val time.Time) Field {
...
 }
 
-// Error constructs a Field that stores err.Error() under the key "error". This is
-// just a convenient shortcut for a common pattern - apart from saving a few
-// keystrokes, it's no different from using zap.String.
+// Error constructs a Field that stores err.Error() under the key "error". If
+// passed a nil error, it returns a no-op field.
+//
+// This is just a convenient shortcut for a common pattern - apart from saving a
+// few keystrokes, it's no different from a nil check and zap.String.
```
prashantv: Should we mention that this is eager?

field.go
akshayjshah: Sure. I'll also change the Field docs so that they don't suggest that laziness is the default.

encoder.go
akshayjshah: Discussed offline. Conclusions:

* it's not worth separating this into two separate interfaces.
* instead, make some assumptions about the implementation of this interface that allow for clearer docs.
* in a future PR, explore some options for simplifying this interface, which will also simplify the docs.

field.go
akshayjshah: Better to just make it lazy, I think.

prashantv: sure, in that case we probably don't want to say it's the same as `zap.String`, maybe that it's similar but this is lazy.

akshayjshah: Yep, updated.

akshayjshah:
Updated.

encoder.go
```diff
@@ -25,12 +25,24 @@ import (
...
 
-// Encoder is a format-agnostic interface for all log field encoders. It's not
-// safe for concurrent use.
+// Encoder is a format-agnostic interface for all log entry marshalers. Since
+// log encoders don't need to support the same wide range of use cases as
+// general-purpose marshalers, it's possible to make them much faster and
+// lower-allocation.
+//
+// Implementations of the KeyValue interface's methods can, of course, freely
+// modify the receiver. However, the Clone, Free, and WriteEntry methods will be
```
prashantv: `Free` can't be called concurrently

prashantv:
lgtm with minor comment

PR #121, opened 2016-08-02
Add a DiscardOutput option

akshayjshah:
In tests, we frequently use `zap.Output(zap.Discard)`. This is noisy and
a bit annoying; supply a canned `zap.DiscardOutput` option instead.

### DIFF
--- a/benchmarks/stdlib_bench_test.go
+++ b/benchmarks/stdlib_bench_test.go
@@ -40,10 +40,12 @@ func BenchmarkStandardLibraryWithoutFields(b *testing.B) {
 }
 
 func BenchmarkZapStandardizeWithoutFields(b *testing.B) {
-	logger, err := zwrap.Standardize(zap.New(
-		zap.NewJSONEncoder(),
-		zap.DebugLevel,
-		zap.Output(zap.Discard)),
+	logger, err := zwrap.Standardize(
+		zap.New(
+			zap.NewJSONEncoder(),
+			zap.DebugLevel,
+			zap.DiscardOutput,
+		),
 		zap.InfoLevel,
 	)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


writer.go
```diff
@@ -26,8 +26,14 @@ import (
...
 )
 
-// Discard is a convenience wrapper around ioutil.Discard.
-var Discard = AddSync(ioutil.Discard)
+var (
+	// Discard is a convenience wrapper around ioutil.Discard.
+	Discard = AddSync(ioutil.Discard)
+	// DiscardOutput is an Option that discards logger output.
+	DiscardOutput = Output(Discard)
+	// DiscardErrorOutput is an Option that discards logger error output.
```
prashantv: I think we should hold off on this till we see a lot of people asking for it, or we notice it in our own tests. `DiscardOutput` makes sense since we're already using it a lot in our tests.

benchmarks/stdlib_bench_test.go
```diff
@@ -40,10 +40,12 @@ func BenchmarkStandardLibraryWithoutFields(b *testing.B) {
 }
 
 func BenchmarkZapStandardizeWithoutFields(b *testing.B) {
-	logger, err := zwrap.Standardize(zap.New(
-		zap.NewJSONEncoder(),
-		zap.DebugLevel,
-		zap.Output(zap.Discard)),
+	logger, err := zwrap.Standardize(
+		zap.New(
+			zap.NewJSONEncoder(),
```
prashantv: (as a separate change) We've duplicated these lines a lot in our benchmarks, and I think it's just adding noise to the tests, we should probably dedupe this with a helper.

akshayjshah: :+1: sounds good.

writer.go
akshayjshah: Done.

prashantv:
lgtm

PR #122, opened 2016-08-02
Add missing license to one file

akshayjshah:
One file snuck in without our license header - fix it!


blampe:
🚢 

Do we need commit hooks for this?

akshayjshah:
Yep, definitely.

PR #116, opened 2016-08-01
Add support for uint and uint64

creiht:
Fixes #110 

Before opening your pull request, please make sure that you've:

- [x] [signed Uber's Contributor License Agreement](https://docs.google.com/a/uber.com/forms/d/1pAwS_-dA1KhPlfxzYLBqK6rsSWwRwH95OCCZrcsY5rk/viewform);
- [x] added tests to cover your changes;
- [x] run the test suite locally (`make test`); and finally,
- [x] run the linters locally (`make lint`).

Thanks for your contribution!

PR #123, opened 2016-08-02
Add a human-optimized text encoder

### LINKED ISSUE #59: Add a TextEncoder for prettier console output
Add a TextEncoder for pretty, non-JSON console output. This is definitely a nice-to-have for the wider community, and it forces us to keep loggers and encoders strongly separate.

akshayjshah:
Add a text encoder that's friendlier for human consumption. By default, the `TextEncoder` output looks like this:

```
[I] 1970-01-01T00:00:00Z Something happened. foo=bar baz=bing
```

The timestamp format is configurable via functional options, and the encoder includes support for unsigned ints (in advance of #116).

Resolves #59.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -151,10 +151,10 @@ func ExampleNew_textEncoderNestedObject() {
 	textLogger := zap.New(zap.NewTextEncoder(
 		zap.TextNoTime(), // drop timestamps in tests.
 	))
-	textLogger.Info("With Marshaler", zap.String("fake", "{"), zap.Marshaler("m", zap.LogMarshalerFunc(m)))
+	textLogger.Info("With Marshaler", zap.Marshaler("m", zap.LogMarshalerFunc(m)))
 
 	// Output:
-	// [I] With Marshaler fake={ m={loggable=yes number=1}
+	// [I] With Marshaler m={loggable=yes number=1}
 }
 
 func ExampleNew_options() {
--- a/example_test.go
+++ b/example_test.go
@@ -105,9 +105,9 @@ func ExampleNest() {
 func ExampleNew() {
 	// The default logger outputs to standard out and only writes logs that are
 	// Info level or higher.
-	logger := zap.New(
-		zap.NewJSONEncoder(zap.NoTime()), // drop timestamps in tests
-	)
+	logger := zap.New(zap.NewJSONEncoder(
+		zap.NoTime(), // drop timestamps in tests
+	))
 
 	// The default logger does not print Debug logs.
 	logger.Debug("This won't be printed.")
@@ -115,6 +115,46 @@ func ExampleNew() {
 
 	// Output:
 	// {"level":"info","msg":"This is an info log."}
+}
+
+func ExampleNew_textEncoder() {
+	// For more human-readable output in the console, use a TextEncoder.
+	textLogger := zap.New(zap.NewTextEncoder(
+		zap.TextNoTime(), // drop timestamps in tests.
+	))
+	textLogger.Info("This is a text log.")
+
+	// Output:
+	// [I] This is a text log.
+}
+
+func ExampleNew_textEncoderSingleCurlyBrace() {
+	textLogger := zap.New(zap.NewTextEncoder(
+		zap.TextNoTime(), // drop timestamps in tests.
+	))
+	textLogger.Info("Test 1.")
+	textLogger.Info("With fields", zap.String("f1", "v1"), zap.String("f2", "v2"))
+	textLogger.Info("With fields", zap.String("f1", "{"), zap.String("f2", "v2"))
+
+	// Output:
+	// [I] Test 1.
+	// [I] With fields f1=v1 f2=v2
+	// [I] With fields f1={ f2=v2
+}
+
+func ExampleNew_textEncoderNestedObject() {
+	m := func(kv zap.KeyValue) error {
+		kv.AddString("loggable", "yes")
+		kv.AddInt("number", 1)
+		return nil
+	}
+	textLogger := zap.New(zap.NewTextEncoder(
+		zap.TextNoTime(), // drop timestamps in tests.
+	))
+	textLogger.Info("With Marshaler", zap.String("fake", "{"), zap.Marshaler("m", zap.LogMarshalerFunc(m)))
+
+	// Output:
+	// [I] With Marshaler fake={ m={loggable=yes number=1}
 }
 
 func ExampleNew_options() {
@@ -198,4 +238,15 @@ func ExampleNewJSONEncoder() {
 		zap.MessageKey("@message"),         // customize the message key
 		zap.LevelString("@level"),          // stringify the log level
 	)
+}
+
+func ExampleNewTextEncoder() {
+	// A text encoder with the default settings.
+	zap.NewTextEncoder()
+
+	// Dropping timestamps is often useful in tests.
+	zap.NewTextEncoder(zap.TextNoTime())
+
+	// If you don't like the default timestamp formatting, choose another.
+	zap.NewTextEncoder(zap.TextTimeFormat(time.RFC822))
 }
--- a/example_test.go
+++ b/example_test.go
@@ -122,39 +122,11 @@ func ExampleNew_textEncoder() {
 	textLogger := zap.New(zap.NewTextEncoder(
 		zap.TextNoTime(), // drop timestamps in tests.
 	))
-	textLogger.Info("This is a text log.")
+
+	textLogger.Info("This is a text log.", zap.Int("foo", 42))
 
 	// Output:
-	// [I] This is a text log.
-}
-
-func ExampleNew_textEncoderSingleCurlyBrace() {
-	textLogger := zap.New(zap.NewTextEncoder(
-		zap.TextNoTime(), // drop timestamps in tests.
-	))
-	textLogger.Info("Test 1.")
-	textLogger.Info("With fields", zap.String("f1", "v1"), zap.String("f2", "v2"))
-	textLogger.Info("With fields", zap.String("f1", "{"), zap.String("f2", "v2"))
-
-	// Output:
-	// [I] Test 1.
-	// [I] With fields f1=v1 f2=v2
-	// [I] With fields f1={ f2=v2
-}
-
-func ExampleNew_textEncoderNestedObject() {
-	m := func(kv zap.KeyValue) error {
-		kv.AddString("loggable", "yes")
-		kv.AddInt("number", 1)
-		return nil
-	}
-	textLogger := zap.New(zap.NewTextEncoder(
-		zap.TextNoTime(), // drop timestamps in tests.
-	))
-	textLogger.Info("With Marshaler", zap.Marshaler("m", zap.LogMarshalerFunc(m)))
-
-	// Output:
-	// [I] With Marshaler m={loggable=yes number=1}
+	// [I] This is a text log. foo=42
 }
 
 func ExampleNew_options() {
--- a/text_encoder.go
+++ b/text_encoder.go
@@ -1,0 +1,230 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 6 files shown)


text_encoder.go
```diff
@@ -0,0 +1,226 @@
...
+	final.bytes = append(final.bytes, ' ')
+	final.bytes = append(final.bytes, t.Format(enc.timeFmt)...)
+}
+
+func (enc *textEncoder) addMessage(final *textEncoder, msg string) {
+	final.bytes = append(final.bytes, ' ')
+	final.bytes = append(final.bytes, msg...)
+}
+
+// A TextOption is used to set options for a text encoder.
```
prashantv: Rather than add separate options for `NoTime` and `TimeFormat` etc, I would prefer to:
 - change the `JSONOption` interface method to be something like `applyJSON` and have the `TextOption` interface use `applyText`. This will allow a single formatter to apply to both.
 - Make `TextFormatter` implement` `TextOption` as well.
 - Now the same `NoTime` and `TimeFormatter` types can be used as both text options and JSON options

This reduces the interface and allows the user to use the same options for both encoders.

example_test.go
```diff
@@ -102,16 +102,23 @@ func ExampleNest() {
...
+	logger := zap.New(zap.NewJSONEncoder(
+		zap.NoTime(), // drop timestamps in tests
+	))
 
 	// The default logger does not print Debug logs.
 	logger.Debug("This won't be printed.")
 	logger.Info("This is an info log.")
 
+	// For more human-readable output in the console, use a TextEncoder.
+	textLogger := zap.New(zap.NewTextEncoder(
```
prashantv: should we have this as a separate example so that the output lines only have a single encoding.

text_encoder.go
```diff
@@ -0,0 +1,226 @@
...
+	return nil
+}
+
+func (enc *textEncoder) truncate() {
+	enc.bytes = enc.bytes[:0]
+}
+
+func (enc *textEncoder) addKey(key string) {
+	lastIdx := len(enc.bytes) - 1
+	if lastIdx >= 0 && enc.bytes[lastIdx] != '{' {
```
prashantv: This seems hacky and incorrect.

E.g., if I do `addString("key", "{")` then the value in `bytes` is `key={` and this will check the last byte, see that it's `{`, and drop the space.

Use a separate bool to track state, and suggest adding failing tests as well.

example_test.go
akshayjshah: Sure, can do.

text_encoder.go
akshayjshah: Sure, I can do that. It makes the JSON options more complex, but it's eminently doable.

text_encoder.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	final.bytes = append(final.bytes, ']')
+}
+
+func (enc *textEncoder) addTime(final *textEncoder, t time.Time) {
+	if enc.timeFmt == "" {
+		return
+	}
+	final.bytes = append(final.bytes, ' ')
+	final.bytes = append(final.bytes, t.Format(enc.timeFmt)...)
```
noxiouz: time.AppendFormat could save some allocations here
https://golang.org/pkg/time/#Time.AppendFormat

akshayjshah: Great point - I didn't even know that function existed.

sectioneight:
This seems useful. Is this still being worked on? Would you like help shepherding it over the finish line?

akshayjshah:
@sectioneight Very much so - I'm out on paternity leave tomorrow, and I'm not sure when I'm going to get back to this.

sectioneight:
Cool, will do. @akshayjshah do you have any local/unpushed changes before I take over this branch?

akshayjshah:
Nope, this is it - all yours.

text_encoder.go
sectioneight: So perhaps I'm missing something, but in a text encoder why do we care about `{` or `}` specially at all? The output isn't supposed to be valid JSON

akshayjshah: Only because I chose `{` and `}` as the delimiters for nested objects.

text_encoder.go
akshayjshah: To document an offline conversation we had quite a while ago:

Consolidating these is possible, but I think it'll make them appreciably more complex. For example, the JSON encoder supports an option that sets the key for the entry's message (e.g., `{"msg": "foo"}` or `{"the_message": "foo"}`). That's not a meaningful concept for the text encoder.

I don't mind whichever direction @sectioneight wants to go with this.

sectioneight: I'll choose the path of least resistance here :)

text_encoder.go
sectioneight: Sorry, I'm a bit tired -- @prashantv could you provide me with the failing test that this code doesn't cover? I'm still trying to wrap my head around what all this logic is for

sectioneight:
One clarifying question open

text_encoder.go
prashantv: Sure,
```go
func ExampleNew_textEncoderBug() {
	textLogger := New(NewTextEncoder(
		TextNoTime(), // drop timestamps in tests.
	))
	textLogger.Info("Test 1.")
	textLogger.Info("With fields", String("f1", "v1"), String("f2", "v2"))
	textLogger.Info("With fields", String("f1", "{"), String("f2", "v2"))

	// Output:
	// [I] Test 1.
	// [I] With Fields f1=v1 f2=v2
	// [I] With Fields f1={ f2=v2
}
```

This currently fails:
```
got:
[I] Test 1.
[I] With fields f1=v1 f2=v2
[I] With fields f1={f2=v2
want:
[I] Test 1.
[I] With Fields f1=v1 f2=v2
[I] With Fields f1={ f2=v2
```

Basically the `addKey` is supposed to add ` key=` (space at the beginning) for keys added to the top level object.

However, since we also support adding nested objects, if the user adds something using a log marshaler, we want the output to look something like:
```
[I] Log message nested={a=1 b=2}
```

To get that to work, the code currently looks at the last character, and if it's `{`, it drops the space in front. However, this assumes that the user's values for other fields cannot end in a `{`, which I'd prefer not to assume.

To fix this properly, it needs the encoder to keep some state for when it's inside of an object (and we're at the first field) to avoid the space.

text_encoder.go
prashantv: Happy to figure out how best to clean up the API in a separate PR.

I'm leaning towards separating time formatters from knowing about keys, There can be separate options for formatting time, and for customizing the time key. That way the options that customize keys (like `MessageKey` and `TimestampKey`) only implement `JSONOption` but the time formatting options `NoTime`, `EpochFormatter`, `RFC3339Formatter` can apply to both JSON encoders and text encoders.

text_encoder.go
sectioneight: Makes perfect sense. Thanks @prashantv!

sectioneight:
Ok @prashantv -- I took a pass at implementing the state. I'm not sure if the semantics are correct yet -- is the `AddMarshaler` the only place we need to care?

text_encoder.go
```diff
@@ -0,0 +1,232 @@
...
+	return &textEncoder{
+		bytes:  make([]byte, 0, _initialBufSize),
+		nested: atomic.NewBool(false),
+	}
+}}
+
+type textEncoder struct {
+	bytes   []byte
+	timeFmt string
+	nested  *atomic.Bool
```
sectioneight: I went with an atomic Bool here -- I'm pretty sure these encoders are supposed to be goroutine safe but I don't see any sync primitives on modifying e.g. `enc.bytes` in the various methods.

Is locking taken care of by the logging and should the encoder not have to worry about it?

example_test.go
```diff
@@ -117,6 +117,46 @@ func ExampleNew() {
...
+		kv.AddInt("number", 1)
+		return nil
+	}
+	textLogger := zap.New(zap.NewTextEncoder(
+		zap.TextNoTime(), // drop timestamps in tests.
+	))
+	textLogger.Info("With Marshaler", zap.Marshaler("m", zap.LogMarshalerFunc(m)))
+
+	// Output:
+	// [I] With Marshaler m={loggable=yes number=1}
```
sectioneight: This sort of feels like it's less useful as an example but as a unit test. However, it doesn't look like the unit test framework is set up to do "end-to-end" (e.g. assert the actual output) and I wanted to show that I handles this case properly.

Is there a better place to put this?

text_encoder.go
prashantv: The encoder doesn't need to be safe for concurrent use when adding fields.

The only slightly confusing piece is that `Clone` and `WriteEntry` may be called concurrently, as described here:
https://godoc.org/github.com/uber-go/zap#Encoder

Since this will only be modified when adding fields, no need for atomic here.

text_encoder.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func (enc *textEncoder) AddString(key, val string) {
+	enc.addKey(key)
+	enc.bytes = append(enc.bytes, val...)
+}
+
+func (enc *textEncoder) AddBool(key string, val bool) {
+	enc.addKey(key)
+	if val {
```
prashantv: Can simplify to just use `AppendBool` here (same as #148)

text_encoder.go
```diff
@@ -0,0 +1,230 @@
...
+	enc.AddUint64(key, uint64(val))
+}
+
+func (enc *textEncoder) AddUint64(key string, val uint64) {
+	enc.addKey(key)
+	enc.bytes = strconv.AppendUint(enc.bytes, val, 10)
+}
+
+func (enc *textEncoder) AddFloat64(key string, val float64) {
+	enc.addKey(key)
```
prashantv: we can simplify to just use `strconv.AppendFloat` here. In the JSON encoder, the output for nan/infinity is quoted, so a normal would get: `"num": 1` but nan would get `"num": "+Inf"`

For the text encoder, we don't need to quote the inf/nan, and `AppendFloat` already handles these correctly,
https://play.golang.org/p/W3FP95QIMc

text_encoder.go
```diff
@@ -0,0 +1,230 @@
...
+		enc.bytes = strconv.AppendFloat(enc.bytes, val, 'f', -1, 64)
+	}
+}
+
+func (enc *textEncoder) AddMarshaler(key string, obj LogMarshaler) error {
+	enc.addKey(key)
+	enc.firstNested = true
+	enc.bytes = append(enc.bytes, '{')
+	err := obj.MarshalLog(enc)
+	enc.bytes = append(enc.bytes, '}')
```
prashantv: We should reset `firstNested` to `false` here. I think a failing test case for this would:
 - add an object where the `MarhsalLog` adds no fields
 - then add a normal field (like string)

So we'd want something like `emptyObj={} something=val` but we might get `emptyObj={}something=val`

text_encoder.go
sectioneight: I removed the atomics after I realized encoders are only used once. Not sure why GH is still showing this...

example_test.go
sectioneight: Moved to `text_logger_test.go`

text_encoder.go
sectioneight: Done

text_encoder.go
sectioneight: Done

text_encoder.go
sectioneight: Good catch! Fixed and added negative test

sectioneight:
Ok @prashantv, I think I've addressed all comments (and nice catch on the nested error!)

sectioneight:
Still getting used to the new github review flow. Are there more comments to be addressed?

example_test.go
```diff
@@ -117,6 +117,46 @@ func ExampleNew() {
...
+	textLogger := zap.New(zap.NewTextEncoder(
+		zap.TextNoTime(), // drop timestamps in tests.
+	))
+	textLogger.Info("This is a text log.")
+
+	// Output:
+	// [I] This is a text log.
+}
+
+func ExampleNew_textEncoderSingleCurlyBrace() {
```
prashantv: Can we remove this example now that the integration test covers it

example_test.go
prashantv: Still see it here -- has it been removed from `example_test.go`?

sectioneight:
@prashantv updated by moving out of example_test -- does your approval mean a self-merge is now appropriate? Or what's your team's process on GitHub?

example_test.go
sectioneight: Yes, it's removed. Ugh, I still don't understand the new GH UI

example_test.go
sectioneight:
```bash
$ grep Marshaler example_test.go
// no results
```

prashantv:
Yep, for contributors who are part of the organization or project, typically approval means self-merge. For external contributors, the reviewer tends to merge.

Since you are in the org, feel free to self-merge! Thanks for wrapping this up @sectioneight

text_encoder.go
sectioneight: @prashantv can you open an issue describing the desired changes to the APIs so we make sure to follow up in the future? I'm one of those weird people that actually likes following up on loose ends

prashantv:
Forgot to mention, suggest "Squash and merge" -- and feel free to edit the title/body of the squashed commit, rather than leaving the default list of all commits.

text_encoder.go
prashantv: Done,
https://github.com/uber-go/zap/issues/149

sectioneight: u r awsm

PR #124, opened 2016-08-02
Add a simple lint check for license headers

akshayjshah:
This simple linter just checks for the copyright statement, not the full
license, but it should be sufficient to catch unintentional oversights.

### DIFF
--- a/check_license.sh
+++ b/check_license.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+text=`cat LICENSE.txt | head -1`
+
+ERROR_COUNT=0
+while read file
+do
+    head -1 ${file} | grep -q "${text}"
+    if [ $? -ne 0 ]; then
+        echo "$file is missing license header."
+        ERROR_COUNT+=1
+    fi
+done < <(git grep -l "" | grep "\.go")
+
+exit $ERROR_COUNT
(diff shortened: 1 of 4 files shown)


blampe:
Better than nothing 👍

check_license.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+text=`cat LICENSE.txt | head -1`
```
prashantv: no need for cat, `head -1 LICENSE.txt`

check_license.sh
```diff
@@ -0,0 +1,15 @@
...
+
+ERROR_COUNT=0
+while read file
+do
+    head -1 ${file} | grep -q "${text}"
+    if [ $? -ne 0 ]; then
+        echo "$file is missing license header."
+        ERROR_COUNT+=1
+    fi
+done < <(git grep -l "" | grep "\.go")
```
prashantv: can probably use `git ls-files '*.go'`

prashantv:
lgtm

check_license.sh
akshayjshah: I wasn't able to find an option to make `git ls-files` recurse through subdirectories. Do you know of one?

check_license.sh
akshayjshah: :+1:

PR #126, opened 2016-08-03
Fix Check not triggering panic/exit on Panic/Fatal

### LINKED ISSUE #125: Panic and Fatal do not currently panic or exit when using Check and the level is disabled
If the logger level is set to ignore all panic/fatal logs (e..g, by using `FatalLevel+1`), then there is an inconsistency in how we handle the logs.

If we do `logger.Panic(...)`, then the log is skipped, but the process crashes.
If we use the `Check` method and pass in `PanicLevel` or `FatalLevel`, then we return a not-OK checked message and we don't actually panic or fatal.

Normally, check will cause a panic on the `Write` if the level is enabled.

I've written some tests to reproduce this behaviour:
https://github.com/uber-go/zap/compare/pv_panic_always?expand=1

I think the best option is for `Check` to always return `true` for `PanicLevel` and `FatalLevel`.

prashantv:
Panic and Fatal should always cause a panic or exit, even if the level
is disabled. This is handled correctly in logger.Panic/Fatal/Log, but
when the CheckedMessage is used, we're not calling panic/exit.

We should always return a valid message from Check when the level is
Fatal or Panic. This means that Panic/Fatal will not be sampled, since
the sampler should not call panic/exit directly, but should defer to the
underlying logger, and there's no way to specify don't log, but call the
panic/exit functions.

Fixes #125

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -89,8 +89,15 @@ func (log *logger) With(fields ...Field) Logger {
 }
 
 func (log *logger) Check(lvl Level, msg string) *CheckedMessage {
-	if !(lvl >= log.Level()) {
-		return nil
+	switch lvl {
+	case PanicLevel, FatalLevel:
+		// Panic and Fatal should always cause a panic/exit, even if the level
+		// is disabled.
+		break
+	default:
+		if lvl < log.Level() {
+			return nil
+		}
 	}
 	return NewCheckedMessage(log, lvl, msg)
 }
--- a/logger_test.go
+++ b/logger_test.go
@@ -193,6 +193,35 @@ func TestJSONLoggerPanic(t *testing.T) {
 	})
 }
 
+func TestJSONLoggerCheckPanic(t *testing.T) {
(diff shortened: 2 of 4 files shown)


logger.go
```diff
@@ -89,8 +89,15 @@ func (log *logger) With(fields ...Field) Logger {
...
 func (log *logger) Check(lvl Level, msg string) *CheckedMessage {
-	if !(lvl >= log.Level()) {
-		return nil
+	switch lvl {
+	case PanicLevel, FatalLevel:
+		// Panic and Fatal should always cause a panic/exit, even if the level
+		// is disabled.
+		break
+	default:
+		if lvl < log.Level() {
```
prashantv: This is just some minor code cleanup, it does not change the behaviour of the `if`.

I think it's easier to read `if a < b` vs `if !(a >= b)`. The same change is also made in zwrap/sample.go.

akshayjshah:
:shipit:

PR #128, opened 2016-08-09
Distinguish between tip and go1.7

billf:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 go:
   - 1.5
   - 1.6
+  - 1.7rc6
   - tip
 env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,6 +3,7 @@ sudo: false
 go:
   - 1.5
   - 1.6
+  - 1.7rc3
```
prashantv: I think they just released 1.7 RC6 today, can we bump this to RC6?

prashantv:
lgtm, just bump to rc6.

billf:
yeah, realized that as soon as i sent it that i didn't check for the latest.

PR #129, opened 2016-08-09
Install test dependencies before 'go vet'

billf:


prashantv:
Thanks for fixing this, can you also reorder the `make test` and `make lint` in the `.travis.yml`? That way, we'll have the lint failures fail early, and since we build the test dependencies, the `go test` command will run faster.

PR #133, opened 2016-08-18
Ultra-fast, eminently lossy NullEncoder

billf:


null_encoder.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"io"
+	"time"
+)
+
+// nullEncoder is an Encoder implementation that throws everything away.
+type nullEncoder struct{}
+
+// NewNullEncoder creates a fast, no-allocation encoder.
```
prashantv: Since there's no options or any customizations for this encoder, does it make sense to just have a single global instance and remove the `NewNullEncoder` constructor?
```go
var NullEncoder Encoder

func init() {
  NullEncoder = nullEncoder{}
}
```

(I like to separate the declaration from the assignment so that the ` = nullEncoder{}` doesn't show up in the godoc)

null_encoder.go
```diff
@@ -0,0 +1,90 @@
...
+func NewNullEncoder() Encoder {
+	return nullEncoder{}
+}
+
+func (nullEncoder) Free() {
+}
+
+// AddString throws away a string key and value.
```
prashantv: I love all these comments, but I don't think they're actually necessary since this is a private type implementing an interface which has godoc already :) You should be able to compress this down to:
```go
func (nullEncoder) AddString(_, _ string)    {}
func (nullEncoder) AddBool(_ string, _ bool) {}
func (nullEncoder) AddInt(_ string, _ int)   {}
[...]
```

(in fact, gofmt will align the `{}` for you to make the compressed version look nice)

null_encoder.go
```diff
@@ -0,0 +1,90 @@
...
+func (nullEncoder) AddUint64(_ string, _ uint64) {
+}
+
+// AddFloat64 throws away a string key and float64 value.
+func (nullEncoder) AddFloat64(_ string, _ float64) {
+}
+
+// AddMarshaler applies a LogMarshaler to nothingness.
+func (enc nullEncoder) AddMarshaler(_ string, obj LogMarshaler) error {
+	// is it part of the contract that we call this?
```
prashantv: Good question, I don't think so. The user's interaction will typically be through `zap.Marshaler([...])` which is lazy, it calls the marshal method if the log message needs to be encoded. However if the log level is disabled or has been sampled away, then the Marshal method is not called.

I think it should be similar for `AddMarshaler` -- the implementation can choose not to call it.

prashantv:
My new favourite encoder (after all, it's the fastest, and has the least code).

Can we add unit tests to avoid the coverage drop (I can't imagine the tests being too exciting unfortunately, but at least we can make sure that nothing was written to the writer no matter which method is called).

I'm also curious what the benchmarks are! This will give us numbers on just the function call / arg passing overhead compared to the actual work we do in our real encoders.

null_encoder.go
akshayjshah: I like that approach too. One thing to consider is that it'll make the GoDoc output a little less nice - with a constructor function, all the different encoder constructors are grouped under the Encoder type. Top-level vars are in a completely different section.

null_encoder.go
akshayjshah: :+1:

null_encoder.go
prashantv: Good point, maybe just `func NullEncoder() Encoder` then? It'll still be backed by a global by it will show under the `Encoder` type?

billf:
```
BenchmarkStandardJSON-8                	 1000000       	      2647 ns/op
BenchmarkZapJSON-8                     	10000000       	       254 ns/op
BenchmarkZapNull-8                     	100000000      	        14.5 ns/op
```
```
BenchmarkJSONLogMarshalerFunc-8        	10000000       	       122 ns/op
BenchmarkNullLogMarshalerFunc-8        	30000000       	        43.2 ns/op
```

null_encoder.go
```diff
@@ -0,0 +1,63 @@
...
+func (nullEncoder) AddInt(_ string, _ int)         {}
+func (nullEncoder) AddInt64(_ string, _ int64)     {}
+func (nullEncoder) AddUint(_ string, _ uint)       {}
+func (nullEncoder) AddUint64(_ string, _ uint64)   {}
+func (nullEncoder) AddFloat64(_ string, _ float64) {}
+
+func (nullEncoder) AddMarshaler(_ string, _ LogMarshaler) error { return nil }
+func (nullEncoder) AddObject(_ string, _ interface{}) error     { return nil }
+
+// Clone copies the current encoder, including any data already encoded.
```
prashantv: can drop comment here too, since it doesn't actually copy the encoded data etc

prashantv:
lgtm, thanks for adding the benchmarks. `BenchmarkNullLogMarshalerFunc` is more expensive than I expected, it looks like all it does is:
 - get a reference to an existing global null encoder
 - call 2 functions which do nothing, but through interfaces (so there's some vtable overhead)

However, since we're using `LogMarshalerFunc`, it allocates a closure to capture the reference to `i`. That allocation ends up taking most of the time.

billf:
should we pull `LogMarshalerFunc` out of both the json and null encoder and instead use a struct implementing `LogMarshaler` allocated outside of the benchmark loop?

since `LogMarshalerFunc`'s implementation cannot ever optimized further, i don't see a reason to benchmark it after such a removal.

prashantv:
Let's merge this as is, and we can move out the `LogMarshalerFunc` benchmark separately.

PR #135, opened 2016-08-19
Enable lint on go1.7

billf:


Makefile
```diff
@@ -33,6 +33,8 @@ else
 	@echo "Not installing golint, since we don't expect to lint on" $(GO_VERSION)
 endif
 
+VET_SKIP_ERRORF=grep -v -e "possible formatting directive in Error call"
```
akshayjshah: Is this a pervasive issue in zap, or is there one particular place where vet complains?

billf: calls to testify's `assert.Error`, which i guess i can just rewrite.

prashantv:
We should keep the checks as `assert.Error` since it reports better error messages if the tests fail.

It seems to only happen if the vendor directory is available:
```bash
$ go tool vet flag_test.go
flag_test.go:61: possible formatting directive in Error call
$ mv vendor vendor_

$ go tool vet flag_test.go

$ mv vendor_ vendor

$ go tool vet flag_test.go
flag_test.go:61: possible formatting directive in Error call
```

It looks like if the type is available, `vet` does extra checking:
https://golang.org/src/cmd/vet/print.go#L589

prashantv:
Should we disable the printf checks when running vet for now? `go tool vet -printf=false [...]`

prashantv:
lgtm, can you add a comment to why we disabled printf checks.

PR #145, opened 2016-09-18
Add support for uintptr and unsafe.Pointer

maverickwoo:

Thanks for your contribution!

### DIFF
--- a/json_encoder.go
+++ b/json_encoder.go
@@ -141,6 +141,10 @@ func (enc *jsonEncoder) AddUint64(key string, val uint64) {
 	enc.bytes = strconv.AppendUint(enc.bytes, val, 10)
 }
 
+func (enc *jsonEncoder) AddUintptr(key string, val uintptr) {
+	enc.AddUint64(key, uint64(val))
+}
+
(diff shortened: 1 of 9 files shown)


json_encoder.go
```diff
@@ -141,6 +141,12 @@ func (enc *jsonEncoder) AddUint64(key string, val uint64) {
 	enc.bytes = strconv.AppendUint(enc.bytes, val, 10)
 }
 
+func (enc *jsonEncoder) AddUintptr(key string, val uintptr) {
+	enc.addKey(key)
+	enc.bytes = append(enc.bytes, '0', 'x')
+	enc.bytes = strconv.AppendUint(enc.bytes, uint64(val), 16)
```
akshayjshah: Unfortunately, this isn't valid JSON - the [section of the spec on Numbers](https://tools.ietf.org/html/rfc7159#section-6) specifically requires that numbers be represented in base 10. Instead of this, let's just delegate to `enc.AddInt64`.

maverickwoo: Thank you for pointing out the JSON spec. :) I have also removed support for `unsafe.Pointer` and squashed the commits.

PR #146, opened 2016-09-21
Validate JSON snippets are not only expected, but valid.

billf:


field_test.go
```diff
@@ -46,9 +47,19 @@ func assertFieldJSON(t testing.TB, expected string, field Field) {
 	enc := newJSONEncoder()
 	defer enc.Free()
 
+	var out interface{}
+	err := json.Unmarshal([]byte("{"+expected+"}"), &out)
+	require.NoError(t, err,
+		"Expected JSON snippet '%s' must be valid for use in an object.", expected)
```
prashantv: nit: use `%q` rather than `'%s'`

field_test.go
```diff
@@ -46,9 +47,19 @@ func assertFieldJSON(t testing.TB, expected string, field Field) {
...
+
 	field.AddTo(enc)
 	assert.Equal(t, expected, string(enc.bytes),
 		"Unexpected JSON output after applying field %+v.", field)
+	eb := append([]byte{'{'}, enc.bytes...)
```
prashantv: once we validate the expected unmarshals with `{` and `}` around it, and we make sure that `expected` is the same as `enc.bytes`, is there any point in trying to unmarshal the `enc.bytes` separately?

billf: none that i can think of. 

i did this section first and then thought "i should also validate that what's passed in as expected is valid as well"

prashantv:
lgtm, thanks @billf

PR #147, opened 2016-09-22
Add a slightly slower, but less verbose, logger

akabos:
```
Benchmark10FieldsSugar-8     1000000          1365 ns/op        2554 B/op         26 allocs/op
Benchmark10Fields-8          2000000           597 ns/op         642 B/op          1 allocs/op
```

This sugar implementation is twice as slow if compared to core, but still twice as fast if compared to nearest competition.

### DIFF
--- a/benchmarks/zap_sugar_bench_test.go
+++ b/benchmarks/zap_sugar_bench_test.go
@@ -43,8 +43,22 @@ func fakeSugarFields() []interface{} {
 	}
 }
 
+var newCoreLoggerDefaults = []zap.Option{
+	zap.ErrorLevel,
+	zap.DiscardOutput,
+}
+
+func newCoreLogger(options ...zap.Option) zap.Logger {
+	options = append(newCoreLoggerDefaults, options...)
+	return zap.New(zap.NewJSONEncoder(), options...)
+}
+
+func newSugarLogger(options ...zap.Option) zap.Sugar {
+	return zap.NewSugar(newCoreLogger(options...))
+}
+
 func BenchmarkZapSugarDisabledLevelsWithoutFields(b *testing.B) {
-	logger := zap.NewSugar(zap.New(zap.NewJSONEncoder(), zap.ErrorLevel, zap.DiscardOutput))
+	logger := newSugarLogger()
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
@@ -55,12 +69,7 @@ func BenchmarkZapSugarDisabledLevelsWithoutFields(b *testing.B) {
 
 func BenchmarkZapSugarDisabledLevelsAccumulatedContext(b *testing.B) {
 	context := fakeFields()
-	logger := zap.NewSugar(zap.New(
-		zap.NewJSONEncoder(),
-		zap.ErrorLevel,
-		zap.DiscardOutput,
-		zap.Fields(context...),
-	))
+	logger := newSugarLogger(zap.Fields(context...))
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
@@ -70,11 +79,7 @@ func BenchmarkZapSugarDisabledLevelsAccumulatedContext(b *testing.B) {
 }
 
 func BenchmarkZapSugarDisabledLevelsAddingFields(b *testing.B) {
-	logger := zap.NewSugar(zap.New(
-		zap.NewJSONEncoder(),
-		zap.ErrorLevel,
-		zap.DiscardOutput,
-	))
+	logger := newSugarLogger()
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
@@ -84,11 +89,7 @@ func BenchmarkZapSugarDisabledLevelsAddingFields(b *testing.B) {
 }
 
 func BenchmarkZapSugarAddingFields(b *testing.B) {
-	logger := zap.NewSugar(zap.New(
-		zap.NewJSONEncoder(),
-		zap.DebugLevel,
-		zap.DiscardOutput,
-	))
+	logger := newSugarLogger()
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
@@ -98,11 +99,7 @@ func BenchmarkZapSugarAddingFields(b *testing.B) {
 }
 
 func BenchmarkZapSugarWithAccumulatedContext(b *testing.B) {
-	logger, _ := zap.NewSugar(zap.New(
-		zap.NewJSONEncoder(),
-		zap.DebugLevel,
-		zap.DiscardOutput,
-	)).With(fakeSugarFields()...)
+	logger, _ := newSugarLogger().With(fakeSugarFields()...)
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
@@ -112,11 +109,7 @@ func BenchmarkZapSugarWithAccumulatedContext(b *testing.B) {
 }
 
 func BenchmarkZapSugarWithoutFields(b *testing.B) {
-	logger := zap.New(
-		zap.NewJSONEncoder(),
-		zap.DebugLevel,
-		zap.DiscardOutput,
-	)
+	logger := newSugarLogger()
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
@@ -127,12 +120,8 @@ func BenchmarkZapSugarWithoutFields(b *testing.B) {
 
 func BenchmarkZapSugarSampleWithoutFields(b *testing.B) {
 	messages := fakeMessages(1000)
-	base := zap.New(
-		zap.NewJSONEncoder(),
-		zap.DebugLevel,
-		zap.DiscardOutput,
-	)
-	logger := zap.NewSugar(zwrap.Sample(base, time.Second, 10, 10000))
+	core := newCoreLogger()
+	logger := zap.NewSugar(zwrap.Sample(core, time.Second, 10, 10000))
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		i := 0
@@ -145,12 +134,8 @@ func BenchmarkZapSugarSampleWithoutFields(b *testing.B) {
 
 func BenchmarkZapSugarSampleAddingFields(b *testing.B) {
 	messages := fakeMessages(1000)
-	base := zap.New(
-		zap.NewJSONEncoder(),
-		zap.DebugLevel,
-		zap.DiscardOutput,
-	)
-	logger := zap.NewSugar(zwrap.Sample(base, time.Second, 10, 10000))
+	core := newCoreLogger()
+	logger := zap.NewSugar(zwrap.Sample(core, time.Second, 10, 10000))
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		i := 0
--- a/benchmarks/zap_sugar_bench_test.go
+++ b/benchmarks/zap_sugar_bench_test.go
@@ -99,7 +99,7 @@ func BenchmarkZapSugarAddingFields(b *testing.B) {
 }
 
 func BenchmarkZapSugarWithAccumulatedContext(b *testing.B) {
-	logger, _ := newSugarLogger().With(fakeSugarFields()...)
+	logger := newSugarLogger().With(fakeSugarFields()...)
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
--- a/benchmarks/zap_sugar_bench_test.go
+++ b/benchmarks/zap_sugar_bench_test.go
@@ -1,0 +1,162 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package benchmarks
+
+import (
+	"testing"
+	"time"
+
+	"github.com/uber-go/zap"
+	"github.com/uber-go/zap/zwrap"
+)
+
+func fakeSugarFields() []interface{} {
+	return []interface{}{
+		errExample,
+		"int", 1,
+		"int64", 2,
+		"float", 3.0,
+		"string", "four!",
+		"stringer", zap.DebugLevel,
+		"bool", true,
+		"time", time.Unix(0, 0),
+		"duration", time.Second,
+		"another string", "done!",
+	}
+}
+
+func BenchmarkZapSugarDisabledLevelsWithoutFields(b *testing.B) {
+	logger := zap.NewSugar(zap.New(zap.NewJSONEncoder(), zap.ErrorLevel, zap.DiscardOutput))
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			logger.Info("Should be discarded.")
+		}
+	})
+}
+
+func BenchmarkZapSugarDisabledLevelsAccumulatedContext(b *testing.B) {
+	context := fakeFields()
+	logger := zap.NewSugar(zap.New(
+		zap.NewJSONEncoder(),
+		zap.ErrorLevel,
+		zap.DiscardOutput,
+		zap.Fields(context...),
+	))
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			logger.Info("Should be discarded.")
+		}
+	})
+}
+
+func BenchmarkZapSugarDisabledLevelsAddingFields(b *testing.B) {
+	logger := zap.NewSugar(zap.New(
+		zap.NewJSONEncoder(),
+		zap.ErrorLevel,
+		zap.DiscardOutput,
+	))
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			logger.Info("Should be discarded.", fakeSugarFields()...)
+		}
+	})
+}
+
+func BenchmarkZapSugarAddingFields(b *testing.B) {
+	logger := zap.NewSugar(zap.New(
+		zap.NewJSONEncoder(),
+		zap.DebugLevel,
+		zap.DiscardOutput,
+	))
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			logger.Info("Go fast.", fakeSugarFields()...)
+		}
+	})
+}
+
+func BenchmarkZapSugarWithAccumulatedContext(b *testing.B) {
+	logger, _ := zap.NewSugar(zap.New(
+		zap.NewJSONEncoder(),
+		zap.DebugLevel,
+		zap.DiscardOutput,
+	)).With(fakeSugarFields()...)
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			logger.Info("Go really fast.")
+		}
+	})
+}
+
+func BenchmarkZapSugarWithoutFields(b *testing.B) {
+	logger := zap.New(
+		zap.NewJSONEncoder(),
+		zap.DebugLevel,
+		zap.DiscardOutput,
+	)
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			logger.Info("Go fast.")
+		}
+	})
+}
+
+func BenchmarkZapSugarSampleWithoutFields(b *testing.B) {
+	messages := fakeMessages(1000)
+	base := zap.New(
+		zap.NewJSONEncoder(),
+		zap.DebugLevel,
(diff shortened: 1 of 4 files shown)


sugar.go
```diff
@@ -0,0 +1,169 @@
...
+				fields = append(fields, Uint64(key, value.(uint64)))
+			case string:
+				fields = append(fields, String(key, value.(string)))
+			case time.Time:
+				fields = append(fields, Time(key, value.(time.Time)))
+			case time.Duration:
+				fields = append(fields, Duration(key, value.(time.Duration)))
+			case fmt.Stringer:
+				fields = append(fields, Stringer(key, value.(fmt.Stringer)))
+			default:
```
billf: this is missing support for the `error` type. 

in the existing zap logger, logging an `error` writes a key named `"error"`. with the field count being used here, we have to decide if we want to support:

```go
sugar.With("k1", 42, errors.New("this automatically goes into 'error'))`
```
or 
```go
sugar.With("k1", errors.New("nothing automatic"), "k2", errors.New("field is an errorType still"))
```

sugar.go
```diff
@@ -0,0 +1,169 @@
...
+				return nil, errors.New("invalid argument type")
+			}
+		}
+	}
+	return fields, nil
+}
+
+// Log ...
+func (s *sugar) Log(lvl Level, msg string, args ...interface{}) error {
+	fields, err := getSugarFields(args...)
```
billf: since this may be an expensive operation due to reflection and, in some cases,  calling receivers on arguments passed in, this should check to see if the log level is valid _before_ calling `getSugarFields`

sugar_bench_test.go
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
akabos: All your points about benchmarks are absolutely valid. But my intention for this file was to have means of comparison between core logger and sugared one. So I basically mirrored `Benchmark10Fields` from `logger_test_bench.go`.

sugar.go
akabos: I suggest sugar logger should accept optional error value as the first of variadic arguments. E.g.:

```
sugar.Warn("message", errors.New("this is a error"), "a", "b")
```

```
{"level":"warn","msg":"message","err":"this is a error","a":"b"}
```
Motivation behind my suggestion is that having more then one errors in single log message is probably pretty rare requirement which don't need to be covered by syntax sugar.

akshayjshah: I'd prefer not to do this, since the common (non-error) case will then require passing a `nil`.

With this API, callers will need to pass both the key and the value for errors: `sugar.With("error", err)`. We'll need to add a case like this:

```
case error:
    fields = append(fields, String(key, value.(error).Error()))
```

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+
+package zap
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+// Sugar is a wrapper around core logger whith less verbose API
```
akshayjshah: nit: "with", not "whith"

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+	// descendants. This makes it easy to change the log level at runtime
+	// without restarting your application.
+	SetLevel(Level)
+
+	// Create a child logger, and optionally add some context to that logger.
+	With(...interface{}) (Sugar, error)
+
+	// Log a message at the given level. Messages include any context that's
+	// accumulated on the logger, as well as any fields added at the log site.
+	Log(Level, string, ...interface{}) error
```
akshayjshah: Let's get rid of this; in an interface that's designed to be terse, I don't think it adds much.

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+		return fields, nil
+	}
+
+	switch args[0].(type) {
+	case error:
+		fields = append(fields, Error(args[0].(error)))
+		noErrArgs = args[1:]
+	default:
+		noErrArgs = args
+	}
```
akshayjshah: I'd prefer to remove this block; treating the first of the variadic arguments differently from the rest is a little too much magic for me.

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+			case time.Time:
+				fields = append(fields, Time(key, value.(time.Time)))
+			case time.Duration:
+				fields = append(fields, Duration(key, value.(time.Duration)))
+			case fmt.Stringer:
+				fields = append(fields, Stringer(key, value.(fmt.Stringer)))
+			case error:
+				return nil, errors.New("error can only be the first argument")
+			default:
+				return nil, errors.New("invalid argument type")
```
akshayjshah: In addition to supporting errors (as @billf mentioned), please make sure to support all the types that zap has fields for. I believe that this switch is missing `uintptr` and `zap.Marshaler`.

Rather than returning an error if the argument doesn't match one of the types zap explicitly supports, fall back to using `zap.Object`, which attempts to use an encoder-appropriate serialization.

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+	for ii, value = range noErrArgs {
+		if (ii % 2) == 0 {
+			switch value.(type) {
+			case string:
+				key = value.(string)
+			default:
+				return nil, errors.New("field name must be string")
+			}
+		} else {
+			switch value.(type) {
```
akshayjshah: Have you seen the `switch value := value.(type)` form of type switches? It's described in [Effective Go](https://golang.org/doc/effective_go.html#type_switch).

Using that form will reduce duplication below, since you won't have to type-cast in each case.

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return NewSugar(s.core.With(fields...)), nil
+}
+
+func getSugarFields(args ...interface{}) ([]Field, error) {
+	var (
+		noErrArgs []interface{}
+		fields    []Field
```
akshayjshah: Let's pre-allocate the right amount of space here - we should be able to do `fields := make([]Field, 0, len(args)/2)`.

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+	case error:
+		fields = append(fields, Error(args[0].(error)))
+		noErrArgs = args[1:]
+	default:
+		noErrArgs = args
+	}
+
+	if (len(noErrArgs) % 2) != 0 {
+		return nil, errors.New("invalid number of arguments")
+	}
```
akshayjshah: Instead of returning an error, let's add a field to the log output describing the issue. In that case, we should ignore the last argument and include it in the error:

```
fields = append(fields, zap.String("numLogArgsError", fmt.Sprintf("odd number of arguments passed to logger, ignoring last arg: %v", args[len(args)-1])))`
```

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+				return nil, errors.New("invalid argument type")
+			}
+		}
+	}
+	return fields, nil
+}
+
+// Log ...
+func (s *sugar) Log(lvl Level, msg string, args ...interface{}) error {
+	if cm := s.core.Check(lvl, msg); cm.OK() {
```
akshayjshah: Awesome - thanks for being careful to use the checked message API!

Since we're updating `getSugarFields` so that it won't need to return an error, let's not have this method (or any of the Info, Debug, Error, etc. method) return errors either.

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+	Log(Level, string, ...interface{}) error
+	Debug(string, ...interface{}) error
+	Info(string, ...interface{}) error
+	Warn(string, ...interface{}) error
+	Error(string, ...interface{}) error
+	Panic(string, ...interface{}) error
+	Fatal(string, ...interface{}) error
+	// If the logger is in development mode (via the Development option), DFatal
+	// logs at the Fatal level. Otherwise, it logs at the Error level.
+	DFatal(string, ...interface{}) error
```
akshayjshah: It seems like all these methods return errors because `getSugarFields` can return an error. I'd like to avoid this, since it's awkward for loggers to return errors - what's the user going to do, log the error?

I've left some comments on `getSugarFields` that should remove the returned error. Once those changes are in, let's change `With`, `Debug`, `Info`, etc. to not return errors either.

sectioneight:
any updates / timeline on this? we'd like to use it. would you like my team to take over this PR and finish addressing comments or were you planning to?

cc @uber-go/service-framework

sectioneight:
@akabos to be clear that question was directed at you -- do you plan on working on this any more, or would you like us to pick it up for you?

anuptalwalkar:
@akabos Are you continuing to work on this PR? We are running on a tight schedule and this work seems really useful and relevant for us. I am happy to take over your work, address comments and fixes for this Logger.

sugar.go
akabos: So how do you suggest treating errors then? Alternative approach is expecting an error in any position and would involve even more magic.

akabos:
@sectioneight @anuptalwalkar I'd like to finish that myself. Please give me couple of days.

anuptalwalkar:
@akabos Thank you so much! I appreciate speeding up the timeline. This is fantastic work:)

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+		return nil, errors.New("invalid number of arguments")
+	}
+
+	for ii, value = range noErrArgs {
+		if (ii % 2) == 0 {
+			switch value.(type) {
+			case string:
+				key = value.(string)
+			default:
+				return nil, errors.New("field name must be string")
```
anuptalwalkar: Just my $0.02- Loggers usually don't throw/return errors and rather gulp them down to the log files. Instead of returning error for this function, you want to just add a string field with a tiny message "unparsable argument" or something. It should be on the devs to notice an fix the logging format later in the game.

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+// Sugar is a wrapper around core logger with less verbose API
+type Sugar interface {
+	// Check the minimum enabled log level.
```
sectioneight: I think this violates golint or vet or one of the tools? The comment should start with "Level", right?

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+	Level() Level
+	// Change the level of this logger, as well as all its ancestors and
+	// descendants. This makes it easy to change the log level at runtime
+	// without restarting your application.
+	SetLevel(Level)
```
sectioneight: Same comment here

sugar.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Create a child logger, and optionally add some context to that logger.
+	With(...interface{}) (Sugar, error)
```
sectioneight: Samesies

sugar.go
akabos: fixed

sugar.go
akabos: fixed

sugar.go
akabos: fixed

sugar.go
akabos: Sorry about that. I'm still very new to Go, so your notice is really appreciated.

Fixed.

sugar.go
akabos: Added support for `zap.Marshaler` and changed fall back from error to `zap.Object`. And I couldn't find a singe mention of `uintptr` in zap codebase.

sugar.go
akabos: Fixed, please see b819160

sugar.go
```diff
@@ -112,31 +112,33 @@ func getSugarFields(args ...interface{}) ([]Field, error) {
...
+				fields = append(fields, Time(key, v))
 			case time.Duration:
-				fields = append(fields, Duration(key, value))
+				fields = append(fields, Duration(key, v))
 			case fmt.Stringer:
-				fields = append(fields, Stringer(key, value))
-			case error:
-				return nil, errors.New("error can only be the first argument")
+				fields = append(fields, Stringer(key, v))
+			case LogMarshaler:
```
akabos: @akshayjshah @anuptalwalkar @sectioneight 

I looked through [https://coveralls.io/builds/8191745/source?filename=sugar.go#L138](coverage report) and eventually found that `LogMarshaler` never get caught. `Object` field is always used instead. I'm going to update test suite to fail in such case, but what's important is that I don't really have a clue why would it happen. Can anyone please comment on the situation?

sugar.go
akabos: Well, I don't know much about how `golint` _should_ work, but apparently it doesn't have any objections about all these comments. Furthermore, they are basically copied over from `zap.Logger` interface definition, see https://github.com/uber-go/zap/blob/master/logger.go#L33

sugar_test.go
```diff
@@ -0,0 +1,226 @@
...
+	logger := New(newJSONEncoder(NoTime()), allOpts...)
+	sugar := NewSugar(logger)
+
+	f(sugar, sink)
+}
+
+type m9e struct {
+	foo int
+	bar bool
+}
```
anuptalwalkar: You need this struct to be json formattable. This will fix the failing TestSugarLogTypes test  - 

`type m9e struct {
	Foo int  `json:"foo"`
	Bar bool `json:"bar"`
}`

sugar.go
anuptalwalkar: @akabos I bet (Just guessing) the LogMarshaler didn't understand unexported struct. I commented on the test below to update m9e to json serializable struct that seemed to fix the test coverage issue you are seeing.

sugar_test.go
akabos: This will fix test but no the issue. When you define struct as json formatable `Object` field start to mimic expected `LogMarshal` output.

sugar.go
akabos: As far as I understand Go exported/unexported logic, these fields are only used by `m9e` own method and the fact they are not unexported should not affect anything.

sugar.go
```diff
@@ -0,0 +1,193 @@
...
+		if (ii % 2) == 0 {
+			switch value.(type) {
+			case string:
+				key = value.(string)
+			default:
+				return nil, errors.New("field name must be string")
+			}
+		} else {
+			switch v := value.(type) {
+			case error:
```
prashantv: I'm not sure about this special casing for error. I think it's a little surprising, complicates the code, and will possibly cause unexpected errors.

For example, if some struct happens to satisfy the error interface (E.g., Thrift exceptions), then they will never be able to use it as a value.

benchmarks/zap_sugar_bench_test.go
```diff
@@ -0,0 +1,162 @@
...
+		"stringer", zap.DebugLevel,
+		"bool", true,
+		"time", time.Unix(0, 0),
+		"duration", time.Second,
+		"another string", "done!",
+	}
+}
+
+func BenchmarkZapSugarDisabledLevelsWithoutFields(b *testing.B) {
+	logger := zap.NewSugar(zap.New(zap.NewJSONEncoder(), zap.ErrorLevel, zap.DiscardOutput))
```
prashantv: we duplicate the `zap.New(...)` quite a lot, maybe a helper could help
```go

func newSugarLogger(options [...]) zap.Sugar
```

and by defualt the options should probably just created a JSON encoder that with discard output

sugar.go
prashantv: Normally comments in Go are `Field is [...]` or `Func does blah`. In this case, I think rather than `Check the minimum enabled log level`, `Level returns the minimum enabled log level.` might be a better comment

For example, look at the comments on this interface in the standard library:
https://golang.org/pkg/net/http/#CookieJar

sugar.go
prashantv: Same as above

sugar.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (s *sugar) Level() Level {
+	return s.core.Level()
+}
+
+func (s *sugar) SetLevel(lvl Level) {
+	s.core.SetLevel(lvl)
+}
+
+func (s *sugar) With(args ...interface{}) (Sugar, error) {
```
prashantv: I'm not sure if returning errors from this method is useful -- users can't really do much. I think perhaps we should just write an error to the `ErrorOutput`?

sugar.go
```diff
@@ -0,0 +1,193 @@
...
+				fields = append(fields, Uint(key, v))
+			case uint64:
+				fields = append(fields, Uint64(key, v))
+			case string:
+				fields = append(fields, String(key, v))
+			case time.Time:
+				fields = append(fields, Time(key, v))
+			case time.Duration:
+				fields = append(fields, Duration(key, v))
+			case fmt.Stringer:
```
prashantv: `fmt.Stringer` should come after `LogMarshaler` -- if the type knows how to serialize itself, I think we should prefer that.

Please add a unit test to ensure that's the case as well.

sugar.go
```diff
@@ -0,0 +1,193 @@
...
+				fields = append(fields, Stringer(key, v))
+			case LogMarshaler:
+				fields = append(fields, Marshaler(key, v))
+			default:
+				fields = append(fields, Object(key, value))
```
prashantv: In this approach, if a type has `String`, there's no way to serialize the fields using the JSON marshaller.

sugar.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+		}
+	}
+	return fields, nil
+}
+
+// Log ...
```
prashantv: incomplete comment. You can delete comments on exported methods on private types -- since the interface already has documentation

sugar.go
```diff
@@ -0,0 +1,193 @@
...
+func (s *sugar) Log(lvl Level, msg string, args ...interface{}) {
+	var (
+		fields []Field
+		err    error
+	)
+	if cm := s.core.Check(lvl, msg); cm.OK() {
+		fields, err = getSugarFields(args...)
+		if err != nil {
+			fields = []Field{Error(err)}
```
prashantv: If the user has an error field, it would get overwritten by this error field.

I think we have to get away from returning errors when converting fields

sugar.go
```diff
@@ -0,0 +1,193 @@
...
+	s.Log(PanicLevel, msg, args...)
+}
+
+func (s *sugar) Fatal(msg string, args ...interface{}) {
+	s.Log(FatalLevel, msg, args...)
+}
+
+func (s *sugar) DFatal(msg string, args ...interface{}) {
+	lvl := ErrorLevel
+	if s.core.(*logger).Development {
```
prashantv: Please use the underlying `Logger`'s `DFatal` method. This method will panic if the passed in logger is wrapped via `Sample`, or some other wrapper which implements the `Logger` interface but is not the base logger.

Can we add a unit test that would have caught this issue?

prashantv:
Thanks for the contribution @akabos 

I think the big issues to figure out are:
 - whether we should do the special-casing for errors, and only allowing them as the first field.
 - dealing with errors when converting between the list of `interface{}` to a list of `zap.Field`

Would appreciate feedback from others as well. My personal opinion is that the special-casing of the first field of `error` is a little surprising, and that we shouldn't return errors during conversion.

sugar.go
akabos: Everything user tried to log would get overwritten with error returned from `getSugarFields`. I'm pretty sure it is valid way of handling invalid arguments.

sugar.go
akabos: I removed LogMarshaler altogether. As I mentioned in previous comments, type switch fails to catch structs implementing this interface for some reason.

sugar.go
akabos: removed

sugar_test.go
akabos: I removed LogMarshaler altogether because type switch fail to catch structs implementing this interface.

benchmarks/zap_sugar_bench_test.go
akabos: fixed

sugar.go
akabos: fixed

sugar.go
akabos: Fixed

sugar.go
akabos: I'm not sure either but I don't see much options here. Also I think this wrapper supposed to be merely syntax sugar. It's not supposed to satisfy all possible requirements and just have to be "ok for most uses".

I'm open to any ideas about handling errors, but as for now this is the only working and (mostly) magic free implementation I can think of.

sugar.go
akabos: Not relevant any more. I made sugar logger to handle internal error in the same manner core logger does.

sugar.go
akabos: I made `DFatal` to relay directly to core logger's `DFatal`. 

But speaking of unit test that would have caught this issue, I really doubt that sugar logger suite should include tests for core logger's functionality.

akabos:
@prashantv 

Thanks for your comments! After latest commits, Sugar logger does not inject own error fields into log messages any more. I agree that it was petty awkward way to handle internal errors.

Instead, I made it to behave exactly as core logger does. I hope this is going to satisfy everyone who expressed concern on the matter.

akabos:
@anuptalwalkar 

By no means I insist on "error only on first place" approach. But this special case is the consequence of the special nature of error field. Among all fields only this one has no name so it has to be dealt with specially in any case. I chose this approach only because it seemed easiest to implement while benefits from allowing error in arbitrary position seemed questionable.

I still think that having "first place" restriction for errors is valid, but I'd like to see this PR merged ASAP and I'm ready to change sugar logger to handle errors in any particular way you see appropriate. Just let me know what exactly you see appropriate.

sugar.go
prashantv: I missed this earlier, but it seems like both @akshayjshah and I think we should remove the first variadic argument can be `error`.

As for the checking the `case error`, I'd consider only allowing that for a field named "error", just in case someone wants to log a type that happens to implement `error` but they acutally want the object logged.

sugar.go
prashantv: Can you explain more about the type switch not catching this -- I think keeping log marshaler is a good idea, and it should work,

sugar.go
prashantv: Thanks for fixing it. I'm not suggesting unit testing the core logger's `DFatal method, but a test which ensures that the `DFatal` method here *doesn't* panic if it's passed an implementation of `zap.Logger` which isn't the underlying `logger` type.

sugar.go
```diff
@@ -0,0 +1,200 @@
...
+	// the Development option), otherwise it logs at the Error level.
+	DFatal(string, ...interface{})
+}
+
+type sugar struct {
+	core Logger
+}
+
+func (s *sugar) internalError(msg string) {
+	s.core.(*logger).internalError(msg)
```
prashantv: There should be no casts from `Logger` to `*logger`, as that will panic if the user passes some sort of wrapped logger (e.g., like one returned from sampling).

prashantv:
@akabos The `Error` field is a helper that we don't have to support. I think it's OK to pass `("error", err)`.

Thanks for fixing the error handling, I think logging it as an internal error is definitely better. We need to figure out a way to do that that won't cause panics, but otherwise the error handling is better.

I don't have strong opinions no whether we should always look for `error` type everywhere, or only for a field named `"error"`.

sugar.go
akabos: `internalError`is not declared for `Logger` interface. I'm not sure if there is another way to access it. Even if I re-implement this method on `Sugar`, `ErrorOutput` is still not declared on `LoggerInterface`. What do you suggest?

prashantv: Feel free to defer this with a TODO, and we can circle back to it, since this will require some thinking about.

For now, make the internal error just do `s.core.Error("Logger error", zap.String("message", msg))` or something.

As a separate PR, we need to think about either exposing the error output, or trying to cast the passed in logger to something that implements an `internalError` method, and if it's not implemented doing the log to the normal error stream.

akshayjshah:
@akabos, I'm carrying forward this work in #185; of course, I'm preserving your commits.

PR #150, opened 2016-09-26
Add singleton and self-configuring logger

### LINKED ISSUE #127: Provide a singleton root logger
Loggers typically provide a singleton root logger for convenience. I'm not seeing that in zap. 

I realize it's trivial to implement in application code, but I'm curious about the rationale behind the omission from zap. Is it planned and just not implemented yet?

sectioneight:
This adds `Standard()` which gives you a default JSON logger.

It also gives you `ConfigureStandard()` to set up your standard logger.

Fixes #127

Note that this is just a really rough sketch at how this could look. I don't use singletons much in my code and am sure there's a more elegant/ergonomic way to do this. Thoughts?

```
Standard returns the standard, singleton logger. Although this behavior is discouraged in many settings
many projects use it regardless and we'd prefer a reference implementation to each service doing it differently.
```

### DIFF
--- a/singleton.go
+++ b/singleton.go
@@ -32,7 +32,9 @@ func Standard() Logger {
 	return singletonLogger
 }
 
-// ConfigureStandard configures the singleton logger
+// ConfigureStandard configures the singleton logger. Note that this does not provide any
+// synchronization guarantees -- it is up to you to configure your logger before calling Standard()
+// and using it if you want to customize the options.
 func ConfigureStandard(enc Encoder, options ...Option) Logger {
 	singletonLogger = New(enc, options...)
 	return singletonLogger
--- a/singleton.go
+++ b/singleton.go
@@ -1,0 +1,49 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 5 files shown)


singleton.go
```diff
@@ -0,0 +1,49 @@
...
+
+import "sync"
+
+var (
+	singletonLogger *logger
+	initOnce        sync.Once
+	configLock      sync.Mutex
+)
+
+// Standard returns the standard, singleton logger. Although this behavior is discouraged in many settings
```
prashantv: I think this is going to get confused with the [StandardLogger](https://godoc.org/github.com/uber-go/zap/zwrap#StandardLogger). I'm not sure what a good alternative is though.

singleton.go
```diff
@@ -0,0 +1,49 @@
...
+// many projects use it regardless and we'd prefer a reference implementation to each service doing it differently.
+func Standard() Logger {
+	if singletonLogger == nil {
```
prashantv: This can race with `ConfigureStandard`. Currently `configLock` protects against calling `ConfigureStandard` in parallel, but there's still a race if someone calls it while trying to read it.

I don't really think we want a lock when trying to use the logger though. My preference is to instead not have any synchronization, and make it clear that `ConfigureStandard` cannot be called in parallel with `Standard`, and initialize it in the declaration, e.g.,
```go
var singletonLogger = ConfigureStandard(NewJSONEncoder(...))
```

We can then remove the lock and the `sync.Once`

singleton.go
```diff
@@ -0,0 +1,49 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zap
+
+import "sync"
+
+var (
+	singletonLogger *logger
```
prashantv: do we need the private logger type here?

singleton.go
sectioneight: Well, I think we could call it `StandardLogger` here as well, and then it's clear if you're asking from the `zap` package to get one of those.

For reference, logrus uses `StandardLogger()` to get the singleton: https://github.com/Sirupsen/logrus/blob/master/exported.go#L12

singleton.go
sectioneight: Nope

sectioneight:
I'm fine with moving to a `log` package, although I think it's a bit annoying to have a name that conflicts with the standard lib.

singleton.go
prashantv: The `StandardLogger` in the `zwrap` package refers to an interface that the standard library logger implements, while this one refers to a singleton implementing the `zap.Logger` interface. Having the same name reused for very different things sounds confusing.

We could rename the logger in `zwrap`, or we could do what log15 does, and call it `Root`:
https://godoc.org/github.com/inconshreveable/log15#Root

I'm also noticing that a lot of logging libraries don't expose a singleton at all. Even the standard library exposes functions for the singleton logger, but not the actual instance.

sectioneight:
So it sounds like the open questions are:

1) Should we provide a singleton at all
2) If so, what the name should be

sectioneight:
talked with @prashantv offline and decided more thought/discussion is required as to where this lives and the exact API and semantics

PR #151, opened 2016-09-26
Add Uintptr to the text encoder

prashantv:


prashantv:
Yep a new thing, but you didn't miss anything, just a race between tests being run for https://github.com/uber-go/zap/pull/123 and https://github.com/uber-go/zap/pull/145

PR #152, opened 2016-09-28
Add Tee to write to multiple sinks

### LINKED ISSUE #144: Support tee'ing to multiple sinks
Both for the output and error output, we should support a `Tee` constructor that works like the `tee` Unix utility, echoing output to multiple sinks. This is particularly useful in development, where we may want to send output to both the console and a file.

sectioneight:
Fixes #144 

- [x] Figure out if errors should fail early or we should try to write/sync all backends
- [x] Decide whether we care about short writes and returning `io.ErrShortWriter`

### DIFF
--- a/tee.go
+++ b/tee.go
@@ -23,7 +23,7 @@ package zap
 import "io"
 
 // Tee creates a WriteSyncer that duplicates its writes and
-// sync calls. It is similar to io.MultiWriter
+// sync calls. It is similar to io.MultiWriter.
 func Tee(writeSyncers ...WriteSyncer) WriteSyncer {
 	// Copy to protect against https://github.com/golang/go/issues/7809
 	ws := make([]WriteSyncer, len(writeSyncers))
--- a/tee.go
+++ b/tee.go
@@ -35,7 +35,7 @@ func Tee(writeSyncers ...WriteSyncer) WriteSyncer {
 
 // See https://golang.org/src/io/multi.go
 // In the case where not all underlying syncs writer all bytes, we return the smallest number of bytes wtirren
-// but still call Write() on allthe underlying syncs.
+// but still call Write() on all the underlying syncs.
 func (t *teeWriteSyncer) Write(p []byte) (int, error) {
 	errs := make([]error, 0, len(t.writeSyncers))
 	nWritten := 0
@@ -57,7 +57,7 @@ func (t *teeWriteSyncer) Sync() error {
 	return wrapMutiError(t.writeSyncers...)
 }
 
-// Run a series of `f`s, collecting and aggregating errors ig presenta`
+// Run a series of `f`s, collecting and aggregating errors if presents
 func wrapMutiError(fs ...WriteSyncer) error {
 	errs := make([]error, 0, len(fs))
 	for _, f := range fs {
--- a/tee.go
+++ b/tee.go
@@ -20,7 +20,7 @@
 
 package zap
 
-import "io"
+import "bytes"
 
 // Tee creates a WriteSyncer that duplicates its writes and
 // sync calls. It is similar to io.MultiWriter.
@@ -34,26 +34,60 @@ func Tee(writeSyncers ...WriteSyncer) WriteSyncer {
 }
 
 // See https://golang.org/src/io/multi.go
+// In the case where not all underlying syncs writer all bytes, we return the smallest number of bytes wtirren
+// but still call Write() on allthe underlying syncs.
 func (t *teeWriteSyncer) Write(p []byte) (int, error) {
+	errs := make([]error, 0, len(t.writeSyncers))
+	nWritten := 0
 	for _, w := range t.writeSyncers {
 		n, err := w.Write(p)
 		if err != nil {
-			return 0, err
+			errs = append(errs, err)
 		}
-		if n != len(p) {
-			return n, io.ErrShortWrite
+		if nWritten == 0 && n != 0 {
+			nWritten = n
+		} else if n < nWritten {
+			nWritten = n
 		}
 	}
-	return len(p), nil
+	return nWritten, newMultiError(errs...)
 }
 
 func (t *teeWriteSyncer) Sync() error {
-	for _, w := range t.writeSyncers {
-		if err := w.Sync(); err != nil {
-			return err
+	return wrapMutiError(t.writeSyncers...)
+}
+
+// Run a series of `f`s, collecting and aggregating errors ig presenta`
+func wrapMutiError(fs ...WriteSyncer) error {
+	errs := make([]error, 0, len(fs))
+	for _, f := range fs {
+		if err := f.Sync(); err != nil {
+			errs = append(errs, err)
 		}
 	}
-	return nil
+	return newMultiError(errs...)
+}
+
+func newMultiError(errs ...error) error {
+	if len(errs) == 0 {
+		return nil
+	}
+	return multiError{
+		errs: errs,
+	}
+}
+
+type multiError struct {
+	errs []error
+}
+
+func (m multiError) Error() string {
+	sb := bytes.Buffer{}
+	for _, err := range m.errs {
+		sb.WriteString(err.Error())
+		sb.WriteString(" ")
+	}
+	return sb.String()
 }
 
 type teeWriteSyncer struct {
--- a/tee.go
+++ b/tee.go
@@ -1,0 +1,61 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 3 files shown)


tee.go
```diff
@@ -0,0 +1,61 @@
...
+
+// See https://golang.org/src/io/multi.go
+func (t *teeWriteSyncer) Write(p []byte) (int, error) {
+	for _, w := range t.writeSyncers {
+		n, err := w.Write(p)
+		if err != nil {
+			return 0, err
+		}
+		if n != len(p) {
+			return n, io.ErrShortWrite
```
sectioneight: NOTE: I'm not sure if this is how we want to handle this error. This is how `io.MultiWriter` is implemented, which is why I took this route, but it also creates a weird binding to the `io` package just to import a sentinel error, which I know are generally frowned upon.

tee.go
```diff
@@ -0,0 +1,61 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zap
+
+import "io"
+
+// Tee creates a WriteSyncer that duplicates its writes and
+// sync calls. It is similar to io.MultiWriter
```
akshayjshah: nit: trailing punctuation?

tee.go
akshayjshah: I'll defer to @prashantv on whether we should couple this to the `io` package.

Separately, should we short-circuit if we fail to write to just one of the writers? In the context of a logging package, I'd expect to at least attempt a write to each of the underlying writers.

tee.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+	return len(p), nil
+}
+
+func (t *teeWriteSyncer) Sync() error {
+	for _, w := range t.writeSyncers {
+		if err := w.Sync(); err != nil {
+			return err
+		}
```
akshayjshah: Likewise here.

tee_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type failingWriter struct {
+	err    error
+	nBytes int
+}
+
+func (f failingWriter) Write(p []byte) (int, error) {
+	return f.nBytes, f.err
+}
```
akshayjshah: There's a small helper package that I've been using to test WriteSyncers - might be helpful here. See https://github.com/uber-go/zap/blob/master/spywrite/writer.go.

tee.go
sectioneight: Done

tee.go
sectioneight: I just copied the semantics of MultiWriter. I can change this, but I think it's an opinion call, so I'd rather wait til you and @prashantv agree on the semantics (here and below).

tee_test.go
sectioneight: Cool. Will update.

tee.go
sectioneight: I have no idea why this comment still shows up. Do you know how to mark it as "done" since I already fixed this?

sectioneight:
@prashantv @akshayjshah updated PR description with open questions. Not sure the best platform for discourse. Want to just leave comments here?

tee.go
prashantv: I'd like to try writing to all the writers, and capture any errors in a multi-error.

sectioneight: Sounds good to me. Do you consider a short write an `io.ErroShortWriter` an error, or should we just return `n`, and if so, which `n`?

prashantv: Since we'll try writing to all of them, I think returning min(n) and no error might be the best option. No strong opinions on that though.

sectioneight: Done

tee.go
sectioneight: Done

tee_test.go
sectioneight: Done

sectioneight:
Alrighty, I believe I've adressed all comments. Please take a look when you got a chance :)

tee.go
```diff
@@ -0,0 +1,95 @@
...
+	ws := make([]WriteSyncer, len(writeSyncers))
+	copy(ws, writeSyncers)
+	return &teeWriteSyncer{
+		writeSyncers: ws,
+	}
+}
+
+// See https://golang.org/src/io/multi.go
+// In the case where not all underlying syncs writer all bytes, we return the smallest number of bytes wtirren
+// but still call Write() on allthe underlying syncs.
```
rndstr: typo nit-pick: `all bytes` `wtirrent` `allthe`

tee.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}
+	return nWritten, newMultiError(errs...)
+}
+
+func (t *teeWriteSyncer) Sync() error {
+	return wrapMutiError(t.writeSyncers...)
+}
+
+// Run a series of `f`s, collecting and aggregating errors ig presenta`
```
rndstr: nit: `… if present`?

sectioneight: lol thanks. you can tell i'm working from home since i type on a kinesis advantage at work...

rndstr: haha I presumed it was late-night induced but that Kinesis looks awesome! it even has pedals?! on a related note I stumbled upon the [SafeType](http://allthingsergo.com/blog/wp-content/uploads/2014/02/IMGP4652.jpg) which seems to be on another level

tee.go
sectioneight: oh em ge i need to go to  sleep. god catch!

tee.go
sectioneight: I don't use the pedals because I stand, but I'm sure at some point we'll live in a future where I'm suspended in zero-gravity and using my ears to actuate key combinations

sectioneight:
@prashantv ping for review

tee.go
```diff
@@ -0,0 +1,95 @@
...
+func newMultiError(errs ...error) error {
+	if len(errs) == 0 {
+		return nil
+	}
+	return multiError{
+		errs: errs,
+	}
+}
+
+type multiError struct {
```
prashantv: nit: can probably just do `type multiError []error` and then when it's used earlier, you can actually use `multiError` directly instead of `[]error` and converting

tee.go
```diff
@@ -0,0 +1,95 @@
...
+	return &teeWriteSyncer{
+		writeSyncers: ws,
+	}
+}
+
+// See https://golang.org/src/io/multi.go
+// In the case where not all underlying syncs writer all bytes, we return the smallest number of bytes wtirren
+// but still call Write() on all the underlying syncs.
+func (t *teeWriteSyncer) Write(p []byte) (int, error) {
+	errs := make([]error, 0, len(t.writeSyncers))
```
prashantv: I don't think we want to allocate `errs` here since the common case is that there'll be no errors, so it's not worth eating the cost of an allocation in the successful case.

tee_test.go
```diff
@@ -0,0 +1,102 @@
...
+func TestTeeFailsShortWrite(t *testing.T) {
+	shorter := spywrite.ShortWriter{}
+	ws := Tee(AddSync(shorter))
+
+	n, err := ws.Write([]byte("test"))
+	assert.NoError(t, err, "Expected fake-success from short write")
+	assert.Equal(t, 3, n, "Expected byte count to return from underlying writer")
+}
+
+func TestWritestoAllSyns_EvenIfFirstErrors(t *testing.T) {
```
prashantv: typo: Syns

tee_test.go
```diff
@@ -0,0 +1,102 @@
...
+	failer := spywrite.FailWriter{}
+	shortWriter := spywrite.ShortWriter{}
+	ws := Tee(AddSync(failer), AddSync(shortWriter))
+
+	n, err := ws.Write([]byte("fail"))
+	assert.Error(t, err, "Expected error from call to a writer that failed")
+	assert.Equal(t, 3, n, "Short writer should drop a byte")
```
prashantv: should we use a successful writer after the fail writer and verify contents instead? (this works too, just seems a little clearer to show that the second one can succeed)

tee_test.go
```diff
@@ -0,0 +1,102 @@
...
+	assert.NoError(t, newMultiError(), "no errors should cause a nil error")
+}
+
+func TestNewMultiError_WrapsStrings(t *testing.T) {
+	err := newMultiError(errors.New("battlestar"), errors.New("galactaca"))
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "battlestar")
+	assert.Contains(t, err.Error(), "galactaca")
+}
+
```
prashantv: should we have a test for `Sync` with multiple writers where the first one fails (similar to the Write)

tee.go
sectioneight: Good point, done.

tee.go
sectioneight: Done (why doesn't GH have a done button?)

tee_test.go
sectioneight: Done

tee_test.go
sectioneight: Done

tee_test.go
sectioneight: Done

tee.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			errs = append(errs, err)
+		}
+		if nWritten == 0 && n != 0 {
+			nWritten = n
+		} else if n < nWritten {
+			nWritten = n
+		}
+	}
+	if len(errs) > 0 {
```
prashantv: nit: I typically add a `asError()` method that does this inside the `multiError` type, suggest using that here and in `Sync`

sectioneight: Good call, done

PR #154, opened 2016-09-29
Fix licence check script

sectioneight:
Bash arithmetic is weird. Also git ls-files needs a pattern.

```
zap - fix-license-check ❯ git ls-files "\.go" | wc -l
       0
zap - fix-license-check ❯ git ls-files "*\.go" | wc -l
      70
```


sectioneight:
The += stuff is particularly funny:

```bash
FOO=0 ; for i in $(seq 10) ; do ; FOO+=1 ; done ; echo $FOO
01111111111
```

PR #155, opened 2016-09-30
Add support for uint32

higebu:

Thanks for your contribution!

### DIFF
--- a/field.go
+++ b/field.go
@@ -36,6 +36,7 @@ const (
 	intType
 	int64Type
 	uintType
+	uint32Type
 	uint64Type
 	uintptrType
 	stringType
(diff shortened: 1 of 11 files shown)


prashantv:
Thanks for your contribution @higebu!

I'm not sure whether we should add support for `uint32` on top of `uint64`. We have the following options for signed integers: `zap.Int` and `zap.Int64`.

For users of `int32`, they can use either `zap.Int` or `zap.Int64` with a cast, and the same is true for `uint32` with `zap.Uint` or `zap.Uint64`.

Technically only `zap.Int64` is required, but since the most common integer type is likely `int`, it made sense to provide `zap.Int` as a helper for common use cases.

Would like to hear other people's thoughts on keeping the API small vs making it easier to use in more use cases. cc @abhinav @akshayjshah

higebu:
@prashantv Thank you for your comment.

I know we can use zap.Int64 for uint32 with a cast. Feel free to merge or reject this PR.

maverickwoo:
>Would like to hear other people's thoughts on keeping the API small vs making it easier to use in more use cases.

I advocate supporting every type in the `builtin` package because these types are part of the Go language and it seems reasonable for users of a logging library to expect the library to support all built-in types.

akshayjshah:
For now, I'd also prefer to keep the API as small as is practical. I'm going to close this, but we'll reconsider expanding our type support once the rest of the API is more stable.

Issue #159, opened 2016-10-26
Logger with encoding differing by output?


jcorbin:
So I see that I can have a logger tee'd, e.g. to stdout and a file.  But I'd really like to encode the differently (e.g. text to stdout, json to the file).

Is this presently possible by, at worse, an (ab)use of Hooks?

What would it take to get support for this, if it's not currently supported?

prashantv:
I think this would require a `TeeLogger` which takes multiple underlying loggers and call all functions on both. That way we can have very separate logging configurations (e.g., console gets everything using the TextEncoder, log file only contains warnings or higher encoded using JSON).

akshayjshah:
This was resolved in #161.

PR #161, opened 2016-10-27
Add TeeLogger

jcorbin:
Towards #159 
- [x] made spy logger honor level, so that I can use it to verify one useful way of using tee logger
- [x] write a basic tee logger; TeeLogger is now considered complete per all prior conversations
- [x] finish Level refactor #180 
- [x] finish CheckedMessage composition support #182


tee_logger.go
```diff
@@ -0,0 +1,135 @@
...
+		log.DFatal(msg, fields...)
+	}
+}
+
+func (ml multiLogger) With(fields ...Field) Logger {
+	clone := &multiLogger{
+		Meta: ml.Meta.Clone(),
+		logs: ml.logs,
+	}
+	addFields(clone.Encoder, fields)
```
jcorbin: Actually, I think this is wrong, and points to the fact that I'm missing a test case for the With path (e.g. over differently encoded legs).

tee_logger.go
```diff
@@ -0,0 +1,135 @@
...
+	case PanicLevel, FatalLevel:
+		// Panic and Fatal should always cause a panic/exit, even if the level
+		// is disabled.
+		break
+	default:
+		if lvl < ml.Level() {
+			return nil
+		}
+	}
+	return NewCheckedMessage(ml, lvl, msg)
```
jcorbin: Maybe should call all sub `.Check`s and return a compound checked message.  Otherwise, this assumes that all `Logger` implementations treat level monotonically... maybe that was an intentional design limitation?

tee_logger.go
```diff
@@ -0,0 +1,135 @@
...
+// NOTE: Panic and Fatal may not work as expected, since the first core logger
+// will itself Panic or Fatal (interrupting the TeeLogger loop).
+func TeeLogger(logs ...Logger) Logger {
+	switch len(logs) {
+	case 0:
+		return nil
+	case 1:
+		return logs[0]
+	default:
+		lvl := logs[0].Level()
```
prashantv: nit: I'd call this `minLevel`

tee_logger.go
```diff
@@ -0,0 +1,135 @@
...
+		for _, log := range logs[1:] {
+			if ll := log.Level(); ll < lvl {
+				lvl = ll
+			}
+		}
+		ml := &multiLogger{
+			Meta: MakeMeta(nil),
```
prashantv: Do we need a meta embedded? I think we're going to be implementing almost all the methods it provides ourselves so not sure if it gives us much.

tee_logger.go
prashantv: I think a `Clone` should be calling `With` on all the underlying loggers and creating a new `multiLogger` with the results of the `With`

tee_logger.go
prashantv: That's a good point, I think we should call `Check` on all the underlying implementations too. This becomes especially important when something like the sampled logger relies on the `Check` being called to figure out whether a log should be written

jcorbin:
- took away the meta embedding
- simplified .With cloning, since we've got a value receiver just like Meta does
- that way when I copied the "pointer to an atomic" pattern from it, I was confident it would have same semantics as prior work
- call all the sub logger .With s
- call all the sub loggger .Check s ; the "natural" path of "a multi-logger returns a multi-checked-message" wasn't readily possible:
  - the interface return value is concrete, so I can't easily return a `type multiCheckedMessage []CheckedMessage` that apes the `.OK / .Write` contract
  - I could've tried to expand the CheckedMessage struct to support multi-ing itself..
  - ...but since it's basically "just" a `.Log`-once-guard, decided to start out by "stopping on first OK" and discarding it

Was the design choice to make `CheckedMessage` a concrete type done explicitly for performance?  That's my working suspicion since an interface value may have needed an alloc?

akshayjshah:
> Was the design choice to make CheckedMessage a concrete type done explicitly for performance? That's my working suspicion since an interface value may have needed an alloc?

Nailed it.

Pending changes in other PRs/as discussed, approach looks great.

jcorbin:
Regrettably, the "TeeLogger_Panic" isn't really as useful as it should be, since `spy.Logger` will never panic (similarly, fatal never exits).

tee_logger.go
```diff
@@ -0,0 +1,139 @@
...
+import "github.com/uber-go/atomic"
+
+// TeeLogger creates a Logger that duplicates its log calls to two or
+// more loggers. It is similar to io.MultiWriter.
+//
+// The returned logger will initially have the minimum level of all passed
+// loggers. Changing the returned logger's level will change the level on all
+// wrapped loggers.
+//
+// With the exception of DFatal, the returned tee logger calls .Log(level, ...)
```
jcorbin: This exception is unfortunate in my view, pointing to "DFatal should be just another log level" in my opinion; we could have the exact same semantics, that `logger.DFatal` gets to make a choice, but `logger.Log` either passes the level through, or maps it to "error" always.

jcorbin:
Updated with an initial WIP (tests pass) of the "biggest rock" from #166: dropped `Logger.SetLevel` and `Logger.Level`, replaced with `DynamicLevel`.

I'll start breaking more incremental PRs out tomorrow.

jcorbin:
Ready for review.

tee_logger.go
```diff
@@ -0,0 +1,120 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zap
+
+// TeeLogger creates a Logger that duplicates its log calls to two or
+// more loggers. It is similar to io.MultiWriter.
+//
+// For each logging level methad (.Debug, .Info, etc), the TeeLogger calls each
```
akshayjshah: spelling: "method"

tee_logger.go
```diff
@@ -0,0 +1,120 @@
...
+// level; each sub-logger's DFatal method dynamically chooses to either cal
+// Error or Fatal.
+//
+// Check returns a CheckedMessage chain of any OK CheckedMessages returned by
+// all sub-loggers. The returned message is OK if any of the sub-messages are.
+// An exception is made for FatalLevel and PanicLevel, where a CheckedMessage
+// is returned against the TeeLogger itself. This is so that
+// tlog.Check(PanicLevel, ...).Write(...) is equivalent to tlog.Panic(...)
+// (likewise for FatalLevel).
+func TeeLogger(logs ...Logger) Logger {
```
akshayjshah: Perhaps just `Tee`? We've got `Sample`, and we'll soon have `Sugar` and `Desugar`, all of which implicitly deal in loggers.

When we get around to teeing outputs, we can use `TeeOutput`.

tee_logger.go
```diff
@@ -0,0 +1,120 @@
...
+// For each logging level methad (.Debug, .Info, etc), the TeeLogger calls each
+// sub-logger's level method.
+//
+// Exceptions are made for the Fatal and Panic methods: the returned logger
+// calls .Log(FatalLevel, ...) and .Log(PanicLevel, ...). Only after all
+// sub-loggers have received the message, then the TeeLogger terminates the
+// process (using os.Exit or panic() per usual semantics).
+//
+// DFatal is handled similarly to Fatal and Panic, since it is not actually a
+// level; each sub-logger's DFatal method dynamically chooses to either cal
```
akshayjshah: spelling: "call"

Also, I'm increasingly convinced that you're right about `DFatal` - it should be an actual level. I've filed #188 to track this.

tee_logger.go
```diff
@@ -0,0 +1,120 @@
...
+	case 0:
+		return nil
+	case 1:
+		return logs[0]
+	default:
+		return multiLogger(logs)
+	}
+}
+
+type multiLogger []Logger
```
akshayjshah: Awesome! Really glad that the composition ended up clean enough that we don't need any extra state at all.

tee_logger.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (ml multiLogger) DFatal(msg string, fields ...Field) {
+	for _, log := range ml {
+		log.DFatal(msg, fields...)
+	}
+}
+
+func (ml multiLogger) With(fields ...Field) Logger {
+	ml = append([]Logger(nil), ml...)
```
akshayjshah: What's the benefit of this over `ml := make(multiLogger, len(ml))`?

tee_logger.go
```diff
@@ -0,0 +1,120 @@
...
+	for i, log := range ml {
+		ml[i] = log.With(fields...)
+	}
+	return multiLogger(ml)
```
akshayjshah: Since `multiLogger` exposes the higher-perf `Logger` API, let's golf away the copy here. Can we use a `for i :=` loop and `ml[i] = ml[i].With(fields...)` inside the loop?

(I'll open a separate PR to do the same throughout the core logger.)

tee_logger_test.go
```diff
@@ -0,0 +1,145 @@
...
+			Level:  zap.PanicLevel,
+			Msg:    "baz",
+			Fields: []zap.Field{},
+		},
+	}, sink2.Logs())
+}
+
+// XXX: we cannot presently write `func TestTeeLogger_Fatal(t *testing.T)`,
+// because we can't have both a spy logger and an exit stub without a
+// dependency cycle.
```
akshayjshah: Hopefully this'll get better when/if we promote the spy logger to a full-fledged observing logger.

test_logger_bench_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	withBenchedTeeLogger(b, func(logger zap.Logger) {
+		b.RunParallel(func(pb *testing.PB) {
+			i := 0
+			for pb.Next() {
+				tt := cases[i]
+				if cm := logger.Check(tt.lvl, tt.msg); cm.OK() {
+					cm.Write(zap.Int("i", i))
+				}
+				i = i % len(cases)
```
akshayjshah: Should we be incrementing `i` in here?

Also, what are the benchmark results here?

tee_logger.go
jcorbin: Only rub was `zap.Tee` was already taken; but I too think that `zap.Tee` should be a logger, while the "output tee" gets named differently, will update

PR #162, opened 2016-10-27
Honor level in spy.Logger

jcorbin:
While writing the test for #161, I noticed that `spy.Logger` doesn't honor level.

If desired, I'll first write a test first (who tests the test fixture before the tests do?).

### DIFF
--- a/spy/logger.go
+++ b/spy/logger.go
@@ -102,49 +102,53 @@ func (l *Logger) Check(lvl zap.Level, msg string) *zap.CheckedMessage {
 
 // Log writes a message at the specified level.
 func (l *Logger) Log(lvl zap.Level, msg string, fields ...zap.Field) {
-	l.sink.WriteLog(lvl, msg, l.allFields(fields))
+	l.log(lvl, msg, fields)
 }
 
 // Debug logs at the Debug level.
 func (l *Logger) Debug(msg string, fields ...zap.Field) {
-	l.sink.WriteLog(zap.DebugLevel, msg, l.allFields(fields))
+	l.log(zap.DebugLevel, msg, fields)
 }
 
 // Info logs at the Info level.
 func (l *Logger) Info(msg string, fields ...zap.Field) {
-	l.sink.WriteLog(zap.InfoLevel, msg, l.allFields(fields))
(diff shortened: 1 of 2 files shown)


prashantv:
If we do this, we'll need to take options for setting the level of the `spy.Logger`

There's failing tests probably because the default log level is dropping logs that need to be captured.

jcorbin:
The failing tests turned out to be just that I hadn't followed the "always panic or fatal, level be damned" state of affairs ;-)  However better support for passing options was already on my mind towards a lazy logger, so I'll put out another PR shortly to make that easier, and do for for the spy logger.

zwrap/sample_test.go
```diff
@@ -180,7 +180,9 @@ func TestSamplerCheckPanicFatal(t *testing.T) {
 
 		assert.Nil(t, sampler.Check(zap.DebugLevel, "foo"), "Expected a nil CheckedMessage at disabled log levels.")
 		for i := 0; i < 5; i++ {
-			if cm := sampler.Check(level, "sample"); cm.OK() {
+			cm := sampler.Check(level, "sample")
+			assert.True(t, cm.OK(), "expected fatal level to always be OK")
```
billf: nit: testify assert returns success, this simplifies to
```go
  if assert.True(t,  cm.OK(), "expected blah") {
    cm.Write(zap.Int("iter", i))
  }

File: spy/logger.go [hunk 2]
```diff
@@ -90,59 +90,80 @@ func (l *Logger) With(fields ...zap.Field) zap.Logger {
...
 }
 
+func (l *Logger) log(lvl zap.Level, msg string, fields []zap.Field) {
+	switch lvl {
+	case zap.PanicLevel, zap.FatalLevel:
+		// Panic and Fatal should always cause a panic/exit, even if the level
+		// is disabled.
+		break
+	default:
+		if !(lvl >= l.Level()) {
```
akshayjshah: Simplify to `if lvl < l.Level()`?

jcorbin:
Indeed, next steps under #166 will move towards and eventually de-dupe the copied logic from logger.  Also, in that work, we can flip the comparison once; for now I'm leaving it as "negated >=" for consistency with the core logger implementation.

Issue #156, opened 2016-10-11
Group test packages into testImport


imkira:
Shouldn't we group test packages (benchmark, asserters, etc.) into `testImport` section as explained in the docs?

https://glide.readthedocs.io/en/latest/glide.yaml/

A simple check with the latest glide did not yield any problems like fetching of unnecessary dependencies, but who knows in the future.

akshayjshah:
Probably, yes! Glide has added quite a few features since we started this project. Feel like opening a PR?

imkira:
Good good! @akshayjshah I just send #165 addressing these and other issues.

akshayjshah:
PR merged - thanks for the contribution!

PR #165, opened 2016-11-07
Fix and update glide.yaml

imkira:
- Removed unused dependencies from glide.yaml and vendor
- Moved test dependencies into testImport directive
- Moved non-imported test dependencies into Makefile
- Added license information into glide.yaml

This is a followup on #156

Please let me know if this is OK, or if I should change something.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,55 +1,70 @@
-hash: 64d5a2692f570c258ae1c30e683872ea1f8e7f79595fef4f84517aed055dbb83
-updated: 2016-06-02T18:44:39.806767976-07:00
+hash: 8aeb29a35adb31f8b46a792ae1b304c2c55f2d10bfe0ca1a4b8ac5330e22decc
+updated: 2016-11-09T16:14:48.657534669+09:00
 imports:
-- name: github.com/apex/log
-  version: a999c1b29c986b1972ec53b15092b63a8051ec83
-  subpackages:
-  - handler/json
-- name: github.com/axw/gocov
-  version: f5b2b5cf8644b8b432436eaee1c8a73d2f4cf86e
-  subpackages:
-  - gocov
 - name: github.com/cactus/go-statsd-client
-  version: 5517f304319b9a066354b84860cb264c89655f69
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -24,11 +24,11 @@ dependencies:
...
 	glide install
 	@echo "Installing test dependencies..."
-	go install ./vendor/github.com/axw/gocov/gocov
-	go install ./vendor/github.com/mattn/goveralls
+	go get -u github.com/axw/gocov/gocov
+	go get -u github.com/mattn/goveralls
 ifdef SHOULD_LINT
 	@echo "Installing golint..."
-	go install ./vendor/github.com/golang/lint/golint
+	go get -u github.com/golang/lint/golint
```
akshayjshah: I'd like to keep these pinned with Glide, rather than installing whatever's on master. It's very irritating to have builds start failing if upstream breaks or introduces new lint rules.

Concretely, this means that `goveralls` and `golint` should be in `glide.yaml`.

imkira:
@akshayjshah I changed as you asked.

I am thinking, we should pin down all repos in `glide.yaml`, either using semver or a commit id, otherwise even though they're pinned in glide.lock that only works for `glide install` and as long we dont do `glide up`, in which case all will be re-pinned in `glide.lock`.

akshayjshah:
Thanks for the update, @imkira.

Let's keep this as-is. We don't want to pin any dependencies in the manifest file (`glide.yaml`), for a few reasons:

1. We're pinning in the lock file just for repeatable test runs and benchmarks. As long as none of our dependencies make breaking API changes, we don't need to lock to a particular release.
2. Adding constraints to our manifest affects all users of zap who also use glide (when building projects, glide attempts to honor the manifests of all included libraries).

imkira:
Gocha! I still think pinning down only testImport should not affect users importing via glide. I think glide ignores `testImports` completely when you depend on another glide package and imports all `import` packages implicitly for you. I haven't tested this though but this is what I would expect it to do for me. Anyway, looks good to me as is.

PR #164, opened 2016-11-07
Move testify under testImport

gliptak:
Before opening your pull request, please make sure that you've:

- [X] [signed Uber's Contributor License Agreement](https://docs.google.com/a/uber.com/forms/d/1pAwS_-dA1KhPlfxzYLBqK6rsSWwRwH95OCCZrcsY5rk/viewform);
- [X] added tests to cover your changes;
- [X] run the test suite locally (`make test`); and finally,
- [X] run the linters locally (`make lint`).

Thanks for your contribution!

https://github.com/uber-go/zap/issues/156

This changes ```glide.lock```. Please let me know if an updated version is to be added into this commit.

PR #167, opened 2016-11-07
Move benchmark dependencies under testImports

billf:
Inspired by #164, move all the dependencies we pull in to run benchmark tests into the test import section. Otherwise, consumers of this library will pull in a bunch of dependencies they don't need.


akshayjshah:
Closing in favor of #165.

Issue #166, opened 2016-11-07
Next Steps wrt level


jcorbin:
Capturing the highlights of an offline conversation with @prashantv and @akshayjshah : 

- [x] make it so that `.Log` doesn't actually `panic` or `os.Exit` for `PanicLevel` and `FatalLevel`; this will allow for higher level loggers like `TeeLogger` to correctly fan out to more than one internal logger

We seem to have at least fundamental agreement that the `SetLevel` and `Level` methods weren't actually necessary; most of the conversation centered around instrumental differences in how else to do it.  The key feature that `SetLevel` is providing is dynamic runtime change (e.g. by the http `levelHandler`).  An initial way to eliminate them may look like:
- [x] create an explicit "dynamic level" zap option #180 
  - a first cut of this may first be specific to the http handler
  - follow up work could generalize, e.g. so that the "dynamic level" option constructor, also returns a "controller" that could be shared by more than one runtime mechanism
- ~~this would also set us up well to factor out a "level logger" higher level logger that implements only the level filter logic; this would allow sharing between things like the spy logger and the primary concrete logger~~ (this turned out to not be necessary; DynamicLevel suffices)

- [x] make CheckedMessage "compose better"; it wasn't entirely clear what the benefits were here, I'm hoping @prashantv can reply here with details since he had a clearer handle on this part than I #182

prashantv:
The `CheckedMessage` composition is basically allowing creation of a `CheckedMessage` from other checked messages.

E.g.,
```go
func MultiCheckedMessage(msgs... *CheckedMessage) *CheckedMessage
```

The behaviour I'd expect is:
 - If all of them are `!OK()`, it returns `nil`
 - If only one is `OK()`, it returns that specific instance
 - If there are multiple non-nil entities, we'll need to have a field `[]*CheckedMessage` inside the `CheckedMessage` struct that contains the non-nil entries. This `CheckedMessage` would just call `Write` on the list of `CheckedMessage` rather than using the logger/level.

Does that sound reasonable?

akshayjshah:
That's more or less what I had in mind. Note that it should be safe to pool the `[]*CheckedMessage` and free it at the end of `Write`, since we already ensure that checked messages aren't re-used.

jcorbin:
The only rub there @prashantv:
[...]

PR #168, opened 2016-11-07
Change Logger.Log specification to "should not actually panic or exit"

jcorbin:
Towards #166

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -50,6 +50,11 @@ type Logger interface {
 
 	// Log a message at the given level. Messages include any context that's
 	// accumulated on the logger, as well as any fields added at the log site.
+	//
+	// Calling Panic should panic(), calling Fatal() should terminate the
+	// process, but calling Log(PanicLevel, ...) or Log(FatalLevel, ...) should
+	// not. It may not be possible for compatibility wrappers to comply with
+	// this last part (e.g. the bark wrapper).
 	Log(Level, string, ...Field)
 	Debug(string, ...Field)
 	Info(string, ...Field)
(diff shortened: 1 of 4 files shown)


jcorbin:
Updated #161 to use this, to resolve its prior note on Fatal and Panic

logger.go
```diff
@@ -50,6 +50,11 @@ type Logger interface {
 
 	// Log a message at the given level. Messages include any context that's
 	// accumulated on the logger, as well as any fields added at the log site.
+	//
+	// Calling Panic should panic(), calling Fatal() should terminate the
```
akshayjshah: nit: "...should panic() *and* calling Fatal()..."

logger_test.go
```diff
@@ -184,19 +184,52 @@ func TestJSONLoggerLog(t *testing.T) {
 		logger.Log(DebugLevel, "foo")
 		assert.Equal(t, `{"level":"debug","msg":"foo"}`, buf.Stripped(), "Unexpected output from Log.")
 	})
+}
 
-	withJSONLogger(t, nil, func(logger Logger, buf *testBuffer) {
-		assert.Panics(t, func() { logger.Log(PanicLevel, "foo") }, "Expected logging at Panic level to panic.")
-		assert.Equal(t, `{"level":"panic","msg":"foo"}`, buf.Stripped(), "Unexpected output from panic-level Log.")
-	})
+func TestJSONLoggerLog_panic(t *testing.T) {
```
akshayjshah: I don't think that we're using this underscore convention in the rest of the test suite; purely for consistency, can we call these `TestJSONLoggerLogPanic` and `...Fatal`?

akshayjshah: Whoops, I lied - we're using this in the bark wrapper code. Mea culpa, do as you wish.

akshayjshah:
lgtm, merge away.

PR #172, opened 2016-11-09
Refactor sampler checking

jcorbin:
Another towards the main point of #166:
- call underlying `Logger.Check`, rather than rely on level logic
- explicate the "always panic/fatal" logic
- return any underlying logger CheckedMessage, so that it would work right for #161 back a compound one

### DIFF
--- a/zwrap/sample.go
+++ b/zwrap/sample.go
@@ -66,6 +66,11 @@ func (c *counters) Reset(key string) {
 // sampler will emit the first N logs in each bucket and every Mth log
 // therafter. Sampling loggers are safe for concurrent use.
 //
+// Panic and Fatal logging are NOT sampled, and will always call the underlying
+// logger to panic() or terminate the process. HOWEVER Log-ing at PanicLevel or
+// FatalLevel, if it happens is sampled and will call the underlying logger Log
+// method, which should NOT panic() or terminate.
+//
 // Per-message counts are shared between parent and child loggers, which allows
 // applications to more easily control global I/O load.
 func Sample(zl zap.Logger, tick time.Duration, first, thereafter int) zap.Logger {
--- a/zwrap/sample.go
+++ b/zwrap/sample.go
@@ -144,15 +144,11 @@ func (s *sampler) Error(msg string, fields ...zap.Field) {
 }
 
 func (s *sampler) Panic(msg string, fields ...zap.Field) {
-	if s.check(zap.PanicLevel, msg) {
-		s.Logger.Panic(msg, fields...)
-	}
+	s.Logger.Panic(msg, fields...)
 }
 
 func (s *sampler) Fatal(msg string, fields ...zap.Field) {
-	if s.check(zap.FatalLevel, msg) {
-		s.Logger.Fatal(msg, fields...)
-	}
+	s.Logger.Fatal(msg, fields...)
 }
 
 func (s *sampler) DFatal(msg string, fields ...zap.Field) {
@@ -162,12 +158,6 @@ func (s *sampler) DFatal(msg string, fields ...zap.Field) {
 }
 
 func (s *sampler) check(lvl zap.Level, msg string) bool {
-	switch lvl {
-	case zap.PanicLevel, zap.FatalLevel:
-		// Ignore sampling for Panic and Fatal, since it should always
-		// cause a panic or exit.
-		return true
-	}
 	if s.Logger.Check(lvl, msg) == nil {
 		return false
 	}
--- a/zwrap/sample.go
+++ b/zwrap/sample.go
@@ -98,10 +98,10 @@ func (s *sampler) With(fields ...zap.Field) zap.Logger {
 }
 
 func (s *sampler) Check(lvl zap.Level, msg string) *zap.CheckedMessage {
-	if !s.check(lvl, msg) {
-		return nil
+	if s.check(lvl, msg) {
+		return zap.NewCheckedMessage(s.Logger, lvl, msg)
 	}
-	return zap.NewCheckedMessage(s.Logger, lvl, msg)
+	return nil
 }
 
 func (s *sampler) Log(lvl zap.Level, msg string, fields ...zap.Field) {
--- a/zwrap/sample.go
+++ b/zwrap/sample.go
@@ -162,6 +162,10 @@ func (s *sampler) check(lvl zap.Level, msg string) bool {
 	if s.Logger.Check(lvl, msg) == nil {
 		return false
 	}
+	return s.sampled(lvl, msg)
+}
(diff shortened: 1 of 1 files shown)


zwrap/sample.go
```diff
@@ -98,70 +98,66 @@ func (s *sampler) With(fields ...zap.Field) zap.Logger {
...
 func (s *sampler) Fatal(msg string, fields ...zap.Field) {
-	if s.check(zap.FatalLevel, msg) {
-		s.Logger.Fatal(msg, fields...)
-	}
+	s.Logger.Fatal(msg, fields...)
 }
 
 func (s *sampler) DFatal(msg string, fields ...zap.Field) {
-	if s.check(zap.ErrorLevel, msg) {
+	if s.Logger.Check(zap.ErrorLevel, msg) != nil && s.sampled(zap.ErrorLevel, msg) {
```
jcorbin: This is working as prior, but suspect imo: it won't "always fatal" in dev, since it gets checked and sampled at the "error" level.

akshayjshah: That's a good point. There's no need to keep the existing behavior; let's either fix it in this diff or file an issue.

zwrap/sample.go
```diff
@@ -98,70 +98,66 @@ func (s *sampler) With(fields ...zap.Field) zap.Logger {
...
 func (s *sampler) Check(lvl zap.Level, msg string) *zap.CheckedMessage {
-	if !s.check(lvl, msg) {
+	cm := s.Logger.Check(lvl, msg)
+	if cm == nil {
 		return nil
 	}
-	return zap.NewCheckedMessage(s.Logger, lvl, msg)
+	switch lvl {
+	case zap.PanicLevel, zap.FatalLevel:
+		break
```
akshayjshah: Can we add a note to the `Sample` doc comment pointing out that `Panic` and `Fatal` level logs are never sampled? This isn't a change in behavior, but I neglected to document it properly.

zwrap/sample.go
```diff
@@ -98,70 +98,66 @@ func (s *sampler) With(fields ...zap.Field) zap.Logger {
...
+	default:
+		if !s.sampled(lvl, msg) {
+			return nil
+		}
+	}
+	return cm
```
akshayjshah: To me, this is simpler as:

```go
switch lvl {
case zap.PanicLevel, zap.FatalLevel:
  return cm
}
if s.sampled(lvl, msg) {
  return cm
}
return nil
```

zwrap/sample.go
```diff
@@ -98,70 +98,66 @@ func (s *sampler) With(fields ...zap.Field) zap.Logger {
...
-	switch lvl {
-	case zap.PanicLevel, zap.FatalLevel:
-		// Ignore sampling for Panic and Fatal, since it should always
-		// cause a panic or exit.
-		return true
-	}
-	if lvl < s.Level() {
-		return false
-	}
+func (s *sampler) sampled(lvl zap.Level, msg string) bool {
```
akshayjshah: Since we're not using it, can we change this method to not take the level?

zwrap/sample.go
jcorbin: Described in #175 for now; deferring rather than add yet one more stacked PR towards in this train ;-)

zwrap/sample.go
jcorbin: Done, recorded my nascent desires for level+msg sampling in #176.

zwrap/sample.go
```diff
@@ -98,70 +103,62 @@ func (s *sampler) With(fields ...zap.Field) zap.Logger {
...
-	return zap.NewCheckedMessage(s.Logger, lvl, msg)
+	if s.sampled(msg) {
+		return cm
+	}
+	return nil
 }
 
 func (s *sampler) Log(lvl zap.Level, msg string, fields ...zap.Field) {
-	if s.check(lvl, msg) {
+	if s.Logger.Check(lvl, msg) != nil && s.sampled(msg) {
```
akshayjshah: Can we special-case panic/fatal logging here as well? Instead of making the sampling behavior dependent on which logger methods are called (`sampler.Panic(msg)` is different from `sampler.Log(zap.PanicLevel, msg`), we should not sample panic or fatal level logs anywhere.

Also, let's use the `OK` method instead of a nil check (same throughout the leveled logging methods).

zwrap/sample.go
```diff
@@ -98,70 +103,62 @@ func (s *sampler) With(fields ...zap.Field) zap.Logger {
...
-	}
+	s.Logger.Panic(msg, fields...)
 }
 
 func (s *sampler) Fatal(msg string, fields ...zap.Field) {
-	if s.check(zap.FatalLevel, msg) {
-		s.Logger.Fatal(msg, fields...)
-	}
+	s.Logger.Fatal(msg, fields...)
 }
```
akshayjshah: Since we've embedded the unsampled logger, do we need `Panic` and `Fatal` on the sampler? We should get delegation for free.

zwrap/sample.go
```diff
@@ -98,70 +103,62 @@ func (s *sampler) With(fields ...zap.Field) zap.Logger {
...
 func (s *sampler) Check(lvl zap.Level, msg string) *zap.CheckedMessage {
-	if !s.check(lvl, msg) {
-		return nil
+	cm := s.Logger.Check(lvl, msg)
+	switch lvl {
+	case zap.PanicLevel, zap.FatalLevel:
+		return cm
 	}
-	return zap.NewCheckedMessage(s.Logger, lvl, msg)
+	if s.sampled(msg) {
```
akshayjshah: I think that we still need a call to `cm.OK` here:

1. We don't want calls to `logger.Check(zap.DebugLevel, "foo")` to count against the sampling for `logger.Check(zap.ErrorLevel, "foo")` if debug-level logging is disabled.
2. There's also no need to incur the locking/atomic/map access costs of `sampled` for disabled levels.

Perhaps just this:
```go
if !cm.OK() || s.sampled(msg) {
  return cm
}
```

zwrap/sample.go
jcorbin: Hmm, it certainly will be the simplest change (and in line with the current doc strings) for now to make `Log`ing at fatal or panic level not sampled.

However, now that we're in this place of "you _can_ log at Fatal or Panic level directly, and it's not terminal", a sampled logger should now downsample such logging to protect downstreams.  That said, I can see the argument for "no, we want that to blow up as horrible unsupported/broken usage"...

jcorbin: Going to try going one step further on `cm.OK()`, and actually call `cm.Write` instead of `s.logger.Log`.

PR #173, opened 2016-11-09
WIP: Dynamiclevel instead of SetLevel

jcorbin:
WIP prefix of #161 that gets us through most of #166 (only CheckedMessage composition is yet TODO).

PRs to peel out of this one before its ready for proper review:
- [x] rework fatal/panic logic #168 
- ~~decouple sampler checking from level logic #172~~ (turns out that this isn't a strict dependency, but stil good to do towards #166 )
- [x] add `Meta.Enabled`, and  unify `Check` implementations (spy, sampler, logger, zbark) through `Meta` #174
- [ ] rework levels to be "just" an implementation of meta enabled #177
- [ ] replace `Meta.SetLevel` / `Meta.Level` with a `DynamicLevel` implementation of meta enabled

### DIFF
--- a/http_handler.go
+++ b/http_handler.go
@@ -35,18 +35,19 @@ type errorResponse struct {
 }
 
 type levelHandler struct {
-	logger Logger
+	lvl AtomicLevel
 }
 
-// NewHTTPHandler returns an HTTP handler that can atomically change the logging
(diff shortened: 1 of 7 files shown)


jcorbin:
Superceded by #180

PR #174, opened 2016-11-09
Factor out Meta.{Check,Enabled}

jcorbin:
Set's the stage to further redefine levels towards #173.

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -79,3 +79,19 @@ func (m Meta) Clone() Meta {
 func (m Meta) Enabled(lvl Level, msg string) bool {
 	return lvl >= m.Level()
 }
+
+// Check returns a CheckedMessage logging the given message is Enabled, nil
+// otherwise.
+func (m Meta) Check(log Logger, lvl Level, msg string) *CheckedMessage {
+	switch lvl {
+	case PanicLevel, FatalLevel:
+		// Panic and Fatal should always cause a panic/exit, even if the level
+		// is disabled.
+		break
+	default:
+		if !m.Enabled(lvl, msg) {
+			return nil
+		}
+	}
+	return NewCheckedMessage(log, lvl, msg)
+}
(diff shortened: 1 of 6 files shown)


meta.go
```diff
@@ -74,3 +74,24 @@ func (m Meta) Clone() Meta {
 	m.Encoder = m.Encoder.Clone()
 	return m
 }
+
+// Enabled returns true if logging a message at a particular level is enabled.
+func (m Meta) Enabled(lvl Level, msg string) bool {
```
akshayjshah: Why take the message as a param?

meta.go
```diff
@@ -74,3 +74,24 @@ func (m Meta) Clone() Meta {
...
+	case PanicLevel, FatalLevel:
+		// Panic and Fatal should always cause a panic/exit, even if the level
+		// is disabled.
+		break
+	default:
+		if !m.Enabled(lvl, msg) {
+			return nil
+		}
+	}
+	return NewCheckedMessage(log, lvl, msg)
```
akshayjshah: Unless there's another use case for it, I'm all for un-exporting `NewCheckedMessage`. Existing callers should all have access to a `Meta`, which they can use instead.

zwrap/sample_test.go
```diff
@@ -181,8 +181,7 @@ func TestSamplerCheckPanicFatal(t *testing.T) {
 			}
 		}
 
-		expected := buildExpectation(level, 0, 1, 2, 3, 4)
-		assert.Equal(t, expected, sink.Logs(), "Unexpected output when sampling with Check.")
+		assert.Equal(t, []spy.Log(nil), sink.Logs(), "Unexpected output when sampling with Check.")
```
akshayjshah: Yowza...previous behavior was totally busted :(

meta.go
jcorbin: Good qn, I think having the msg only encourages sampling-like things; will drop.

PR #177, opened 2016-11-10
Indirect level checking

jcorbin:
Makes each level value a static `LevelEnabler`.

I had to introduce a lock to avoid the race condition, before we introduce `{Atomic,Dynamic}Level` in #173.  If that's too distasteful, we can just jump to reviewing #173 instead.


jcorbin:
Abandoning; consolidating with #173

Issue #143, opened 2016-09-17
Use bit-shifting instead of modulus in sampling


akshayjshah:
We can make the sampling wrapper ever so slightly faster if we use bit-shifting operations instead of the modulus operator. This may require rounding the sampling rates to the nearest (or next-lowest) power of two.

gliptak:
Do you refer to https://github.com/uber-go/zap/blob/master/zwrap/sample.go#L172 here? Thanks

akshayjshah:
Yep, that's the line!

akshayjshah:
After @gliptak explored this in #178, the hoped-for performance gains didn't materialize. Closing this, since the gains aren't worth the API limitations.

PR #178, opened 2016-11-11
Use bit-wise and operation to calculate sampling (#143)

gliptak:

Thanks for your contribution!

https://github.com/uber-go/zap/issues/143

### DIFF
--- a/zwrap/sample.go
+++ b/zwrap/sample.go
@@ -64,36 +64,44 @@ func (c *counters) Reset(key string) {
 // Sample returns a sampling logger. The logger maintains a separate bucket
 // for each message (e.g., "foo" in logger.Warn("foo")). In each tick, the
 // sampler will emit the first N logs in each bucket and every Mth log
-// therafter. Sampling loggers are safe for concurrent use.
+// thereafter (rounded down to nearest power of 2). Sampling loggers are safe for concurrent use.
 //
 // Per-message counts are shared between parent and child loggers, which allows
 // applications to more easily control global I/O load.
 func Sample(zl zap.Logger, tick time.Duration, first, thereafter int) zap.Logger {
+	power := 1
+	for {
+		thereafter >>= 1
+		if thereafter <= 0 {
+			break
+		}
+		power <<= 1
+	}
 	return &sampler{
-		Logger:     zl,
-		tick:       tick,
-		counts:     &counters{counts: make(map[string]*atomic.Uint64)},
-		first:      uint64(first),
-		thereafter: uint64(thereafter),
+		Logger: zl,
+		tick:   tick,
+		counts: &counters{counts: make(map[string]*atomic.Uint64)},
+		first:  uint64(first),
+		power:  uint64(power - 1),
 	}
 }
 
 type sampler struct {
 	zap.Logger
 
-	tick       time.Duration
-	counts     *counters
-	first      uint64
-	thereafter uint64
+	tick   time.Duration
+	counts *counters
+	first  uint64
+	power  uint64
 }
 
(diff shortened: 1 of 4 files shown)


zwrap/sample.go
```diff
@@ -64,36 +64,44 @@ func (c *counters) Reset(key string) {
 // Sample returns a sampling logger. The logger maintains a separate bucket
 // for each message (e.g., "foo" in logger.Warn("foo")). In each tick, the
 // sampler will emit the first N logs in each bucket and every Mth log
-// therafter. Sampling loggers are safe for concurrent use.
+// thereafter (rounded down to nearest power of 2). Sampling loggers are safe for concurrent use.
```
akshayjshah: Thanks for fixing this misspelling :)

zwrap/sample.go
```diff
@@ -64,36 +64,44 @@ func (c *counters) Reset(key string) {
...
-		Logger:     zl,
-		tick:       tick,
-		counts:     &counters{counts: make(map[string]*atomic.Uint64)},
-		first:      uint64(first),
-		thereafter: uint64(thereafter),
+		Logger: zl,
+		tick:   tick,
+		counts: &counters{counts: make(map[string]*atomic.Uint64)},
+		first:  uint64(first),
+		power:  uint64(power - 1),
```
akshayjshah: If `thereafter` isn't a power of two, let's also log an informative error using the passed-in logger. The message should include the desired value and the value we've rounded down to.

That way users get some feedback that they're misconfiguring their logger.

zwrap/sample_test.go
```diff
@@ -97,7 +97,7 @@ func TestSampler(t *testing.T) {
 		for i := 1; i < 10; i++ {
 			tt.logFunc(sampler, i)
 		}
-		expected := buildExpectation(tt.level, 1, 2, 5, 8)
+		expected := buildExpectation(tt.level, 1, 2, 4, 6, 8)
```
akshayjshah: For completeness, let's also add a test that passes an exact power-of-two value for `thereafter` (which ensures that we're handling the rounding-down logic properly).

akshayjshah:
One more important point: can you also add a benchmark for this? I'd like to compare against the master branch and make sure that the performance improvement is worth the somewhat surprising behavior.

gliptak:
@akshayjshah Which benchmark method do you recommend to clone?
https://github.com/uber-go/zap/tree/master/benchmarks
Also how/where do you run the benchmarks?

zwrap/sample_test.go
gliptak: added

zwrap/sample.go
gliptak: using zl or onto stderr?

akshayjshah: Using `zl` - I'd like to make sure that the message ends up in the user's application logs.

gliptak:
Master (modified to ```1<<13```):

```
BenchmarkZapSampleWithoutFields-4             	20000000	        77.4 ns/op	       0 B/op	       0 allocs/op
```

My branch:
```
BenchmarkZapSampleWithoutFieldsPowerOfTwo-4   	20000000	        75.2 ns/op	       0 B/op	       0 allocs/op
```

gliptak:
@akshayjshah Are you good with the benchmark?

akshayjshah:
@gliptak Thanks for adding the benchmark! Unfortunately, I don't think it's worth introducing this API limitation (sampling only powers of two) for a 2ns improvement.

We do very much appreciate your contribution, though - if you're interested in adding more to zap, I can tag you on some proposals that are less speculative.

gliptak:
@akshayjshah Sure, please point me to other proposals.

PR #180, opened 2016-11-14
Replace Logger.{Level,SetLevel} with DynamicLevel

jcorbin:
Depends on #172, supercedes #173.

After this PR, the only part of #166 left is the CheckedMessage composition work.

### DIFF
--- a/http_handler.go
+++ b/http_handler.go
@@ -35,18 +35,19 @@ type errorResponse struct {
 }
 
 type levelHandler struct {
-	logger Logger
+	lvl AtomicLevel
 }
 
-// NewHTTPHandler returns an HTTP handler that can atomically change the logging
-// level at runtime. Keep in mind that changing a logger's level also affects that
-// logger's ancestors and descendants.
+// NewHTTPHandler returns an HTTP handler that can atomically change a dynamic
+// logging level at runtime. Keep in mind that changing a dynamic logging level
+// affects not just the logger(s) that it was passed to, but also any
+// sub-loggers that have been or will be created With them.
 //
 // GET requests return a JSON description of the current logging level. PUT
 // requests change the logging level and expect a payload like:
 //   {"level":"info"}
-func NewHTTPHandler(logger Logger) http.Handler {
-	return &levelHandler{logger: logger}
+func NewHTTPHandler(lvl AtomicLevel) http.Handler {
+	return &levelHandler{lvl: lvl}
 }
 
 func (h *levelHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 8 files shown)


jcorbin:
Rebased, ready for review

http_handler.go
```diff
@@ -35,18 +35,19 @@ type errorResponse struct {
 }
 
 type levelHandler struct {
-	logger Logger
+	lvl AtomicLevel
```
akshayjshah: Why not just add `ServeHTTP` to the `AtomicLevel` type? It doesn't add any overhead if it's not used, and it removes one type and wrapper from this now-crowded package.

level.go
```diff
@@ -105,3 +121,41 @@ func (l *Level) UnmarshalText(text []byte) error {
 	}
 	return nil
 }
+
+// Enabled return true if the given level is at or above this level.
```
akshayjshah: return -> returns

level.go
```diff
@@ -105,3 +121,41 @@ func (l *Level) UnmarshalText(text []byte) error {
...
+func (l Level) Enabled(lvl Level) bool {
+	return lvl >= l
+}
+
+// DynamicLevel creates an atomically changeable dynamic logging level.  The
+// returned level can be passed as a logger option just like a concrete level.
+//
+// The value's SetLevel() method may be called later to change the enabled
+// logging level of all loggers that were passed the value (either explicitly,
+// or by creating sub-loggers with Logger.With).
```
akshayjshah: nice - this is a much clearer explanation than the parent/child terminology I was using.

meta.go
```diff
@@ -37,37 +33,25 @@ type Meta struct {
 	Hooks       []Hook
 	Output      WriteSyncer
 	ErrorOutput WriteSyncer
-
-	lvl *atomic.Int32
+	LevelEnabler
```
akshayjshah: Can we keep embeds at the top, separated from the other fields with a blank line?

akshayjshah:
Great - appreciate the quick turnaround after a delayed review. Feel free to merge after resolving the conflict.

anuptalwalkar:
A quick question on this PR. You seem to have removed the Level access on Logger object. How will a user know what log level is set on the created/passed zap.Logger object?

jcorbin:
@anuptalwalkar the idea now is that there isn't necessary a "the level of a logger object" concept; I'd respond to your question of "how?" with a clarifying "why?":
- if what you're trying to do is do your own filtering style logic, then you should be just calling `Logger.Check`
- otherwise, you may want to be using zap `zap.DynamicLevel`, which you can ask "what's the effective logging level"?

If those don't meet your needs, can you describe your use case a little more? (maybe in a new issue?)

anuptalwalkar:
I see. my concerns were around setting up the Logger correctly. So now I can add `zap.DynamicLevel` as an option, but I can't add `zap.DynamicLevel().SetLevel(zap.WarnLevel)` as an option.

anuptalwalkar:
Nevermind actually. I understand now. I have updated my changes as well https://github.com/uber-go/fx/pull/81

PR #181, opened 2016-11-15
Improve internal error handling

jcorbin:
- add timestamp to error output
- tag with "cause" (e.g. "hook" vs "encoder")
- move method to Meta so that other implementation can use it

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -100,6 +100,6 @@ func (m Meta) Check(log Logger, lvl Level, msg string) *CheckedMessage {
 // InternalError prints an internal error message to the configured
 // ErrorOutput
 func (m Meta) InternalError(cause string, err error) {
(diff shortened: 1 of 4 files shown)


meta.go
```diff
@@ -95,3 +96,10 @@ func (m Meta) Check(log Logger, lvl Level, msg string) *CheckedMessage {
 	}
 	return NewCheckedMessage(log, lvl, msg)
 }
+
+// InternalError prints an internal error message to the configured
+// ErrorOutput
```
akshayjshah: Small nits, feel free to land once they're addressed:

1. Please add a period to the end of the sentence.
2. Mind adding an additional sentence explaining that this method is *only* for errors caused by bugs in the logger implementation? Any user-caused errors should just be logged to the application log.

PR #182, opened 2016-11-15
Add CheckedMessage Chaining

jcorbin:
Final part of #166 towards #161 makes CheckedMessage composition friendly.

### DIFF
--- a/checked_message.go
+++ b/checked_message.go
@@ -97,12 +97,15 @@ func (m *CheckedMessage) Write(fields ...Field) {
 //
 // The returned message will first Write to any prior loggers before logging to
 // the newly passed logger.
-func (m *CheckedMessage) Chain(logger Logger, lvl Level, msg string) *CheckedMessage {
-	if m == nil {
-		return logger.Check(lvl, msg)
-	}
-	if next := NewCheckedMessage(logger, lvl, msg); next.OK() {
-		m.push(next)
+func (m *CheckedMessage) Chain(ms ...*CheckedMessage) *CheckedMessage {
+	for _, m2 := range ms {
+		if m2.OK() {
+			if m.OK() {
+				m.push(m2)
+			} else {
+				m = m2
+			}
+		}
 	}
 	return m
 }
--- a/checked_message.go
+++ b/checked_message.go
@@ -30,32 +30,12 @@ type CheckedMessage struct {
 	used   atomic.Uint32
 	lvl    Level
 	msg    string
-	cms    []*CheckedMessage
-}
 
-// MultiCheckedMessage combines one or more CheckedMessages into one, keeping
-// only those that return true from OK().
-func MultiCheckedMessage(cms ...*CheckedMessage) *CheckedMessage {
-	cms = compactCheckedMessages(cms)
-	switch len(cms) {
-	case 0:
-		return nil
-	case 1:
-		return cms[0]
-	default:
-		return &CheckedMessage{cms: cms}
-	}
-}
-
-func compactCheckedMessages(cms []*CheckedMessage) []*CheckedMessage {
-	i := 0
-	for j := 0; j < len(cms); j++ {
-		if cms[j].OK() {
-			cms[i] = cms[j]
-			i++
-		}
-	}
-	return cms[:i]
+	// singly linked list built by Chain
+	next *CheckedMessage // carried by each part of Chain-ed list
+	tail *CheckedMessage // non-nil only in the head of a Chain-ed list
+	// NOTE: suffixes are NOT valid lists, since they lack a tail pointer; tail
+	// pointer exist solely as an optimization for calling head.Chain.
 }
 
 // NewCheckedMessage constructs a CheckedMessage. It's only intended for use by
@@ -89,13 +69,6 @@ func (m *CheckedMessage) Write(fields ...Field) {
 		return
 	}
 
-	if m.cms != nil {
-		for _, cm := range m.cms {
-			cm.Write(fields...)
-		}
-		return
-	}
-
 	switch m.lvl {
 	case DebugLevel:
 		m.logger.Debug(m.msg, fields...)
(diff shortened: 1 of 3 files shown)


checked_message.go
```diff
@@ -30,6 +30,51 @@ type CheckedMessage struct {
 	used   atomic.Uint32
 	lvl    Level
 	msg    string
+	cms    []*CheckedMessage
+}
+
+// MultiCheckedMessage combines one or more CheckedMessages into one, keeping
+// only those that return true from OK().
+func MultiCheckedMessage(cms ...*CheckedMessage) *CheckedMessage {
```
jcorbin: Note Well Reviewers:
- I'm hoping to get away with this "MultiCheckedMessage owns its varargs slice" tradeoff...
- ...rather than needing to copy, e.g. like "zap.Tee"

jcorbin:
E.g. benchmark results over this branch:

- 1d49292 Add MultiCheckedMessage
```
BenchmarkMultiCheckedMessage-8                  20000000               123 ns/op             133 B/op          2 allocs/op
BenchmarkMultiCheckedMessage_sliceLoggers-8     10000000               159 ns/op             149 B/op          3 allocs/op
PASS
ok      github.com/uber-go/zap  4.370s
```

- 6a0b1ae MultiCheckedMessage: refactor to compact and re-use cms
```
BenchmarkMultiCheckedMessage-8                  20000000               130 ns/op             144 B/op          3 allocs/op
BenchmarkMultiCheckedMessage_sliceLoggers-8     10000000               124 ns/op             144 B/op          3 allocs/op
PASS
ok      github.com/uber-go/zap  4.112s
```

- 3afed5e CheckedMessage: shift to singly-linked chaining
```
BenchmarkCheckedMessage_Chain-8                 20000000               121 ns/op             128 B/op          2 allocs/op
BenchmarkCheckedMessage_Chain_sliceLoggers-8    10000000               126 ns/op             128 B/op          2 allocs/op
PASS
ok      github.com/uber-go/zap  3.920s
```

- 30a969a CheckedMessage.Chain: take variadic list; refactor and broaden test
```
BenchmarkCheckedMessage_Chain-8                 20000000               122 ns/op             106 B/op          1 allocs/op
BenchmarkCheckedMessage_Chain_sliceLoggers-8    10000000               124 ns/op             106 B/op          1 allocs/op
PASS
ok      github.com/uber-go/zap  3.957s
```

checked_message.go
```diff
@@ -74,6 +85,44 @@ func (m *CheckedMessage) Write(fields ...Field) {
...
+
+// Chain builds a chain of checked messages over several loggers. It returns an
+// OK message if either the prior message, or the new one returned by
+// logger.Check is OK.
+//
+// The returned message will first Write to any prior loggers before logging to
+// the newly passed logger.
+func (m *CheckedMessage) Chain(ms ...*CheckedMessage) *CheckedMessage {
+	for _, m2 := range ms {
+		if m2.OK() {
```
akshayjshah: Can reduce nesting here with

```go
if !m2.OK() {
  continue
}
```

checked_message.go
```diff
@@ -74,6 +85,44 @@ func (m *CheckedMessage) Write(fields ...Field) {
...
+		}
+	}
+	return m
+}
+
+func (m *CheckedMessage) push(next *CheckedMessage) {
+	if m.tail != nil {
+		m.tail.next = next
+	} else if m.next != nil {
+		panic("invalid CheckedMessage linked list; did we loose our head?")
```
akshayjshah: Rather than panic, can we use `m.logger.DFatal` to log an error? I don't expect this to come up, but I'd really hate for a small error on our part to crash application processes in production.

Also, s/loose/lose.

checked_message_test.go
```diff
@@ -60,3 +60,140 @@ func TestCheckedMessageIsSingleUse(t *testing.T) {
...
+				desc:       "( B OK, C OK ), A OK",
+			},
+		}
+
+		for i, tt := range tests {
+			cm := tt.cms[0].Chain(tt.cms[1:]...)
+			assert.Equal(t, cm.OK(), tt.expectedOK, "%s: expected cm.OK()", tt.desc)
+			cm.Write(Int("i", i))
+		}
+		assert.Equal(t, expected, buf.Lines(), "expected output from MultiCheckedMessage")
```
akshayjshah: Could we include the expected output in the test structs, `Reset`ing the output buffer on each iteration? Might make it easier to identify the failing test case when we break something.

checked_message.go
```diff
@@ -30,6 +30,12 @@ type CheckedMessage struct {
 	used   atomic.Uint32
 	lvl    Level
 	msg    string
+
+	// singly linked list built by Chain
+	next *CheckedMessage // carried by each part of Chain-ed list
+	tail *CheckedMessage // non-nil only in the head of a Chain-ed list
+	// NOTE: suffixes are NOT valid lists, since they lack a tail pointer; tail
+	// pointer exist solely as an optimization for calling head.Chain.
```
akshayjshah: s/exist/exists

akshayjshah:
For shame! Thou art missing the all-important license header.

checked_message.go
```diff
@@ -74,6 +85,43 @@ func (m *CheckedMessage) Write(fields ...Field) {
...
+		}
+	}
+	return m
+}
+
+func (m *CheckedMessage) push(next *CheckedMessage) {
+	if m.tail != nil {
+		m.tail.next = next
+	} else if m.next != nil {
+		m.logger.DFatal("invalid CheckedMessage linked list; did we loose our head?", String("original", m.msg))
```
akshayjshah: homophone nit: s/loose/lose.

checked_message_bench_test.go
```diff
@@ -0,0 +1,96 @@
+package zap_test
```
akshayjshah: The linter is failing the builds b/c this file needs a license header.

checked_message_test.go
```diff
@@ -60,3 +60,143 @@ func TestCheckedMessageIsSingleUse(t *testing.T) {
...
+			// singleton ok cases
+			{
+				cms: []*CheckedMessage{
+					nil,
+					loga.Check(InfoLevel, "apple"),
+				},
+				expectedOK: true,
+				desc:       "nil, A OK",
+				expected: []string{
+					`{"level":"info","msg":"apple","name":"A","i":3}`,
```
akshayjshah: Now that we're clearing the output buffer for each test, do we still need the loop index in the messages? It'll be easier to add new tests or reorder them later if we omit `i`.

checked_message_bench_test.go
```diff
@@ -0,0 +1,96 @@
...
+		{zap.InfoLevel, "fwiw"},
+		{zap.ErrorLevel, "hey!"},
+	}
+
+	p := atomic.NewInt32(0)
+
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		i := 0
+		j := p.Inc()
```
prashantv: nit: instead of casting to `int(j)` twice below, you can cast it once here, `j := int(p.Inc())`

checked_message_bench_test.go
prashantv: any reason why this is in a separate package? i think `json_encoder_bench_test.go` and `null_encoder_bench_test.go` are both in the `zap` package.

For the logger benchmarks, they are good examples of user interaction so it's nice to have them in a separate package and import `zap`, but I think this should just be part of the `zap` package since it's internal benchmarking

checked_message_bench_test.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func BenchmarkCheckedMessage_Chain_sliceLoggers(b *testing.B) {
+	logs := []zap.Logger{
+		zap.New(
+			zap.NullEncoder(),
+			zap.InfoLevel,
+			zap.DiscardOutput,
+		),
+		zap.New(
```
prashantv: should we move the common logic into a private function?
```go
newBenchmarkLogger(options ...zap.Option)
```

It would pass the `NullEncoder()` and `DiscardOutput` by default

checked_message_test.go
jcorbin: Well, i is providing value in that it's testing that .Write passes along fields to all chained Writes correctly, and that those field values weren't just contextually shared somehow before the Check().Write() site.

Issue #136, opened 2016-08-20
JSON Array support


kazegusuri:
zap supports nested JSON Objects with `zap.Nest` or `zap.Marshaler`, but does not support JSON Array yet.  Current workaround is to use `zap.Object` and implement `MarshalJSON` to output JSON Array. How about built-in support for JSON Array like `zap.Array`?

akshayjshah:
That's definitely on our radar. For now, you can use `zap.Object` - if you don't mind falling back to the standard library's `json.Marshal`, you don't even have to implement `MarshalJSON`.

The best way I can think of to support JSON arrays (while keeping performance good) is to add specific helpers for each slice type. We'd end up with `zap.Ints`, `zap.Floats`, `zap.Durations`, etc.

Are you interested in working on this feature?

jcorbin:
@akshayjshah , what do you think of the following approach:
```
diff --git a/keyvalue.go b/keyvalue.go
index cda24f5..862ebb3 100644
--- a/keyvalue.go
+++ b/keyvalue.go
@@ -38,4 +38,26 @@ type KeyValue interface {
 	// allocation-heavy. Consider implementing the LogMarshaler interface instead.
 	AddObject(key string, value interface{}) error
 	AddString(key, value string)
+	AddArray(key, valueAdder func(Values) error)
+}
+
+// Values is an encoding agnostic interface to add array data to a keyed value
+// within a logging context. Like slices, Values aren't safe for concurrent use
+// (though typical use shouldn't require locks).
+//
+// XXX example
+type Values interface {
+	AddBool(value bool)
+	AddFloat64(value float64)
+	AddInt(value int)
+	AddInt64(value int64)
+	AddUint(value uint)
+	AddUint64(value uint64)
+	AddUintptr(value uintptr)
+	AddMarshaler(marshaler LogMarshaler) error
+	// AddObject uses reflection to serialize arbitrary objects, so it's slow and
+	// allocation-heavy. Consider implementing the LogMarshaler interface instead.
+	AddObject(value interface{}) error
+	AddString(value string)
+	AddArray(valueAdder func(Values) error) // XXX to taste depending on how recursive you're feeling this morning ;-)
 }
```

akshayjshah:
This is in-progress on the dev branch: #211

akshayjshah:
Further support in #258.

akshayjshah:
Landed on dev.

PR #183, opened 2016-11-16
WIP: Add Caller/Callers fields

jcorbin:
This PR is essentially a proposal-as-prototype-code, rather than than proposal-as-issue; how about:
- replacing `zap.Stack` with `zap.Callers`
- simplifying the `zap.AddCaller` hook to just add a `zap.Caller` field, rather than mutate msg...
- ...oh and also let the user pass a skip offset to `zap.AddCaller` so that it can get it right under logger wrappers

Other questions include:
- what if we added proper array support, rather than the hacked object? (see also #136)
- maybe we should have explicit PC(s) marshaling support in the the KeyValue interface itself...
- ...this would let the text encoder recover a more human friendly stacktrace-esque form
- ...maybe the Error field should also always (modulo an Option?) add a callers field?

### DIFF
--- a/field.go
+++ b/field.go
@@ -24,6 +24,8 @@ import (
 	"encoding/base64"
 	"fmt"
 	"math"
+	"runtime"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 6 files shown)


jcorbin:
Superceded by #201, will open an issue to discuss what missing functionality we want to support going forward.

PR #184, opened 2016-11-18
WIP: Sentry logger

glibsm:
This Sentry Logger can be used in combination with the forthcoming TeeLogger to easily report errors, or other types of logs as well, to Sentry.

It is written in a separate package `zap/sentry`, so the users of zap do not inherit a transitive dependence, but I think that quite a few if the current users might be interested in this Sentry integration.

It allows for somewhat flexible setup about how and when to send sentry packets through `Capturer` interface. Provided with it are two examples: memory one used for testing and a non-blocking one (without checking the outcome of the packet send). If speed is no main concern, another `BlockingCapturer` could be added in the future that will wait on the outcome of the packet send.

TODO: Benchmark?

### DIFF
--- a/sentry/capturer.go
+++ b/sentry/capturer.go
@@ -22,7 +22,7 @@ package sentry
 
 import raven "github.com/getsentry/raven-go"
 
-// Capturer knows what to do with a Sentry packet
+// Capturer knows what to do with a Sentry packet.
 //
 // Allows for a variety of implementations of how to send Sentry packets.
 // For more performance sensisitve systems, it might make sense to batch
@@ -39,12 +39,12 @@ func (m *memCapturer) Capture(p *raven.Packet) {
 	m.packets = append(m.packets, p)
 }
 
-// NonBlockingCapturer does not wait for the result of Sentry packet sending
+// NonBlockingCapturer does not wait for the result of Sentry packet sending.
 type NonBlockingCapturer struct {
 	*raven.Client
 }
 
-// Capture will fire off a packet without checking the error channel
+// Capture will fire off a packet without checking the error channel.
 func (s *NonBlockingCapturer) Capture(p *raven.Packet) {
 	s.Client.Capture(p, nil)
 }
--- a/sentry/capturer.go
+++ b/sentry/capturer.go
@@ -23,28 +23,25 @@ package sentry
 import raven "github.com/getsentry/raven-go"
 
 // Capturer knows what to do with a Sentry packet.
-//
-// Allows for a variety of implementations of how to send Sentry packets.
-// For more performance sensitive systems, it might make sense to batch
-// rather than opening up a connection on each send.
 type Capturer interface {
-	Capture(p *raven.Packet)
+	Capture(p *raven.Packet) error
 }
 
 type memCapturer struct {
 	packets []*raven.Packet
 }
 
-func (m *memCapturer) Capture(p *raven.Packet) {
+func (m *memCapturer) Capture(p *raven.Packet) error {
 	m.packets = append(m.packets, p)
+	return nil
 }
 
-// NonBlockingCapturer does not wait for the result of Sentry packet sending.
-type NonBlockingCapturer struct {
+type nonBlockingCapturer struct {
 	*raven.Client
 }
 
 // Capture will fire off a packet without checking the error channel.
-func (s *NonBlockingCapturer) Capture(p *raven.Packet) {
+func (s *nonBlockingCapturer) Capture(p *raven.Packet) error {
 	s.Client.Capture(p, nil)
+	return nil
 }
--- a/sentry/capturer.go
+++ b/sentry/capturer.go
@@ -25,11 +25,14 @@ import raven "github.com/getsentry/raven-go"
 // Capturer knows what to do with a Sentry packet.
 type Capturer interface {
 	Capture(p *raven.Packet) error
+	Close()
 }
 
 type memCapturer struct {
 	packets []*raven.Packet
 }
+
+func (m *memCapturer) Close() {}
 
 func (m *memCapturer) Capture(p *raven.Packet) error {
 	m.packets = append(m.packets, p)
@@ -40,8 +43,12 @@ type nonBlockingCapturer struct {
 	*raven.Client
 }
 
+func (nb *nonBlockingCapturer) Close() {
+	nb.Client.Close()
+}
+
 // Capture will fire off a packet without checking the error channel.
-func (s *nonBlockingCapturer) Capture(p *raven.Packet) error {
-	s.Client.Capture(p, nil)
+func (nb *nonBlockingCapturer) Capture(p *raven.Packet) error {
+	nb.Client.Capture(p, nil)
 	return nil
 }
--- a/sentry/capturer.go
+++ b/sentry/capturer.go
@@ -1,0 +1,50 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE
+
+package sentry
+
+import raven "github.com/getsentry/raven-go"
+
+// Capturer knows what to do with a Sentry packet
+//
+// Allows for a variety of implementations of how to send Sentry packets.
+// For more performance sensisitve systems, it might make sense to batch
+// rather than opening op a connection on each send.
+type Capturer interface {
+	Capture(p *raven.Packet)
+}
+
+type memCapturer struct {
+	packets []*raven.Packet
+}
+
+func (m *memCapturer) Capture(p *raven.Packet) {
+	m.packets = append(m.packets, p)
+}
+
+// NonBlockingCapturer does not wait for the result of Sentry packet sending
+type NonBlockingCapturer struct {
+	*raven.Client
+}
+
+// Capture will fire off a packet without checking the error channel
+func (s *NonBlockingCapturer) Capture(p *raven.Packet) {
+	s.Client.Capture(p, nil)
+}
--- a/sentry/capturer.go
+++ b/sentry/capturer.go
@@ -25,8 +25,8 @@ import raven "github.com/getsentry/raven-go"
 // Capturer knows what to do with a Sentry packet.
 //
 // Allows for a variety of implementations of how to send Sentry packets.
-// For more performance sensisitve systems, it might make sense to batch
-// rather than opening op a connection on each send.
+// For more performance sensitive systems, it might make sense to batch
+// rather than opening up a connection on each send.
 type Capturer interface {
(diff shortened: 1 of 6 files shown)


sentry/sentry.go
```diff
@@ -0,0 +1,212 @@
...
+		traceContextLines: _traceContextLines,
+		extra:             zwrap.KeyValueMap{},
+	}
+
+	for _, option := range options {
+		option(l)
+	}
+
+	// Default to live sentry packet capturer
+	if l.Capturer == nil {
```
jcorbin: How about a `Client` option instead of the dsn string?
- then you wouldn't need an `error` return here, since it's the caller's responsibility to create and pass a raven client
- other capturer forms could be handled by other options (e.g. a `BatchedClient` option)

sentry/sentry.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+		l.Capturer = &NonBlockingCapturer{Client: client}
+	}
+
+	return l, nil
+}
+
+// WithCapturer allows to specify which packet capturer should be used
+func WithCapturer(c Capturer) func(l *Logger) {
```
jcorbin: This `With` prefix isn't a pattern that the core zap options follow, maybe we should drop it here too?

sentry/sentry.go
```diff
@@ -0,0 +1,212 @@
...
+	return func(l *Logger) {
+		l.Capturer = c
+	}
+}
+
+// WithMinLevel provides a slice of levels for which sentry packaet should be sent
```
jcorbin: It's not actually a slice anymore, artifact of prior form?

Ideally, the sentry logger should use the new unified level options introduced by #180; however the sentry logger here doesn't currently take any of the core zap options for its embedded `Meta`.

sentry/sentry.go
```diff
@@ -0,0 +1,212 @@
...
+// takes place
+func WithTraceEnabled(enabled bool) func(l *Logger) {
+	return func(l *Logger) {
+		l.traceEnabled = enabled
+	}
+}
+
+// WithTraceCfg allows to change the number of skipped frames, number of context lines and
+// list of go prefixes that are considered "in-app", i.e. "github.com/uber-go/zap"
+func WithTraceCfg(skip, context int, prefixes []string) func(l *Logger) {
```
jcorbin: These could (probably should) just be three separate options.

sentry/sentry.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Error sends Sentry information provided minimum threshold is met
+func (l *Logger) Error(msg string, fields ...zap.Field) {
+	l.log(zap.ErrorLevel, msg, fields)
+}
+
+// Panic sends Sentry information provided minimum threshold is met
+func (l *Logger) Panic(msg string, fields ...zap.Field) {
+	l.log(zap.PanicLevel, msg, fields)
```
jcorbin: should `panec()` after log

sentry/sentry.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Fatal sends Sentry information provided minimum threshold is met
+func (l *Logger) Fatal(msg string, fields ...zap.Field) {
+	l.log(zap.FatalLevel, msg, fields)
```
jcorbin: Should `os.Exit(1)` after log.

@akshayjshah maybe we should move the `_exit` symbol into a sub-package like `github.com/uber-go/zap/internal` so that it can be shared by other zap-included concrete implementations for testing? (as bonus, this would be one less reason why tests couldn't be in their `*_test` package form)

sentry/sentry.go
```diff
@@ -0,0 +1,212 @@
...
+	// Consider pre-allocing the whole new resulting map, but that may be even more expensive
+	for _, f := range fields {
+		f.AddTo(l.extra)
+	}
+	return l
+}
+
+// Notify sentry if the log level meets minimum threshold
+func (l *Logger) log(lvl zap.Level, msg string, fields []zap.Field) {
+	if lvl >= l.minLevel {
```
jcorbin: Should be using `Meta.Enabled`, if we could do away with the `minLevel` field here and just use `Meta`'s leveling support

sentry/sentry.go
glibsm: I started that way, then consolidated :(

sentry/sentry.go
glibsm: have never heard of `panEc` ;)
will make the change

sentry/sentry.go
glibsm: There is one small problem with this @jcorbin : it means that people using this logger would have to import raven-go as well as zap/sentry (which also imports raven-go) into their projects. If the two api versions don't match then there be trouble. 

I wonder if this is a safer approach, even though it requires some more error validation.

sentry/sentry.go
```diff
@@ -0,0 +1,214 @@
...
+		l.traceAppPrefixes = prefixes
+	}
+}
+
+// Extra stores additional information for each Sentry event.
+func Extra(extra map[string]interface{}) Option {
+	return func(l *Logger) {
+		l.extra = extra
+	}
+}
```
akshayjshah: For congruence with `zap.Fields`, consider calling this `Fields` and taking a variadic number of `zap.Fields` as arguments. We'll likely want to pass the same information in both places, so representing it the same way will be convenient.

sentry/sentry.go
```diff
@@ -0,0 +1,214 @@
...
+	traceAppPrefixes  []string
+
+	// Additional information that should be attached to each packet, i.e. customerUUID
+	extra zwrap.KeyValueMap
+}
+
+// Option pattern for logger creation.
+type Option func(l *Logger)
+
+// New Sentry logger.
```
akshayjshah: Please document the default settings.

sentry/sentry.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+}
+
+// TraceEnabled allows to change wether (the somewhat expensive, relatively)
+// stack trace generation takes place.
+func TraceEnabled(enabled bool) Option {
+	return func(l *Logger) {
+		l.traceEnabled = enabled
+	}
+}
```
akshayjshah: Since the default is to enable tracing, consider making this `DisableTraces()`.

sentry/sentry.go
```diff
@@ -0,0 +1,214 @@
...
+	// Consider pre-allocing the whole new resulting map, but that may be even more expensive
+	for _, f := range fields {
+		f.AddTo(l.extra)
+	}
+	return l
+}
+
+// Notify sentry if the log level meets minimum threshold.
+func (l *Logger) log(lvl zap.Level, msg string, fields []zap.Field) {
+	if lvl >= l.minLevel {
```
akshayjshah: Prefer to structure your code to minimize indentation (similar to https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow) - in this case, return early if the level doesn't meet the min, which allows the meat of the method to be de-dented.

sentry/sentry.go
```diff
@@ -0,0 +1,214 @@
...
+		// append all the fields from the current log message to the stored ones
+		extra := l.extra
+		for _, f := range fields {
+			f.AddTo(extra)
+		}
```
akshayjshah: Again, this is mutating the receiver, which isn't what we want.

sentry/sentry.go
```diff
@@ -0,0 +1,214 @@
...
+
+		if l.traceEnabled {
+			trace := raven.NewStacktrace(l.traceSkipFrames, l.traceContextLines, l.traceAppPrefixes)
+			if trace != nil {
+				packet.Interfaces = append(packet.Interfaces, trace)
+			}
+		}
+
+		l.Capture(packet)
```
akshayjshah: Fine for now. You may want to talk to the upstream `raven-go` maintainers to see if they're willing to expose a more performant API. Perhaps `CaptureBytes([]byte)`, so we can use zap's JSONEncoder and incrementally serialize this payload, passing the JSON-encoded bytes here?

/cc @sectioneight

sentry/capturer.go
```diff
@@ -0,0 +1,50 @@
...
+// Allows for a variety of implementations of how to send Sentry packets.
+// For more performance sensisitve systems, it might make sense to batch
+// rather than opening op a connection on each send.
+type Capturer interface {
+	Capture(p *raven.Packet)
+}
+
+type memCapturer struct {
+	packets []*raven.Packet
+}
```
akshayjshah: Is anyone going to use the `memCapturer` outside the tests in this package? If not, let's un-export all these types and remove the `WithCapturer` option - your tests can just mutate `logger.Capturer`.

sentry/sentry.go
```diff
@@ -0,0 +1,214 @@
...
+		zap.InfoLevel:  raven.INFO,
+		zap.WarnLevel:  raven.WARNING,
+		zap.ErrorLevel: raven.ERROR,
+		zap.PanicLevel: raven.ERROR,
+		zap.FatalLevel: raven.FATAL,
+	}
+)
+
+// Logger automatically sends logs above a certain threshold to Sentry.
+type Logger struct {
```
akshayjshah: Do we need to export this type, or can `New` just return a `zap.Logger`? If possible, I'd prefer that - you can cast back to a `*logger` in your tests.

sentry/capturer.go
```diff
@@ -0,0 +1,50 @@
...
+
+package sentry
+
+import raven "github.com/getsentry/raven-go"
+
+// Capturer knows what to do with a Sentry packet.
+//
+// Allows for a variety of implementations of how to send Sentry packets.
+// For more performance sensisitve systems, it might make sense to batch
+// rather than opening op a connection on each send.
```
glibsm: Yeah I don't know what's vscode's problem. I've fixed it once before to run spell check in comments, seems like I need to look into it again.

sentry/sentry.go
glibsm: @akshayjshah I'm not for a moment suggesting that there will be two versions of a binary running around, I was simply talking about glide and dependency management.

This change would mean that people using `zap/sentry` would also have to import `raven-go` directly in order to use the sentry logger. That means they *could* lock in the version of raven-go in their glide.yaml, which then can lead to trouble. We run into this situation pretty frequently with uberfx, since we're acting as a middle man for a lot of other libraries.

If both of you don't think that it is going to be a usability issue, I have no problem with making the change otherwise.

sentry/sentry.go
```diff
@@ -0,0 +1,214 @@
...
+// DFatal is a nop.
+func (l *Logger) DFatal(msg string, fields ...zap.Field) {}
+
+// With returns Sentry logger with additional context.
+func (l *Logger) With(fields ...zap.Field) zap.Logger {
+	// Consider pre-allocing the whole new resulting map, but that may be even more expensive
+	for _, f := range fields {
+		f.AddTo(l.extra)
+	}
+	return l
```
glibsm: Totally. I'll make sure it's covered by tests

glibsm: > Most of the time, fields added to that logger won't ever be sent to Sentry

@akshayjshah why do you think that's the case? I think it's quite the opposite - we want to provide as much context as possible to Sentry, which includes all the `with*` fields.

sentry/sentry.go
glibsm: Good tip, thank you.

sentry/sentry.go
glibsm: Yes, that is definitely on the list. 

There are a bunch of PRs/issues into raven-go along the same lines: not very flexible external API. We plan to help with the situation if we can.

Even though, I still think using their lower-level API to capture a packet is better than doing it by hand ourselves, but could be convinced either way.

sentry/sentry.go
akshayjshah: Most of the time, we log at `InfoLevel` and below; in those cases, any work done in the Sentry logger is pure overhead, since we typically use Sentry only for errors and above.

sentry/sentry.go
akshayjshah: Let's see what the benchmarks look like.

sentry/sentry.go
glibsm: I see what you mean now, makes sense.

You are talking about lowering the overhead as much as possible for the Sentry logger when it actually doesn't need to send any packets or be active in any way, I will put some more thought into it and add some benches for various cases.

sentry/sentry.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// Fatal sends Sentry information provided minimum threshold is met, then exits.
+func (l *Logger) Fatal(msg string, fields ...zap.Field) {
+	l.log(zap.FatalLevel, msg, fields)
+	os.Exit(1)
+}
+
+// DFatal is a nop.
+func (l *Logger) DFatal(msg string, fields ...zap.Field) {}
```
glibsm: I think @jcorbin had some ideas around it.

The question is: what happens when TeeLogger calls out to multiple loggers that fatal/panic. Is it only the first one that gets a chance to act? Is the order somehow defined explicitly?

akshayjshah: There are some issues filed about DFatal, but we'll migrate the Sentry logger along with all the others. Tee loggers don't call the leveled logger methods - they use `.Log`.

HJFinch:
Hey there,
do you have any information why this was closed/deleted? The topic itself would be very interesting and it would be a great feature.

PR #185, opened 2016-11-21
Add SugaredLogger

### LINKED ISSUE #138: Add a slightly slower, but less verbose, logger
`zap.Logger` is really fast, but it's quite verbose - users must repeatedly reference the `zap` package. This is a good trade-off for performance-critical applications, but it's not good for most users. Consider the impression that this snippet from the README leaves on potential users and contributors:

``` go
logger.Info("Failed to fetch URL.",
  zap.String("url", url),
  zap.Int("attempt", tryNum),
  zap.Duration("backoff", sleepFor),
)
```

The wrapper provided by the `zbark` sub-package is great for our internal use, since we're using the `bark.Logger` interface all over. However, it's also unpleasantly verbose:

``` go
logger.WithFields(bark.Fields{
    "url": url,
    "attempt": tryNum,
    "backoff": sleepFor,
}).Info("Failed to fetch URL.")
```

To show off how easy it is to write a clean wrapper on top of the core `zap.Logger` type, spur adoption, and generally make a better impression on first-time users, I'd like to include a higher-level `zap.SugaredLogger` (name TBD - we could even rename the current logger to `CoreLogger` and call the new thing `Logger`). I'm open to other ideas, but I'd like something similar to the `log15` API:

``` go
logger := zap.Sugar(coreLogger)
logger.Info("Failed to fetch URL",
    "url", url,
    "attempt", tryNum,
    "backoff", sleepFor,
)
anotherCoreLogger := zap.Desugar(logger)
```

akshayjshah:
This PR carries forward the work that @akabos did in #147. Once complete, this will resolve #138.

Relevant benchmarks for logrus, the regular zap logger, and the sugared logger:

```
BenchmarkLogrusAddingFields-4                         	  100000	     10219 ns/op	    5726 B/op	      77 allocs/op
BenchmarkLogrusWithAccumulatedContext-4               	  200000	      7760 ns/op	    3919 B/op	      61 allocs/op
BenchmarkLogrusWithoutFields-4                        	 1000000	      2317 ns/op	    1321 B/op	      25 allocs/op

BenchmarkZapAddingFields-4                            	 1000000	      1287 ns/op	     768 B/op	       4 allocs/op
BenchmarkZapWithAccumulatedContext-4                  	 5000000	       341 ns/op	      64 B/op	       2 allocs/op
BenchmarkZapWithoutFields-4                           	 5000000	       363 ns/op	      64 B/op	       2 allocs/op

BenchmarkZapSugarAddingFields-4                       	  500000	      4160 ns/op	    1443 B/op	      36 allocs/op
BenchmarkZapSugarWithAccumulatedContext-4             	 3000000	       408 ns/op	      80 B/op	       3 allocs/op
BenchmarkZapSugarWithoutFields-4                      	 5000000	       400 ns/op	      80 B/op	       3 allocs/op
```

By these benchmarks, we're doing pretty well - the sugared logger is similar to logrus in developer ergonomics, but is much, much faster and lower-allocation. However, the sugared field-adding API does introduce quite a bit of allocation overhead.

### DIFF
--- a/field.go
+++ b/field.go
@@ -201,7 +201,7 @@ func Any(key string, value interface{}) zapcore.Field {
 	case time.Duration:
 		return Duration(key, val)
 	case fmt.Stringer:
-		return String(key, val.String())
+		return Stringer(key, val)
 	default:
 		return Reflect(key, val)
 	}
--- a/field.go
+++ b/field.go
@@ -156,3 +156,53 @@ func Reflect(key string, val interface{}) zapcore.Field {
 func Nest(key string, fields ...zapcore.Field) zapcore.Field {
 	return zapcore.Field{Key: key, Type: zapcore.ObjectMarshalerType, Interface: zapcore.Fields(fields)}
 }
+
+// Any takes a key and an arbitrary value and chooses the best way to represent
+// them as a field, falling back to a reflection-based approach only if
+// necessary.
+func Any(key string, value interface{}) zapcore.Field {
+	switch val := value.(type) {
+	case zapcore.ObjectMarshaler:
+		return Object(key, val)
+	case bool:
+		return Bool(key, val)
+	case float64:
+		return Float64(key, val)
+	case float32:
+		return Float64(key, float64(val))
+	case int:
+		return Int(key, val)
+	case int64:
+		return Int64(key, val)
+	case int32:
+		return Int64(key, int64(val))
+	case int16:
+		return Int64(key, int64(val))
+	case int8:
+		return Int64(key, int64(val))
+	case uint:
+		return Uint(key, val)
+	case uint64:
+		return Uint64(key, val)
+	case uint32:
+		return Uint64(key, uint64(val))
+	case uint16:
+		return Uint64(key, uint64(val))
+	case uint8:
+		return Uint64(key, uint64(val))
+	case uintptr:
+		return Uintptr(key, val)
+	case string:
+		return String(key, val)
+	case error:
+		return String(key, val.Error())
+	case time.Time:
+		return Time(key, val)
+	case time.Duration:
+		return Duration(key, val)
+	case fmt.Stringer:
+		return String(key, val.String())
+	default:
+		return Reflect(key, val)
+	}
+}
--- a/field.go
+++ b/field.go
@@ -164,6 +164,8 @@ func Any(key string, value interface{}) zapcore.Field {
 	switch val := value.(type) {
 	case zapcore.ObjectMarshaler:
 		return Object(key, val)
+	case zapcore.ArrayMarshaler:
+		return Array(key, val)
 	case bool:
 		return Bool(key, val)
 	case float64:
@@ -194,12 +196,12 @@ func Any(key string, value interface{}) zapcore.Field {
 		return Uintptr(key, val)
 	case string:
 		return String(key, val)
-	case error:
-		return String(key, val.Error())
 	case time.Time:
 		return Time(key, val)
 	case time.Duration:
 		return Duration(key, val)
+	case error:
+		return String(key, val.Error())
 	case fmt.Stringer:
 		return Stringer(key, val)
 	default:
--- a/sugar.go
+++ b/sugar.go
@@ -77,17 +77,39 @@ func (s *sugar) With(args ...interface{}) (Sugar, error) {
 	return NewSugar(s.core.With(fields...)), nil
 }
 
+const (
+	expectFirst = iota
+	expectString
+	expectValue
+)
+
 func getSugarFields(args ...interface{}) ([]Field, error) {
-	if (len(args) % 2) != 0 {
+	var (
+		noErrArgs []interface{}
+		fields    []Field
+
+		ii    int
+		key   string
+		value interface{}
+	)
+
+	if len(args) == 0 {
+		return fields, nil
+	}
+
+	switch args[0].(type) {
+	case error:
+		fields = append(fields, Error(args[0].(error)))
+		noErrArgs = args[1:]
+	default:
+		noErrArgs = args
+	}
+
+	if (len(noErrArgs) % 2) != 0 {
 		return nil, errors.New("invalid number of arguments")
 	}
-	var (
-		fields []Field
-		ii     int
-		key    string
-		value  interface{}
-	)
-	for ii, value = range args {
+
+	for ii, value = range noErrArgs {
 		if (ii % 2) == 0 {
 			switch value.(type) {
 			case string:
@@ -117,6 +139,8 @@ func getSugarFields(args ...interface{}) ([]Field, error) {
 				fields = append(fields, Duration(key, value.(time.Duration)))
 			case fmt.Stringer:
 				fields = append(fields, Stringer(key, value.(fmt.Stringer)))
+			case error:
+				return nil, errors.New("error can only be the first argument")
 			default:
 				return nil, errors.New("invalid argument type")
 			}
--- a/sugar.go
+++ b/sugar.go
@@ -112,31 +112,33 @@ func getSugarFields(args ...interface{}) ([]Field, error) {
 				return nil, errors.New("field name must be string")
 			}
 		} else {
-			switch value := value.(type) {
-			case bool:
-				fields = append(fields, Bool(key, value))
-			case float64:
-				fields = append(fields, Float64(key, value))
-			case int:
-				fields = append(fields, Int(key, value))
-			case int64:
-				fields = append(fields, Int64(key, value))
-			case uint:
-				fields = append(fields, Uint(key, value))
-			case uint64:
-				fields = append(fields, Uint64(key, value))
-			case string:
-				fields = append(fields, String(key, value))
-			case time.Time:
-				fields = append(fields, Time(key, value))
-			case time.Duration:
-				fields = append(fields, Duration(key, value))
-			case fmt.Stringer:
-				fields = append(fields, Stringer(key, value))
+			switch v := value.(type) {
 			case error:
 				return nil, errors.New("error can only be the first argument")
+			case bool:
+				fields = append(fields, Bool(key, v))
+			case float64:
+				fields = append(fields, Float64(key, v))
+			case int:
+				fields = append(fields, Int(key, v))
+			case int64:
+				fields = append(fields, Int64(key, v))
+			case uint:
+				fields = append(fields, Uint(key, v))
+			case uint64:
+				fields = append(fields, Uint64(key, v))
+			case string:
+				fields = append(fields, String(key, v))
+			case time.Time:
+				fields = append(fields, Time(key, v))
+			case time.Duration:
+				fields = append(fields, Duration(key, v))
+			case fmt.Stringer:
+				fields = append(fields, Stringer(key, v))
+			case LogMarshaler:
+				fields = append(fields, Marshaler(key, v))
 			default:
-				return nil, errors.New("invalid argument type")
+				fields = append(fields, Object(key, value))
 			}
 		}
 	}
--- a/sugar.go
+++ b/sugar.go
@@ -192,9 +192,9 @@ func (s *sugar) Fatal(msg string, args ...interface{}) {
 }
 
 func (s *sugar) DFatal(msg string, args ...interface{}) {
-	lvl := ErrorLevel
-	if s.core.(*logger).Development {
-		lvl = FatalLevel
+	fields, err := getSugarFields(args...)
+	if err != nil {
+		s.internalError(err.Error())
 	}
-	s.Log(lvl, msg, args...)
+	s.core.DFatal(msg, fields...)
 }
--- a/sugar.go
+++ b/sugar.go
@@ -1,0 +1,169 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zap
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+// Sugar is a wrapper around core logger whith less verbose API
+type Sugar interface {
+	// Check the minimum enabled log level.
+	Level() Level
+	// Change the level of this logger, as well as all its ancestors and
+	// descendants. This makes it easy to change the log level at runtime
+	// without restarting your application.
+	SetLevel(Level)
+
+	// Create a child logger, and optionally add some context to that logger.
+	With(...interface{}) (Sugar, error)
+
+	// Log a message at the given level. Messages include any context that's
+	// accumulated on the logger, as well as any fields added at the log site.
+	Log(Level, string, ...interface{}) error
+	Debug(string, ...interface{}) error
+	Info(string, ...interface{}) error
+	Warn(string, ...interface{}) error
+	Error(string, ...interface{}) error
+	Panic(string, ...interface{}) error
+	Fatal(string, ...interface{}) error
+	// If the logger is in development mode (via the Development option), DFatal
+	// logs at the Fatal level. Otherwise, it logs at the Error level.
+	DFatal(string, ...interface{}) error
+}
+
+type sugar struct {
+	core Logger
+}
+
+// NewSugar is a constructor for Sugar
+func NewSugar(core Logger) Sugar {
+	return &sugar{core}
+}
+
+func (s *sugar) Level() Level {
+	return s.core.Level()
+}
+
+func (s *sugar) SetLevel(lvl Level) {
+	s.core.SetLevel(lvl)
+}
+
+func (s *sugar) With(args ...interface{}) (Sugar, error) {
+	fields, err := getSugarFields(args...)
+	if err != nil {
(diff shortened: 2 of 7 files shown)


jcorbin:
This is great, I think it may even deduplicate effort between uberfx and other ongoing internal work.

My only thoughts on a high level about its API are: it seems like we have strictly two things going on here:
1. adding sugared methods for adding fields in a less verbose way
2. providing it a way back to dynamic message string  building

I'd really like to see us keep those concerns split (e.g.SugaredLogger embedded by FormattedLogger?) My primary concerns are around making it clear to other core zap pieces when they can and cannot assume that message is a manageable cardinality (e.g. the sampler would at worst degrade to no weighting, but we could do something even more graceful like degrade to an LRU counts table).

anuptalwalkar:
@jcorbin I agree with not duplicating this effort. UberFx's current interface is conforming with SugarLogger. Once this PR is ready for merge,  My plan is to cleanup UberFx log and embed sugar logger. That way we will get unified experience whether we are using UberFx or zap.

akabos:
Any progress on that? Is there anything I can assist with?

sugar.go
```diff
@@ -0,0 +1,214 @@
...
+//  )
+// is the equivalent of
+//   coreLogger.With(
+//     String("hello", "world"),
+//     String("failure", "oh no"),
+//     Int("count", 42),
+//     Object("user", User{name: "alice"}),
+//   )
+func (s *SugaredLogger) With(args ...interface{}) *SugaredLogger {
+	return &SugaredLogger{core: s.core.With(sweetenFields(args, s.core)...)}
```
jcorbin: Pulling my thread along about `original, mine *Logger`: I guess then that `With` should only with `mine` and set `s.original = nil`, leaving it up to an eventual desugar to fault in if someone wants to unwrap a contextified sugared logger.

jcorbin: Also, I feel that this method should just be "sugar" for `WithFields`: `return s.WithFields(sweetenFields(args, s.core)...)`

sugar.go
```diff
@@ -0,0 +1,214 @@
...
+// Fatalf uses fmt.Sprintf to construct a dynamic message and logs it at
+// FatalLevel, then calls os.Exit(1). It doesn't add to the message's
+// structured context.
+func (s *SugaredLogger) Fatalf(template string, args ...interface{}) {
+	if ce := s.core.Check(FatalLevel, fmt.Sprintf(template, args...)); ce != nil {
+		ce.Write()
+	}
+}
+
+func sweetenFields(args []interface{}, errLogger Logger) []zapcore.Field {
```
jcorbin: Hmm, this is so close to being a more useful `func ZipAny(...interface{}) ([]zapcore.Field, error)`.

Towards a counterfactual then, how bad would it be to not have this sugaredlogger, when logging sites could be:
```
log.Info("stuff", zap.ZipAny(
    "this", 2,
    "andThen", 3,
    "andFurthermore", map[string]string{"cantStop": "meNow!"},
)...)

log.Debug(fmt.Sprintf("I just %s", "can't stop myself"))
```

In the real "approved" case, there's only one "zap" in play, and in the regrettable case, they clearly know what they've bought into...

(this counterfactual up against my prior "how really does this (un)wrapping work" meme)

sugar.go
```diff
@@ -0,0 +1,214 @@
...
+
+// Sugar converts a Logger to a SugaredLogger.
+func Sugar(core Logger) *SugaredLogger {
+	// TODO: increment caller skip.
+	return &SugaredLogger{core}
+}
+
+// Desugar unwraps a SugaredLogger, exposing the original Logger.
+func Desugar(s *SugaredLogger) Logger {
+	// TODO: decrement caller skip.
```
akshayjshah: TODO is to use whatever `WithOptions` thing we end up settling on.

sugar.go
akshayjshah: Sure, that's reasonable.

sugar.go
akshayjshah: Happy to consider `ZipAny` later on; in general, though, I think that there's an important place for a logger that makes these semi-magical behaviors the default. In contexts where performance isn't a primary concern, I'd like to have as little noise as possible around logging sites.

sugar.go
```diff
@@ -0,0 +1,214 @@
...
+// WithFields adds structured fields to the logger's context, just like the
+// standard Logger's With method.
+func (s *SugaredLogger) WithFields(fs ...zapcore.Field) *SugaredLogger {
+	return &SugaredLogger{core: s.core.With(fs...)}
+}
+
+// Debug logs a message and some key-value pairs at DebugLevel. Keys and values
+// are treated as they are in the With method.
+func (s *SugaredLogger) Debug(msg interface{}, keysAndValues ...interface{}) {
+	if ce := s.core.Check(DebugLevel, sweetenMsg(msg)); ce != nil {
```
prashantv: Why duplicate these check + call to Write for each level? A single helper (that will probably be inlined) should clean up this implementation:
```go

func (s *SugaredLogger) log(level zap.Level, msg interface{}, keysAndValues []interface{})
```

sugar.go
```diff
@@ -0,0 +1,214 @@
...
+
+// WithFields adds structured fields to the logger's context, just like the
+// standard Logger's With method.
+func (s *SugaredLogger) WithFields(fs ...zapcore.Field) *SugaredLogger {
+	return &SugaredLogger{core: s.core.With(fs...)}
+}
+
+// Debug logs a message and some key-value pairs at DebugLevel. Keys and values
+// are treated as they are in the With method.
+func (s *SugaredLogger) Debug(msg interface{}, keysAndValues ...interface{}) {
```
prashantv: Why is `msg` an `interface{}`

sugar.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	return fields
+}
+
+func sweetenMsg(msg interface{}) string {
+	if str, ok := msg.(string); ok {
+		return str
+	}
+	return fmt.Sprint(msg)
+}
```
prashantv: I really dislike this "message is an interface" pattern.
 - You eat an extra allocation on every single log converting the `string` to an `interface{}`
 - Apart from `error`, I don't see many use cases for supporting arbitrary types as messages
 - No other logging library I know allows the message to be an `interface{}`. Either it only takes objects as `interface{}` where all objects are treated the same, or it takes a message `string` and some set of key-value pairs (possibly using `interface{})`.

field.go
```diff
@@ -156,3 +156,53 @@ func Reflect(key string, val interface{}) zapcore.Field {
 func Nest(key string, fields ...zapcore.Field) zapcore.Field {
 	return zapcore.Field{Key: key, Type: zapcore.ObjectMarshalerType, Interface: zapcore.Fields(fields)}
 }
+
+// Any takes a key and an arbitrary value and chooses the best way to represent
+// them as a field, falling back to a reflection-based approach only if
+// necessary.
+func Any(key string, value interface{}) zapcore.Field {
+	switch val := value.(type) {
+	case zapcore.ObjectMarshaler:
```
prashantv: Why no `ArrayMarshaler` here?

field.go
```diff
@@ -156,3 +156,53 @@ func Reflect(key string, val interface{}) zapcore.Field {
...
+		return Uint64(key, uint64(val))
+	case uint16:
+		return Uint64(key, uint64(val))
+	case uint8:
+		return Uint64(key, uint64(val))
+	case uintptr:
+		return Uintptr(key, val)
+	case string:
+		return String(key, val)
+	case error:
```
prashantv: suggest putting this next to `fmt.Stringer` to group interfaces together

sugar.go
```diff
@@ -0,0 +1,214 @@
...
+
+package zap
+
+import (
+	"fmt"
+
+	"go.uber.org/zap/zapcore"
+)
+
+const oddNumberErrMsg = "Passed an odd number of keys and values to SugaredLogger, ignoring last."
```
prashantv: nit: underscore prefix for unexported global which we seem to be using elsewhere in zap.

sugar.go
```diff
@@ -0,0 +1,214 @@
...
+// Fatalf uses fmt.Sprintf to construct a dynamic message and logs it at
+// FatalLevel, then calls os.Exit(1). It doesn't add to the message's
+// structured context.
+func (s *SugaredLogger) Fatalf(template string, args ...interface{}) {
+	if ce := s.core.Check(FatalLevel, fmt.Sprintf(template, args...)); ce != nil {
+		ce.Write()
+	}
+}
+
+func sweetenFields(args []interface{}, errLogger Logger) []zapcore.Field {
```
prashantv: Why is this function not on `*Sugar`. The only field in `Sugar` is a logger, and this function takes that as an argument.

sugar.go
```diff
@@ -0,0 +1,214 @@
...
+	if len(args) == 0 {
+		return nil
+	}
+	if len(args)%2 == 1 {
+		errLogger.DPanic(oddNumberErrMsg, Any("ignored", args[len(args)-1]))
+	}
+
+	fields := make([]zapcore.Field, len(args)/2)
+	for i := range fields {
+		key := sweetenMsg(args[2*i])
```
prashantv: I think we should enforce that keys are strings.

sugar.go
akshayjshah: That breaks the caller skipping.

prashantv: The caller skipping should allow skipping n frames, not just one. It seems a little arbitrary to say a logger can only skip a single call frame.

sugar.go
akshayjshah: Because we don't have sum types?

sugar.go
akshayjshah: I've had a number of very long discussions about this, and I'm generally convinced that there's utility here. We already have a logger that's type-safe, low-allocation, and fast. For the sugared logger, I think there's value in supporting `error` and `fmt.Stringer` in addition to strings. Given the limitations of the type system and reasonable APIs, eating an allocation and accepting `interface{}` seems like a reasonable tradeoff.

sugar.go
akshayjshah: Sure, can move it.

sugar.go
akshayjshah: In order:

* That's true. However, if you're using the sugared logger, I don't think a single allocation is breaking the bank.
* `fmt.Stringer` is the other obvious interface we should support, but yes, that's probably it. `logger.Error(err)` is so common that this seems worthwhile to me. This also makes debug-level logging much easier in development, since you can easily log arbitrary structs.
* Logging in most dynamically typed languages works this way, and that shapes many developers' expectations.

field.go
akshayjshah: Just on oversight, since I was working on this before that branch landed.

field.go
akshayjshah: Sure.

akshayjshah:
In summary: we've had this discussion with a number of people, in a number of different contexts. *For the sugared logger*, I think that the improved ergonomics of accepting `interface{}` where possible are worth the cost. This makes logging errors in production and arbitrary things in development much simpler.

The other reasonable option is `logger.Info(string, map[string]interface{})`, with `zap.Map` or `zap.SugaredFields` offered as a type alias for `map[string]interface{}`. I find the wordiness of this API and the intrusiveness of the zap package quite unappealing.

We can, of course, always just offer both. There's not much code required.

sugar.go
akshayjshah: The problem isn't the number of frames skipped, it's the incongruence between the number of frames skipped when calling `CheckedEntry.Write` and `logger.Info`. Putting another layer of indirection in here dries up the logger, but prevents us from exposing a checking API. I'd intended to expose such an API; until that happens, I'll add the extra indirection back in.

prashantv:
> An intermediate position is logger.Error(string, keysAndVals ...interface{}). I'm not wholly opposed to this, but I do really hate logger.Error(err.Error()) all over the place.

Yeah this is the approach I'd prefer. Agreed that `logger.Error(err.Error())` is not great, but it's something that the compiler will complain about, and would be better done as `logger.Error("some context", "error", err)`.

Supporting `fmt.Stringer` for a message doesn't seem that useful, and I don't think anything other than a string makes sense for field keys.

Using `interface{}` has a few issues:
 - messages can now be `nil`, which seems like a bug.
 - harder to catch places where the message is accidentally replaced with some other random object (e.g., a bad merge)

Are there may other logging libraries that accept the equivalent of `interface{}` for messages and keys and stringify it? The only ones I can think of accept `interface{}` when accepting a list of objects, and they're either all merged using `fmt.Sprint` or use `fmt.Sprintf` with a template.

E.g., This logger looks very similar to log15, which allows passing context as `...interface{}`, but it expects messages to be strings, and that field keys are strings:
https://github.com/inconshreveable/log15/blob/master/format.go#L146

I don't think the value of supporting automatic conversion of any type to strings is worth the potential for issues that the compiler can't help with.

sugar.go
prashantv: * I agree that a single allocation isn't a huge deal, but it is another drawback of this approach.
 * I'm not sure I agree that a `Stringer` is useful. Can you provide use cases from real world code where users are calling `.String()` and passing it as a message right now?
 * There is no single message that is passed as a `interface{}`.
   * logrus supports a list of `interface{}` that it then puts together, which is more similar to the standard library `log` package. This pattern is very different to the SugaredLogger, since a log line like `log.Info("dial failed", hostname, err)` will just append everything together with `logrus` or `log`, but with the `SugaredLogger`, it will treat `hostname` as a key. Considering how differently these work, I'm not sure it's a good comparison.
   * log15 expects messages to be strings, see the [Logger type](https://godoc.org/gopkg.in/inconshreveable/log15.v2#Logger).
   * The `go-kit/log` [Logger](https://godoc.org/github.com/go-kit/kit/log#Logger) doesn't even take a message, it only takes key-value pairs.

sugar.go
prashantv: Ahh, that makes sense. One other option is to have the sugar logger maintain 2 loggers internally, one with +2 skip, and one with +1 skip, and have the checking API use the one with less skip.

Definitely not ideal though. Considering this and @jcorbin's comment on https://github.com/uber-go/zap/pull/247#pullrequestreview-17100761, I wonder if per-call site skip makes sense to consider.

ansel1:
logxi functions look like `Info(s string, args ...interface{})`, and generally require `args` to be balanced pairs of strings/interface{}.  But it handles the special case of an `args` with a length of 1 (a single logged "thing" with no key).  It's a compromise somewhere between `Info(err)` and `Info("error occurred", "error", err)`:

    Info("error occurred", err)  // or...
    Info("user struct", user)

The key for a "bare" arg defaults to an underscore, so the entry ends up looking like:

    Error occurred     _=error message

akshayjshah:
To me, the point under debate here is how type-safe it's necessary for the sugared logger to be, and how much we're willing to sacrifice in ergonomics for that safety. My broad point, which is firmly borne out by the spectrum of existing logging libraries, is that type safety just isn't a priority for typical logging APIs. In the absolute worst case, we'll end up logging a message with a handful of oddly-named fields, which is certainly not the end of the world.

Given that, I don't see why we wouldn't make a best effort to support a wide variety of types. Errors are the most obvious case: while it's obviously feasible for users to call `Error()` themselves, it's needlessly irritating. This is the same reason logging libraries provide formatting helpers, instead of forcing users to call `fmt.Sprintf` directly.

The same is true regarding field names. Sure, we can log a `DPanic` when we encounter a non-string key, but why wouldn't we then make a best effort to also something useful? We're guaranteed that we won't affect any other fields, so I don't see the harm. It's certainly better than dropping a field.

It's also worth remembering that we *already have* a fully type-safe, performance-optimized logger.  Anyone who cares deeply about log site correctness or shaving off a single allocation here or there will already be using that logger. *I* plan on using that logger most of the time. The sugared logger is explicitly intended to provide ergonomics and ease of use. Since the effect of passing problematic messages or keys is (a) minimal, and (b) easily caught in development with `DPanic`, I don't find the safety concerns compelling.

Let's discuss this in person and come to a conclusion tomorrow.

sugar.go
akshayjshah: * Excellent - glad we agree that, in this context, the perf impact of a single `interface{}` conversion isn't something to be concerned about.
* Appeals to data are great. Typically, the onus of proof falls on the minority position - in this case, that's the argument for compile-time safety at log sites. Of course, any quantification of this is difficult because the problem is circular: developers don't explicitly stringify their messages because they overwhelmingly choose logging libraries that do so for them.
* It seems to me that the strong consensus in logging, both in Go and in other languages, is to sacrifice some degree of compile-time type-checking for convenience. This is why `Sprint`-style interfaces are so common in Go. After your (accurate) correction regarding the `log15` API, I know of three libraries that have strongly-typed `string` arguments as messages - `log15`, `logxi`, and `apex/log`. The standard library, `glog`, `logrus`, `go-kit`, `go-logging`, `xlog`, `seelog`, `go-playground/log`, `sasbury/logging`, `logsip`, and `spacelog` are all built around loosely-typed APIs. I'm quite confident that the majority of libraries I don't know about are also loosely-typed. They're loosely-typed because it's relatively cheap to accept `interface{}`s instead of strings, it's convenient, and the consequences of passing un-printable or otherwise problematic values are not particularly dire.

prashantv:
Your main argument for supporting message as an `interface{}` is that a large number of existing loggers allow it. Almost every library you listed has a **significant** difference from `SugaredLogger`, they accept `...interface{}` and stringify all objects and output a space-separated list, for all of the following: [glog](https://godoc.org/github.com/golang/glog#Error), [logrus](https://godoc.org/github.com/sirupsen/logrus#Info), [go-logging](https://godoc.org/github.com/op/go-logging#Logger.Info), [xlog](https://godoc.org/github.com/hlandau/xlog#Logger.Info), [seelog](https://godoc.org/github.com/cihub/seelog#Info), [go-playground/log](https://godoc.org/github.com/go-playground/log#Info), [sasburry/logging](https://godoc.org/github.com/sasbury/logging#Info), [logsip](https://godoc.org/github.com/iamthemuffinman/logsip#Info), [spacelog](https://godoc.org/github.com/spacemonkeygo/spacelog#Logger.Info)

There is no special "message", or "field key" positional argument in any these interfaces, all arguments are treated exactly the same. It's fine to reorder the objects, and still have valid output. The same is definitely not true for the `SugaredLogger` interface. Given the significant difference in how arguments are handled, I don't think it makes any sense to use the existing interfaces as support for the "message should be `interface{}`" argument. The existing interfaces support the idea that all `...interface{}` arguments should be treated the same.

Most of the logging libraries that *do* treat messages and key names separately use a string on the other hand ([log15](https://godoc.org/gopkg.in/inconshreveable/log15.v2#Info), [logxi](https://godoc.org/github.com/mgutz/logxi/v1#Info), [apex/log](https://godoc.org/github.com/apex/log#Logger.WithField), [go-playground/log](https://godoc.org/github.com/go-playground/log#Field).

The [go-kit logger](https://godoc.org/github.com/go-kit/kit/log#Logger) treats keys and values in a similar way to `SugaredLogger`, but it also doesn't have the concept of a message.

> My broad point, which is firmly borne out by the spectrum of existing logging libraries, is that type safety just isn't a priority for typical logging APIs. In the absolute worst case, we'll end up logging a message with a handful of oddly-named fields, which is certainly not the end of the world.

Structured logging is also not a priority based on existing libraries and existing code. However, the API we're debating is intended for structured logging, and this interface is more error-prone for structured logging. Existing logging library interfaces are not error-prone for their use case of logging a list of objects.

akshayjshah:
As I said earlier, I don't think further back-and-forth in these comments is helpful. Let's discuss this in person.

PR #186, opened 2016-11-22
Error handling in output sync and AtomicLevel

jcorbin:
- noticed around recent `AtomicLevel.ServeHTTP` work
- the sync error is probably an unlikely edge case, but nice to have
- some of them were overly pedantic, so this just silences the warning

### DIFF
--- a/http_handler.go
+++ b/http_handler.go
@@ -23,6 +23,7 @@ package zap
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 )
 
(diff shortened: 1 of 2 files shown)


http_handler.go
```diff
@@ -40,32 +41,36 @@ func (lvl AtomicLevel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	enc := json.NewEncoder(w)
-
 	switch r.Method {
 
 	case "GET":
 		current := lvl.Level()
-		enc.Encode(payload{Level: &current})
+		if err := enc.Encode(payload{Level: &current}); err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			io.WriteString(w, fmt.Sprintf("AtomicLevel.ServeHTTP internal error: %v", err))
```
prashantv: we can use `fmt.Fprintf(w, ....)` here

http_handler.go
```diff
@@ -40,32 +41,36 @@ func (lvl AtomicLevel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				Error: "Must specify a logging level.",
+			})
 			return
 		}
-
 		lvl.SetLevel(*req.Level)
-		enc.Encode(req)
+		if err := enc.Encode(req); err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			io.WriteString(w, fmt.Sprintf("AtomicLevel.ServeHTTP internal error: %v", err))
```
prashantv: why do we use `errorResponse` earlier, but not here?

PR #191, opened 2016-11-28
Pool CheckedMessage (free after Write)

jcorbin:


jcorbin:
Initial benchmark against master (at 11e8d8f6b3c9c574462f8278f3f2b7eb56b12e27):
```
BenchmarkCheckedMessage_Chain-8                 20000000                85.7 ns/op           106 B/op          1 allocs/op
BenchmarkCheckedMessage_Chain_sliceLoggers-8    20000000                96.6 ns/op           106 B/op          1 allocs/op
PASS
ok      github.com/uber-go/zap  4.858s
```

And against 9b9bbde4b64099f47f4b5d20787799d02d9d3f41 (this PR) :
```
BenchmarkCheckedMessage_Chain-8                 20000000                73.8 ns/op            42 B/op          0 allocs/op
BenchmarkCheckedMessage_Chain_sliceLoggers-8    20000000                77.1 ns/op            42 B/op          0 allocs/op
PASS
ok      github.com/uber-go/zap  3.194s
```

checked_message.go
```diff
@@ -20,7 +20,13 @@
...
 package zap
 
-import "github.com/uber-go/atomic"
+import (
+	"sync"
+
+	"github.com/uber-go/atomic"
+)
+
+var cmPool sync.Pool
```
akshayjshah: I think that the code below would be a little clearer if you defined a `New` method on the pool that returned `&CheckedMessage{}`.

Also, please underscore-prefix global vars (per the in-house style guide - I've probably been a bit inconsistent in zap, but I'll clean up eventually).

checked_message.go
```diff
@@ -41,6 +47,25 @@ type CheckedMessage struct {
 // NewCheckedMessage constructs a CheckedMessage. It's only intended for use by
 // wrapper libraries, and shouldn't be necessary in application code.
 func NewCheckedMessage(logger Logger, lvl Level, msg string) *CheckedMessage {
+	for {
+		val := cmPool.Get()
+		if val == nil {
+			break
+		}
```
akshayjshah: I don't think that this nil check is necessary (once we have a `New` on `cmPool`); we never return nil CMs to the pool, and we should be able to vigorously exercise these code paths in tests.

checked_message.go
```diff
@@ -87,6 +112,8 @@ func (m *CheckedMessage) Write(fields ...Field) {
 	}
 
 	m.next.Write(fields...)
+	m.lvl, m.next, m.tail = invalidLevel, nil, nil
```
akshayjshah: Not strictly necessary, but a nice safety nonetheless. If we're going to do this, let's pull this logic out into a small `reset` method and also clear out the message and logger instance.

checked_message.go
```diff
@@ -41,6 +47,25 @@ type CheckedMessage struct {
...
+		val := cmPool.Get()
+		if val == nil {
+			break
+		}
+		m := val.(*CheckedMessage)
+		if n := m.used.Load(); n > 1 {
+			// we've already logged (and survived!) a DFatal log in the second
+			// CheckedMessage.Write, so just skip it and move on
+			continue
+		}
```
akshayjshah: I'm not sure that `used` is providing us the ironclad guarantees it used to - it's entirely possible to use a checked message once, keep a reference to it, and concurrently pull it out of the pool in another location. The second reference passes this test, since we've only used the message once, and then we hit line 61 and both callers see the new message. Now the original call site (which is misbehaving and re-using the checked message) is polluted with the message and level of the new call site. Both call sites have references to the same checked message, with a fresh atomic, so it's a race to see who gets to use it.

I don't know of a good pattern to avoid this sort of problem with pooling - it's inherently fiddly and error-prone, which is why we avoided it in the current implementation. I'm inclined to remove `used` altogether and make the documentation for `Write` and `Check` much, much more emphatic. (If you can think of a way to preserve a guarantee of safety, I'm all ears.)

checked_message.go
jcorbin: Well, resetting at least the next and tail pointers are necessary before re-use; I'd rather get them "going" rather than "coming".

Especially now that the lack of atomic obviates the need for the entire pointer-value-stomp under NewCheckedMessage...

jcorbin:
Benchmark update (on my machine):

On master (at 11e8d8f6b3c9c574462f8278f3f2b7eb56b12e27):
```
BenchmarkCheckedMessage_Chain-8                 20000000               101 ns/op             106 B/op          1 allocs/op
BenchmarkCheckedMessage_Chain_sliceLoggers-8    20000000               108 ns/op             106 B/op          1 allocs/op
PASS
ok      github.com/uber-go/zap  4.391s
```

This PR (as of dd7e783a949f8b535e2de6a45408ede885a89981):
```
BenchmarkCheckedMessage_Chain-8                 20000000                82.0 ns/op            42 B/op          0 allocs/op
BenchmarkCheckedMessage_Chain_sliceLoggers-8    20000000                87.2 ns/op            42 B/op          0 allocs/op
PASS
ok      github.com/uber-go/zap  3.548s
```

checked_message.go
```diff
@@ -41,34 +48,25 @@ type CheckedMessage struct {
...
 // wrapper libraries, and shouldn't be necessary in application code.
 func NewCheckedMessage(logger Logger, lvl Level, msg string) *CheckedMessage {
-	return &CheckedMessage{
-		logger: logger,
-		lvl:    lvl,
-		msg:    msg,
+	for {
+		m := _CMPool.Get().(*CheckedMessage)
+		m.logger, m.lvl, m.msg = logger, lvl, msg
+		return m
```
akshayjshah: I don't think we need the enclosing loop here - what is it for?

checked_message.go
```diff
@@ -20,14 +20,21 @@
 
 package zap
 
-import "github.com/uber-go/atomic"
+import "sync"
+
+var _CMPool = sync.Pool{
```
akshayjshah: I *think* that the typical rule for initialisms would make this `_cmPool`. https://github.com/golang/go/wiki/CodeReviewComments#initialisms

checked_message.go
```diff
@@ -41,34 +48,25 @@ type CheckedMessage struct {
...
+// Write logs the pre-checked message with the supplied fields. It will call
+// the underlying level method (Debug, Info, Warn, Error, Panic, and Fatal) for
+// the defined levels; the Log method is only called for unknown logging
+// levels.
 //
-// Write will call the underlying level method (Debug, Info, Warn, Error,
-// Panic, and Fatal) for the defined levels; the Log method is only called for
-// unknown logging levels.
+// It MUST be called at most once, since Write will return the *CheckedMessage
+// to an internal pool for potentially immediate re-use.
```
akshayjshah: Excellent. Perhaps add a scary final sentence explicating the results of re-use - corrupted memory, undefined behavior, race conditions, etc.

Basically, I don't want to assume that all users of the library understand that "returned to an internal pool" means "pending apocalypse and nightmare debugging."

prashantv: IMO, a comment is a good start, but not enough. We can add some validation through a single bool in the CheckedMessage that will give a lot more safety by catching bad behaviour.

- Add a bool `safeToWrite`
- When we get from the pool, set `safeToWrite = true`
- When we `Write`, ensure `safeToWrite == true` and then set `safeToWrite = false` before returning to the pool.

If a user tries to call `Write` twice, and the message hasn't been reused yet, then the check will fail since `safeToWrite` is `false`.

If the `CheckedMessage` is reused, and someone tries to reuse it, we'll have 2 `Write` calls, and it'll fail again similar to the above case.

The bool doesn't need to be atomic, since any concurrent access will cause a data race which is what should happen, there's no reason for correct code to concurrently use use the CheckedMessage.

It should be pretty much free, since it's a single bool. We could probably do something similar with an int too, but I think the bool approach is simpler.

prashantv: Yep it is very similar -- it turns out `used` didn't need to be atomic, since valid use wouldn't have allowed concurrent access to the value.

akshayjshah: I'm fine with adding a best-effort approach like this; let's be sure to leave some comments in the checking code that clearly indicate that it's best-effort and shouldn't be relied upon.

jcorbin:
master @ 11e8d8f6b3c9c574462f8278f3f2b7eb56b12e27:
```
BenchmarkCheckedMessage_Chain-8                 20000000               107 ns/op             106 B/op          1 allocs/op
BenchmarkCheckedMessage_Chain_sliceLoggers-8    10000000               107 ns/op             106 B/op          1 allocs/op
PASS
ok      github.com/uber-go/zap  3.487s
master...origin/master
```

latest PR e98d7d0070efde8ea6f3f5ea1b118b1db94f987f:
```
BenchmarkCheckedMessage_Chain-8                 20000000                75.4 ns/op            42 B/op          0 allocs/op
BenchmarkCheckedMessage_Chain_sliceLoggers-8    20000000                81.9 ns/op            42 B/op          0 allocs/op
PASS
ok      github.com/uber-go/zap  3.307s
```

checked_message_test.go
```diff
@@ -47,17 +47,18 @@ func TestJSONLoggerCheck(t *testing.T) {
...
-		cm := logger.Check(InfoLevel, "single-use")
-		cm.Write() // ok
-		cm.Write() // first re-use logs error
-		cm.Write() // second re-use is silently ignored
-		assert.Equal(t, expected, buf.Lines(), "Expected re-using a CheckedMessage to log an error.")
+func TestCheckedMessageUnsafeWrite(t *testing.T) {
+	withJSONLogger(t, opts(InfoLevel), func(logger Logger, buf *testBuffer) {
+		stub := stubExit()
+		defer stub.Unstub()
+		cm := logger.Check(InfoLevel, "bob lob law blog")
```
prashantv: it's bob loblaw's law blog!
http://arresteddevelopment.wikia.com/wiki/Bob_Loblaw

checked_message.go
```diff
@@ -20,16 +20,24 @@
...
 
 // A CheckedMessage is the result of a call to Logger.Check, which allows
 // especially performance-sensitive applications to avoid allocations for disabled
 // or heavily sampled log levels.
 type CheckedMessage struct {
-	logger Logger
-	used   atomic.Uint32
-	lvl    Level
-	msg    string
+	safeToWrite bool
```
prashantv: why reorder the fields? if `safeToWrite` is next to `Level`, the struct should take up less space.

PR #194, opened 2016-11-30
WIP: Fix time across composed log sites

jcorbin:
Currently with a zap.Tee logger:
- each sub logger calls time.Now itself at moderately different times
- so does every (*CheckedMessage).Write in a chain

The approach here is to pass a single `time.Time` along to `Logger.Log`.

I can easily separate the entry refactor work (it's already in a separate branch `imp_entry`) if desired.


checked_message.go
```diff
@@ -83,10 +91,12 @@ func (m *CheckedMessage) Write(fields ...Field) {
 	case FatalLevel:
 		m.logger.Fatal(m.msg, fields...)
 	default:
-		m.logger.Log(m.lvl, m.msg, fields...)
+		m.logger.Log(t, m.lvl, m.msg, fields...)
```
akshayjshah: Can we leave a TODO to use `logger.Log` and propagate time in the above cases too? Probably best to address that when we make it easier to write new loggers (i.e., when we're 100% sure that all `logger.Panic` methods just call `logger.Log` and the `panic`).

jcorbin:
Tabling this PR for now, as it has reveal need for an interface reworking to manage the emerging complexity of logger / message composition.

akshayjshah:
This is now part of #201, right? If so, can we close this one?

jcorbin:
Right, this one is superceded.

Issue #70, opened 2016-05-27
Add a Sentry hook


akshayjshah:
This isn't a blocking issue for 1.0, but we should provide a Sentry hook (either in a subpackage or a completely separate repo). It'll be useful for internal users and as an example of hook-based external integrations.

belak:
I'd be happy to work on this, but it looks like this would need to be added using `hook` and/or `Option`, both of which aren't public.

sectioneight:
What's the priority on this? I've heard rumors we're moving off Sentry internally

akshayjshah:
Priority here depends entirely on whether we're planning to migrate our internal midi-repo to zap. We integrate with Sentry via a logger hook, so we'd need a similar mechanism for zap.

Even if we don't actually ship a Sentry hook, it'd be good to rough one out - that way we're sure that the public API is sufficient to support this kind of use case. For external users, integrating with Sentry/Rollbar/Bugsnag/etc is often important.

akshayjshah:
Closing this, since we've decided to go in another direction internally.

I'd strongly prefer not to accumulate plugins for every exception/error reporting service (Sentry, Bugsnag, Rollbar, etc.) in zap - these should ship as external hooks or wrappers.

belak:
We recently cleaned up our internal logging and decided to start using raven-go directly anyway, as it gives us much more control than a logging plugin, particularly because of stack traces and other complications.

Anyway, Hook is a thing now anyway, so it should be simple to implement this hook if a user wants it.

PR #201, opened 2016-12-06
Refactor zap around Facility

jcorbin:
Current state (as of 2016-12-23: done / in review


akshayjshah:
Haven't traced the code flow in detail, but `Entry` definitely escapes to the heap; let's make sure we pool it and have a very clear story for where it's returned to the pool (preferably one that's entirely controlled by the Logger, so implementors of our various interfaces don't have to get pooling right).

entry.go
```diff
@@ -20,43 +20,34 @@
...
+// XXX compat with *CheckedMessage, drop?
+func (e *Entry) OK() bool {
+	return e != nil
 }
 
-// Fields returns a mutable reference to the entry's accumulated context.
-func (e *Entry) Fields() KeyValue {
-	return e.enc
+// Write writes the entry any logger reference stored. This only works if the
+// Entry was returned by Logger.Check.
```
jcorbin: Will go the route of a CheckedEntry that embeds a Facility and Entry.

facility.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// Facility is a destination for log entries. It can have pervasive fields
+// added with With(). Log should not be called if Enabled returns false.
+type Facility interface {
+	With(...Field) Facility
+	Enabled(Entry) bool
+	Log(Entry, ...Field)
+	// XXX idea on how we could restore internal enoding error repuorting:
+	//     SetErrorOutput(ws WriteSyncer)
```
jcorbin: Great point above, all we need to do is make it `Log(Entry, ...Field) error`, then reporting isn't the facilities's concern.

facility.go
```diff
@@ -0,0 +1,86 @@
...
+
+type ioFacility struct {
+	Encoder
+	Output WriteSyncer
+}
+
+func newIOFacility(enc Encoder, out WriteSyncer) *ioFacility {
+	if out == nil {
+		out = newLockedWriteSyncer(os.Stdout)
+	}
```
jcorbin: Fair point, may be best moved to `func WriterFacility(Encoder, io.Writer) Facility`

jcorbin:
Will introduce `*CheckedEntry` which we can pool and/or verify its heap escape velocity; then `Entry` will only ever be passed and received by value.

jcorbin:
I'll internalize the `Logger` type to be just `logger` again, and then re-introduce the `Logger` type as an interface.

After that, the next big rocks to be pushed:
- [ ] Entry => CheckedEntry split
- [ ] Drop the LevelEnabler from logger; demote it from `Option` status
- [ ] At least use LevelEnabler in most concrete Facility types; take LevelEnabler as part of the Facility contract if necessary for composition

As you can see in this work: https://github.com/uber-go/zap/pull/201/commits/f764dcfe4fd14f40dd0960d48a7debf1deeb05af#diff-133c628d7af54d3a9d7bf19081e3ed57R29 , I already bumped up against this need with the tee-spy tests.

jcorbin:
Updated the lead comment with current status notes.

jcorbin:
Past tests, ready for review; some high order bits:

### `Facility` could be further broken up

Leaf `Facility` implementations (e.g. `spyFacility` and `ioFacility`) only really need to implement `WriteEntry`; they simply embed a `LevelEnabler` with boilerplate `Log` and `Check` implementations. It may be worth it to factor out an `EntryWriter` interface, and then implement say a `LeveledEntryWriter` type that would be used by constructor funcs (e.g. `spy.New` and `zap.WriterFacility`).

This would benefit from moving all facility parts into a sub-package, so that we can stop confusing "log users" with types for "log implementors".

#### Naming

As always, the hard problem. I'm especially unhappy with the rampant use of `zap.WriterFacility`

#### Options

Maybe things get better if we re-introduce options like `zap.Output` now in terms of "it sets the facility using WriterFacility".  These could even be per encoding, e.g. maybe `zap.JSONOutput(io.Writer, LevelEnabler, ...JSONOption)`; corollary, maybe we should chase the text v json unification, and make that `OutputOption`, and then we could have `LevelEnabler` implement it, and do away with all the "default level passing" noise.

### What remains of `Hooks`

They still exist, but are now value-oriented and unexported.

Going forward, we plan to claw `spy.Facility` back into core (current proposed name being `zap.Observer`) for "hooks that only look and the entry, but don't change it".  This would get them field visibility also (e.g. this would make it possible to implement a "sentry hook" per #70 trivially, modulo performance concerns around the `map[string]interface{}` pattern).

Support for mutative external hooks has been deemed "not supported", and won't be coming back.

We may decide to go further, special casing caller/stack adding in the `logger` implementation.

### Dropped `Logger.Log`

Since composition now happens at the Facility level, rather than the Logger level, we no longer have a supported use case for "programatic level" passing; speak up now if any valid use cases occur to you dear Reviewers.

### Dropped `zap.Meta`

Implementing `zap.Logger` is now considered an edge case; the primary path should be implementing a `zap.Facility`. To that end `Meta` has gone away, subsumed by the unexported concrete `logger`.

logger.go
```diff
@@ -48,91 +49,137 @@ type Logger interface {
...
+	development bool
+	hooks       []hook
+	errorOutput WriteSyncer
 }
 
-func (log *logger) With(fields ...Field) Logger {
-	clone := &logger{
-		Meta: log.Meta.Clone(),
+// New returns a new logger with sensible defaults: logging at InfoLevel,
+// development mode off, errors writtten to stdand error, and logs JSON encoded
```
rogpeppe: errors written to standard error

spy/facility.go
```diff
@@ -0,0 +1,115 @@
...
+	var logs []Log
+	s.Lock()
+	logs = append(logs, s.logs...)
+	s.Unlock()
+	return logs
+}
+
+// Facility implements a zap.Facility that captures Log records.
+type Facility struct {
+	sync.Mutex
```
rogpeppe: I don't believe this should be embedded (mutexes should almost always be unexported), and it's not used anyway.

spy/facility.go
```diff
@@ -0,0 +1,115 @@
...
+	context []zap.Field
+}
+
+// With creates a sub spy facility so that all log records recorded under it
+// have the given fields attached.
+func (sf *Facility) With(fields ...zap.Field) zap.Facility {
+	return &Facility{
+		LevelEnabler: sf.LevelEnabler,
+		sink:         sf.sink,
+		context:      append(sf.context, fields...),
```
rogpeppe: This is dangerous and can allow With to overwrite fields from another
facility.
For example, consider:

	f, sink := spy.New(zap.DebugLevel)
	l1 := zap.New(f).With(zap.Int("l1 both", 1)).With(zap.Int("l1 both 2", 1))
	l2 := l1.With(zap.Int("l2 both", 2))
	l3 := l2.With(zap.Int("l3 only", 3))
	l4 := l2.With(zap.Int("l4 only", 4))
	l3.Info("l3")
	l4.Info("l4")

The l3.Info log will print the "l4 only" field because the context
has been overwritten by the append to the shared slice.

I think you probably want:

      context:      append(sf.context[0:len(sf.context):len(sf.context)], fields...),

to force the whole slice to be copied each time.

marshaler_example_test.go
```diff
@@ -62,7 +62,7 @@ func ExampleMarshaler() {
 		},
 	}
 
-	logger := zap.New(zap.NewJSONEncoder(zap.NoTime()))
+	logger := zap.New(zap.WriterFacility(zap.NewJSONEncoder(zap.NoTime()), nil, zap.InfoLevel))
```
rogpeppe: IMHO this doesn't really seem like an improvement to the API.

How about leaving zap.New the same as it is, as that's what most clients will probably want, and add a new function that can create a logger from a facility when more control is required?

rogpeppe:
Apologies for the drive-by review, but I couldn't resist. I've left some comments, and I'd tend agree with @akshayjshah about the Facility name - it doesn't really suggest anything. I wonder whether "Writer" might be better - this is the basic type which implements log writing after all.

I'd like to see better docs for the Facility type - from the interface, the distinction between Log and Write is really not clear - do we really need both? I'd suggest lose the ... arguments too and using explicit slice types, as variadic arguments are a convenience for direct callers and AFAICS Facility instances should only ever be used by middleware.

spy/facility.go
```diff
@@ -0,0 +1,115 @@
...
+// A Log is an encoding-agnostic representation of a log message.
+type Log struct {
+	Level  zap.Level
+	Msg    string
+	Fields []zap.Field
+}
+
+// A Sink stores Log structs.
+type Sink struct {
+	sync.Mutex
```
rogpeppe: I'd prefer it if this wasn't embedded. I can't see that it's ever a good idea for external callers to be able to lock the Sink's mutex.

spy/facility.go
jcorbin: Great point, not sure how this stayed in ;-)

jcorbin:
Great points @rogpeppe, on the `zap.New` ergonomics front, we totally agree and are planning to something along those lines combined with cleaning up our current package sprawl into some sort of sub-package structure; realize that this PR is against the dev branch, not master, so we may not get all the way there in this PR.

On the "what even is facility, isn't it just a writer" front, also completely agreed: this is essentially what I was saying under my "Facility could be further broken up" github comment above ;-)

I'll walk back the variadics and drop that pesky mutex straightaway; waiting for Akshay tho before moving in any direction on the naming / facility-sizing / ergonomics front tho.

marshaler_example_test.go
jcorbin: Agreed; this PR is against dev, and takes us to a temporary "uncanny valley" on API ergonomics. Either this PR will grow in scope to move all the chairs and rebuild the public API layer, or we'll do that in a Soon ™ to come additional dev-based PR.

akshayjshah:
Interim update - I'm actually looking at this and staging comments. Please don't tack on any more commits! (Yes, I realize that I explicitly asked for this glacier of a PR 😿  ).

zwrap/sample_test.go
```diff
@@ -185,20 +178,21 @@ func TestSamplerCheck(t *testing.T) {
...
-		for i := 0; i < 5; i++ {
-			if cm := sampler.Check(level, "sample"); assert.True(t, cm.OK(), "expected %v level to always be OK", level) {
-				cm.Write(zap.Int("iter", i))
-			}
-		}
-
-		assert.Equal(t, []spy.Log(nil), sink.Logs(), "Unexpected output when sampling with Check.")
-	}
-}
+// TODO: restore this test, now that panic and fatal are actually terminal
```
jcorbin: I can test panic easily enough, but fatal would require access to the unexpected exit stubbing; suggestions?

entry.go
```diff
@@ -20,7 +20,48 @@
...
-import "time"
+import (
+	"fmt"
+	"sync"
+	"time"
+)
+
+var _cePool = sync.Pool{
+	New: func() interface{} {
+		return &CheckedEntry{}
```
jcorbin: Sure, but my comment was really asking: we should benchmark and see if its worth it to special case space for 1, 2, or 3 facilities out side of the slice so that we can support arbitrary. I.E. using a static `[2]Facility` plus an `int` field (or nil checks) to use it before the slice.

jcorbin:
Addressed feedback and rebased on latest dev.

jcorbin:
Aside from minor changes, this last push contained:
- dropped hook support: addCaller and addStack are now in the canon (please consider the new TODO inside the logger struct)
- nuked `WriterFacility` default to stdout
- made `WriterFacility` take a `WriteSyncer`
- axed `Facility.Log`; just have `logger.Log` use `Facility.Check` all the time
  - this compounded for another win around the new addCaller / addStack shape: we avoid collecting if the `Log` is actually a noop
- brought back `safeToWrite` inside of `CheckedEntry`

I punted pre-allocating or carving out any static `Facility` space inside `CheckedEntry` for future work.

entry.go
```diff
@@ -32,10 +69,117 @@ type Entry struct {
...
+	should      CheckWriteAction
+	facs        []Facility
+	// TODO: we could provide static spac for the first N facilities to avoid
+	// allocations in common cases
+}
+
+// Write writes the entry to any Facility references stored, returning any
+// errors, and returns the CheckedEntry reference to a pool for immediate
+// re-use.
+func (ce *CheckedEntry) Write(fields ...Field) error {
```
jcorbin: I want to point out to reviewers this subtle semantic change:
- prior `(*CheckedMessage).Write` was void return, it internally noted any error
- new `(*CheckedEntry).Write` returns the error

Is this okay or should I restore the prior behavior that users of `Logger.Check().Write()` don't have to handle errors?

akshayjshah: I'd prefer to restore the previous behavior. To me, it's unreasonable to expect users to handle errors from the logging library.

We can and should pass around errors internally, but I think we should hide them from the user.

akshayjshah: However, let's take this up in a follow-on PR. This one's gone on long enough :)

akshayjshah:
@jcorbin - I've tacked on a few commits that clean up some typos and fix up a few uncontroversial bits and pieces. Review at your leisure.

Once tests pass, I'm going to merge this PR so that we can move on. Let's address any remaining issues in follow-on PRs (particularly the issue you raised about `CheckedEntry.Write` returning errors).

PR #205, opened 2016-12-07
Add filter logger

shafreeck:
This PR adds a filter logger which is inspired by `Tee`， The `Filter` function returns a logger with multiple loggers and send log to the loggers that matching specific level.

---

Thanks for your contribution!

### DIFF
--- a/filter_bench_test.go
+++ b/filter_bench_test.go
@@ -1,3 +1,23 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
 package zap
 
 import (
--- a/filter_bench_test.go
+++ b/filter_bench_test.go
@@ -1,0 +1,32 @@
+package zap
+
+import (
+	"testing"
+)
+
+func BenchmarkFilter(b *testing.B) {
+	log := New(NewTextEncoder(), DiscardOutput, DebugLevel)
+
+	logger := Filter(
+		LeveledLogger{DebugLevel, log},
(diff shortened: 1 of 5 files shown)


jcorbin:
Awesome, glad to see others thinking in the same vein about how to compose zap Loggers :-)

Supporting uses cases like this is what the `LevelEnabler` is intended for (example below); if you need to get something done now, using it will be a more maintainable path than a filtered logger given our ongoing push to refactor zap (ala #201).

```
//// Outside zap

package main

// justLevel returns a LevelEnabler that only enables one level; tbd whether
// this convenience also belongs in zap or not.
func justLevel(lvl zap.Level) LevelEnabler {
    return zap.LevelEnablerFunc(func(l zap.Level) bool { reurn l == lvl })
}

log := zap.Tee(
    zap.New(..., justLevel(zap.InfoLevel), // this one gets info logs
    zap.New(..., justLevel(zap.WarnLevel), // this one gets warn logs
    zap.New(..., zap.ErrorLevel, // this one gets warn logs
)

//// TODO: we're missing this in zap right now, I've been wanting to add it:

package zap

// LevelEnablerFunc is a convenient way to implement LevelEnabler around an
// anonymous function. It is also a valid Option to pass to a logger.
type LevelEnablerFunc func(Level) bool

// This allows an LevelEnablerFunc to be used as an option.
func (f LevelEnablerFunc) apply(m *Meta) { m.Enabler = f }

// Enabled calls the wrapped function.
func (f LevelEnablerFunc) Enabled(lvl Level) bool { return f(lvl) }
```

shafreeck:
Combining `Tee` with `LevelEnabler` is cool. I have considered it before choosing to implement the `Filter`. The only problem of `Tee` is it may waste some non necessary calls. 

I think `Filter` and `Tee` have different designing goals. As there names imply, `Tee` writes an entry to all subloggers(yeah, the sub loggers can filter it by level, but this has costs to pay) and `Filter` writes only when its level matched. 

As a user, I may prefer the `Filter` way even zap don't supply it natively, after all, we choose zap for its outstanding performance.

prashantv:
Thanks for the contribution @shafreeck!

Can we compare the performance of `Filter` compared to `Tee` + `LevelEnabler`?

I want to make sure that `Filter` has a reasonable performance benefit over the other approach.

shafreeck:
@prashantv 

I added a benchmark for filter. You can run it with `go test -bench Filter`.

The result I got here is:
```
Usage of /tmp/go-build260380227/github.com/uber-go/zap/_test/zap.test:
PASS
BenchmarkFilter-2                                 300000              4008 ns/op
BenchmarkFilterUsingTeeWithLevelEnabler-2         100000             15552 ns/op
ok      github.com/uber-go/zap  3.205s
```

You can see the `Tee with LevelEnabler` cost about 4 times time than `Filter`, this may because I created 4 loggers and `Tee` would call every each and `Filter` just call one of them.

filter_bench_test.go
```diff
@@ -0,0 +1,52 @@
...
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		logger.Error("filter")
+	}
+}
+
+func BenchmarkFilterUsingTeeWithLevelEnabler(b *testing.B) {
+	log := New(NewTextEncoder(), DiscardOutput, DebugLevel)
+
+	logger := Tee(log, log, log, log)
```
akshayjshah: Thanks for adding this benchmark, @shafreeck. However, this isn't setting up a fair comparison - the `FilterLogger` only writes out each message once, while this logger writes each message four times. Of course, it's 4x slower.

Try using the approach @jcorbin outlined, which will more closely match your `FilterLogger` benchmark - each message will be written to only one of the loggers.

shafreeck: My fault. I forgot the LevelEnabler

shafreeck:
@akshayjshah 

Now, I got the result. There is indeed not too much difference.

```
Usage of /tmp/go-build217266664/github.com/uber-go/zap/_test/zap.test:
PASS
BenchmarkFilter-2                                 300000              4185 ns/op
BenchmarkFilterUsingTeeWithLevelEnabler-2         500000              4387 ns/op
ok      github.com/uber-go/zap  3.772s
```

There is no need to add `Filter` to confuse people.  I will close this PR.

BTW: Looking forward to add `LevelEnablerFunc `

Issue #203, opened 2016-12-06
Rename project to go.uber.org/zap


breerly:
I've setup zap on go.uber.org, so that we can take advantage of this import path.

I'd like to get ahead of this breaking change before we 1.0, and before YARPC takes on zap as a hard-dependency.

akshayjshah:
A+, let's do this before 1.0.

akshayjshah:
Landed on the dev branch in #206, which will take effect as part of the 1.0 release.

PR #206, opened 2016-12-08
Move zap and atomic to go.uber.org and update vendor

peter-edge:
https://github.com/uber-go/zap/issues/203

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ go:
   - 1.6
   - 1.7
   - tip
+go_import_path: go.uber.org/zap
 env:
   global:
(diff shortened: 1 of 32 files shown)


akshayjshah:
Cool - thanks for the PR, Peter. I'm happy to land this, but I'd like to do so on the `dev` branch instead of master. When we're ready for a v1 release candidate, we can merge dev into master. (Basically, this library has accumulated a surprising number of users, and I'd like to avoid churning the API and import paths before v1.)

akshayjshah:
Once you've signed the CLA (or talk to the internal folks who manage the whitelist), merge away.

peter-edge:
Whitelisted!

PR #208, opened 2016-12-09
Add the LevelEnablerFunc and test case

arthurkiller:
This PR added the LevelEnablerFunc to give a convenient way creatng a LevelEnabler.
I have tried to give the test case to make sure it could be covered
#205 #201 
Hope it can help!

### DIFF
--- a/level_test.go
+++ b/level_test.go
@@ -21,10 +21,24 @@
 package zap
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
+
+func TestLevelEnablerFunc(t *testing.T) {
+	//testing for the level enabler, I will generate a logger with the given
+	//level and it should be print out
+	opts := []Option{Fields(Int("foo", 42)), LevelEnablerFunc(func(l Level) bool { return l == DebugLevel })}
+	withJSONLogger(t, opts, func(log Logger, buf *testBuffer) {
+		log.Debug("@debug", Int("logger", 0))
+		log.Info("@info", Int("logger", 0))
+		assert.Equal(t, []string{
+			`{"level":"debug","msg":"@debug","foo":42,"logger":0}`,
+		}, strings.Split(buf.Stripped(), "\n"))
+	})
+}
 
 func TestLevelString(t *testing.T) {
 	tests := map[Level]string{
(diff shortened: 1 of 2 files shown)


level_test.go
```diff
@@ -21,11 +21,25 @@
...
+func TestLevelEnablerFunc(t *testing.T) {
+	//testing for the level enabler, I will generate a logger with the given
+	//level and it should be print out
+	opts := []Option{Fields(Int("foo", 42)), LevelEnablerFunc(func(l Level) bool { return l == DebugLevel })}
+	withJSONLogger(t, opts, func(log Logger, buf *testBuffer) {
+		log.Debug("@debug", Int("logger", 0))
+		log.Info("@info", Int("logger", 0))
+		assert.Equal(t, []string{
+			`{"level":"debug","msg":"@debug","foo":42,"logger":0}`,
+		}, strings.Split(buf.Stripped(), "\n"))
```
jcorbin: just use `buf.Lines()` which does the split for you

level_test.go
```diff
@@ -21,11 +21,25 @@
...
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
+func TestLevelEnablerFunc(t *testing.T) {
+	//testing for the level enabler, I will generate a logger with the given
```
jcorbin: None of the other tests have comments like these; we should either brush up the grammar and formatting, or drop it (my preference is for the latter).

arthurkiller:
I have removed the comment and sorry for breaking your format dude lol

rkerno:
Guys, just thinking, rather than introduce the cost of a function call and associated logic to determine whether the level will be processed by the encoder, why not just use a bit mask that has a bit associated with each level and then AND to the user's specified log level to make the determination?

arthurkiller:
Could you please give a more specific implementation? 
In my opinion this is the most cost less way to added a new feature based on this project.

akshayjshah:
Hi @rkerno - that would certainly be more efficient, but it's quite a bit less flexible. We'd like the `LevelEnabler` interface to allow not only level-based checks, but also more exotic use cases. For an example, see the sampling wrapper.

PR #210, opened 2016-12-13
Test and benchmark reflection-based []int and []string encoding

phungs:
These tests show the current int and string array performance based off the Master branch.  We will run these tests again in the JSON Array support PR.

### DIFF
--- a/json_encoder_test.go
+++ b/json_encoder_test.go
@@ -140,6 +140,11 @@ func TestJSONEncoderFields(t *testing.T) {
 		{"marshaler", `"k":{}`, func(e Encoder) {
 			assert.Error(t, e.AddMarshaler("k", loggable{false}), "Expected an error calling MarshalLog.")
 		}},
+		{"ints", `"k":[1,2,3,4,5,6,7,8,9,10]`, func(e Encoder) { e.AddObject("k", []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}) }},
+		{"strings", `"k":["bar 1","bar 2","bar 3","bar 4","bar 5","bar 6","bar 7","bar 8","bar 9","bar 10"]`,
+			func(e Encoder) {
+				e.AddObject("k", []string{"bar 1", "bar 2", "bar 3", "bar 4", "bar 5", "bar 6", "bar 7", "bar 8", "bar 9", "bar 10"})
+			}},
 		{"arbitrary object", `"k":{"loggable":"yes"}`, func(e Encoder) {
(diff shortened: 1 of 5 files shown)


json_encoder_test.go
```diff
@@ -140,6 +140,11 @@ func TestJSONEncoderFields(t *testing.T) {
 		{"marshaler", `"k":{}`, func(e Encoder) {
 			assert.Error(t, e.AddMarshaler("k", loggable{false}), "Expected an error calling MarshalLog.")
 		}},
+		{"ints", `"k":[1,2,3,4,5,6,7,8,9,10]`, func(e Encoder) { e.AddObject("k", []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}) }},
+		{"strings", `"k":["bar 1","bar 2","bar 3","bar 4","bar 5","bar 6","bar 7","bar 8","bar 9","bar 10"]`,
+			func(e Encoder) {
+				e.AddObject("k", []string{"bar 1", "bar 2", "bar 3", "bar 4", "bar 5", "bar 6", "bar 7", "bar 8", "bar 9", "bar 10"})
+			}},
```
akshayjshah: These tests exercise the right code, but I don't see the value in having the test data be so verbose. Could we use slices with only two or three elements?

text_encoder_test.go
```diff
@@ -87,6 +87,11 @@ func TestTextEncoderFields(t *testing.T) {
 		{"marshaler", "k={}", func(e Encoder) {
 			assert.Error(t, e.AddMarshaler("k", loggable{false}), "Expected an error calling MarshalLog.")
 		}},
+		{"ints", "k=[1 2 3 4 5 6 7 8 9 10]", func(e Encoder) { e.AddObject("k", []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}) }},
+		{"strings", `k=[bar 1 bar 2 bar 3 bar 4 bar 5 bar 6 bar 7 bar 8 bar 9 bar 10]`,
+			func(e Encoder) {
+				e.AddObject("k", []string{"bar 1", "bar 2", "bar 3", "bar 4", "bar 5", "bar 6", "bar 7", "bar 8", "bar 9", "bar 10"})
+			}},
```
akshayjshah: Likewise here.

json_encoder_test.go
phungs: Good point. Done.

text_encoder_test.go
phungs: Done.

Issue #189, opened 2016-11-24
TextEncoder format is difficult to parse


rogpeppe:
The TextEncoder doesn't quote field values, which means that
if fields contain spaces or equal characters, it's not possible in general to
say where the field boundaries are.

Also, because it just uses fmt.Sprintf("%+v") to format objects,
information will be lost if the objects contain pointers or have custom
JSON marshaling methods.

The right answer might be to encode to JSON anyway, but post-process
into a more human-readable format - it would be nice if the
TextEncoder output contained all the info that would be in the usual
JSONEncoder output.

akshayjshah:
Agreed - the current `TextEncoder` is still pretty rough. I'm a bit leery of using `MarshalJSON` if the output isn't actually JSON, but we should certainly use `MarshalText` where available.

What are your thoughts on multiline logging output? Internally, we only use the text output for human consumption (usually tee'd to standard out in development and testing), so something YAML-ish is appealing.
```
[I] 1970-01-01T12:00:00Z This is an info message.
     "foo": "bar"
     "baz": {"complex": "object"}
     "someList": ["a", "b", "c"]
```

This is rather difficult to machine-parse, but is nice for humans. It also makes it easier to format stack traces and other complex fields.

rogpeppe:
> I'm a bit leery of using MarshalJSON if the output isn't actually JSON

I know what you mean, but I think it might be for the best, otherwise
you are never going to get information parity between the different
encoder outputs, because many types will implement MarshalJSON
but not MarshalText, and vice versa, and I think that having the
text output represent the actual information that's going into the
log is probably quite important.

I'd (for example) to be able to post-process a JSONEncoder log output
back into a readable format as if it had been printed by TextEncoder directly,
but couldn't do that if they print different things.

As for multi-line output, it would be good experiment with a few different formats
on some real logs to see how they look. It may end up looking nice, or perhaps
the extra lines will make for a log that harder to read because of all the extra
vertical space used.

tesujimath:
I'd like to vote against multi-line output, at least unless there's always a single-line output available.  Grepping through a logfile is incredibly useful, and hopeless for multi-line output.

Also, TextEncoder is nice for human legible output as it is.  I suggest that any attempt to make it reversible is missing the point, and you should be using JSON encoder.

akshayjshah:
Addressed in #267.

akshayjshah:
Landed on dev.

PR #211, opened 2016-12-13
Support []int and []string in JSON encoder

phungs:
I've implemented zap.Ints and zap.Strings and submitting this PR for discussion around API and output.

For API, see examples in logger_bench_test.go.

For output, the JSON and Text encoder outputs are equal to the current output when using zap.Object( []int / []string ).  See json_encoder_test.go and text_encoder_test.go.


phungs:
Comparison of the json_array_pre branch (same as master) vs this branch using -benchtime 500ms, -cpu 1 and 20-35 iterations for each code branch, done before the rebase to json_array_pre.

It appears the new branch is generally faster, but I would take that with a grain of salt.  I would say only the Ints and Strings array tests are much faster, since they have a very wide delta (i.e >= 30%) and both bounds (±) are small/med (i.e. <= ~10%).

The other tests probably perform the same.  For tests where the delta and both bounds are small (i.e <= 5%) it's probably a wash/benchmark noise since the changes in this branch didn't affect those benchmarks.  And for a test like StringerField where the bounds are 12% and 3% and the delta is -9.27%, I would say it's inconclusive.

```
$ benchstat bm_json_array_pre_stat_cpu1_1 bm_json_array_obj_stat_cpu1_1
name                               old time/op    new time/op    delta
CheckedMessage_Chain                  238ns ± 5%     225ns ± 3%   -5.60%  (p=0.000 n=32+18)
CheckedMessage_Chain_sliceLoggers     241ns ± 5%     232ns ± 4%   -3.94%  (p=0.000 n=33+20)
JSONLogMarshalerFunc                  118ns ± 6%     117ns ± 6%     ~     (p=0.075 n=32+20)
ZapJSONInts                          1.15µs ± 4%    0.51µs ± 5%  -55.80%  (p=0.000 n=31+20)
ZapJSONStrings                       1.68µs ± 9%    0.56µs ± 3%  -66.49%  (p=0.000 n=30+20)
ZapJSON                               650ns ± 7%     637ns ± 2%   -1.97%  (p=0.026 n=34+19)
StandardJSON                         6.57µs ± 9%    5.91µs ± 2%  -10.02%  (p=0.000 n=33+19)
NullLogMarshalerFunc                 44.3ns ± 9%    42.5ns ± 4%   -4.12%  (p=0.000 n=32+20)
ZapNull                              39.5ns ± 7%    38.6ns ± 5%   -2.26%  (p=0.008 n=32+20)
NoContext                             545ns ±10%     536ns ± 4%     ~     (p=0.220 n=34+20)
BoolField                             652ns ± 6%     628ns ± 4%   -3.67%  (p=0.000 n=34+20)
Float64Field                          815ns ± 7%     768ns ± 2%   -5.74%  (p=0.000 n=33+15)
IntField                              669ns ± 6%     642ns ± 2%   -4.01%  (p=0.000 n=35+20)
Int64Field                            699ns ± 9%     644ns ± 1%   -7.93%  (p=0.000 n=34+17)
StringField                           667ns ± 7%     635ns ± 3%   -4.75%  (p=0.000 n=35+19)
StringerField                         748ns ±12%     679ns ± 3%   -9.27%  (p=0.000 n=35+20)
TimeField                             743ns ± 9%     675ns ± 2%   -9.26%  (p=0.000 n=35+17)
DurationField                         693ns ± 5%     664ns ± 2%   -4.10%  (p=0.000 n=35+19)
ErrorField                            682ns ± 6%     659ns ± 3%   -3.40%  (p=0.000 n=33+19)
StackField                           8.38µs ± 5%    8.12µs ± 2%   -3.08%  (p=0.000 n=35+19)
MarshalerField                        876ns ± 4%     851ns ± 3%   -2.87%  (p=0.000 n=35+20)
**
IntsField                            1.55µs ± 8%    0.96µs ± 7%  -38.26%  (p=0.000 n=35+20)
StringsField                         2.07µs ± 7%    0.98µs ± 3%  -52.62%  (p=0.000 n=35+19)
**
ObjectField                          2.51µs ± 4%    2.47µs ± 3%   -1.87%  (p=0.000 n=34+20)
AddCallerHook                        1.51µs ± 4%    1.45µs ± 2%   -3.88%  (p=0.000 n=33+19)
10Fields                             1.40µs ± 4%    1.39µs ± 2%     ~     (p=0.359 n=31+20)
100Fields                            6.72µs ± 7%    6.56µs ± 3%   -2.38%  (p=0.002 n=35+20)
Tee_Check                            1.27µs ±10%    1.23µs ± 6%   -2.83%  (p=0.021 n=35+20)

name                               old alloc/op   new alloc/op   delta
CheckedMessage_Chain                  42.0B ± 0%     42.0B ± 0%     ~     (all equal)
CheckedMessage_Chain_sliceLoggers     42.0B ± 0%     42.0B ± 0%     ~     (all equal)
JSONLogMarshalerFunc                  16.0B ± 0%     16.0B ± 0%     ~     (all equal)
ZapJSONInts                            296B ± 0%       80B ± 0%  -72.97%  (p=0.000 n=35+20)
ZapJSONStrings                         520B ± 0%      160B ± 0%  -69.23%  (p=0.000 n=35+20)
ZapJSON                               0.00B          0.00B          ~     (all equal)
StandardJSON                         1.92kB ± 0%    1.92kB ± 0%     ~     (all equal)
NullLogMarshalerFunc                  16.0B ± 0%     16.0B ± 0%     ~     (all equal)
ZapNull                               0.00B          0.00B          ~     (all equal)
NoContext                             0.00B          0.00B          ~     (all equal)
BoolField                             64.0B ± 0%     64.0B ± 0%     ~     (all equal)
Float64Field                          64.0B ± 0%     64.0B ± 0%     ~     (all equal)
IntField                              64.0B ± 0%     64.0B ± 0%     ~     (all equal)
Int64Field                            64.0B ± 0%     64.0B ± 0%     ~     (all equal)
StringField                           64.0B ± 0%     64.0B ± 0%     ~     (all equal)
StringerField                         68.0B ± 0%     68.0B ± 0%     ~     (all equal)
TimeField                             64.0B ± 0%     64.0B ± 0%     ~     (all equal)
DurationField                         64.0B ± 0%     64.0B ± 0%     ~     (all equal)
ErrorField                            64.0B ± 0%     64.0B ± 0%     ~     (all equal)
StackField                           1.09kB ± 0%    1.09kB ± 0%     ~     (all equal)
MarshalerField                        64.0B ± 0%     64.0B ± 0%     ~     (all equal)
**
IntsField                              360B ± 0%      176B ± 0%  -51.11%  (p=0.000 n=35+20)
StringsField                           584B ± 0%      256B ± 0%  -56.16%  (p=0.000 n=35+20)
**
ObjectField                            584B ± 0%      584B ± 0%     ~     (all equal)
AddCallerHook                         48.0B ± 0%     48.0B ± 0%     ~     (all equal)
10Fields                               640B ± 0%      640B ± 0%     ~     (all equal)
100Fields                            1.17kB ± 0%    1.17kB ± 0%     ~     (all equal)
Tee_Check                             64.0B ± 0%     64.0B ± 0%     ~     (all equal)

name                               old allocs/op  new allocs/op  delta
CheckedMessage_Chain                   0.00           0.00          ~     (all equal)
CheckedMessage_Chain_sliceLoggers      0.00           0.00          ~     (all equal)
JSONLogMarshalerFunc                   1.00 ± 0%      1.00 ± 0%     ~     (all equal)
ZapJSONInts                            4.00 ± 0%      1.00 ± 0%  -75.00%  (p=0.000 n=35+20)
ZapJSONStrings                         5.00 ± 0%      1.00 ± 0%  -80.00%  (p=0.000 n=35+20)
ZapJSON                                0.00           0.00          ~     (all equal)
StandardJSON                           38.0 ± 0%      38.0 ± 0%     ~     (all equal)
NullLogMarshalerFunc                   1.00 ± 0%      1.00 ± 0%     ~     (all equal)
ZapNull                                0.00           0.00          ~     (all equal)
NoContext                              0.00           0.00          ~     (all equal)
BoolField                              1.00 ± 0%      1.00 ± 0%     ~     (all equal)
Float64Field                           1.00 ± 0%      1.00 ± 0%     ~     (all equal)
IntField                               1.00 ± 0%      1.00 ± 0%     ~     (all equal)
Int64Field                             1.00 ± 0%      1.00 ± 0%     ~     (all equal)
StringField                            1.00 ± 0%      1.00 ± 0%     ~     (all equal)
StringerField                          2.00 ± 0%      2.00 ± 0%     ~     (all equal)
TimeField                              1.00 ± 0%      1.00 ± 0%     ~     (all equal)
DurationField                          1.00 ± 0%      1.00 ± 0%     ~     (all equal)
ErrorField                             1.00 ± 0%      1.00 ± 0%     ~     (all equal)
StackField                             2.00 ± 0%      2.00 ± 0%     ~     (all equal)
MarshalerField                         1.00 ± 0%      1.00 ± 0%     ~     (all equal)
**
IntsField                              5.00 ± 0%      3.00 ± 0%  -40.00%  (p=0.000 n=35+20)
StringsField                           6.00 ± 0%      3.00 ± 0%  -50.00%  (p=0.000 n=35+20)
**
ObjectField                            7.00 ± 0%      7.00 ± 0%     ~     (all equal)
AddCallerHook                          1.00 ± 0%      1.00 ± 0%     ~     (all equal)
10Fields                               1.00 ± 0%      1.00 ± 0%     ~     (all equal)
100Fields                              3.00 ± 0%      3.00 ± 0%     ~     (all equal)
Tee_Check                              1.00 ± 0%      1.00 ± 0%     ~     (all equal)
```

json_encoder.go
```diff
@@ -95,13 +95,17 @@ func (enc *jsonEncoder) Free() {
 	jsonPool.Put(enc)
 }
 
+func (enc *jsonEncoder) wrapString(val string) {
```
akshayjshah: Perhaps `addAndQuoteString`? I'd expect `wrapString` to return the wrapped value.

field_test.go
```diff
@@ -178,14 +178,26 @@ func TestIntsField(t *testing.T) {
...
 	assertFieldJSON(t, `"foo":[]`, Object("foo", []string{}))
 	assertFieldJSON(t, `"foo":["bar 1"]`, Object("foo", []string{"bar 1"}))
 	assertFieldJSON(t, `"foo":["bar 1","bar 2","bar 3"]`, Object("foo", []string{"bar 1", "bar 2", "bar 3"}))
+
+	assertFieldJSON(t, `"foo":[]`, Strings("foo", []string{}))
+	assertFieldJSON(t, `"foo":["bar 1"]`, Strings("foo", []string{"bar 1"}))
+	assertFieldJSON(t, `"foo":["bar 1","bar 2","bar 3"]`, Strings("foo", []string{"bar 1", "bar 2", "bar 3"}))
+
 	assertCanBeReused(t, Object("foo", []string{"bar 1", "bar 2", "bar 3"}))
+	assertCanBeReused(t, Strings("foo", []string{"bar 1", "bar 2", "bar 3"}))
```
akshayjshah: There's already a `TestObjectField` below. I'd like to:

1. Remove use of `ObjectField` from `TestIntsField` and `TestStringsField`.
2. Add `TestIntsFieldQuick` and `TestStringsFieldQuick`. Those tests should use `testing/quick` to ensure that the output generated by `ObjectField` and our typed array fields is the same for arbitrary slices.

keyvalue.go
```diff
@@ -38,4 +38,8 @@ type KeyValue interface {
 	// allocation-heavy. Consider implementing the LogMarshaler interface instead.
 	AddObject(key string, value interface{}) error
 	AddString(key, value string)
+
+	// Slices
+	AddInts(key string, value []int)
+	AddStrings(key string, value []string)
```
akshayjshah: nit: can we put `AddInts` immediately after `AddInt`, and likewise for `AddStrings`? I think that the type signatures are enough documentation - there's no need for an additional comment.

null_encoder_test.go
```diff
@@ -54,6 +54,10 @@ func TestNullEncoderFields(t *testing.T) {
 		{"marshaler", func(e Encoder) {
 			assert.NoError(t, e.AddMarshaler("k", loggable{true}), "Unexpected error calling MarshalLog.")
 		}},
+		{"ints", func(e Encoder) { e.AddInts("k", []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}) }},
+		{"strings", func(e Encoder) {
+			e.AddStrings("k", []string{"bar 1", "bar 2", "bar 3", "bar 4", "bar 5", "bar 6", "bar 7", "bar 8", "bar 9", "bar 10"})
+		}},
```
akshayjshah: Again, let's trim down the size of the test data.

text_encoder_test.go
```diff
@@ -87,11 +87,16 @@ func TestTextEncoderFields(t *testing.T) {
 		{"marshaler", "k={}", func(e Encoder) {
 			assert.Error(t, e.AddMarshaler("k", loggable{false}), "Expected an error calling MarshalLog.")
 		}},
-		{"ints", "k=[1 2 3 4 5 6 7 8 9 10]", func(e Encoder) { e.AddObject("k", []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}) }},
-		{"strings", `k=[bar 1 bar 2 bar 3 bar 4 bar 5 bar 6 bar 7 bar 8 bar 9 bar 10]`,
+		{"obj_ints", "k=[1 2 3 4 5 6 7 8 9 10]", func(e Encoder) { e.AddObject("k", []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}) }},
+		{"obj_strings", `k=[bar 1 bar 2 bar 3 bar 4 bar 5 bar 6 bar 7 bar 8 bar 9 bar 10]`,
```
akshayjshah: Same as above re: human-readability.

json_encoder_test.go
```diff
@@ -140,11 +140,16 @@ func TestJSONEncoderFields(t *testing.T) {
 		{"marshaler", `"k":{}`, func(e Encoder) {
 			assert.Error(t, e.AddMarshaler("k", loggable{false}), "Expected an error calling MarshalLog.")
 		}},
-		{"ints", `"k":[1,2,3,4,5,6,7,8,9,10]`, func(e Encoder) { e.AddObject("k", []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}) }},
-		{"strings", `"k":["bar 1","bar 2","bar 3","bar 4","bar 5","bar 6","bar 7","bar 8","bar 9","bar 10"]`,
+		{"obj_ints", `"k":[1,2,3,4,5,6,7,8,9,10]`, func(e Encoder) { e.AddObject("k", []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}) }},
+		{"obj_strings", `"k":["bar 1","bar 2","bar 3","bar 4","bar 5","bar 6","bar 7","bar 8","bar 9","bar 10"]`,
```
akshayjshah: The initial strings should be a human-readable description of the test case - there's no need for them to be a single word. How about `"[]int, using reflection"`?

json_encoder.go
```diff
@@ -168,6 +172,52 @@ func (enc *jsonEncoder) AddMarshaler(key string, obj LogMarshaler) error {
 	return err
 }
 
+func (enc *jsonEncoder) arrayBegin() {
```
akshayjshah: Can we call these `addArrayBegin`, etc.? To my eye, methods with noun names look like simple getters. Since these mutate the receiver, I'd prefer them to have verb names.

json_encoder.go
```diff
@@ -168,6 +172,52 @@ func (enc *jsonEncoder) AddMarshaler(key string, obj LogMarshaler) error {
...
+	if l > 0 {
+		enc.wrapString(vals[0])
+
+		for i := 1; i < l; i++ {
+			enc.arraySep()
+			enc.wrapString(vals[i])
+		}
+	}
+	enc.arrayEnd()
+}
```
akshayjshah: We can reduce duplication and nesting in these methods with a small type alias.

```
type jsonArrayIndex int

func (a arrayIndex) AddSeparator(enc *jsonEncoder) {
  if a == 0 {
    return
  }
  enc.bytes = append(enc.bytes, ',')
}
```

Then the encoder methods should become a bit simpler:

```
func (enc *jsonEncoder) AddInts(key string, vals []int) {
  enc.addKey(key)
  enc.arrayBegin()
  for i := range vals {
    jsonArrayIndex(i).AddSeparator(enc)
    enc.bytes = strconv.AppendInt(enc.bytes, int64(vals[i]), 10)
  }
  enc.arrayEnd()
}
```

Likewise for the `TextEncoder`. (In a future PR, we should rethink how we're handling text encoding. The duplication between text and JSON encoding is atrocious.)

akshayjshah: I'm all for performance, especially along the hot path, but I don't think that this is worth the repetition. In a [benchmark of these two approaches](https://github.com/akshayjshah/gobits/tree/master/zapideas), each integer comparison is ~0.1ns - that's about as close to free as we get in Go.

(If we wanted to go this far, which we don't, there's more to be gained by implementing a base-10 specific version of `AppendInt` - we could shave off three comparisons and some array accesses.)

json_encoder.go
jcorbin: +1, verb the names

json_encoder.go
jcorbin: Good point on the premature micro-op @akshayjshah , but I'm not sold that introducing a type alias wins us anything for now (the loc count even smells same-ish to me). How about we just eliminate the special-cased handling of index `0` by making them normal `for i := 0; ...`, and leave this sort of micro perf to future work?

json_encoder.go
phungs: Done.

json_encoder.go
phungs: Done.

json_encoder.go
phungs: I've reverted the loop unrolling micro-op in favor of a cleaner range iteration and abstracted the separator encoding decision into the encoder's addArraySep(i int) function.  This was done for both JSON and Text encoders.

field_test.go
phungs: Good suggestion and done.  Note that the TestStringsFieldQuick currently fails if it is passed a string containing "&".  Here's some test output with an input of []string{"&"}:

mis-match
   sobj: "foo":["\u0026"]
  sstrs: "foo":["&"]

Akshay or I will look for a solution to this.

keyvalue.go
phungs: Done.

null_encoder_test.go
phungs: Done.

text_encoder_test.go
phungs: Done.

json_encoder_test.go
phungs: Done.

json_encoder.go
```diff
@@ -168,6 +172,46 @@ func (enc *jsonEncoder) AddMarshaler(key string, obj LogMarshaler) error {
...
+func (enc *jsonEncoder) addArrayBegin() {
+	enc.bytes = append(enc.bytes, '[')
+}
+
+func (enc *jsonEncoder) addArrayEnd() {
+	enc.bytes = append(enc.bytes, ']')
+}
+
+func (enc *jsonEncoder) addArraySep(i int) {
+	if i == 0 {
```
jcorbin: nit: shorter to just do the `!= 0` case.

json_encoder.go
```diff
@@ -168,6 +172,46 @@ func (enc *jsonEncoder) AddMarshaler(key string, obj LogMarshaler) error {
...
+	}
+	enc.bytes = append(enc.bytes, ',')
+}
+
+// AddInts adds a string key and integer slice to the encoder's fields. The key
+// is JSON-escaped.
+func (enc *jsonEncoder) AddInts(key string, vals []int) {
+	enc.addKey(key)
+	enc.addArrayBegin()
+
```
jcorbin: nit: drop empty line.

json_encoder.go
```diff
@@ -168,6 +172,46 @@ func (enc *jsonEncoder) AddMarshaler(key string, obj LogMarshaler) error {
...
+	for i, val := range vals {
+		enc.addArraySep(i)
+		enc.bytes = strconv.AppendInt(enc.bytes, int64(val), 10)
```
jcorbin: ideally `AddInt` and `AddInts` would share a kernel just like `AddString` and `AddStrings` do.

json_encoder.go
phungs: I considered both approaches but choose the early return as it separates the validation from the logic and would probably be the preferable form if logic was > 1 line (i.e. textEncoder's addTime).  I'll keep this as-is unless @akshayjshah also feels strongly that it should be changed.

json_encoder.go
phungs: done.

json_encoder.go
phungs: Done (also changed in text_encoder.go)

akshayjshah:
@jcorbin and @phungs: I tacked on a few small commits in Scott's absence. Take a look and flag anything objectionable.

jcorbin:
Updates lgtm @akshayjshah

jcorbin:
I share your concerns @prashantv , can you compare/contras your suggestion with mine on https://github.com/uber-go/zap/issues/136#issuecomment-261014480 ?

prashantv:
That's very similar to what I had in mind @jcorbin, the only difference was using an interface rather than a `func(Values)` (since you can easily wrap a function as an interface, whereas a function requires allocation of a closure everytime)

It's a much larger change since it's adding a new top-level interface with a lot of methods to implement, but we could start small (just `Int` and `String` in the `Values` interface to see how it works)

json_encoder.go
```diff
@@ -167,6 +165,28 @@ func (enc *jsonEncoder) AddMarshaler(key string, obj LogMarshaler) error {
...
+	enc.endArray()
+}
+
+// AddStrings adds a string key and string slice to the encoder's fields. The key
+// and the slice elements are JSON-escaped.
+func (enc *jsonEncoder) AddStrings(key string, vals []string) {
+	enc.beginArray(key)
+	for i := range vals {
+		enc.addArraySep(i)
+		enc.addAndQuoteString(vals[i])
```
phungs: Is there a benefit of using vals[i] over val (from for i, val := range vals)?  The latter is a bit more error-proof.

text_encoder.go
```diff
@@ -106,6 +110,42 @@ func (enc *textEncoder) AddMarshaler(key string, obj LogMarshaler) error {
...
+	for i, val := range vals {
+		enc.addArraySep(i)
+		enc.addInt64(int64(val))
+	}
+	enc.addArrayEnd()
+}
+
+func (enc *textEncoder) AddStrings(key string, vals []string) {
+	enc.addKey(key)
+	enc.addArrayBegin()
```
phungs: For consistency, we should mirror the recent changes from json_encoder.go into text_encoder.go: arrayBegin()/arrayEnd() and maybe use of vals[i].  I have some bandwidth and can help with that.

json_encoder.go
akshayjshah: Yes. It avoids a copy.

text_encoder.go
akshayjshah: Don't think it's worth it, since we're also chasing #189 - there's a good chance that we're going to pivot the text encoder to use the JSON encoder.

akshayjshah:
Interesting. This is, I think, primarily an issue for integer types and times, where we'd like to share common implementations of `AppendInt64` and `AppendUint64`.

If we were going to stop with JSON and text, I'd be more interested in these approaches. However, I think it's also valuable to (post-1.0) include a msgpack encoder - to me, it's the obvious fast encoding choice for loosely-typed objects. In that case, I'd much prefer to *not* serialize `[]uint8` as an array of int64s, since I can get significant space savings by explicitly handling each integer type.

In general, I don't expect there to be a large number of encoder implementations (even including external implementations), since there just aren't that many encodings appropriate for unstructured application logs. Given that, I'd prefer that this interface remain large and explicit. A smaller interface that makes assumptions about the desired implementation of the encoder is, IMO, penny-wise and pound-foolish.

Of course, this doesn't mean that I'm willing to tolerate tons of duplication in the *implementation* of the large number of JSON array-encoding methods.

akshayjshah:
Alternate approach explored in #239.

akshayjshah:
Closing in favor of the approach in #239.

PR #214, opened 2016-12-16
Implement flag.{Value,Getter} for zap.Level

jcorbin:
This makes it easy to add a `flag.Var` to control a program's log level.

### DIFF
--- a/level.go
+++ b/level.go
@@ -141,6 +141,34 @@ func (l *Level) UnmarshalText(text []byte) error {
 	return nil
 }
 
+// Set sets the level for the flag.Value interface.
+func (l *Level) Set(s string) error {
+	switch string(s) {
+	case "debug":
+		*l = DebugLevel
+	case "info":
+		*l = InfoLevel
+	case "warn":
+		*l = WarnLevel
+	case "error":
+		*l = ErrorLevel
+	case "dpanic":
+		*l = DPanicLevel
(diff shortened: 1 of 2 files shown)


akshayjshah:
At a quick glance, test failures seem like they're caused by different error message text in different versions of Go. Let's try to make the test independent of the exact error text.

If that's completely impractical, I'm okay dropping support for 1.5 and 1.6 in the dev branch; when we hit 1.0, Go 1.8 will be the stable release. At that point, supporting 1.7 and 1.8 is reasonable.

jcorbin:
Heh, actually I wish there a way to opt back in to the "always print even my 'zero value' default" that went away after 1.6 it seems... but that might also be "smell" that our level type has a non-trivial zero value (unless you really agree that it's "obvious" that a "-level value" defaults to "info" ;-) ).

jcorbin:
Updated to not make any assertions about flag help; only left a minimal error message assertion.

level.go
```diff
@@ -141,6 +141,34 @@ func (l *Level) UnmarshalText(text []byte) error {
 	return nil
 }
 
+// Set sets the level for the flag.Value interface.
+func (l *Level) Set(s string) error {
+	switch string(s) {
```
akshayjshah: `s` is already a string; why the cast here?

PR #225, opened 2017-01-04
Refactor internal error handling

### LINKED ISSUE #223: CheckedEntry.Write should handle any error
Currently it's `func(Entry, ...Field) error`; rather it should have void return type, and deal with the error. To do this, CheckedEntry will need to carry an error output write syncer to report the write error to.

jcorbin:
- handle write errors in CheckedEntry (which is the only remaining write path since we simplified logger over its own Check method)
- de-generic-ify the remaining logger  internal error case for caller adding

This introduces a regrettable API asymmetry:
- fac.Write returns an error
- but fac.Check().Write() eats it

I think we should go further (eventually or in this pr to reviewers' taste), but I haven't found a compelling path yet... something around:
- maybe we should have `type CheckedMessage struct {*CheckedEntry; ErrorOutput WriteSyncer}` and that's what the [Ll]ogger level uses... but then we have to also pool those separately in addition to...
- maybe `(*CheckedEntry).Write` should still return an error, but only if `ErrorOutput` is nil?

Fixes #223

### DIFF
--- a/entry.go
+++ b/entry.go
@@ -154,8 +154,6 @@ func (ce *CheckedEntry) Write(fields ...Field) {
 	case WriteThenPanic:
 		panic(msg)
 	}
-
-	return
 }
 
 // AddFacility adds a facility that has agreed to log this entry. It's intended
--- a/entry.go
+++ b/entry.go
@@ -101,6 +101,7 @@ const (
 // Write, and MUST NOT be retained after calling their Write() method.
 type CheckedEntry struct {
 	Entry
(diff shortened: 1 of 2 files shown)


entry.go
```diff
@@ -144,7 +155,7 @@ func (ce *CheckedEntry) Write(fields ...Field) error {
 		panic(msg)
 	}
 
-	return errs.asError()
+	return
```
akshayjshah: Don't need the bare return here.

PR #226, opened 2017-01-04
Keep Fields an optionFunc

akshayjshah:
Remove the `FieldsT` type in favor of the more terse optionFunc. Since options are only passed infrequently, the overhead of creating a function closure isn't a concern.


akshayjshah:
Gotcha - figured that there was a reason that didn't survive the large refactoring ;)

PR #227, opened 2017-01-04
Add finer grained caller annotation support

jcorbin:
@akshayjshah let me know if you're okay with adding the new field support or not; if so I'll write tests, if not we'll just stop at the new option for now.

Towards #40

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -111,6 +111,10 @@ func (log *logger) Check(lvl Level, msg string) *CheckedEntry {
 }
 
 func (log *logger) check(lvl Level, msg string) *CheckedEntry {
+	// check must always be called directly by another `logger` entry function
+	// (e.g. Check, Log, Info, etc)
+	const callerSkipOffset = 2
+
 	// Create basic checked entry thru the facility; this will be non-nil if
 	// the log message will actually be written somewhere.
 	ent := Entry{
@@ -144,7 +148,7 @@ func (log *logger) check(lvl Level, msg string) *CheckedEntry {
 	ce.ErrorOutput = log.errorOutput
 
 	if log.addCaller {
-		ce.Entry.Caller = MakeEntryCaller(runtime.Caller(log.callerSkip + 2))
+		ce.Entry.Caller = MakeEntryCaller(runtime.Caller(log.callerSkip + callerSkipOffset))
 		if !ce.Entry.Caller.Defined {
 			fmt.Fprintf(log.errorOutput, "%v addCaller error: failed to get caller\n", time.Now().UTC())
 			log.errorOutput.Sync()
--- a/logger.go
+++ b/logger.go
@@ -107,6 +107,10 @@ func (log *logger) With(fields ...Field) Logger {
 }
 
 func (log *logger) Check(lvl Level, msg string) *CheckedEntry {
+	return log.check(lvl, msg)
+}
+
+func (log *logger) check(lvl Level, msg string) *CheckedEntry {
 	// Create basic checked entry thru the facility; this will be non-nil if
 	// the log message will actually be written somewhere.
 	ent := Entry{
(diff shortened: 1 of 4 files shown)


jcorbin:
Okay, dropped the field, and added a test for the new option.  N.B. how I avoided coupling to line numbers by moving utility code into a new file; if you'd rather just use line numbers, we can go back to bb641cd.

logger.go
```diff
@@ -144,7 +144,7 @@ func (log *logger) Check(lvl Level, msg string) *CheckedEntry {
 	ce.ErrorOutput = log.errorOutput
 
 	if log.addCaller {
-		ce.Entry.Caller = MakeEntryCaller(runtime.Caller(log.callerSkip))
+		ce.Entry.Caller = MakeEntryCaller(runtime.Caller(log.callerSkip + 2))
```
akshayjshah: Can we preserve the `defaultCallerSkip` constant instead of putting a literal `2` here?

logger.go
```diff
@@ -159,16 +159,50 @@ func (log *logger) Check(lvl Level, msg string) *CheckedEntry {
...
+	if ce := log.check(PanicLevel, msg); ce != nil {
+		ce.Write(fields...)
+	}
+}
+
+func (log *logger) Fatal(msg string, fields ...Field) {
+	if ce := log.check(FatalLevel, msg); ce != nil {
+		ce.Write(fields...)
+	}
+}
```
akshayjshah: Yuck. I suppose this is a necessary evil to keep the skip level consistent.

logger_test.go
```diff
@@ -371,12 +327,44 @@ func TestJSONLoggerSyncsOutput(t *testing.T) {
...
-			`"caller":"[^"]+/logger_test.go:[\d]+","msg":"Callers\."`,
-			buf.Stripped(), "Expected to find package name and file name in output.")
-	})
+			testCase.pat,
+			lines[i],
+			"Expected to find package name and file name in %s output.",
+			testCase.name)
+	}
+
+	assert.Empty(t, errSink.String(), "Expected error sink to be empty.")
```
akshayjshah: Can we structure this so that each test case includes the logger and the expected message, and in the test loop we

```
tt.logger.Info("foo")
tt.logger.Check(InfoLevel, "foo").Write()
```

and then assert that we got the expected message twice? Seems like that would more clearly associate the loggers and the expectations.

options.go
```diff
@@ -64,6 +64,16 @@ func AddCaller() Option {
 	})
 }
 
+// AddCallerSkip configures the Logger to annotate each message with the filename
+// and line number of a caller outside of zap; which caller in the stack is
+// controlled by the skip argument. If more than one AddCallerSkip option is
+// provided, then the maximum skip amount is used.
```
akshayjshah: nit: "the *sum* of the skip amounts is used." Also, to me the comment suggests that this option also enables caller annotations - may want to clarify.

logger.go
jcorbin: Well, I'd argue for leaving it in place:
- it truly is path dependent on how `logger` calls `check`
- originally I started out with it as a third arg to the new `check` method...
- ...which was either 2 or 3 depending on the entry path call site to `check`
- however, after I modified all the named entries (Debug, Info, Warn, etc) to call `check` instead of `Log`, it turned out that all the sites were passing the same `2` skip arg

I'd be happy to re-introduce a constant, but:
- it doesn't deserve to be package level any more
- rather it would be at the top of `check`, and called something like `skipOffset`

I'll do that since it'll dovetail with any future needs if we ever need to re-promote it to a third `check` arg.

logger.go
jcorbin: Yes it is, and this actually was an untested bug in prior (and I'm still lacking a case for it actually):
- Log is an entry point...
- ...but so are Debug, Info, etc.

So all the named entries need +1 more offset than Log ... but we have no support for dynamically modifying skip (and we're trying very hard to avoid adding such).

logger_test.go
jcorbin: Will do, this code is a bit more repetitive than prior because it's a "contraction" back away from a moment of too much indirection; I'll start layering back some ;-)

logger.go
jcorbin: Nvm the point about `Log is an entry point`: it's not, we dropped it in #201, but the implementation lived on here; will drop.

Issue #224, opened 2017-01-03
Consider separating leaf and middleware Facilities


akshayjshah:
(Relevant only to the `dev` branch. See the large "Facilities could be further broken up" comment on #201.)

Facilities are either (1) middleware for other facilities (e.g., sampling or teeling) or (2) leaves, like the `ioFacility`. Consider separating these types so that implementors need to write less boilerplate.

jcorbin:
Prior thought on this direction describe in https://github.com/uber-go/zap/pull/201#issuecomment-266893762

akshayjshah:
We ended up punting on this.

PR #238, opened 2017-01-11
Factor EntryWriter out of Facility

jcorbin:
Towards #224 .

Also simplified sampler while I was "in the room".

### DIFF
--- a/zapcore/facility.go
+++ b/zapcore/facility.go
@@ -27,14 +27,23 @@ import (
 	"go.uber.org/atomic"
 )
 
+// EntryWriter is a destination for log entttries.  It can have pervasive
+// fields added with With().
+type EntryWriter interface {
+	Write(Entry, []Field) error
+	With([]Field) EntryWriter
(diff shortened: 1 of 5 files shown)


zapcore/facility.go
```diff
@@ -254,22 +272,16 @@ func (s *sampler) With(fields []Field) Facility {
...
+	if n := s.counts.Inc(ent.Message); n > s.first {
+		if n == s.first+1 {
+			time.AfterFunc(s.tick, func() { s.counts.Reset(ent.Message) })
+		}
+		if (n-s.first)%s.thereafter != 0 {
+			return ce
+		}
 	}
-	return (n-s.first)%s.thereafter == 0
+	return s.Facility.Check(ent, ce)
```
akshayjshah: Nice, this is definitely a better way to structure this.

akshayjshah:
RC1 is landing on master tonight; we decided not to land this refactoring.

PR #239, opened 2017-01-11
Initial support for arrays

akshayjshah:
Add basic support for logging arrays. To limit the scope of this commit,
it introduces only arrays of ObjectMarshalers, arrays of arrays, and
arrays of bools. It should, however, clarify how I intend to support
arrays of other types.

Of note, I opted not to have `zap.Bools` return a `Field`. Instead, it
returns a `zapcore.ArrayMarshaler`. This lets us build arrays of
heterogenous arrays. For most users, this will be papered over by the
upcoming `zap.Any(key string, val interface{}) zapcore.Field` function,
which will basically be a huge type switch.

(Along the way, I broke out our multi-error implementation into a separate
internal package.)

This starts to address #136, and is an alternative to the approach in #211.

From the small benchmarks included in this PR:
```
BenchmarkBoolsArrayMarshaler-4   	 1000000	      1028 ns/op	    1088 B/op	       3 allocs/op
BenchmarkBoolsReflect-4          	  500000	      3293 ns/op	    2280 B/op	       8 allocs/op
```
All allocations in the ArrayMarshaler benchmark come from allocating byte slices for encoders.

### DIFF
--- a/array.go
+++ b/array.go
@@ -20,56 +20,18 @@
 
 package zap
 
-import (
-	"go.uber.org/zap/internal/multierror"
-	"go.uber.org/zap/zapcore"
-)
+import "go.uber.org/zap/zapcore"
 
-// TODO: actually add an example showing off how to use these wrappers with the
-// Array field constructor.
-
-// Arrays wraps a slice of ArrayMarshalers so that it satisfies the
-// ArrayMarshaler interface. See the Array function for a usage example.
-func Arrays(as []zapcore.ArrayMarshaler) zapcore.ArrayMarshaler {
-	return arrayMarshalers(as)
+// Array constructs a field with the given key and ArrayMarshaler. It provides
+// a flexible, but still type-safe and efficient, way to add array-like types
+// to the logging context. The struct's MarshalLogArray method is called lazily.
+func Array(key string, val zapcore.ArrayMarshaler) zapcore.Field {
+	return zapcore.Field{Key: key, Type: zapcore.ArrayMarshalerType, Interface: val}
 }
 
-// Objects wraps a slice of ObjectMarshalers so that it satisfies the
-// ArrayMarshaler interface. See the Array function for a usage example.
-func Objects(os []zapcore.ObjectMarshaler) zapcore.ArrayMarshaler {
-	return objectMarshalers(os)
-}
-
-// Bools wraps a slice of bools so that it satisfies the ArrayMarshaler
-// interface. See the Array function for a usage example.
-func Bools(bs []bool) zapcore.ArrayMarshaler {
-	return bools(bs)
-}
-
-type arrayMarshalers []zapcore.ArrayMarshaler
-
-func (as arrayMarshalers) MarshalLogArray(arr zapcore.ArrayEncoder) error {
-	var errs *multierror.Error
-	for i := range as {
-		if as[i] == nil {
-			continue
-		}
-		errs = errs.Append(arr.AppendArray(as[i]))
-	}
-	return errs.AsError()
-}
-
-type objectMarshalers []zapcore.ObjectMarshaler
-
-func (os objectMarshalers) MarshalLogArray(arr zapcore.ArrayEncoder) error {
-	var errs *multierror.Error
-	for i := range os {
-		if os[i] == nil {
-			continue
-		}
-		errs = errs.Append(arr.AppendObject(os[i]))
-	}
-	return errs.AsError()
+// Bools constructs a field that carries a slice of bools.
+func Bools(key string, bs []bool) zapcore.Field {
+	return Array(key, bools(bs))
 }
 
 type bools []bool
--- a/array.go
+++ b/array.go
@@ -1,0 +1,82 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zap
+
+import (
+	"go.uber.org/zap/internal/multierror"
+	"go.uber.org/zap/zapcore"
+)
+
+// TODO: actually add an example showing off how to use these wrappers with the
+// Array field constructor.
+
+// Arrays wraps a slice of ArrayMarshalers so that it satisfies the
+// ArrayMarshaler interface. See the Array function for a usage example.
+func Arrays(as []zapcore.ArrayMarshaler) zapcore.ArrayMarshaler {
+	return arrayMarshalers(as)
+}
+
+// Objects wraps a slice of ObjectMarshalers so that it satisfies the
+// ArrayMarshaler interface. See the Array function for a usage example.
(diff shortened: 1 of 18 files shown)


zapcore/encoder.go
```diff
@@ -22,6 +22,31 @@ package zapcore
...
+	AddFloat64(key string, value float64)
+	AddInt64(key string, value int64)
+	AddUint64(key string, value uint64)
+	AddObject(key string, marshaler ObjectMarshaler) error
+	AddArray(key string, marshaler ArrayMarshaler) error
+	// AddReflected uses reflection to serialize arbitrary objects, so it's slow
+	// and allocation-heavy.
+	AddReflected(key string, value interface{}) error
+	AddString(key, value string)
+}
```
akshayjshah: Moved most of this  here from `object_encoder.go`, since it seemed silly to have two interface-only files for encoders.

array.go
```diff
@@ -0,0 +1,82 @@
...
+	"go.uber.org/zap/internal/multierror"
+	"go.uber.org/zap/zapcore"
+)
+
+// TODO: actually add an example showing off how to use these wrappers with the
+// Array field constructor.
+
+// Arrays wraps a slice of ArrayMarshalers so that it satisfies the
+// ArrayMarshaler interface. See the Array function for a usage example.
+func Arrays(as []zapcore.ArrayMarshaler) zapcore.ArrayMarshaler {
```
prashantv: I don't think `Arrays` and `Objects` is going to be that useful to users.

If someone is implementing the `ArrayMarshaler` interface, then they probably care about performance, and wouldn't want to build up a separate new slice of `[]zapcore.ArrayMarshaler`.

What use case are you expecting for `Arrays` and `Objects`?

array_test.go
```diff
@@ -0,0 +1,85 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zap
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
prashantv: nit: import ordering

array.go
```diff
@@ -0,0 +1,82 @@
...
+func Bools(bs []bool) zapcore.ArrayMarshaler {
+	return bools(bs)
+}
+
+type arrayMarshalers []zapcore.ArrayMarshaler
+
+func (as arrayMarshalers) MarshalLogArray(arr zapcore.ArrayEncoder) error {
+	var errs *multierror.Error
+	for i := range as {
+		if as[i] == nil {
```
prashantv: If we want to keep `arrayMarshalers` around, we should consider whether ignoring `nil` is the right behaviour. Should it just be an error? I think I'm learning towards remove `Arrays` and `Objects` and keep things minimal. It's easy enough for others to implement and add later.

If we really want to keep this type, can we add unit tests for the `== nil` branch here and in `Objects`?

internal/multierror/multierror.go
```diff
@@ -0,0 +1,77 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package multierror provides a simple way to treat a collection of errors as
+// a single error.
+package multierror
+
+import "go.uber.org/zap/buffers"
```
prashantv: is this package not internal on purpose? I really think this should be internal. Otherwise one bad user of `buffers` will cause all sorts of weird data races / corrupt log messages. I really don't think we want any other library to be able to put buffers into our pool.

internal/multierror/multierror.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+// Append adds an error to the collection. Adding a nil error is a no-op.
+func (e *Error) Append(err error) *Error {
+	if err == nil {
+		return e
+	}
+	if e == nil {
+		e = &Error{}
```
prashantv: This is a strange pattern, you have an `Append` that works on a pointer, sometimes modifies the value inside, sometimes returns a new pointer. Given that, why do you need this method on the pointer? Why not always work on the value? Passing around a `[]error` by value is not expensive, makes the semantics cleaner, and reduces the need for `*Error`

That'll probably also avoid an allocation when there is an error

internal/multierror/multierror_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+package multierror
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
```
prashantv: One slightly odd test that I think is worth having -- you really don't want `Error` to implement the `error` interface, since that could cause odd issues. So we could have a test like:
```go
var v interface{} = Error{}
_, ok := v.(error)
require.False(t, ok)
```

zapcore/json_encoder_test.go
```diff
@@ -136,6 +199,57 @@ func TestJSONEncoderFields(t *testing.T) {
 	}
 }
 
+func TestJSONEncoderArrayTypes(t *testing.T) {
```
prashantv: Can you add unit tests for:
 - empty array
 - array with different types

array_test.go
```diff
@@ -0,0 +1,85 @@
...
+			}),
+			[]interface{}{
+				[]interface{}{true},
+				[]interface{}{false},
+			},
+		},
+		{
+			"objects",
+			Objects([]zapcore.ObjectMarshaler{
+				zapcore.ObjectMarshalerFunc(func(_ zapcore.ObjectEncoder) error { return nil }),
```
prashantv: here and below, since you don't use any args, you can just make this `func(zapcore.ObjectEncoder) error { .. }`

array.go
akshayjshah: Sure, I can add some tests. I'm going to have to come through lots and lots of `zapcore` and improve test coverage over the next week.

array.go
akshayjshah: `Arrays` is, IMO, useful precisely because most users won't be implementing `ArrayMarshaler` themselves - they'll be using the umpteen wrapper types we're providing, like `Bools`. Since we're already going down a road where we support mixed-type arrays, I'd like to take that to its logical conclusion.

I'll email you about `Objects` - relies on some internal examples.

internal/multierror/multierror.go
akshayjshah: Yep, makes sense. My bad.

zapcore/json_encoder_bench_test.go
```diff
@@ -63,8 +63,8 @@ func BenchmarkZapJSON(b *testing.B) {
 		for pb.Next() {
 			enc := newJSONEncoder(cfg)
 			enc.AddString("str", "foo")
-			enc.AddInt64("int", 1)
-			enc.AddInt64("int64", 1)
+			enc.AddInt64("int64-1", 1)
```
jcorbin: What's up with this, it seems unrelated?

zapcore/memory_encoder_test.go
```diff
@@ -27,22 +27,45 @@ import (
...
+			e.AppendBool(true)
+			return nil
+		}))
+		return nil
+	})), "Expected AddArray to succeed.")
+	// Nested objects and arrays.
+	assert.NoError(t, enc.AddArray("turduckens", turduckens(2)), "Expected AddObject to succeed.")
+	assert.NoError(t, enc.AddObject("turducken", turducken{}), "Expected AddObject to succeed.")
 
+	wantTurducken := MapObjectEncoder{
```
jcorbin: +100

internal/multierror/multierror.go
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
akshayjshah: My understanding is that copyright applies to the creative work as a whole, which is the zap project; since that project started in 2016, all copyright starts there. (Imagine that Walt Disney updates a few of the line arcs in the original Mickey Mouse drawing. Those lines don't have a separate copyright date.)

Of course, I could easily be wrong.

internal/multierror/multierror.go
```diff
@@ -0,0 +1,71 @@
...
+// Error wraps a []error to implement the error interface.
+type Error struct {
+	errs errSlice
+}
+
+// AsError converts the collection to a single error value.
+//
+// Note that failing to use AsError will almost certainly lead to bugs with
+// non-nil interfaces containing nil concrete values.
+func (e Error) AsError() error {
```
akshayjshah: That is indeed the use case. While I agree that error handling is often repetitive, I think we're often encouraged to program with error values. Where possible, I'd like to make things less repetitive. (I'm treating `bufio.Reader` as my spirit animal here.)

zapcore/json_encoder_bench_test.go
akshayjshah: Yep, the key was just incorrect - it's not an int, but an int64.

zapcore/memory_encoder_test.go
akshayjshah: 🦃

zapcore/json_encoder_bench_test.go
akshayjshah: Yep, the key was just wrong. Noticed and fixed in passing.

PR #241, opened 2017-01-12
Move buffers package to internal, add some test coverage

prashantv:
This came up in https://github.com/uber-go/zap/pull/239#discussion_r95737392

We should never be exposing our buffer pool, since one bad user of the buffer pool could cause problems in logs (such as corrupt data), or cause data races to show up within zap.

I noticed we had no test coverage here, so added a basic test with concurrency, and validates that the returned buffer has 0 length.


akshayjshah:
Sure, makes sense - I didn't think carefully about the impact bad actors could have here. I pushed a commit to fix some small nits in your test - PTAL.

prashantv:
Looks good, thanks for fixing up the missing waitgroup!

Issue #220, opened 2017-01-03
Re-introduce observers/read-only hooks


akshayjshah:
On the dev branch, we've removed hook support entirely. We should consider re-introducing read-only hooks; in particular, we *may* want hooks to be able to read fields.

The best way to do this may be to take the current testing spy, pivot it into a production-ready observer, and make it fast enough to reasonably use in production.

akshayjshah:
Landed on dev as part of #244.

PR #244, opened 2017-01-12
Re-introduce read-only hooks as a facility

akshayjshah:
Re-introduce simple read-only hooks as a facility wrapper. This doesn't fully
satisfy our desire for an observing facility, but it's simple and doesn't
introduce any new external APIs before v1.0.

This should address the immediate needs from #220. We can continue to iterate on a full-fledged observer and introduce it in a post-1.0 point release.

### DIFF
--- a/zapcore/hook.go
+++ b/zapcore/hook.go
@@ -30,9 +30,10 @@ type hooked struct {
 // Hooked wraps a facility and runs a collection of user-defined callback hooks
 // each time a message is logged.
 func Hooked(fac Facility, hooks ...func(Entry) error) Facility {
+	funcs := append([]func(Entry) error{}, hooks...)
 	return &hooked{
 		Facility: fac,
-		funcs:    hooks,
+		funcs:    funcs,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


zapcore/hook.go
```diff
@@ -0,0 +1,59 @@
...
+import "go.uber.org/zap/internal/multierror"
+
+type hooked struct {
+	Facility
+	funcs []func(Entry) error
+}
+
+// Hooked wraps a facility and runs a collection of user-defined callback hooks
+// each time a message is logged.
+func Hooked(fac Facility, hooks ...func(Entry) error) Facility {
```
jcorbin: copy the variadic, or take a slice

zapcore/hook.go
```diff
@@ -0,0 +1,59 @@
...
+func (h *hooked) With(fields []Field) Facility {
+	return &hooked{
+		Facility: h.Facility.With(fields),
+		funcs:    h.funcs,
+	}
+}
+
+func (h *hooked) Write(ent Entry, _ []Field) error {
+	var errs multierror.Error
+	for _, f := range h.funcs {
```
jcorbin: Worth it to just do a `for i := range` rather than copy the function pointer out?

zapcore/hook_test.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	for _, tt := range tests {
+		fac, logs := NewObserver(tt.facilityLevel, 1)
+		intField := makeInt64Field("foo", 42)
+		ent := Entry{Message: "bar", Level: tt.entryLevel}
+
+		var called bool
+		f := func(e Entry) error {
+			called = true
```
jcorbin: counter instead of bool and/or assert only called once?

PR #245, opened 2017-01-12
Refactor Encoder.WriteEntry => Encoder.EncodeEntry

jcorbin:
Encoders should, well, encode. Writing to a sink is a concern for the facility.

### DIFF
--- a/zapcore/facility_test.go
+++ b/zapcore/facility_test.go
@@ -113,7 +113,7 @@ func TestWriterFacilitySyncsOutput(t *testing.T) {
 	}
 }
 
-func TestWriterFacilityWriteEntryFailure(t *testing.T) {
+func TestWriterFacilityWriteFailure(t *testing.T) {
 	fac := WriterFacility(
 		NewJSONEncoder(testJSONConfig()),
 		Lock(AddSync(&testutils.FailWriter{})),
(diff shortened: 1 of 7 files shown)


zapcore/facility_test.go
```diff
@@ -113,7 +113,7 @@ func TestWriterFacilitySyncsOutput(t *testing.T) {
 	}
 }
 
-func TestWriterFacilityWriteEntryFailure(t *testing.T) {
```
akshayjshah: Can we add another test case here with the short writer from the testutils package?

zapcore/json_encoder.go
```diff
@@ -159,20 +144,17 @@ func (enc *jsonEncoder) AppendBool(val bool) {
...
+	final := &jsonEncoder{
+		JSONConfig: enc.JSONConfig,
+		bytes:      buffers.Get(),
 	}
 
-	final := jsonPool.Get().(*jsonEncoder)
-	final.truncate()
 	final.bytes = append(final.bytes, '{')
 	enc.LevelFormatter(ent.Level).AddTo(final)
 	enc.TimeFormatter(ent.Time).AddTo(final)
```
akshayjshah: Just for sanity, we should probably use `final.LevelFormatter`. Same for the time and message formatters.

zapcore/facility_test.go
```diff
@@ -119,7 +119,17 @@ func TestWriterFacilityWriteFailure(t *testing.T) {
...
 	)
+	err := fac.Write(Entry{}, nil)
+	// Should log the error.
+	assert.Error(t, err, "Expected writing Entry to fail.")
+}
 
+func TestWriterFacilityShortWrite(t *testing.T) {
+	fac := WriterFacility(
+		NewJSONEncoder(testJSONConfig()),
+		Lock(AddSync(&testutils.ShortWriter{})),
```
akshayjshah: Separate note: we don't need the the `AddSync` anymore.

PR #247, opened 2017-01-16
Alternate, map-based sugared logger

akshayjshah:
This PR is a sibling to #185; only one of the two should be merged. It explores a different sugaring API, using `map[string]interface{}` to pass fields and `string` to pass messages.

This avoids allocating to pass strings as `interface{}` (for both the message and the field keys). In our `BenchmarkZapSugarAddingFields` benchmark, this saves roughly 1 microsecond and 20 small allocations at each log site. However, it actually allocates more memory overall, increases verbosity at log sites, decreases flexibility, and necessitates at least seven additional methods on the sugared logger.

IMO, this isn't a good tradeoff.

```
BenchmarkLogrusAddingFields-4                         	  200000	      9601 ns/op	    5727 B/op	      77 allocs/op
BenchmarkLogrusWithAccumulatedContext-4               	  200000	      7622 ns/op	    3919 B/op	      61 allocs/op
BenchmarkLogrusWithoutFields-4                        	 1000000	      2222 ns/op	    1321 B/op	      25 allocs/op

BenchmarkZapAddingFields-4                            	 1000000	      1260 ns/op	     768 B/op	       4 allocs/op
BenchmarkZapWithAccumulatedContext-4                  	 5000000	       325 ns/op	      64 B/op	       2 allocs/op
BenchmarkZapWithoutFields-4                           	 5000000	       339 ns/op	      64 B/op	       2 allocs/op

BenchmarkZapSugarAddingFields-4                       	  500000	      2924 ns/op	    1729 B/op	      15 allocs/op
BenchmarkZapSugarWithAccumulatedContext-4             	 5000000	       370 ns/op	      64 B/op	       2 allocs/op
BenchmarkZapSugarWithoutFields-4                      	 5000000	       378 ns/op	      64 B/op	       2 allocs/op
```

### DIFF
--- a/benchmarks/zap_sugar_bench_test.go
+++ b/benchmarks/zap_sugar_bench_test.go
@@ -43,8 +43,22 @@ func fakeSugarFields() []interface{} {
 	}
 }
 
+var newCoreLoggerDefaults = []zap.Option{
+	zap.ErrorLevel,
+	zap.DiscardOutput,
+}
+
+func newCoreLogger(options ...zap.Option) zap.Logger {
+	options = append(newCoreLoggerDefaults, options...)
+	return zap.New(zap.NewJSONEncoder(), options...)
+}
+
+func newSugarLogger(options ...zap.Option) zap.Sugar {
+	return zap.NewSugar(newCoreLogger(options...))
+}
+
 func BenchmarkZapSugarDisabledLevelsWithoutFields(b *testing.B) {
-	logger := zap.NewSugar(zap.New(zap.NewJSONEncoder(), zap.ErrorLevel, zap.DiscardOutput))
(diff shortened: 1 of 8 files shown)


prashantv:
That is a lot of methods, I don't think we want that large of an API.

How about an approach that is more similar to log15, where you always take varargs, and the user can choose to pass a single `Ctx` map, or they can pass a bunch of field/values. Then the API for each level should only need:
```go
Panic(msg string, ctx... interface{})
Panicf(fmt string, args... interface{})
```

akshayjshah:
Fixed the caller skip - totally an oversight on my part. Don't think that this requires a facility for dynamic skips; the current situation is, IMO, simple and quite tolerable.

PR #248, opened 2017-01-18
Add benchmark for stacktrace generation

glibsm:
The "relatively speaking" comment on stack generation
has peaked my interest, so I wanted to see what is the impact
of generating a trace compared to a regular log.

Without and with results:

```
BenchmarkZapAddingFields-8                       3000000               606 ns/op             769 B/op          4 allocs/op
BenchmarkZapAddingFieldsAndStacktrace-8           100000             16526 ns/op            1957 B/op          6 allocs/op
```

### DIFF
--- a/benchmarks/zap_bench_test.go
+++ b/benchmarks/zap_bench_test.go
@@ -170,6 +170,20 @@ func BenchmarkZapAddingFields(b *testing.B) {
 	})
 }
 
+func BenchmarkZapAddingFieldsAndStacktrace(b *testing.B) {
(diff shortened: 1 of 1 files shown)


akshayjshah:
There's already a benchmark for taking stack traces: https://github.com/uber-go/zap/blob/dev/logger_bench_test.go#L127.

This benchmark is structured slightly differently - do you feel that it's adding much extra information over the existing one?

glibsm:
I missed that this was benchmarked elsewhere - I don't think this adds anything new, it just re-states that we're spending about ~1us to get the trace

PR #249, opened 2017-01-19
Log timestamps in localtime for TextEncoder.

### LINKED ISSUE #232: Prefer localtime for TextEncoder
Currently the TextEncoder outputs timestamps in UTC.  I suggest changing this to localtime, in the interest of being more human-legible.

This is a one-line change to logger.log(), which won't affect the JSONEncoder, since that is of course timezone-independent.

I considered whether this should be an option, but I don't think it's worth it, as I'm not really seeing a use for UTC over localtime for human-legible output.

Will you accept such a pull request?

tesujimath:
Resolves #232.  Since others are asking for this, it seemed worth sharing the one-line change.  Didn't seem worth making people wait until 1.0 for such a simple and useful improvement.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -126,7 +126,7 @@ func (log *logger) log(lvl Level, msg string, fields []Field) {
 		return
 	}
 
-	t := time.Now().UTC()
+	t := time.Now().Local()
 	if err := log.Encode(log.Output, t, lvl, msg, fields); err != nil {
 		log.InternalError("encoder", err)
 	}


akshayjshah:
Unfortunately, this just creates the opposite issue - now users are forced to use local time for text logs. Rather than forcing users into local time, we should instead make the text encoder configurable. It's easy enough for a user-supplied formatter function to receive a `time.Time` in UTC and localize it to a time zone of their choice.

If you're willing to make that change, I'm happy to review and accept it; it's just going to be lost in the 1.0 release in a few weeks. As-is, though, we can't merge this.

tesujimath:
By logging in localtime it is already configurable, by means of the system timezone. So if you're a global operation, you can set the timezone of all your servers to be the same, and everything matches. The system logfiles all use localtime by the way. UTC just seems not useful to me.

akshayjshah:
I'm pointing out that this is a breaking change, and it doesn't introduce a way for users to recover the original behavior. I'm not willing to merge breaking changes on master, since we're close to releasing a 1.0 and I'd rather not force users to update multiple times.

akshayjshah:
Fixed on dev, landing on master tonight.

PR #250, opened 2017-01-19
Add support for named loggers

### LINKED ISSUE #130: Add support for named loggers
It can often be convenient for logs to have a name that is part of each log line.  For example, if multiple systems are writing to the same log file, or for pastes of log messages to make it obvious what system the log line came from.

I would like to propose adding a name argument to the logger.New and logger.With methods.  When a child log is created the two log names are concatenated together with a '.'. 

So for the example in the docs:

`{"level":"warn","msg":"Log without structured data..."}`

Would be something like:

`{"name":"myapp","level":"warn","msg":"Log without structured data..."}`

or for a child logger:

`{"name":"myapp.child","level":"warn","msg":"Log without structured data..."}`

akshayjshah:
This resolves #130.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 	"time"
 
 	"go.uber.org/zap/zapcore"
@@ -31,19 +32,26 @@ import (
 
 func defaultEncoderConfig() zapcore.JSONConfig {
 	msgF := func(msg string) zapcore.Field {
-		return zapcore.Field{Type: zapcore.StringType, String: msg, Key: "msg"}
+		return String("msg", msg)
 	}
 	timeF := func(t time.Time) zapcore.Field {
 		millis := t.UnixNano() / int64(time.Millisecond)
-		return zapcore.Field{Type: zapcore.Int64Type, Integer: millis, Key: "ts"}
+		return Int64("ts", millis)
 	}
 	levelF := func(l zapcore.Level) zapcore.Field {
-		return zapcore.Field{Type: zapcore.StringType, String: l.String(), Key: "level"}
+		return String("level", l.String())
+	}
+	nameF := func(n string) zapcore.Field {
+		if n == "" {
+			return Skip()
+		}
+		return String("name", n)
 	}
 	return zapcore.JSONConfig{
 		MessageFormatter: msgF,
 		TimeFormatter:    timeF,
 		LevelFormatter:   levelF,
+		NameFormatter:    nameF,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


logger.go
```diff
@@ -109,18 +120,26 @@ func New(fac zapcore.Facility, options ...Option) Logger {
...
 
+func (log *logger) Named(s string) Logger {
+	if s == "" {
+		return log
+	}
+	l := log.clone()
+	if log.name == "" {
+		l.name = s
+	} else {
+		l.name = strings.Join([]string{l.name, s}, ".")
```
prashantv: Instead of having the logger always use `.` as a separator, what about if we pass name around as a `[]string`, and let the `NameFormatter` decide how to convert that `[]string` into a `zapcore.Field`?

logger.go
```diff
@@ -226,3 +246,15 @@ func (log *logger) Fatal(msg string, fields ...zapcore.Field) {
 func (log *logger) Facility() zapcore.Facility {
 	return log.fac
 }
+
+func (log *logger) clone() *logger {
```
prashantv: It looks like this is a straight up copy. We have a couple of other options here that avoid listing every field (and possibly missing out on a new field that's added later):
```go
func (log logger) clone() *logger {
  // Since this is a method on the value, a copy was already created and passed in as log.
  return &log
}
```

```go
func (log *logger) clone() *logger {
  copy := *log
  return &copy
}
```

logger.go
akshayjshah: That's a really good point. Will change.

logger.go
akshayjshah: Also a good point.

logger.go
akshayjshah: Hm. We could...but it'll typically cost an allocation on the hot path, since we'll need to join the `[]string` into a single period-separated string before writing out the entry. We'll incur this cost on every write, even if the name hasn't changed and there are no new fields.

Overall, I don't think it's worth it. Users who are really dead-set on this can use strings.Replace in their `NameFormatter`.

sugar.go
```diff
@@ -74,7 +74,14 @@ func Desugar(s *SugaredLogger) Logger {
...
 // panics in development and errors in production.
 func (s *SugaredLogger) With(args ...interface{}) *SugaredLogger {
-	return &SugaredLogger{core: s.core.With(s.sweetenFields(args)...)}
+	return s.WithFields(s.sweetenFields(args)...)
+}
+
+// WithFields adds structured fields to the logging context, much like the
+// base Logger. It allows the sugared logger to use more specialized
+// fields (like Stack).
+func (s *SugaredLogger) WithFields(fs ...zapcore.Field) *SugaredLogger {
```
jcorbin: Did you mean to just slip this one into this PR? It seems like a whole different thing; also it's attached to a larger ball of wax:
- if we're going to cede that it's useful toe have a `FooFields` for `Foo=With`...
- ...why not also for `Foo` in `{Debug,Info,Warn,Error,DPanic,Panic,Fatal}`?
- At that point, why don't we just collapse `SugaredLogger` into `Logger`, and dispense with all the wrapping?
- The user story would then become something like: "most people use Foof / Foow, perf-conscious users use FooFields"; maybe even something like:
  - "Don't care about logging? Use Foow"
  - "Want structured logs? Use Foof"
  - "Really care about logs? Library author? Use FooFields"

PR #251, opened 2017-01-24
Double down on loosely-typed fields for SugaredLogger

akshayjshah:
Some of zap's strongly-typed fields rely on the field constructors to do some
work, so it's difficult for users to replicate their effects using the variadic
key-value pairs that the sugared logger prefers. (`Stack()` is a good example of
this.)

To accommodate these use cases, the sugared logger has a `WithFields` method,
but this isn't very satisfying; for symmetry, we'd also want `InfoWithFields`,
`DebugWithFields`, etc.

This PR doubles down on developer ergonomics (over compile-time safety) by
making `With`, `Debugw`, `Infow`, etc. able to accept a mix of zapcore.Fields
and loosely-typed key-value pairs. This obviates the `WithFields` method.

### DIFF
--- a/sugar.go
+++ b/sugar.go
@@ -74,9 +74,9 @@ func Desugar(s *SugaredLogger) Logger {
 //
 // Note that the keys in key-value pairs should be strings. In development,
 // passing a non-string key panics. In production, the logger is more
-// forgiving: a separate error is logged, but the key is coerced to a string
-// with fmt.Sprint and execution continues. Passing an orphaned key triggers
-// similar behavior: panics in development and errors in production.
+// forgiving: a separate error is logged, but the key-value pair is skipped and
+// execution continues. Passing an orphaned key triggers similar behavior:
+// panics in development and errors in production.
 func (s *SugaredLogger) With(args ...interface{}) *SugaredLogger {
 	return &SugaredLogger{core: s.core.With(s.sweetenFields(args)...)}
 }
--- a/sugar.go
+++ b/sugar.go
@@ -49,39 +49,35 @@ func Desugar(s *SugaredLogger) Logger {
 	return s.core
 }
 
-// With adds a variadic number of key-value pairs to the logging context.
-// Even-indexed arguments are treated as keys and zipped with the odd-indexed
-// values using the Any field constructor.
+// With adds a variadic number of fields to the logging context. It accepts a
+// mix of strongly-typed zapcore.Field objects and loosely-typed key-value
+// pairs. When processing pairs, the first element of the pair is used as the
+// field key and the second as the field value.
 //
 // For example,
 //   sugaredLogger.With(
 //     "hello", "world",
 //     "failure", errors.New("oh no"),
+//     Stack(),
 //     "count", 42,
 //     "user", User{name: "alice"},
 //  )
 // is the equivalent of
-//   coreLogger.With(
+//   baseLogger.With(
 //     String("hello", "world"),
 //     String("failure", "oh no"),
+//     Stack(),
 //     Int("count", 42),
 //     Object("user", User{name: "alice"}),
 //   )
 //
-// Note that the keys should be strings. In development, passing a non-string
-// key panics. In production, the logger is more forgiving: a separate error is
-// logged, but the key is coerced to a string with fmt.Sprint and execution
-// continues. Passing an odd number of arguments triggers similar behavior:
-// panics in development and errors in production.
+// Note that the keys in key-value pairs should be strings. In development,
+// passing a non-string key panics. In production, the logger is more
+// forgiving: a separate error is logged, but the key is coerced to a string
+// with fmt.Sprint and execution continues. Passing an orphaned key triggers
+// similar behavior: panics in development and errors in production.
 func (s *SugaredLogger) With(args ...interface{}) *SugaredLogger {
-	return s.WithFields(s.sweetenFields(args)...)
-}
-
-// WithFields adds structured fields to the logging context, much like the
-// base Logger. It allows the sugared logger to use more specialized
-// fields (like Stack).
-func (s *SugaredLogger) WithFields(fs ...zapcore.Field) *SugaredLogger {
-	return &SugaredLogger{core: s.core.With(fs...)}
+	return &SugaredLogger{core: s.core.With(s.sweetenFields(args)...)}
 }
 
 // Debug uses fmt.Sprint to construct and log a message.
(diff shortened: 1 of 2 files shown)


sugar.go
```diff
@@ -212,25 +208,35 @@ func (s *SugaredLogger) sweetenFields(args []interface{}) []zapcore.Field {
...
-	}
 
-	fields := make([]zapcore.Field, len(args)/2)
-	for i := range fields {
-		keyIdx := 2 * i
-		val := args[keyIdx+1]
-		key, ok := args[keyIdx].(string)
+	// Allocate enough space for the common case; if users only pass structured
+	// fields, we'll need to grow the slice.
+	fields := make([]zapcore.Field, 0, len(args)/2)
```
jcorbin: hmmm... I'd argue for using `len(args)` as the default; i.e. conservatively over-estimate, precisely so that if you start converting to "better" structured fields, you don't get a surprising perf/allocs "cliff" behavior.

sugar.go
```diff
@@ -212,25 +208,35 @@ func (s *SugaredLogger) sweetenFields(args []interface{}) []zapcore.Field {
...
+	var i int
```
jcorbin: clearer as `i := 0` imo (nit)

sugar.go
```diff
@@ -212,25 +208,35 @@ func (s *SugaredLogger) sweetenFields(args []interface{}) []zapcore.Field {
...
+	for i < len(args) {
```
jcorbin: In fact, I'd mildly prefer `for i := 0; i < len(ags); {`

sugar.go
```diff
@@ -212,25 +208,35 @@ func (s *SugaredLogger) sweetenFields(args []interface{}) []zapcore.Field {
...
 			s.core.DPanic(
 				_nonStringKeyErrMsg,
-				Int("position", keyIdx),
-				Any("key", args[keyIdx]),
+				Int("position", i),
+				Any("key", args[i]),
 				Any("value", val),
 			)
-			key = fmt.Sprint(args[keyIdx])
+			key = fmt.Sprint(args[i])
```
jcorbin: Is this the right form of degradation? A more conservative stance would be to `break`, so that we don't accidentally put a (maybe very large) value in as a "key". Basically, this case is similar to the parity fail case above, and I think we should "fail open" rather than "fail best effort" in production.

sugar_test.go
```diff
@@ -44,6 +44,13 @@ func TestSugarWith(t *testing.T) {
 		{[]interface{}{}, []zapcore.Field{}, nil},
 		{[]interface{}{"should ignore"}, []zapcore.Field{}, []observer.LoggedEntry{ignored}},
 		{[]interface{}{"foo", 42, "true", "bar"}, []zapcore.Field{Int("foo", 42), String("true", "bar")}, nil},
+		{[]interface{}{Int("foo", 42)}, []zapcore.Field{Int("foo", 42)}, nil},
+		{[]interface{}{Int("foo", 42), "should ignore"}, []zapcore.Field{Int("foo", 42)}, []observer.LoggedEntry{ignored}},
+		{
+			[]interface{}{"first", "field", Int("foo", 42), "baz", "quux", "should ignore"},
+			[]zapcore.Field{String("first", "field"), Int("foo", 42), String("baz", "quux")},
+			[]observer.LoggedEntry{ignored},
```
jcorbin: I'm trying to sort out whether these cases test the development panic path, or the production path, but I can't figure it out just from these declarations:
- I think at this point we need to have a `desc string` field in each case
- bonus points if we switch to named-field struct literals, as this is getting a bit unweildy to read
- whichever path (dev v prod) is covered, cover the other one too, even if it means we need a whole other `func Test*` or a new dimension in our case structs

sugar.go
```diff
@@ -212,25 +208,35 @@ func (s *SugaredLogger) sweetenFields(args []interface{}) []zapcore.Field {
...
-	for i := range fields {
-		keyIdx := 2 * i
-		val := args[keyIdx+1]
-		key, ok := args[keyIdx].(string)
+	// Allocate enough space for the common case; if users only pass structured
+	// fields, we'll need to grow the slice.
+	fields := make([]zapcore.Field, 0, len(args)/2)
+	var i int
+	for i < len(args) {
+		if f, ok := args[i].(zapcore.Field); ok {
```
jcorbin: Can we make the state machine structure clearer please? Something like:
```
for i < len(args) {
  // check for inlined zap Fields
  if f, ok := ... { ... }

  // parity check: we need an even number of key/val pairs (at least since the last zap Field)
  if i == len(args)-1 { ... }]

  // consume a key/val pair
  key, val := args[i], args[i+1]

  // process valid string key
  if keyString, ok := key.(string); ok {
    fields = append(fields, Any(keyString, val))
    i += 2
   continue
  }

  // we have a "key" that isn't a string, crash in development...
  s.core.DPanic(...)

  // ...but in production, stop field conversion here, since this could be a parity error (e.g. a forgotten key, and not the optimistic "you forgot to string-convert a key")
  break
}
```

You could, of course, flip the `string` cast ok predicate back to the negative.

Also see if you think:
```
key := args[i]
i++

val := args[i]
i++
```

is any better; the biggest downside I can see is a need to `i-2` or `i-1` inside the DPanic site, depending on how you intersperse the key/val extraction with the key-string check.

In short: I bet a lot of people will end up reading this source code, since it's one step beneath a common entry point for any `godef` chasers; let's optimize its readability, not just to convince ourselves that it's correct / robust.

sugar_test.go
akshayjshah: Added a description to each scenario and switched to keyed literals.

The tests cover both the production and development paths, since they just assert that we're logging at `DPanicLevel`. The tests for `DPanicLevel` verify that we're panicing and erroring appropriately.

sugar_test.go
```diff
@@ -31,33 +31,88 @@ import (
...
+			errLogs:  []observer.LoggedEntry{ignored("dangling")},
+		},
+		{
+			desc:     "one non-string key",
+			args:     []interface{}{"foo", 42, true, "bar"},
+			expected: []zapcore.Field{Int("foo", 42)},
+			errLogs:  []observer.LoggedEntry{nonString(invalidPair{2, true, "bar"})},
+		},
+		{
+			desc:     "pairs, structured fields, non-string keys, and a dangling key",
```
jcorbin: other cases that occur to me:
- dangling non-string value
- other things that aren't strings as keys, like structs, interface values, etc
- more interleaving, like `pair(s), structured field(s), pair(s), structured field(s)`

I hesitate to suggest programmatically generating test cases, especially in a world where we still can't use sub-tests (I dearly look forward to a post go 1.6 support world ;-) ).

akshayjshah: Added some more tests.

PR #257, opened 2017-01-31
Add test coverage for using levels in a FlagSet

akshayjshah:
Levels implement `flag.Value`, so they should be usable in the standard lib's
flag sets. However, this wasn't covered by tests.

In the final push for docs, we'll need examples for these use cases.

### DIFF
--- a/flag_test.go
+++ b/flag_test.go
@@ -30,12 +30,42 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+type flagTestCase struct {
+	args      []string
+	wantLevel zapcore.Level
+	wantErr   bool
+}
+
+func (tc flagTestCase) RunImplicitSet(t testing.TB) {
+	origCommandLine := flag.CommandLine
+	flag.CommandLine = flag.NewFlagSet("test", flag.ContinueOnError)
+	flag.CommandLine.SetOutput(ioutil.Discard)
+	defer func() { flag.CommandLine = origCommandLine }()
+
+	level := LevelFlag("level", InfoLevel, "")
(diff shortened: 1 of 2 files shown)


flag_test.go
```diff
@@ -30,12 +30,42 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+type flagTestCase struct {
+	args      []string
+	wantLevel zapcore.Level
+	wantErr   bool
+}
+
+func (tc flagTestCase) RunImplicitSet(t testing.TB) {
```
jcorbin: weird that these func names are "exported" on a non-exported type.

flag_test.go
```diff
@@ -30,12 +30,42 @@ import (
...
+	}
+}
+
 func TestLevelFlag(t *testing.T) {
-	tests := []struct {
-		args      []string
-		wantLevel zapcore.Level
-		wantErr   bool
-	}{
+	tests := []flagTestCase{
```
jcorbin: It doesn't fit in your current test harness shape, but: test two differently named levels, and that they don't collide (e.g. the avoided pointer sharing in the `LevelFlag` function, I could see that easily getting broken by a refactor).

This is a realistic use case, e.g. varying console vs file logging levels.

flag_test.go
akshayjshah: True.

I keep wishing that the Go community had some convention for intra-package visibility hints, but I don't think that's happening anytime soon.

flag_test.go
akshayjshah: Good point, I'll add tests for that.

PR #258, opened 2017-01-31
Add encoder support for most built-in types

akshayjshah:
As part of finalizing the encoder interface, this PR adds methods to support nearly all the built-in types. This greatly expands the interface, but it's desirable for binary encodings (like msgpack) that will want to treat each size of integer differently. Since we're adding so many methods to the encoders anyways, I rounded things out by supporting all the built-in types except `error`.

Overall, I think that this is a good tradeoff - I expect that there will be many `Facility` and `WriteSyncer` implementations in the wild, but relatively few `Encoder`s.

This is a (large) precursor to fleshing out our support for array types.


ansel1:
Design question: 

In zap, Field holds values in one of a few strongly typed fields, which avoids converting them into interface{}, which I gather has a performance benefit.

Then Field.AddTo() has a big switch statement on FieldType, to route the value to the appropriate encoder function.  This makes the Encoder interface very wide.

Why not have Encoder.AddField(f *Field), then let the encoder implementation worry about the switch statement?  Seems like that would have several benefits:

- much simpler Encoder interface
- encoders can choose how granular they want to be about handling FieldTypes.  Some encoders may handle all number types in a uniform way, and therefore can have a simpler switch case:

        switch field.Type {
            case: zap.Int64Type, zap.Float64Type, zap.Uint64Type, ...etc:

- methods could be added to Field to massage values into types (e.g. "Field.ValAsInt64").  This would make encoders simpler to write, and consolidate type-massaging code into one place.  These methods would also give encoders a fallback mechanism for handle new FieldType's which they don't know about yet (forward compatibility)
- encoders which are not sensitive to performance can begin life with a very simple AddField() implementation which just massages everything into a string or something.  Then optimizing for performance can be done iteratively later.

akshayjshah:
@ansel1 all good questions. Let me take them one at a time...

> In zap, Field holds values in one of a few strongly typed fields, which avoids converting them into interface{}, which I gather has a performance benefit.

Yep, exactly - converting most built-in types (string, int, uint64, etc.) to `interface{}` costs an allocation, which adds up really fast. (See [Russ Cox's post on interfaces](https://research.swtch.com/interfaces) for details.) We can always layer more magic on top, but we want the base logger to be as fast and low-allocation as possible.

> Why not have Encoder.AddField(f *Field), then let the encoder implementation worry about the switch statement? Seems like that would have several benefits:

You've nailed some of the benefits - it definitely allows encoder implementations to be slimmer. However, it comes with some significant drawbacks too:

- It makes the encoder interface harder to statically verify. The current interface can be fully checked by the compiler, and won't expand later on - there's no need for the developer to worry about field types from the future.
- It makes implementing `ObjectMarshaler` more irritating for end users. To me, it's very convenient for the encoder passed to `MarshalLogObject` to be usable as-is, without constructing fields. Given the choice, I'd rather make life a bit more difficult for a handful of encoder authors but more pleasant for a (hopefully!) large number of users.
- Most importantly, it doesn't reduce the surface area of `ArrayEncoder`. Fundamentally, fields represent key-value pairs, and arrays are just values. We *could* start special-casing fields without keys, but I like the conceptual clarity of the current approach. If encoders have to implement `AppendInt`, `AppendBool`, etc., I'd rather go all-in and make adding key-value pairs explicit too.

To me, these drawbacks outweigh the benefits. What do you think?

For encoders that aren't particularly performance-sensitive, embedding a `MapObjectEncoder` handles all the boilerplate for you (though it doesn't make the iterative perf improvement you're describing easy). Out of curiosity, what sort of encoder do you have in mind?

ansel1:
Hmm...perhaps.  I feel like there's a lot of repetition now around code to handle all the primitive types (all the field constructors, then the field-to-encoder code, then the encoder interface).  It's a lot of boilerplate.  I think you could handle the ArrayEncoder and ObjectMarshaler points above by extracting a Value struct from Field.  So a Field is just a Value + a name.  Encoder interface would be AddField(), ArrayEncoder would be AddValue().  But then, perhaps you'd need to add zap.XXXValue() functions to mirror all the zap.XXXField() functions, which wouldn't be great.  

On the other hand, you're arguing that implementors of ObjectMarshaller (and ArrayMarshaller, etc) would rather deal the ginormous ObjectEncoder interface, rather than re-using the zap.XXXField() functions they are already used to.  It does imply they need to learn two APIs instead of one.

Another alternative is to put a shim between ObjectMarshallers and Encoders, which adapts the ObjectEncoder interface to the much narrower Encoder interface (by translating all those AddXXX() calls to zap.XXXField() calls, then passing them on to Encoder.AddField()).

Just thinking about that, it again strikes me that you have two APIs that do the same thing.  Both APIs are for added very specifically typed values with names to something.  One is the Fields-based API, the other is the ObjectEncoder API.  Just feels like duplication.

ansel1:
Oh, you asked what I was trying to do.  I'm playing around with writing a text encoder.  Seems like that was dropped from dev branch.

akshayjshah:
Ah, I see. The text encoder (at least as planned) is making a triumphant return, and it'll share most of its code with the JSON encoder.

akshayjshah:
> I feel like there's a lot of repetition now around code to handle all the primitive types (all the field constructors, then the field-to-encoder code, then the encoder interface). It's a lot of boilerplate.
> ...
> Just thinking about that, it again strikes me that you have two APIs that do the same thing. Both APIs are for added very specifically typed values with names to something. One is the Fields-based API, the other is the ObjectEncoder API. Just feels like duplication.

I suppose I just don't see it that way. To me, all the field constructors solve a particular problem: mapping a large variety of types into a small union struct while avoiding heap allocations. Encoders solve a different problem: efficiently and incrementally serializing values. The `Field.AddTo` method maps between the two worlds, and only needs to be written once. We could certainly orient the encoder API around fields, but that seems to me like it'd just replace multiple methods with a large switch in each encoder.

I am, of course, willing to be convinced if you want to open a PR and show off the brevity gains ;)

ansel1:
Eh, you called my bluff :)  

For me, a switch statement in the encoder seems like a more elegant solution.  It's easier to code for the encoder author and more tersely expresses common handling code for multiple field types.  but it's pretty close to "6 of one, half a dozen of the other".  I'm playing with encoders, so I guess my sympathies lie there at the moment.

Though I don't find the argument that the compiler verifiability is a win in this case.  Seems like enshrining those types into the API will make it hard to evolve (because it would break all existing encoders).  

For example, what if later you want to add an additional core type, like time.Duration.  This currently gets massaged downward to a plain number, which is unfortunate, because the encoder now can't do anything unique with durations.  In the text encoder, it would be really nice to print durations as their human-friendly strings.  

With a Field-oriented Encoder interface, encoders could opt to handle that field type.  Or they could fall back on some default behavior.

I suppose you can evolve the interface approach too, by adding optional encoder interfaces.  i.e. all encoders must implement the base Encoder interface, but later can optionally implement the DurationEncoder interface if they want to handle durations in a unique manner.

akshayjshah:
I'm fully in agreement with making `time.Time` and `time.Duration` first-class; they're currently in a weird half-supported state. They're somewhat special - more below.

In general, I think that the extensibility support we have makes it pretty easy to support new types later. Let's say that we launched 1.0 and then discovered that we desperately needed support for the `os.FileInfo` type. How would we accomplish this?

```
func FileInfo(key string, info os.FileInfo) zapcore.Field {
  return Object(key, fileinfo(info))
}

type fileinfo os.FileInfo

func (info fileinfo) MarshalLogObject(enc zapcore.ObjectEncoder) error {
  enc.AddString("name", info.Name())
  enc.AddInt64("size", info.Size())
  enc.AddBool("isDirectory", info.IsDir())
  return nil
}
```

At this point, we've added support for a new type, and it immediately works across all existing encoders. Since all encoders support all the builtin types, they're reliable substrates for further abstraction.

Pros of the current approach:
- new field types are immediately supported by all encoders
- all encoders behave consistently (i.e., encoder authors can't forget to handle a field type)

Cons:
- as noted, the encoder interface is broad.
- can't add first-class encoder support for new types without pain (e.g., upcasting interfaces like your `DurationEncoder`)

Perhaps I've done a bad job articulating my aversion to using fields in the encoder, so let me try again. Architecturally, I made encoders intentionally low-level - they're only one step above `*os.File`. They're also close to encodings (e.g., JSON, msgpack) - they define how to serialize primitive types, but don't know about complex types like users and `FileInfo`. Loggers (in zap, that's `zapcore.Facility`, `Entry`, and `Field`) convert domain objects into the simple language of encoders. Part of my resistance to `Encoder.Add(Field)` is that it reaches "up the stack."

`time.Time` and `time.Duration` are special for a few reasons: they're very common in a logging context, everyone wants to serialize them differently, and they easily convert to int64s. Because of that, it's worth adding first-class support - we can put them into the field union as nanos-since-epoch and nanos-elapsed, and we can let end-users configure how they'd like them serialized.

akshayjshah:
I'm going to land this PR just so that I can keep stacking diffs; I'm looking forward to continuing this discussion, though - it's valuable feedback on design. Happy to keep talking in these comments, or we can move to an issue.

ansel1:
Yeah, you sold me.  Good points.

akshayjshah:
🎉  Glad we can agree! 🎉 

Appreciate the design push-back, though. I'm separately considering ways to make it easier for text-based encoders (JSON, some ad-hoc console-friendly thing, LTSV, etc) share more of the boring byte-shuffling code. I'll mention you in the PR - I'd like your input.

PR #260, opened 2017-02-01
Add support for remaining array types

### LINKED ISSUE #136: JSON Array support

akshayjshah:
Add wrappers for common array types. This fixes #136.

(Note that this is stacked on top of #258.)

### DIFF
--- a/array.go
+++ b/array.go
@@ -20,7 +20,11 @@
 
 package zap
 
-import "go.uber.org/zap/zapcore"
+import (
+	"time"
+
+	"go.uber.org/zap/zapcore"
+)
 
 // Array constructs a field with the given key and ArrayMarshaler. It provides
 // a flexible, but still type-safe and efficient, way to add array-like types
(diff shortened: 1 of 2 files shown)


array.go
```diff
@@ -34,6 +38,110 @@ func Bools(key string, bs []bool) zapcore.Field {
...
+}
+
+// Errors constructs a field that carries a slice of errors.
+func Errors(key string, errs []error) zapcore.Field {
+	return Array(key, errArray(errs))
+}
+
+// Durations constructs a field that carries a slice of time.Durations. Like
+// Duration, it represents each duration as an integer number of nanoseconds.
+func Durations(key string, ds []time.Duration) zapcore.Field {
```
jcorbin: It's especially bizarre that we have this, but not the singular (nor special encoder support for it); in the vein of my comment on #258 .

akshayjshah:
Indeed!

Issue #221, opened 2017-01-03
Re-introduce terse logger instantiation


akshayjshah:
On the dev branch, we've broken the options-style constructors. We've also increased the number of types that may need to take options.

We should re-introduce a means to construct loggers conveniently but take care to allow extensions to work too.

akshayjshah:
Landed on dev in #274.

PR #264, opened 2017-02-02
Make times and durations first-class in encoders

### LINKED ISSUE #149: Clean up API for timestamp formatting in JSON encoder and text encoder
See https://github.com/uber-go/zap/pull/123#r73206733 for context.

Currently, we have EpochFormatter, NoTime, RFC3339Formatter, TextNoTime, and TextTimeFormat.

We can create a separate option for setting the timestamp key in the JSON encoder, and share the time formatters between the JSON encoder and the text encoder.

### LINKED ISSUE #259: Add first-class encoder support for times and durations
Remove the `Time` field constructor in favor of more-explicit `Epoch` (floating-point seconds since epoch) and `ISO8601` constructors.

If necessary, we can also include `EpochMillis` (int64 milliseconds since epoch).

akshayjshah:
This PR resolves #259 in two steps.

The first commit is a pre-factoring. It makes encoder configuration more modular, splitting configuration of message/name/timestamp keys from the formatting of the values. (This was enabled by our work on array support.) It paves the way to fix #149 too. Two things to note about that change:

1. It highlights our need for a user-facing defaults + functional options paradigm (#221). Once we have one of those, most of the long struct declarations introduced here will go away.
1. In zapcore, I didn't attempt to guard against nil formatters; I'd rather let the user-facing zap package define safe defaults (also part of #221).

After pre-factoring, the second commit is quite small. It makes durations and times first-class types in encoders (including arrays of each).


benchmarks/zap_bench_test.go
```diff
@@ -54,27 +54,15 @@ var _jane = user{
...
-		TimeFormatter:    timeF,
-		LevelFormatter:   levelF,
-		NameFormatter:    nameF,
+	return zapcore.NewJSONEncoder(zapcore.EncoderConfig{
+		MessageKey:        "msg",
+		LevelKey:          "level",
+		TimeKey:           "ts",
+		CallerKey:         "caller",
+		StacktraceKey:     "stacktrace",
+		TimeFormatter:     func(t time.Time, enc zapcore.ArrayEncoder) { enc.AppendInt64(t.UnixNano() / int64(time.Millisecond)) },
```
jcorbin: I feel like these functions could be useful, maybe even exported, vars in zapcore, e.g.: `func TimeEncodedAsEpochMS(t time.Time, enc zapcore.ArrayEncoder) { enc.AppendInt64(t.UnixNano() / int64(time.Millisecond)) }`.  Also, I feel that such a default should be floating point ms, not truncated int ms.

zapcore/encoder.go
```diff
@@ -20,6 +20,21 @@
...
+
+// An EncoderConfig allows users to configure the concrete encoders supplied by
+// zapcore.
+type EncoderConfig struct {
+	// Set the keys used for each log entry.
+	MessageKey, LevelKey, TimeKey, NameKey, CallerKey, StacktraceKey string
+	// Configure the primitive representations of common complex types. For
+	// example, some users may want all time.Times serialized as floating-point
+	// seconds since epoch, while others may prefer ISO8601 strings.
+	LevelFormatter    func(Level, ArrayEncoder)
```
jcorbin: nit: why not keep with the "Encoder" nomenclature, and call these e.g. `LevelEncoder`?

Or, since they're raw functions, the lest KoN'd form would be `func EncodeLevel(Level, ArrayEncoder)`.

zapcore/json_encoder.go
```diff
@@ -60,10 +52,10 @@ type jsonEncoder struct {
 // libraries will ignore duplicate key-value pairs (typically keeping the last
 // pair) when unmarshaling, but users should attempt to avoid adding duplicate
 // keys.
-func NewJSONEncoder(cfg JSONConfig) Encoder {
+func NewJSONEncoder(cfg EncoderConfig) Encoder {
```
jcorbin: I guess it's mildly weird that we store a pointer to a passed value, rather than letting the caller pass a pointer (e.g. so that multiple json-encoded facilities can share the same config). Doubly so if this new config achieves it's namesake of being the config for all encoders, for sharing b/w console and json.

zapcore/json_encoder.go
```diff
@@ -193,51 +195,61 @@ func (enc *jsonEncoder) AddUint8(k string, v uint8)         { enc.AddUint64(k, u
...
 	}
 
 	final.bytes = append(final.bytes, '{')
-	final.LevelFormatter(ent.Level).AddTo(final)
-	final.TimeFormatter(ent.Time).AddTo(final)
-	final.NameFormatter(ent.LoggerName).AddTo(final)
-	if ent.Caller.Defined {
+	if final.LevelKey != "" {
+		final.addKey(final.LevelKey)
+		final.LevelFormatter(ent.Level, final)
```
jcorbin: nit: any of `FormatLevel`, `EncodeLevel`, or `AppendLevel` would fit better as the internal field name.

zapcore/json_encoder.go
```diff
@@ -193,51 +195,61 @@ func (enc *jsonEncoder) AddUint8(k string, v uint8)         { enc.AddUint64(k, u
...
+		final.addKey(enc.MessageKey)
+		final.AppendString(ent.Message)
 	}
-	final.MessageFormatter(ent.Message).AddTo(final)
 	if len(enc.bytes) > 0 {
-		if len(final.bytes) > 1 {
-			// All the formatters may have been no-ops.
-			final.bytes = append(final.bytes, ',')
-		}
+		final.separateElements()
```
jcorbin: nit: `appendElementSeparator` (or `addElementSeparator` if you prefer)

benchmarks/zap_bench_test.go
akshayjshah: Agreed, but I'd like to defer exporting such functions until we do config; I suspect that the two will be closely related.

I'm all for changing the default (agreed that what we have is weird). I'll do that in a separate PR because it introduces lots of test assertion changes.

zapcore/encoder.go
akshayjshah: `EncodeFoo` sounds good. `LevelEncoder` sounds like `JSONEncoder`, but is a whole different beast.

zapcore/json_encoder.go
akshayjshah: Accepting a pointer to config suggests, to me, that the config is optional. In this case, it's not.

In the case when we have both JSON and console encoders running, I don't think they'll share configs - they'll want different time and level formatters, at least.

zapcore/json_encoder.go
akshayjshah: 👍

zapcore/json_encoder.go
akshayjshah: :+1:

PR #265, opened 2017-02-02
Add support for logging to isolated namespaces

### LINKED ISSUE #132: Support nested loggers that can only add context to a nested field
Copying from https://github.com/uber-go/zap/issues/81#issuecomment-235629205

Currently a logger always has a reference to the top level fields. This is more likely to cause field clashes as the logger is passed between multiple libraries. 

We could add support for a logger that only had a reference to a nested field:

``` go
logger := New(...)
_ = newConnection(logger.Nested("conn"))
_ = newRequest(logger.Nested("request"))
```

The nested logger would not be able to add anything to the top level fields, so you could pass in the logger to other components (internal or external) without having to worry about clashes.

akshayjshah:
Add an `OpenNamespace` method to encoders and create a `NamespaceType` field. This allows applications to force sub-components and third-party libraries to log into an isolated namespace, preventing key collisions. The approach taken here emphasizes that fields are just messages to encoders, and it combines well with #251 to let the sugared logger also support namespacing.

Along the way, this PR removes the `Nest` field constructor and encoder method, since they haven't proved useful in production and will inevitably create confusion with namespaces. If users need that functionality, they can still use `Object("foo", Fields{someField, someOtherField})`.

This fixes #132.

### DIFF
--- a/zapcore/json_encoder.go
+++ b/zapcore/json_encoder.go
@@ -39,7 +39,8 @@ const (
 
 type jsonEncoder struct {
 	*EncoderConfig
-	bytes []byte
+	bytes          []byte
+	openNamespaces int
 }
 
 // NewJSONEncoder creates a fast, low-allocation JSON encoder. The encoder
@@ -102,6 +103,12 @@ func (enc *jsonEncoder) AddReflected(key string, obj interface{}) error {
 	enc.addKey(key)
 	enc.bytes = append(enc.bytes, marshaled...)
 	return nil
+}
+
+func (enc *jsonEncoder) OpenNamespace(key string) {
+	enc.addKey(key)
+	enc.bytes = append(enc.bytes, '{')
+	enc.openNamespaces++
 }
 
 func (enc *jsonEncoder) AddString(key, val string) {
@@ -211,21 +218,18 @@ func (enc *jsonEncoder) AppendUint8(v uint8)                { enc.AppendUint64(u
 func (enc *jsonEncoder) AppendUintptr(v uintptr)            { enc.AppendUint64(uint64(v)) }
 
 func (enc *jsonEncoder) Clone() Encoder {
-	clone := &jsonEncoder{EncoderConfig: enc.EncoderConfig}
+	clone := *enc
 	clone.bytes = append(buffers.Get(), enc.bytes...)
-	return clone
+	return &clone
 }
 
 func (enc *jsonEncoder) EncodeEntry(ent Entry, fields []Field) ([]byte, error) {
-	final := &jsonEncoder{
-		EncoderConfig: enc.EncoderConfig,
-		bytes:         buffers.Get(),
-	}
+	final := *enc
+	final.bytes = append(buffers.Get(), '{')
 
-	final.bytes = append(final.bytes, '{')
 	if final.LevelKey != "" {
 		final.addKey(final.LevelKey)
-		final.EncodeLevel(ent.Level, final)
+		final.EncodeLevel(ent.Level, &final)
 	}
 	if final.TimeKey != "" {
 		final.AddTime(final.TimeKey, ent.Time)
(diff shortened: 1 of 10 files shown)


zapcore/json_encoder.go
```diff
@@ -207,15 +214,16 @@ func (enc *jsonEncoder) AppendUint8(v uint8)                { enc.AppendUint64(u
 func (enc *jsonEncoder) AppendUintptr(v uintptr)            { enc.AppendUint64(uint64(v)) }
 
 func (enc *jsonEncoder) Clone() Encoder {
-	clone := &jsonEncoder{JSONConfig: enc.JSONConfig}
+	clone := &jsonEncoder{JSONConfig: enc.JSONConfig, openNamespaces: enc.openNamespaces}
```
jcorbin: can we simplify to the form:
```
clone := *enc
...
return &clone
```
?

zapcore/json_encoder.go
```diff
@@ -207,15 +214,16 @@ func (enc *jsonEncoder) AppendUint8(v uint8)                { enc.AppendUint64(u
...
 	clone.bytes = append(buffers.Get(), enc.bytes...)
 	return clone
 }
 
 func (enc *jsonEncoder) EncodeEntry(ent Entry, fields []Field) ([]byte, error) {
 	final := &jsonEncoder{
-		JSONConfig: enc.JSONConfig,
-		bytes:      buffers.Get(),
+		JSONConfig:     enc.JSONConfig,
```
jcorbin: maybe also:
```
final := *enc
final.bytes = buffers.Get()
```

zapcore/json_encoder_impl_test.go
```diff
@@ -191,6 +191,18 @@ func TestJSONEncoderObjectFields(t *testing.T) {
 				assert.Error(t, e.AddReflected("k", noJSON{}), "Unexpected success JSON-serializing a noJSON.")
 			},
 		},
+		{
+			desc: "namespace",
```
jcorbin: probably useful in covering more cases, such as:
- `ns1`
- `ns1, val`
- `ns1, ns2, val`
- `ns1, val, ns2, val2 // this is the one you have now`

I guess if you want fewer cases, let's have: `ns1, ns2, val, ns3, val2, ns4` since it captures:
- a ns with none of its own fields
- an empty leaf ns
- still capture values at multiple levels

zapcore/json_encoder_test.go
```diff
@@ -100,3 +100,17 @@ func TestJSONEncodeEntry(t *testing.T) {
 		)
 	})
 }
+
+func TestJSONEncodeEntryClosesNamespaces(t *testing.T) {
```
jcorbin: I'm confused as to why this isn't a case in the above?

zapcore/memory_encoder.go
```diff
@@ -23,86 +23,105 @@ package zapcore
 // MapObjectEncoder is an ObjectEncoder backed by a simple
 // map[string]interface{}. It's not fast enough for production use, but it's
 // helpful in tests.
-type MapObjectEncoder map[string]interface{}
+type MapObjectEncoder struct {
+	Fields map[string]interface{}
```
jcorbin: Can I get some comments here, I think the way it works:
```
// Fields is all of the encoded fields; the top-level
Fields ...

// cur is where we're currently adding fields, maybe to a sub-scope
cur ...
```

zapcore/json_encoder_test.go
akshayjshah: The previous test doesn't actually write an Entry, since that simplifies most of the assertions.

zapcore/json_encoder_impl_test.go
akshayjshah: :+1:

zapcore/json_encoder.go
akshayjshah: Sure.

zapcore/memory_encoder.go
akshayjshah: :+1:

zapcore/json_encoder_test.go
akshayjshah: It's a little gnarly. Prefer to leave as-is.

PR #266, opened 2017-02-02
Add zapglobal package to allow zap to be used as a global logger

peter-edge:


akshayjshah:
All the logger methods are safe to call concurrently, so the only reason to need locks is to reconfigure the logger. What do you think of instead putting that burden on the application developer? I don't think it's nuts to require that `func main()` configure the global logger (if necessary) before starting any other goroutines.

This would let us have have `zap.L` and `zap.S` be a global `zap.Logger` and `zap.SugaredLogger`, respectively - that saves us from needing a separate package just to house these top-level floating functions.

peter-edge:
That's fine, I'll send a new (2 line probably) PR tomorrow or tonight.

peter-edge:
Also I completely agree in general it's a terrible antipattern, but it's one I have a hard time fighting against.

Also if you like DI, check out my go copy of guice :) go.pedge.io/inject

PR #267, opened 2017-02-03
Add a console encoder

akshayjshah:
Re-introduce the text encoder, but be a bit clearer - it's a *console* encoder, designed for humans looking at a TTY.

The console encoder serializes the Entry's timestamp, level, logger name, caller, message, context, and stacktrace like this:

```
2017-02-01T11:59:59Z  DEBUG  main.ringpop@gossip.go:56 Pinged peer. {"hostport": "1.2.3.4:5678"}
2017-02-01T12:00:00Z  PANIC  main.ringpop@gossip.go:72 Nil peer! {"some": "context"}
syscall.Syscall(0x3, 0x4, 0xc4204961bb, 0x445, 0x1, 0x0, 0x0)
	/usr/local/Cellar/go/1.7.4_2/libexec/src/syscall/asm_darwin_amd64.s:16 +0x5
syscall.read(0x4, 0xc4204961bb, 0x445, 0x445, 0x5a, 0x0, 0x0)
	/usr/local/Cellar/go/1.7.4_2/libexec/src/syscall/zsyscall_darwin_amd64.go:973 +0x55
syscall.Read(0x4, 0xc4204961bb, 0x445, 0x445, 0x5a, 0x0, 0x0)
	/usr/local/Cellar/go/1.7.4_2/libexec/src/syscall/syscall_unix.go:161 +0x49
```

This approach keeps the core Entry content (message, timestamp, level, etc.) in a plain-text, tab-aligned format, prints stacktraces as familiar multi-line dumps, and doesn't attempt to invent an ad-hoc encoding for the structured context. Instead, just use JSON (with some spacing for readability) - it already strikes a pretty good balance between human- and machine-friendliness.

The end result re-uses nearly all of the existing JSON encoder. While it's slower than zap's JSON encoder, it's way faster than the standard library.

```
BenchmarkZapJSON-4   	         2000000	       854 ns/op	    1312 B/op	       6 allocs/op
BenchmarkZapConsole-4   	 1000000	      1699 ns/op	    1532 B/op	      13 allocs/op
BenchmarkStandardJSON-4   	  300000	      4039 ns/op	    1920 B/op	      38 allocs/op
```

### DIFF
--- a/zapcore/console_encoder.go
+++ b/zapcore/console_encoder.go
@@ -24,9 +24,25 @@ import (
 	"bytes"
 	"fmt"
 	"strconv"
+	"sync"
 
 	"go.uber.org/zap/internal/buffers"
 )
+
+var _sliceEncoderPool = sync.Pool{
+	New: func() interface{} {
+		return &sliceArrayEncoder{elems: make([]interface{}, 0, 2)}
+	},
+}
+
+func getSliceEncoder() *sliceArrayEncoder {
+	return _sliceEncoderPool.Get().(*sliceArrayEncoder)
+}
+
+func putSliceEncoder(e *sliceArrayEncoder) {
+	e.elems = e.elems[:0]
+	_sliceEncoderPool.Put(e)
+}
 
 type consoleEncoder struct {
 	*jsonEncoder
@@ -56,7 +72,10 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) ([]byte, error) {
 	// We don't want the date and level to be quoted and escaped (if they're
 	// encoded as strings), which means that we can't use the JSON encoder. The
 	// simplest option is to use the memory encoder and fmt.Fprint.
-	arr := &sliceArrayEncoder{elems: make([]interface{}, 0, 2)}
+	//
+	// If this ever becomes a performance bottleneck, we can implement
+	// ArrayEncoder for our plain-text format.
+	arr := getSliceEncoder()
 	if c.TimeKey != "" {
 		c.EncodeTime(ent.Time, arr)
 	}
@@ -69,6 +88,7 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) ([]byte, error) {
 		}
 		fmt.Fprint(line, arr.elems[i])
 	}
+	putSliceEncoder(arr)
 
 	// Compose the logger name and caller info into a call site and add it.
 	c.writeCallSite(line, ent.LoggerName, ent.Caller)
--- a/zapcore/console_encoder.go
+++ b/zapcore/console_encoder.go
@@ -1,0 +1,138 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zapcore
+
+import (
+	"bytes"
+	"fmt"
+	"strconv"
+
+	"go.uber.org/zap/internal/buffers"
+)
+
+type consoleEncoder struct {
+	*jsonEncoder
+}
+
+// NewConsoleEncoder creates an encoder whose output is designed for human -
+// rather than machine - consumption. It serializes the core log entry data
+// (message, level, timestamp, etc.) in a plain-text format and leaves the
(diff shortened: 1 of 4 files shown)


ansel1:
counter argument for using the jsonEncoder for the fields:

We log a lot of multi-line fields at debug level.  For example, dumping http requests and responses (inc. bodies), .pem files, routing tables, and stacktraces (not stacktraces from the log site, but from error sites).

For machine-processed logs, encoding the line breaks via json or LTSV such that all logs are a single line makes a lot of sense.  But not console logs.  As you've conceded to, with the multi-line log-site stacktraces you print.

I'd propose console logs should preserve multiline fields as-is, making json encoding inappropriate.  I'd suggest using quasi-LTSV encoding for the fields.  By "quasi", I mean it would relax some of LTSV's requirements, like encoding newlines and tabs.  So it wouldn't really be machine-parseable LTSV, but would be more human-friendly than json.

In fact, I wonder if you could just tweak this implementation to allow any Encoder to be used for the fields (rather than hard coding the jsonEncoder).  When the LTSV encoder is ported to zap 1.0, it can be swapped in by the developer in place of the json encoder.

akshayjshah:
@ansel1 I'm totally sympathetic to that use case. However, this is one of the places where every team and company are going to have strong preferences - we're just not going to find a solution that everyone's happy with.

My goal for this encoder is to have something that's *mostly* single-line, apart from stack traces at the log site, while being somewhat more readable than plain JSON. Ideally, I'd also like `go test -v ./... | grep someFailureMessage` to be useful, which makes multiline output less attractive.

That said, I'm open to *also* including a multi-line encoder that satisfies the use cases you're describing. We can always ship that after 1.0 as `NewMultilineEncoder` or something similar. When/if we get to that point, we can refactor the current console encoder to enable code reuse.

jcorbin:
@akshayjshah +1 on the shape of output, I'll come back and do an actual code review Soon ™

zapcore/console_encoder.go
```diff
@@ -0,0 +1,158 @@
...
+	// If this ever becomes a performance bottleneck, we can implement
+	// ArrayEncoder for our plain-text format.
+	arr := getSliceEncoder()
+	if c.TimeKey != "" {
+		c.EncodeTime(ent.Time, arr)
+	}
+	if c.LevelKey != "" {
+		c.EncodeLevel(ent.Level, arr)
+	}
+	for i := range arr.elems {
```
akshayjshah: Because these are user-supplied functions, we don't necessarily have both a time and a level - it's easy enough to pass `func(time.Time, ArrayEncoder) {}`.

I should add tests for these situations, though.

zapcore/console_encoder_bench_test.go
```diff
@@ -0,0 +1,50 @@
...
+func BenchmarkZapConsole(b *testing.B) {
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			enc := NewConsoleEncoder(humanEncoderConfig())
+			enc.AddString("str", "foo")
+			enc.AddInt64("int64-1", 1)
+			enc.AddInt64("int64-2", 2)
+			enc.AddFloat64("float64", 1.0)
+			enc.AddString("string1", "\n")
+			enc.AddString("string2", "💩")
```
akshayjshah: They're specifically testing astral-plane unicode handling - `string2` is a pile of poo, `string3` is the unsure face, and `string4` is the say-no-evil monkey.

zapcore/console_encoder.go
```diff
@@ -0,0 +1,158 @@
...
+		line.WriteByte(':')
+		line.WriteString(strconv.FormatInt(int64(caller.Line), 10))
+	}
+}
+
+func (c consoleEncoder) writeContext(line *bytes.Buffer, extra []Field) {
+	context := c.jsonEncoder.Clone().(*jsonEncoder)
+	defer buffers.Put(context.bytes)
+
+	for i := range extra {
```
akshayjshah: :+1:

zapcore/console_encoder.go
```diff
@@ -0,0 +1,158 @@
...
+		extra[i].AddTo(context)
+	}
+	context.closeOpenNamespaces()
+	if len(context.bytes) == 0 {
+		return
+	}
+
+	if line.Len() > 0 {
```
akshayjshah: Sure, makes sense.

zapcore/console_encoder.go
```diff
@@ -0,0 +1,158 @@
...
+// NewConsoleEncoder creates an encoder whose output is designed for human -
+// rather than machine - consumption. It serializes the core log entry data
+// (message, level, timestamp, etc.) in a plain-text format and leaves the
+// structured context as JSON.
+//
+// Note that although the console encoder doesn't use the keys specified in the
+// encoder configuration, it will omit any element whose key is set to the empty
+// string.
+func NewConsoleEncoder(cfg EncoderConfig) Encoder {
+	j := NewJSONEncoder(cfg).(*jsonEncoder)
```
akshayjshah: :+1:

zapcore/console_encoder.go
akshayjshah: Predictably, the tests exposed the fact that I hadn't thought through this all the way. Fixed both the JSON and console encoders to handle this situation reasonably.

zapcore/encoder_test.go
```diff
@@ -218,97 +257,59 @@ func TestJSONEncoderConfiguration(t *testing.T) {
...
-		buf, err := enc.EncodeEntry(Entry{
-			LoggerName: "",
-			Level:      InfoLevel,
-			Message:    `hello\`,
-			Time:       epoch,
-		}, nil)
-		assert.NoError(t, err, "EncodeEntry returned an unexpected error.")
-		assert.Equal(
-			t,
-			`{"level":"info","ts":0,"msg":"hello\\","foo":"bar"}`+"\n",
```
akshayjshah: Escaping in general is, but I'll amend the test above to make sure we're also escaping the message.

zapcore/console_encoder_bench_test.go
prashantv: Haha, thanks for the detailed info. GitHub is acting a little strange, I can see the emoji on the main "Conversation" tab,
image

And yet it just disappears on the "Files Changed" tab:
image

zapcore/console_encoder.go
prashantv: Ahh, makes sense

zapcore/console_encoder_bench_test.go
akshayjshah: I'll file a ticket. Defcon 1, emoji support is compromised!

akshayjshah: Hrm. Turns out that it's the Sourcegraph extension - everything's fine if you disable that extension and reload the page.

PR #273, opened 2017-02-03
JSON-serialize pointers as hex

akshayjshah:
This is a small change to encoder `uintptr` as hex, which is a bit more familiar and matches the standard lib's stacktraces. This seems like an improvement to me, but I can certainly be convinced to abandon it.


akshayjshah:
Meh - I'm not too attached to this, and I'd rather keep the text and JSON representations the same (since we're just using JSON in the console encoder).

PR #274, opened 2017-02-04
Re-introduce terse configuration

akshayjshah:
This PR re-introduces a nice, terse way to configure sane-by-default loggers. It also introduces a configuration struct for loggers, since many applications want to drive logger configuration from config instead of code.


zapcore/encoder.go
```diff
@@ -80,16 +182,32 @@ type ObjectEncoder interface {
 // arrays even though they aren't typical in Go. Like slices, ArrayEncoders
 // aren't safe for concurrent use (though typical use shouldn't require locks).
 type ArrayEncoder interface {
+	// Built-in types
+	PrimitiveArrayEncoder
```
akshayjshah: I'll break out this change into a separate PR - it was an oversight on my part. Passing an `ArrayEncoder` to a `TimeEncoder` opens us up to infinite recursion :/

config.go
```diff
@@ -0,0 +1,213 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"go.uber.org/zap/zapcore"
+)
+
+// SamplingConfig sets a sampling strategy for the logger. Sampling caps the
+// global CPU and I/O load that logging puts on your process while preserving a
```
jcorbin: I'd prefer to soften this to "while attempting to preserve a".

config.go
```diff
@@ -0,0 +1,213 @@
...
+type Config struct {
+	// Level is the minimum enabled logging level. Note that this is a dynamic
+	// level, so calling Config.Level.SetLevel will atomically change the log
+	// level of all loggers descended from this config. The zero value is
+	// InfoLevel.
+	Level AtomicLevel `json:"level",yaml:"level"`
+	// Development puts the logger in development mode, which changes the
+	// behavior of DPanicLevel and takes stacktraces more liberally.
+	Development bool `json:"development",yaml:"development"`
+	// DisableCaller stops annotating logs with the calling function's file
```
jcorbin: Say more about when it happens if not enabled, always? Only for warn and up?

config.go
```diff
@@ -0,0 +1,213 @@
...
+	// name and line number.
+	DisableCaller bool `json:"disable_caller",yaml:"disable_caller"`
+	// DisableStacktrace completely disables automatic stacktrace capturing.
```
jcorbin: Say more about when it happens if not enabled, always? Only for error and up?

config.go
```diff
@@ -0,0 +1,213 @@
...
+//
+// It doesn't do anything that can't be done with New, Options, and the various
+// zapcore.WriteSyncer and zapcore.Facility wrappers, but it's a simpler way to
+// toggle common options.
+type Config struct {
+	// Level is the minimum enabled logging level. Note that this is a dynamic
+	// level, so calling Config.Level.SetLevel will atomically change the log
+	// level of all loggers descended from this config. The zero value is
+	// InfoLevel.
+	Level AtomicLevel `json:"level",yaml:"level"`
```
jcorbin: How about a:
```
Level Level
AllowLevelChange bool
```

And defaulting to static levels?

config.go
```diff
@@ -0,0 +1,213 @@
...
+	"time"
+
+	"go.uber.org/zap/zapcore"
+)
+
+// SamplingConfig sets a sampling strategy for the logger. Sampling caps the
+// global CPU and I/O load that logging puts on your process while preserving a
+// representative subset of your logs.
+//
+// See zapcore.Sample for details.
```
prashantv: Should we mention that these values are per-second?

config.go
```diff
@@ -0,0 +1,213 @@
...
+	// level, so calling Config.Level.SetLevel will atomically change the log
+	// level of all loggers descended from this config. The zero value is
+	// InfoLevel.
+	Level AtomicLevel `json:"level",yaml:"level"`
+	// Development puts the logger in development mode, which changes the
+	// behavior of DPanicLevel and takes stacktraces more liberally.
+	Development bool `json:"development",yaml:"development"`
+	// DisableCaller stops annotating logs with the calling function's file
+	// name and line number.
+	DisableCaller bool `json:"disable_caller",yaml:"disable_caller"`
```
prashantv: Any reason for snake case in JSON? I tend to use the google JSON style guide which suggests `camelCase`
https://google.github.io/styleguide/jsoncstyleguide.xml#Property_Name_Format

I'm not sure what's appropriate for YAML, don't think there's any standard, I've seen `hyphernated-case` as well as `camelCase` and `snake_case`

config.go
```diff
@@ -0,0 +1,213 @@
...
+	// DisableStacktrace completely disables automatic stacktrace capturing.
+	DisableStacktrace bool `json:"disable_stacktrace",yaml:"disable_stacktrace"`
+	// Sampling sets a sampling policy. A nil SamplingConfig disables sampling.
+	Sampling *SamplingConfig `json:"sampling",yaml:"sampling"`
+	// Encoding sets the logger's encoding. Valid values are "json" and
```
jcorbin: Interesting, I would've expected more of a:
```
EnableConsole bool // maybe disable instead of enable
DisableStandardErrors bool
FileEncoding string
OutputPaths []string
ErrorOutputPaths []string
```

The idea being:
- stdout is always console encoded
- files are always json encoded
- errors at least go to stderr, and maybe also a file

config.go
prashantv: Can you add information about when stack traces are captured (looks like error or higher for prod, and warn or higher for dev)

config.go
```diff
@@ -0,0 +1,213 @@
...
+	dyn := DynamicLevel()
+	dyn.SetLevel(DebugLevel)
+
+	return Config{
+		Level:       dyn,
+		Development: true,
+		Encoding:    "console",
+		EncoderConfig: zapcore.EncoderConfig{
+			// Keys can be anything except the empty string.
+			TimeKey:        "T",
```
jcorbin: These are some surprising defaults imo, but I guess not that the console encoder uses them much?

Apropos of my above remarks, I think I'd still expect a "development" or "staging" deploy to log json to files, and I'd still want to see warn and up on stdout for things like docker logs in production...

config.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// Build constructs a logger from the Config and Options.
+func (cfg Config) Build(opts ...Option) (*Logger, error) {
+	sink, errSink, err := cfg.openSinks()
+	if err != nil {
+		return nil, err
+	}
+
+	enc, err := cfg.buildEncoder()
```
jcorbin: How about building the encoder first, before opening the sinks to minimize open file handles under error branch?

config.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return New(
+		zapcore.WriterFacility(enc, sink, cfg.Level),
+		cfg.buildOptions(errSink)...,
+	).WithOptions(opts...), nil
```
jcorbin: Much utility in not doing the `WithOptions` if empty?

config.go
```diff
@@ -0,0 +1,213 @@
...
+
+	if cfg.Sampling != nil {
+		opts = append(opts, WrapFacility(func(fac zapcore.Facility) zapcore.Facility {
+			return zapcore.Sample(fac, time.Second, int(cfg.Sampling.Initial), int(cfg.Sampling.Thereafter))
+		}))
+	}
+
+	if len(cfg.InitialFields) > 0 {
+		fs := make([]zapcore.Field, 0, len(cfg.InitialFields))
+		for k, v := range cfg.InitialFields {
```
jcorbin: Is that a `map[string]interface{}`? Shouldn't we provide a stable field order by sorting the keys?

config.go
```diff
@@ -0,0 +1,213 @@
...
+	sink, err := Open(cfg.OutputPaths...)
+	if err != nil {
+		return nil, nil, err
+	}
+	errSink, err := Open(cfg.ErrorOutputPaths...)
+	if err != nil {
+		return nil, nil, err
+	}
+	return sink, errSink, nil
+}
```
jcorbin: I'd feel mildly better if we could ensure closing happens under errors... but actually doesn't Go setup a finalizer for that by default? Maybe my concerns are baseless?

config_test.go
```diff
@@ -0,0 +1,86 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			temp, err := ioutil.TempFile("", "zap-prod-config-test")
+			require.NoError(t, err, "Failed to create temp file.")
+			defer os.Remove(temp.Name())
+
+			tt.cfg.OutputPaths = []string{temp.Name()}
```
jcorbin: I always feel wary when I see a test fixture mutation...

config.go
```diff
@@ -0,0 +1,213 @@
...
+			LevelKey:       "level",
+			NameKey:        "logger",
+			CallerKey:      "caller",
+			MessageKey:     "msg",
+			StacktraceKey:  "stacktrace",
+			EncodeLevel:    zapcore.LowercaseLevelEncoder,
+			EncodeTime:     zapcore.EpochTimeEncoder,
+			EncodeDuration: zapcore.SecondsDurationEncoder,
+		},
+		OutputPaths:      []string{"stdout"},
```
prashantv: Should the default output for logs be stderr?
logrus seems to use stderr by default: https://godoc.org/github.com/sirupsen/logrus#Logger
As does the standard go `log` package.

(Ideally production configs would actually use a file, but that's not a great default)

config_test.go
```diff
@@ -0,0 +1,86 @@
...
+			tt.cfg.EncoderConfig.TimeKey = "" // no timestamps in tests
+			tt.cfg.InitialFields = map[string]interface{}{"k": "v"}
```
jcorbin: And if you add a second field, this becomes unstable (due to map ordering per above comment)

config.go
```diff
@@ -0,0 +1,213 @@
...
+			LevelKey:       "L",
+			NameKey:        "N",
+			CallerKey:      "C",
+			MessageKey:     "M",
+			StacktraceKey:  "S",
+			EncodeLevel:    zapcore.CapitalLevelEncoder,
+			EncodeTime:     zapcore.ISO8601TimeEncoder,
+			EncodeDuration: zapcore.StringDurationEncoder,
+		},
+		OutputPaths:      []string{"stdout"},
```
prashantv: same as earlier, might want to use stderr for all logs

level.go
```diff
@@ -87,3 +87,19 @@ func (lvl AtomicLevel) Level() zapcore.Level {
 func (lvl AtomicLevel) SetLevel(l zapcore.Level) {
 	lvl.l.Store(int32(l))
 }
+
+// UnmarshalText unmarshals the text to a DynamicLevel. It uses the same text
+// representation as zapcore.Level ("debug", "info", "warn", "error", "dpanic",
+// "panic", and "fatal").
+func (lvl *AtomicLevel) UnmarshalText(text []byte) error {
+	if lvl.l == nil {
+		lvl.l = atomic.NewInt32(int32(InfoLevel))
```
jcorbin: Prefer to defer thus until after we've parsed without err.

level.go
```diff
@@ -87,3 +87,19 @@ func (lvl AtomicLevel) Level() zapcore.Level {
...
+	}
+
+	var l zapcore.Level
+	if err := l.UnmarshalText(text); err != nil {
+		return err
+	}
+	lvl.SetLevel(l)
```
jcorbin: Prefer to:
```
if lvl.l == nil {
  lvl.l = atomic.NewInt32(int32(l))
} else {
  lvl.SetLevel(l)
}
```

config.go
```diff
@@ -0,0 +1,213 @@
...
+		opts = append(opts, Development())
+	}
+
+	if !cfg.DisableCaller {
+		opts = append(opts, AddCaller())
+	}
+
+	stackLevel := ErrorLevel
+	if cfg.Development {
+		stackLevel = WarnLevel
```
prashantv: optional: can move the `Development()` option append here

config.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	if !cfg.DisableStacktrace {
+		opts = append(opts, AddStacktrace(stackLevel))
+	}
+
+	if cfg.Sampling != nil {
+		opts = append(opts, WrapFacility(func(fac zapcore.Facility) zapcore.Facility {
```
prashantv: (just a question) is there a difference between wrapping the facility when passing it into `New` vs calling `WrapFacility`?

logger.go
```diff
@@ -105,6 +99,16 @@ func (log *Logger) Named(s string) *Logger {
 	return l
 }
 
+// WithOptions clones the current Logger, applies the supplied Options, and
+// returns the result. It's safe to use concurrently.
```
jcorbin: "... and returns the clone." (... "cloned child logger." ?)

writer.go
```diff
@@ -0,0 +1,64 @@
...
+// WriteSyncer.
+//
+// Passing no paths returns a no-op WriteSyncer. The special paths "stdout" and
+// "stderr" are interpreted as os.Stdout and os.Stderr, respectively.
+func Open(paths ...string) (zapcore.WriteSyncer, error) {
+	if len(paths) == 0 {
+		return zapcore.AddSync(ioutil.Discard), nil
+	}
+	writers, err := open(paths)
+	return zapcore.Lock(zapcore.MultiWriteSyncer(writers...)), err
```
prashantv: Should `MultiWriteSyncer` always lock?

writer.go
```diff
@@ -0,0 +1,64 @@
...
+// Open is a high-level wrapper that takes a variadic number of paths, opens or
+// creates each of the specified files, and combines them into a locked
+// WriteSyncer.
+//
+// Passing no paths returns a no-op WriteSyncer. The special paths "stdout" and
+// "stderr" are interpreted as os.Stdout and os.Stderr, respectively.
+func Open(paths ...string) (zapcore.WriteSyncer, error) {
+	if len(paths) == 0 {
+		return zapcore.AddSync(ioutil.Discard), nil
+	}
```
prashantv: can we special case `len(paths) == 1` and return the `WriteSyncer` directly

writer.go
```diff
@@ -0,0 +1,64 @@
...
+	for _, path := range paths {
+		switch path {
+		case "stdout":
+			writers = append(writers, os.Stdout)
+			continue
+		case "stderr":
+			writers = append(writers, os.Stderr)
+			continue
+		}
+		f, err := os.OpenFile(path, os.O_RDWR|os.O_APPEND|os.O_CREATE, 0644)
```
jcorbin: if this were a `default:`, you wouldn't need those `continue`s

prashantv: Since we never read the file, I think we can use `O_WRONLY`.

For the mask, it is then umasked, so create uses `0666`:
https://golang.org/src/os/file.go?s=7623:7662#L248

We should probably do the same

zapcore/sampler.go
```diff
@@ -65,7 +65,13 @@ func fnv32a(s string) uint32 {
 	return hash
 }
 
-// Sample creates a facility that samples incoming entries.
+// Sample creates a Facility that samples incoming entries, which caps the CPU
+// and I/O load of logging while preserving a representative subset of your
+// logs.
```
prashantv: should we mention the "optimized for performance over correctness" here as well?

config.go
akshayjshah: :+1:

zapcore/sampler.go
```diff
@@ -65,7 +65,13 @@ func fnv32a(s string) uint32 {
 	return hash
 }
 
-// Sample creates a facility that samples incoming entries.
+// Sample creates a Facility that samples incoming entries, which caps the CPU
+// and I/O load of logging while preserving a representative subset of your
```
jcorbin: nit: "attempts to preserve a"

config.go
akshayjshah: Seems like everyone just ports the convention from their primary programming language. FB's API is snake-cased, Twitter's API is snake-cased, and the portions of the AWS APIs I've used are snake-cased.

I don't really care much though. Happy to change if you have a strong preference.

config.go
akshayjshah: Sure, sounds good. This is yet another place where's Go chose to go down an unusual route :(

We'll have to disagree about prod logs - I still think they should log to standard out, and it's the process manager's job to pipe that to a file. That's the [12-factor standard](https://12factor.net/logs).

config.go
akshayjshah: Nope. This was just a bit easier organizationally - otherwise creating the facility needed its own helper method.

writer.go
akshayjshah: Not sure exactly what you're asking here. Since we're opening paths, the WriteSyncers in question are files, so we always need a lock.

In general, though, I don't think MultiWriteSyncers should lock.

writer.go
akshayjshah: :+1:

writer.go
akshayjshah: Yes, but then I'd need another level of nesting. To me, this is similar to an early return. Reasonable?

akshayjshah: :+1:

zapcore/sampler.go
akshayjshah: Sure. I was saving that until the sampler fix lands, but this is a perfectly fine time to add it.

zapcore/sampler.go
akshayjshah: :+1:

config.go
akshayjshah: :+1:

config.go
akshayjshah: Why? The performance overhead is near-zero, and I'd always recommend using a dynamic level.

config.go
akshayjshah: Good point.

config.go
akshayjshah: Great minds do indeed think alike. 👍  👍

config.go
akshayjshah: I'm glad that we can support that setup, but I'm less of a fan of making that part of the quickstart. I remembered discussing it with you and played with it for a while, and I found it pretty jarring. IMO it's just too unusual.

If and when we add new encoders, I'd also like this config struct to support them. Again, I have msgpack in production in mind.

config.go
akshayjshah: The console encoder only cares about empty/non-empty - the actual values aren't used.

Terminology varies, but I'd call anything that's "deployed" a production config. In production, you may have canary, staging, and stable, but it's all production. This is for use while developers are writing code, running unit tests, and exploring on their local machines.

config.go
akshayjshah: Sure.

config.go
akshayjshah: Sure, can do - might as well save an encoder clone.

config.go
akshayjshah: Since we're handling a lot of different things in here, I'd rather handle each of the config struct's fields in a separate little block.

config.go
akshayjshah: Sure, can do.

config_test.go
akshayjshah: Happy to assert that the field isn't set beforehand.

config_test.go
akshayjshah: Yep, good point.

config.go
akshayjshah: Certainly possible; I'll just have the `Open` function also return a `close` closure. I suppose some people will want to continue operating even if they can't open their log files, so we should be careful.

level.go
akshayjshah: :+1:

config.go
akshayjshah: :+1:

logger.go
akshayjshah: Done.

config.go
akshayjshah: Yep.

akshayjshah:
Addressed all comments except:

1. Casing of JSON/YAML keys in config struct. Not opposed to changing, but the widely-used APIs I'm familiar with prefer snake case.
2. Different encodings for different outputs. In practice, feels jarring to me; also complicates config struct and future extensibility story.

config.go
jcorbin: Ahh okay, thanks for clarifying what "development" means.

akshayjshah: Hm. Can't tell if that's mild sarcasm or not :)

Is there a better name we can use here? Not too late to rename "development" throughout the codebase.

jcorbin: Nope, wasn't meant to be sarcastic at all, what I meant:
- to me, the word "development" in zap is 90% about "will DPanic panic?'
- I can see a sensible argument for wanting a staging or canary cluster to go either way on that point
- however I'd always expect a such-deployed cluster to always log to similar channels and encoded the same way as production, the only exception being when running tests, and otherwise "local"/non-deployed development

So I'm not sure if "development" is the best name or not for "non deployed config default", and whether or not we want the ability to vary DPanic's disposition within a canned "deployed config".

Issue #275, opened 2017-02-05
Lossy nanosecond timestamps with EpochFormatter


errnoh:
While doing nanosecond level logging I noticed that some of the events were happening in wrong order.
After investigating I noticed that the EpochFormatter is doing lossy conversion due to float64 rounding errors:

(see example [here](https://play.golang.org/p/y8tRt8ebY_) )

time: 1486215887.584922537
zap : 1486215887.584922552

time: 1486215941.624710199
zap : 1486215941.624710083

While RFC3339Formatter strips the sub-second information. (time.RFC3339Nano is there for nanoseconds)

akshayjshah:
Fixed in your PR - thanks again!

PR #276, opened 2017-02-05
Add RFC3339NanoFormatter and UnixNanoFormatter

errnoh:
Adds support for two different methods of logging in nanosecond precision without losing information. 

Both are new TimeFormatters, so that no existing code will break. RFC3339NanoFormatter is similar to RFC3339Formatter but uses `time.RFC3339Nano` instead of `time.RFC3339`, and UnixNanoFormatter uses `time.Time.UnixNano()` int64 output that is fast to both serialize and read back in time.Time format.

See #275 for the issue in question.

### DIFF
--- a/json_options.go
+++ b/json_options.go
@@ -59,11 +59,27 @@ func EpochFormatter(key string) TimeFormatter {
 	})
 }
 
+// UnixNanoFormatter encodes the entry time as a single int64 with nanosecond
+// precision under the provided key.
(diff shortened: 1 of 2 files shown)


akshayjshah:
LGTM. Thanks for the contribution, @errnoh! Keep in mind that we're getting close to releasing a stable 1.0 of zap, and many of these APIs will change. I'll make sure to keep a handful of nanosecond-precision options available with the new APIs, too.

PR #277, opened 2017-02-06
Add benchmarks for lion and glide update

peter-edge:
I had a version of these when I was first looking at zap vs lion - this (and/or additional benchmarking) would be really useful for me as I'm looking at zap :)

Let me know if this is ok, I know it's an extra dependency but I've made sure it's only for testing. This also does a full glide update so that the glide.lock hash is correct.


akshayjshah:
Will look in detail later on, but I'm fine with adding lion JSON & text benchmarks for comparison.

I'm less sold on including the proto toolchain - it's interesting, but not directly comparable since zap doesn't do proto-encoded logs. Would it be easier to do that on the lion side?

peter-edge:
Short: Sure that would be fine, I'd argue this won't be a problem but you're the maintainer and I defer to you obviously :)

Long: The reason I added is it's more interesting in terms of an actual comparison, especially if we work on a structured format for zap different than JSON - JSON was just something I added for the lion equivalent of SugaredLogger. My thinking was the difference is minimal - especially if the version of go.pedge.io/lion and github.com/golang/protobuf are not updated in glide.lock, there won't be any updates needed, and if there are, all that is needed is protoc (which nowadays is easy to install). In a subsequent PR, I was going to add a Docker development environment as the SOT for testing/vendoring/etc (see https://github.com/uber-go/sally/pull/25), and proto compilation could easily be in there.

PR #278, opened 2017-02-07
Add a no-op Facility

akshayjshah:
This is a portion of our terse config and global logger story; for both, we'll
want a no-op logger, and therefore, a no-op Facility.

### DIFF
--- a/zapcore/facility.go
+++ b/zapcore/facility.go
@@ -41,6 +41,19 @@ type Facility interface {
 	Write(Entry, []Field) error
 }
 
+type nopFacility struct{}
+
+// NopFacility returns a no-op Facility.
+func NopFacility() Facility { return nopFacility{} }
+
+func (nopFacility) Enabled(Level) bool { return false }
+
+func (n nopFacility) With([]Field) Facility { return n }
+
+func (n nopFacility) Check(_ Entry, ce *CheckedEntry) *CheckedEntry { return ce }
+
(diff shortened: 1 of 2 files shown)


zapcore/facility.go
```diff
@@ -41,6 +41,19 @@ type Facility interface {
 	Write(Entry, []Field) error
 }
 
+type nopFacility struct{}
+
+// NopFacility returns a no-op Facility.
+func NopFacility() Facility { return nopFacility{} }
+
+func (nopFacility) Enabled(Level) bool { return false }
+
```
prashantv: optional: if you remove the blank lines between the methods, go will align the bodies. since the implementations are not complex, think it might be worth removing the extra whitespace

zapcore/facility.go
```diff
@@ -41,6 +41,19 @@ type Facility interface {
...
+func (n nopFacility) With([]Field) Facility { return n }
+
+func (n nopFacility) Check(_ Entry, ce *CheckedEntry) *CheckedEntry { return ce }
```
prashantv: nit: no need for named receiver here since it's unused

zapcore/facility.go
akshayjshah: Sounds good. I usually only do that when the methods have some structural similarity, but it's probably an improvement here too.

zapcore/facility.go
akshayjshah: 👍

PR #279, opened 2017-02-07
Export Level-, Time-, and DurationEncoders

akshayjshah:
This is another precursor to terse, config-based logger construction. Since we'd
like users to be able to hydrate a logger from JSON or YAML config, we must
provide a way for them to specify how they'd like levels, times, and durations
to be encoded. We'll do so by exporting a small selection of encoders and
implementing text.Unmarshaler.

Along the way, we fix a current bug that allows user-supplied encoders to recur
infinitely.

Note that this changes the default time encoding to floating-point seconds since
epoch, which is IMO a better default than our current integer
milliseconds-since-epoch. I plan to keep this the default for non-development
loggers.

### DIFF
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -103,7 +103,12 @@ func (e *TimeEncoder) UnmarshalText(text []byte) error {
 // A DurationEncoder serializes a time.Duration to a primitive type.
 type DurationEncoder func(time.Duration, PrimitiveArrayEncoder)
 
-// NanosDurationEncoder serializes a time.Duration to an int64 number of
+// SecondsDurationEncoder serializes a time.Duration to a floating-point number of seconds elapsed.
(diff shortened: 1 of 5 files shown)


zapcore/encoder.go
```diff
@@ -20,19 +20,130 @@
...
+
+// UnmarshalText unmarshals text to a DurationEncoder. "string" is unmarshaled
+// to StringDurationEncoder, and anything else is unmarshaled to
+// NanosDurationEncoder.
+func (e *DurationEncoder) UnmarshalText(text []byte) error {
+	switch string(text) {
+	case "string":
+		*e = StringDurationEncoder
+	default:
+		*e = NanosDurationEncoder
```
jcorbin: Hmm, why not a floating-point-second encoder, and default like for time?

akshayjshah: Sure, can do. I don't have much of a preference here.

PR #280, opened 2017-02-07
Update Tee to use no-op Facility

akshayjshah:
The facility tee should be using the no-op introduced in #278. Added a small unit test to verify behavior when passed one or zero inputs.

### DIFF
--- a/zapcore/tee.go
+++ b/zapcore/tee.go
@@ -23,13 +23,14 @@ package zapcore
 import "go.uber.org/zap/internal/multierror"
 
 // Tee creates a Facility that duplicates log entries into two or more
-// facilities; if you call it with less than two, you get back the one facility
-// you passed (or nil in the pathological case).
+// Facilities.
+//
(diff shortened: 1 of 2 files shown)


zapcore/tee.go
```diff
@@ -23,13 +23,14 @@ package zapcore
 import "go.uber.org/zap/internal/multierror"
 
 // Tee creates a Facility that duplicates log entries into two or more
-// facilities; if you call it with less than two, you get back the one facility
-// you passed (or nil in the pathological case).
+// Facilities.
+//
+// Calling it with a single Facility returns the input unchanged, and calling
+// it with no input returns a no-op Facility.
 func Tee(facs ...Facility) Facility {
```
jcorbin: On a related note, should we elide any `NopFacility` members in facs? Maybe open issue to track the concept?

akshayjshah: Possibly? I'll open an issue to track that (and similar issues for the writer tee).

PR #281, opened 2017-02-07
Add a buffering wrapper for WriteSyncers

akshayjshah:
We'll likely want this as part of the default production setup.


zapcore/write_syncer.go
```diff
@@ -83,6 +84,62 @@ func (w writerWrapper) Sync() error {
...
 
+type bufferedWriter struct {
+	size int // in bytes
+	buf  []byte
+	ws   WriteSyncer
+}
+
+// Buffer creates a WriteSyncer that buffers the specified number of bytes in
+// memory. Buffered WriteSyncers aren't safe for concurrent use, and should be
+// wrapped by Lock in production.
```
jcorbin: nit: the "in production" end sounds weird to me, why not just stop at saying "aren't safe for concurrent use."?  Whether or not something is used concurrently doesn't seem to me to be function of its deployment environment ;-)

zapcore/write_syncer.go
```diff
@@ -83,6 +84,62 @@ func (w writerWrapper) Sync() error {
...
+	return &bufferedWriter{
+		size: size,
+		buf:  buffers.Get(),
+		ws:   ws,
+	}
+}
+
+func (b *bufferedWriter) Write(bs []byte) (int, error) {
+	// Attempt to buffer.
+	if len(b.buf)+len(bs) <= b.size {
```
jcorbin: edge case nit: it seem so to me that we should should flush as soon as we hit '== size' rather than waiting for the "size+1'th" byte.

zapcore/write_syncer_impl_test.go
```diff
@@ -0,0 +1,191 @@
...
+		flushed string                        // written to underlying WriteSyncer
+		n       int
+		err     error
+	}{
+		{
+			desc:    "new write fills buffer",
+			initial: "ab",
+			try:     "cd",
+			buf:     "abcd",
+			flushed: "",
```
jcorbin: yeah, see this is the weird part to me, the buffer is now "full", but it won't flush until next write.

zapcore/write_syncer_impl_test.go
```diff
@@ -0,0 +1,191 @@
...
+			initial: "ab",
+			try:     "cde",
+			buf:     "cde",
+			flushed: "ab",
+			n:       3,
+		},
+		{
+			desc:    "new write is too big to buffer",
+			initial: "ab",
+			try:     "cdefg",
```
jcorbin: have another case where the try is just "abc" from this case, it should still be "too big" because of "remaining space" where-as a 5-length string would've never fit irrespective of initial

zapcore/write_syncer_impl_test.go
```diff
@@ -0,0 +1,191 @@
...
+		assert.Equal(t, tt.buf, string(b.buf), "Unexpected buffer contents (%s).", tt.desc)
+		assert.Equal(t, tt.flushed, sink.String(), "Unexpected bytes written to underlying WriteSyncer (%s).", tt.desc)
+	}
+}
+
+func TestBufferedWriterSync(t *testing.T) {
+	sink := &bytes.Buffer{}
+	underlying := AddSync(sink)
+
+	b := Buffer(4, underlying).(*bufferedWriter)
```
jcorbin: moar cases, including:
- empty -> sync
- two writes -> sync
?

zapcore/write_syncer_test.go
```diff
@@ -18,110 +18,95 @@
...
-func TestMultiWriteSyncerSync_AllCalled(t *testing.T) {
-	failed, second := &testutils.Buffer{}, &testutils.Buffer{}
-
-	failed.SetError(errors.New("disposal broken"))
-	ws := MultiWriteSyncer(failed, second)
-
-	assert.Error(t, ws.Sync(), "Expected first sink to fail")
-	assert.True(t, failed.Called(), "Expected first sink to have Sync method called.")
-	assert.True(t, second.Called(), "Expected call to Sync even with first failure.")
+func TestBufferedWriterPropagatesFailures(t *testing.T) {
```
jcorbin: Wait a minute, how does this differ from the above tests? Why can't we just have this one?

zapcore/write_syncer_test.go
```diff
@@ -18,110 +18,95 @@
...
+			n:              3,
+			expectWriteErr: "failed",
+			expectSyncErr:  "failed; no sync",
+		},
+	}
+
+	for _, tt := range tests {
+		buf := Buffer(4, tt.ws)
+		_, err := buf.Write([]byte(tt.initial))
+		assert.NoError(t, err, "Unexpected error writing initial buffer contents (%q).", tt.desc)
```
jcorbin: Probably worth using subtests (my heuristic is "at the point you feel compelled to interpolate a desc string") ;-)

zapcore/write_syncer.go
akshayjshah: Really? That seems like an off-by-one error.

If you told me that you could hold four apples but always drop the fourth apple, I'd call you a liar ;)

zapcore/write_syncer_impl_test.go
akshayjshah: That's the test immediately preceding this one. They're testing two separate code paths.

akshayjshah:
Oops, closed by mistake.

To me, this doesn't seem limited to files. For example, it'd be nice to have a buffering wrapper for a Kafka `WriteSyncer`, instead of requiring that the Kafka client implement its own buffering.

I'm not opposed to using `bufio`, but its error-handling strategy seems undesirable to me. Do we really want to short-circuit all writes after encountering a single error?

zapcore/write_syncer.go
akshayjshah: :+1: good point.

zapcore/write_syncer_test.go
akshayjshah: Fair enough. Still getting used to that as an option.

zapcore/write_syncer_test.go
akshayjshah: This test can't assert anything about the status of the internal buffer. Regardless, if we're going to use bufio, all this is out.

akshayjshah:
Abandoning this (for now, at least). We can return to it later if it proves indispensable.

PR #282, opened 2017-02-08
Explicitly test go1.8

billf:
A release branch has been cut, but we are only testing against 'tip'.


akshayjshah:
`gimme` doesn't appear to support this. Are you sure we don't need `1.8rc2` or something similar?

PR #285, opened 2017-02-09
Fix sampler

akshayjshah:
Use the timestamp on each Entry to fix our sampler. This introduces another
atomic operation in the sampling path, but it eliminates our use of the
runtime's timer heap.

Before:
```
BenchmarkZapSampleWithoutFields-4       20000000                56.7 ns/op             0 B/op          0 allocs/op
```

After:
```
BenchmarkZapSampleWithoutFields-4               20000000                63.7 ns/op             0 B/op          0 allocs/op
```

### DIFF
--- a/zapcore/sampler_test.go
+++ b/zapcore/sampler_test.go
@@ -21,12 +21,12 @@
 package zapcore_test
 
 import (
-	"fmt"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"go.uber.org/zap/internal/observer"
 	"go.uber.org/zap/testutils"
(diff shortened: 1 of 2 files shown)


zapcore/sampler_test.go
```diff
@@ -40,103 +40,97 @@ func fakeSampler(lvl LevelEnabler, tick time.Duration, first, thereafter int) (F
...
+		require.Equal(t, "", entry.Message, "Message wasn't created by writeSequence.")
+		require.Equal(t, 1, len(entry.Context), "Unexpected number of fields.")
+		require.Equal(t, lvl, entry.Level, "Unexpected level.")
+		f := entry.Context[0]
+		require.Equal(t, "iter", f.Key, "Unexpected field key.")
+		require.Equal(t, Int64Type, f.Type, "Unexpected field type")
+		seen[i] = f.Integer
 	}
-	return expected
+	assert.Equal(t, seq, seen, "Unexpected sequence logged at level %v.", lvl)
```
akshayjshah: This approach makes the failure output from testify much more useful - instead of seeing a huge diff of Entry structs, we just see the diff in sequence numbers.

zapcore/sampler_test.go
```diff
@@ -40,103 +40,97 @@ func fakeSampler(lvl LevelEnabler, tick time.Duration, first, thereafter int) (F
...
+		// Ensure that counts aren't shared between levels.
+		probeLevel := DebugLevel
+		if lvl == DebugLevel {
+			probeLevel = InfoLevel
+		}
+		for i := 0; i < 10; i++ {
+			writeSequence(sampler, 1, probeLevel)
+		}
+		// Clear any output.
+		logs.TakeAll()
```
jcorbin: glad to see it wasn't just my crazy use case ;-)

zapcore/sampler_test.go
```diff
@@ -40,103 +40,97 @@ func fakeSampler(lvl LevelEnabler, tick time.Duration, first, thereafter int) (F
...
-	for i := 1; i < 10; i++ {
-		writeIter(first, i, InfoLevel)
-	}
-	second := sampler.With([]Field{makeInt64Field("child", 2)})
-	// The new child logger should share the same counters, so we don't expect to
-	// write these logs.
-	for i := 10; i < 20; i++ {
-		writeIter(second, i, InfoLevel)
+	// If we log five or fewer messages every tick, none of them should be
+	// sampled.
```
jcorbin: "sampled" to me means that we see them, to me this should say "none of them should be dropped by the sampler"?

akshayjshah: Sure, will change.

prashantv:
I have an alternate approach in progress that doesn't swap the atomic for every log, and avoids the divide. However, I wanted to validate in benchmarks that it's actually more performant, and noticed something: The tick period for the benchmark is so high that we're not testing the "reset" case in the benchmarks (which by default run for one second).

I reduced it to `time.Millisecond`, and the results of the benchmark results are significantly different:

Before:
```
BenchmarkZapSampleWithoutFields-8       50000000                30.4 ns/op             1 B/op          0 allocs/op
```

After:
```
BenchmarkZapSampleWithoutFields-8        5000000               268 ns/op              80 B/op          1 allocs/op
```

This also caught another issue that we can fix separately: 
https://github.com/uber-go/zap/issues/288

prashantv:
Paired with @akshayjshah offline to figure out the cause -- it turns out that using `time.Millisecond` as the tick causes the current sampler to oversample, as the reset is run in a separate goroutine and is queued behind the benchmarking work.

Opened #290 with an alternate approach for sampling based on atomics that has some advantages over the approach in this PR:
 - Uses a sliding window rather than fixed windows
 - In the normal case, will load an extra atomic, but won't write to an extra atomic until the reset period has occurred

I've ensured the tests from this branch are copied forward and pass on the alternate implementation.

PR #289, opened 2017-02-11
Fix allocation when returning []byte to pool

### LINKED ISSUE #288: Buffer pool is causing an allocation on Put
The internal buffers package has a `Put` method with a signature of `Put([]byte)`

Casting a `[]byte` to an `interface{}` as we put the slice back in the underlying `sync.Pool` currently causes an allocation.

We should consider a wrapper type or using a `*[]byte` on `Put`

akshayjshah:
In order to pool byte slices without incurring an allocation (by casting into
`interface{}`), we need to switch to using a pointer type. The natural candidate
is the standard library's `bytes.Buffer`; it's a bit slower than straight
slices, but that's probably okay. However, we use `strconv.AppendFloat` and
friends to avoid allocations in the encoder, and those functions won't work with
`byte.Buffer`.

Instead, this PR adds a small `Buffer` type that implements just the bits of
`bytes.Buffer` and `strconv` that we need. It renames the private `buffers`
package to `bufferpool` to clarify the difference between the two.

Before:

```
BenchmarkZapWithoutFields-4      3000000               436 ns/op              80 B/op          2 allocs/op
```

After:

```
BenchmarkZapWithoutFields-4      5000000               331 ns/op              32 B/op          1 allocs/op
```

This fixes #288.

### DIFF
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -49,7 +49,7 @@ func (b *Buffer) AppendString(s string) {
 
 // AppendInt appends an integer to the underlying buffer (assuming base 10).
 func (b *Buffer) AppendInt(i int64) {
-	b.bs = strconv.AppendInt(b.bs, int64(i), 10)
+	b.bs = strconv.AppendInt(b.bs, i, 10)
 }
 
 // AppendUint appends an unsigned integer to the underlying buffer (assuming
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -1,0 +1,102 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package buffer provides a thin wrapper around a byte slice. Unlike the
+// standard library's bytes.Buffer, it supports a portion of the strconv
+// package's zero-allocation formatters.
+package buffer
+
+import "strconv"
+
+const _size = 1024 // by default, create 1 KiB buffers
+
+// Buffer is a thin wrapper around a byte slice.
+type Buffer struct {
+	bs []byte
+}
+
+// New creates a new Buffer of the default size.
+func New() *Buffer {
+	return &Buffer{make([]byte, 0, _size)}
+}
+
+// AppendByte writes a single byte to the Buffer.
+func (b *Buffer) AppendByte(v byte) {
+	b.bs = append(b.bs, v)
+}
+
+// AppendString writes a string to the Buffer.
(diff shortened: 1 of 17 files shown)


internal/bufferpool/bufferpool.go
```diff
@@ -18,32 +18,30 @@
...
-// Package buffers provides strongly-typed functions to interact with a shared
-// pool of byte slices.
-//
-// It's used heavily inside zap, but callers may also take advantage of it;
-// it's particularly useful when implementing json.Marshaler, text.Marshaler,
-// and similar interfaces.
-package buffers
+// Package bufferpool provides strongly-typed functions to interact with a shared
+// pool of byte buffers.
+package bufferpool
```
akshayjshah: The `Buffer` type needs to be non-internal, since it's part of the exported interfaces. The pool, however, should remain internal.

zapcore/console_encoder.go
```diff
@@ -103,52 +102,52 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) ([]byte, error) {
 	// If there's no stacktrace key, honor that; this allows users to force
 	// single-line output.
 	if ent.Stack != "" && c.StacktraceKey != "" {
-		line.WriteByte('\n')
-		line.WriteString(ent.Stack)
+		line.AppendByte('\n')
```
akshayjshah: Unfortunately, this triggers a variety of irritating `go vet` checks that attempt to enforce compliance with some interfaces in the standard lib.

buffer/buffer_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}{
+		{"AppendByte", func() { buf.AppendByte('v') }, "v"},
+		{"AppendString", func() { buf.AppendString("foo") }, "foo"},
+		{"AppendIntPositive", func() { buf.AppendInt(42) }, "42"},
+		{"AppendIntNegative", func() { buf.AppendInt(-42) }, "-42"},
+		{"AppendUint", func() { buf.AppendUint(42) }, "42"},
+		{"AppendBool", func() { buf.AppendBool(true) }, "true"},
+		{"AppendFloat64", func() { buf.AppendFloat(3.14, 64) }, "3.14"},
+		// Intenationally introduce some floating-point error.
+		{"AppendFloat32", func() { buf.AppendFloat(float64(float32(3.14)), 32) }, "3.14"},
```
akshayjshah: This tests that the `bitSize` parameter is being plumbed through correctly.

https://play.golang.org/p/ohu6TG6Ll6

buffer/buffer_test.go
```diff
@@ -0,0 +1,91 @@
...
+		{"AppendWrite", func() { buf.Write([]byte("foo")) }, "foo"},
```
akshayjshah: :+1:

zapcore/console_encoder.go
```diff
@@ -64,8 +63,8 @@ func (c consoleEncoder) Clone() Encoder {
 	return consoleEncoder{c.jsonEncoder.Clone().(*jsonEncoder)}
 }
 
-func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) ([]byte, error) {
-	line := bytes.NewBuffer(buffers.Get())
+func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer, error) {
```
akshayjshah: Primary advantage is less concrete: it more cleanly separates encoders, which turn objects into bytes, and Facilities/WriteSyncers. It also preserves some degree of API similarity with the marshaler types in the standard library, which typically return `([]byte, error)`.

If you're staunchly against this, though, I can certainly switch.

buffer/buffer.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// AppendString writes a string to the Buffer.
+func (b *Buffer) AppendString(s string) {
+	b.bs = append(b.bs, s...)
+}
+
+// AppendInt appends an integer to the underlying buffer (assuming base 10).
+func (b *Buffer) AppendInt(i int64) {
+	b.bs = strconv.AppendInt(b.bs, int64(i), 10)
```
akshayjshah: :+1:

buffer/buffer_test.go
prashantv: Ok cool, makes sense

zapcore/console_encoder.go
prashantv: I'm not very against it -- I do like the separation between encoders and writers, just wanted to make sure we were getting some other advantages for the cost of another public type.

zapcore/console_encoder.go
prashantv: That might not be so bad since then we'll implement `WriteString` with a return of `(int, error) which will allow `io.WriteString` to efficiently write a string to the buffer without converting to `[]byte`.

internal/bufferpool/bufferpool.go
prashantv: Ah fair enough.

zapcore/console_encoder.go
akshayjshah: It's actually the `WriteByte` method that gets flagged: https://golang.org/cmd/vet/#hdr-Methods

The returned error is particularly irritating because it then triggers `errcheck` if you don't use it.

PR #291, opened 2017-02-14
Rename Facility to Core

akshayjshah:
Rename zapcore.Facility to zapcore.Core, which is a bit more evocative.
zapcore.Logger is probably a better name, but is likely to create more confusion
among first-time users and GoDoc readers.

After looking at the end result, I'm inclined to stop here and keep Logger and
SugaredLogger as they are. Elevating SugaredLogger to the primary type ended up
(to me) confusing the story of what's the fundamental logger and what's the
API sugaring layer.


peter-edge:
I'm not sure I love `Core`, but I'm almost at a loss for what to call it. `Facility` is still nice, what about `Pusher` or `Endpoint`?

akshayjshah:
I'd prefer to go with a name that clearly indicates that it's the primary type in the `zapcore` package. `Pusher` and `Endpoint` sound a bit too much like networking/RPC to me, and `Facility` is a bit vague.

Agreed that this is hard. IMO `zapcore.Logger` is the most precise, but will be *very* confusing for users.

zapcore/hook.go
```diff
@@ -23,35 +23,42 @@ package zapcore
...
 	return &hooked{
-		Facility: h.Facility.With(fields),
-		funcs:    h.funcs,
+		Core:  h.Core.With(fields),
+		funcs: h.funcs,
 	}
 }
 
 func (h *hooked) Write(ent Entry, _ []Field) error {
+	// Since we've already called
```
prashantv: incomplete comment

internal/observer/observer.go
```diff
@@ -94,11 +94,10 @@ type observer struct {
 	sink func(LoggedEntry) error
 }
 
-// New creates a new facility that buffers logs in memory (without any
-// encoding). It's particularly useful in tests, though it can serve a variety
-// of other purposes as well. This constructor returns the facility itself and
-// a function to retrieve the observed logs.
-func New(enab zapcore.LevelEnabler, sink func(LoggedEntry) error, withContext bool) zapcore.Facility {
+// New creates a new Core that buffers logs in memory (without any encoding).
+// It's particularly useful in tests, though it can serve a variety of other
```
prashantv: "thought i can serve a variety of other purposes" -- seems a little vague. If we don't have any other use cases to call out, is it worth having this comment?

logger.go
```diff
@@ -47,30 +47,34 @@ type Logger struct {
 	callerSkip int
 }
 
-// New constructs a new Logger from the provided Facility and options. Passing
-// a nil Facility results in a no-op Logger.
-func New(fac zapcore.Facility, options ...Option) *Logger {
-	if fac == nil {
-		fac = zapcore.NopFacility()
+// New constructs a new Logger from the provided zapcore.Core and Options. Passing
+// a nil zapcore.Core returns a no-op Logger.
```
prashantv: instead of saying it's a noop logger, should we say we fallback to a noop core? (the logger itself is still functional in terms of calling hooks, still panics, etc)

options.go
```diff
@@ -34,25 +34,27 @@ func (f optionFunc) apply(log *Logger) {
 	f(log)
 }
 
-// WrapFacility wraps or replaces the logger's underlying facility.
-func WrapFacility(f func(zapcore.Facility) zapcore.Facility) Option {
+// WrapCore wraps or replaces the Logger's underlying Core.
```
prashantv: should we say `zapcore.Core` in the comment (to match the comment for `Core()`)

zapcore/core.go
```diff
@@ -0,0 +1,107 @@
...
+// NewCore creates a Core that writes logs to a WriteSyncer.
+func NewCore(enc Encoder, ws WriteSyncer, enab LevelEnabler) Core {
+	return &ioCore{
+		LevelEnabler: enab,
+		enc:          enc,
+		out:          Lock(ws),
+	}
+}
+
+type ioCore struct {
```
prashantv: i like how simple the basic core implementation turned out

zapcore/entry.go
```diff
@@ -104,67 +104,70 @@ type Entry struct {
...
-// is done. Actions are ordered in increasing severity.
+// CheckWriteAction indicates what action to take after Write is done. Actions
+// are ordered in increasing severity.
 type CheckWriteAction uint8
 
 const (
-	// WriteThenNoop is the default behavior to do nothing speccial after write.
+	// WriteThenNoop indicates that nothing special needs to be done. It's the
+	// default behavior.
 	WriteThenNoop = CheckWriteAction(iota)
```
prashantv: While this works fine, the standard way to define constants of a custom type is `WriteThenNoop CheckWriteAction = iota`.

See examples,
https://golang.org/pkg/time/#Weekday
https://golang.org/pkg/reflect/#ChanDir
https://golang.org/pkg/reflect/#Kind

Since this shows up in documentation, I'd prefer if we used the standard syntax to avoid any confusion.

akshayjshah: Sure, sounds good.

PR #300, opened 2017-02-15
Improve error messages when using old import paths

akshayjshah:
Add the magic comments required to provide better error messages when users try
to import "github.com/uber-go/zap" and ".../zapcore".


jcorbin:
TIL (and for future travelers): https://golang.org/cmd/go/#hdr-Import_path_checking

PR #302, opened 2017-02-15
Add benchmarks for lion's JSON logger

akshayjshah:
Add benchmarks for `go.pedge.io/lion`, but only benchmark the JSON logger.

At some point, we may want to add protobuf benchmarks, but I'd rather not introduce dependencies on the proto compiler just for tests.

Current results (from my machine, comparable to benchmarks in README):

```
testing: warning: no tests to run
BenchmarkWithoutFields/go.pedge.io/lion-4                        1000000              1738 ns/op            1224 B/op         10 allocs/op
BenchmarkAccumulatedContext/go.pedge.io/lion-4                    300000              6403 ns/op            3978 B/op         36 allocs/op
BenchmarkAddingFields/go.pedge.io/lion-4                          200000              9992 ns/op            5999 B/op         62 allocs/op
```

### DIFF
--- a/benchmarks/lion_test.go
+++ b/benchmarks/lion_test.go
@@ -1,0 +1,31 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


benchmarks/lion_test.go
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
peter-edge: 2017 :)

peter-edge:
LGTM

PR #306, opened 2017-02-17
Use Println and Infoln for lion benchmarks

peter-edge:
Small thing I missed on the last review, my apologies - the way to do non-formatted printing of a string in lion is to use Println/Infoln instead of Printf/Infof, the former calls fmt.Sprint while the latter calls fmt.Sprintf, which I think is a little more expensive.


akshayjshah:
I'm happy to do whatever you think will give lion the best performance. In general, though, `printf` is faster for static strings, since it avoids the overhead of converting `string` to `interface{}` (which promotes the string to the heap).

Let me know if you'd still like this merged.

peter-edge:
Ah ok, nevermind.

PR #307, opened 2017-02-17
Add colored LevelEncoders

### LINKED ISSUE #234: Colored log output
I'm wondering if there are any plans to implement colorized log output in this library. While it might not be something one would enable in a production environment, for local development it can be quite nice. Alternatives like logrus provide this sort of functionality.

Just based on the public API, it seems like this could be implemented as a [TextOption](https://godoc.org/github.com/uber-go/zap#TextOption). If it isn't slated for development, would you be open to a contribution with it?

peter-edge:
Add a small internal package to handle TTY output coloring, and use that package to define some color-aware LevelEncoders.

Fixes #234.

image

### DIFF
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -21,7 +21,6 @@
 package zapcore
 
 import (
-	"strings"
 	"time"
 
 	"go.uber.org/zap/buffer"
@@ -36,18 +35,44 @@ func LowercaseLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
 	enc.AppendString(l.String())
 }
 
+// LowercaseColorLevelEncoder serializes a Level to a lowercase string and adds coloring.
+// For example, InfoLevel is serialized to "info" and colored blue.
+func LowercaseColorLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
+	s, ok := _levelToLowercaseColorString[l]
+	if !ok {
+		s = _unknownLevelColor.Add(l.String())
+	}
+	enc.AppendString(s)
+}
+
 // CapitalLevelEncoder serializes a Level to an all-caps string. For example,
 // InfoLevel is serialized to "INFO".
 func CapitalLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
-	enc.AppendString(strings.ToUpper(l.String()))
+	enc.AppendString(l.CapitalString())
+}
+
+// CapitalColorLevelEncoder serializes a Level to an all-caps string and adds color.
+// For example, InfoLevel is serialized to "INFO" and colored blue.
+func CapitalColorLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
+	s, ok := _levelToCapitalColorString[l]
+	if !ok {
+		s = _unknownLevelColor.Add(l.CapitalString())
+	}
+	enc.AppendString(s)
 }
 
 // UnmarshalText unmarshals text to a LevelEncoder. "capital" is unmarshaled to
-// CapitalLevelEncoder, and anything else is unmarshaled to LowercaseLevelEncoder.
+// CapitalLevelEncoder, "coloredCapital" is unmarshaled to CapitalColorLevelEncoder,
+// "colored" is unmarshaled to LowercaseColorLevelEncoder, and anything else
+// is unmarshaled to LowercaseLevelEncoder.
 func (e *LevelEncoder) UnmarshalText(text []byte) error {
 	switch string(text) {
 	case "capital":
 		*e = CapitalLevelEncoder
+	case "capitalColor":
+		*e = CapitalColorLevelEncoder
+	case "color":
+		*e = LowercaseColorLevelEncoder
 	default:
 		*e = LowercaseLevelEncoder
 	}
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ export GO15VENDOREXPERIMENT=1
 BENCH_FLAGS ?= -cpuprofile=cpu.pprof -memprofile=mem.pprof -benchmem
 PKGS ?= $(shell glide novendor)
 # Many Go tools take file globs or directories as arguments instead of packages.
-PKG_FILES ?= *.go zapcore benchmarks buffer testutils internal/bufferpool internal/exit internal/multierror internal/observer
+PKG_FILES ?= *.go zapcore benchmarks buffer testutils internal/bufferpool internal/exit internal/multierror internal/observer internal/color
 
 # The linting tools evolve with each Go version, so run them only on the latest
 # stable release.
--- a/internal/color/color.go
+++ b/internal/color/color.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package color adds coloring functionality for TTY output.
+package color
+
+import "fmt"
+
+// Foreground colors.
+const (
+	Black Color = iota + 30
+	Red
+	Green
+	Yellow
+	Blue
+	Magenta
+	Cyan
+	White
+)
+
+// Color represents a text color.
+type Color uint8
+
+// Add adds the coloring to the given string.
+func (c Color) Add(s string) string {
+	return fmt.Sprintf("\x1b[%dm%s\x1b[0m", uint8(c), s)
+}
--- a/internal/color/color_test.go
+++ b/internal/color/color_test.go
@@ -1,0 +1,36 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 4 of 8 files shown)


peter-edge:
The code for the above output:

```go
package main

import (
	"fmt"
	"os"

	"go.uber.org/zap"
)

func main() {
	if err := do(); err != nil {
		fmt.Fprintln(os.Stderr, err.Error())
		os.Exit(1)
	}
}

func do() error {
	logger, err := zap.NewDevelopment()
	if err != nil {
		return err
	}
	logger.Debug("test", zap.Any("foo", foo{One: "one", Two: "two"}))
	logger.Info("test", zap.Any("foo", foo{One: "one", Two: "two"}))
	logger.Warn("test", zap.Any("foo", foo{One: "one", Two: "two"}))
	logger.Error("test", zap.Any("foo", foo{One: "one", Two: "two"}))
	return nil
}

type foo struct {
	One string
	Two string
}
```

zapcore/encoder.go
```diff
@@ -36,18 +55,36 @@ func LowercaseLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
 	enc.AppendString(l.String())
 }
 
+// ColoredLowercaseLevelEncoder serializes a Level to a lowercase string and adds coloring.
+// For example, InfoLevel is serialized to "info".
+func ColoredLowercaseLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
+	enc.AppendString(levelToColor[l].Sprint(l.String()))
```
akshayjshah: `Sprintf` instead, which saves an allocation. We should also cache the colorized output for the known levels, which should shave off the other allocation.

This also needs to handle map misses; are nil `*color.Color`s safe to use?

zapcore/encoder.go
```diff
@@ -36,18 +55,36 @@ func LowercaseLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
...
 // CapitalLevelEncoder serializes a Level to an all-caps string. For example,
 // InfoLevel is serialized to "INFO".
 func CapitalLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
 	enc.AppendString(strings.ToUpper(l.String()))
 }
 
+// ColoredCapitalLevelEncoder serializes a Level to an all-caps string and adds color.
+// For example, InfoLevel is serialized to "INFO".
+func ColoredCapitalLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
+	enc.AppendString(levelToColor[l].Sprint(strings.ToUpper(l.String())))
```
akshayjshah: Same here.

zapcore/encoder.go
```diff
@@ -36,18 +55,36 @@ func LowercaseLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
...
 // UnmarshalText unmarshals text to a LevelEncoder. "capital" is unmarshaled to
-// CapitalLevelEncoder, and anything else is unmarshaled to LowercaseLevelEncoder.
+// CapitalLevelEncoder, "colored_capital" is unmarshaled to ColoredCapitalLevelEncoder,
+// "colored" is unmarshaled to ColoredLowercaseLevelEncoder, and anything else
+// is unmarshaled to LowercaseLevelEncoder.
 func (e *LevelEncoder) UnmarshalText(text []byte) error {
 	switch string(text) {
 	case "capital":
 		*e = CapitalLevelEncoder
+	case "colored_capital":
```
akshayjshah: Camel case for consistency with YAML/JSON keys.

zapcore/encoder.go
```diff
@@ -24,9 +24,28 @@ import (
...
 
+	"github.com/fatih/color"
+
 	"go.uber.org/zap/buffer"
 )
 
+var (
+	// TODO(pedge): add a test to verify all levels covered? It would be nice
+	// to maintain a slice of all levels in level.go for testing, even if there
+	// is already _minLevel and _maxLevel
```
akshayjshah: Don't mind putting one in `testutils`, though whatever code deals with levels should handle unexpected levels too.

zapcore/encoder.go
```diff
@@ -24,9 +24,28 @@ import (
...
+		InfoLevel:   color.New(color.FgBlue),
+		WarnLevel:   color.New(color.FgYellow),
+		ErrorLevel:  color.New(color.FgRed),
+		DPanicLevel: color.New(color.FgRed),
+		PanicLevel:  color.New(color.FgRed),
+		FatalLevel:  color.New(color.FgRed),
+	}
+)
+
+// TODO(pedge): we can probably just cache the level strings
```
akshayjshah: They're constants, so this doesn't allocate.

peter-edge:
go-isatty does about what you'd expect - it disables colors if stdout does not refer to a terminal. go-colorable handles stdout for colors if on windows, but with how we use fatih/color, it's effectively never used. It's weird that both of these operate on os.Stdout as a constant (or maybe I'm missing something), but that's how how it is.

Additionally when it comes to stability, note that fatih/color vendors go-isatty and go-colorable itself, and I think it's a bug in glide that these are separately vendored in zap. The code from fatih/color will use it's own vendored versions, so we only need to worry about updates to fatih/color.  Of note, I've used this package in lion for a long time and have never had any issues.

We can try to re-implement some of the logic in fatih/color that we need, which isn't that complicated, that is one option, and having read over the code more closely, is the option I think I'd recommend. We might get some performance wins too (adding color isn't really hard to do). Adding color is actually a really simple thing, we don't need all the functionality of fatih/color to do it.

peter-edge:
I pushed a commit that does all coloring internally. Note that coloring is not disabled if there is a not a tty, but I'm not sure this is a problem for now - we can add a check if the output is a tty, but this will mean a lot of changes down the stack (find out if we're using a WriteSyncer, find out if the WriteSyncer is actually a file, do the logic in go-isatty effectively) and your setup of EncoderConfig should take this into account. All comments should be addressed through this commit.

jcorbin:
Glad to see the internal pivot; I do think we want `isatty` logic still tho, at least part of `zap. NewDevelopmentConfig`?

Either way, I'm excited to see color support coming to zap 👍

zapcore/encoder.go
```diff
@@ -36,18 +64,45 @@ func LowercaseLevelEncoder(l Level, enc PrimitiveArrayEncoder) {
...
+}
+
+func encodeWithColor(l Level, enc PrimitiveArrayEncoder, s string) {
+	colorCode, ok := levelToColorCode[l]
+	if !ok {
+		enc.AppendString(s)
+		return
+	}
+	enc.AppendString(fmt.Sprintf("\x1b[%dm%s\x1b[0m", colorCode, s))
+}
```
akshayjshah: Can we move the coloring logic (color codes, cached strings, etc.) to a separate file? This one's getting a bit big.

If we start coloring other output, we should make this a proper internal package.

config.go
```diff
@@ -129,7 +129,7 @@ func NewDevelopmentConfig() Config {
 			CallerKey:      "C",
 			MessageKey:     "M",
 			StacktraceKey:  "S",
-			EncodeLevel:    zapcore.CapitalLevelEncoder,
+			EncodeLevel:    zapcore.ColoredCapitalLevelEncoder,
```
akshayjshah: I'd really prefer to keep the default un-colored - it's just safer.

prashantv:
I'd also prefer to avoid any extra dependencies here. For the `isatty` logic, I think we can `stat` stdin to check if it's a terminal:
```go
func useColor() bool {
  stat, err := os.Stdin.Stat()
  if err != nil {
    return false
  }
  return stat.Mode() & os. ModeCharDevice != 0
}
```

There'll be a ton of edge cases which I don't think we want to deal with -- if there's a terminal, we can assume it supports colour (and the user is always able to override this).

akshayjshah:
Interesting - that's a good idea that I can get behind. Perhaps `NewDevelopmentConfig` can stat stdin and choose an appropriate level encoder.

peter-edge:
I'm actually going to argue against checking for color for now, and just make the default no color - I've had more situations where I end up wanting to force coloring than the reverse (with docker for example), and I think we can punt on this, especially if the default is no color.

config.go
peter-edge: Done

peter-edge:
I updated this to use an internal package, to have an AllLevels variable in zapcore, and to have cached color level strings. I think that other than a comment about testing the color encoding (which seems more trivial now), everything is taken care of @akshayjshah .

peter-edge:
I also cached the uncolored strings as well.

peter-edge:
Here's the same code now updated to set the LevelEncoder to use color:

```go
package main

import (
	"fmt"
	"os"

	"go.uber.org/zap"
	"go.uber.org/zap/zapcore"
)

func main() {
	if err := do(); err != nil {
		fmt.Fprintln(os.Stderr, err.Error())
		os.Exit(1)
	}
}

func do() error {
	config := zap.NewDevelopmentConfig()
	config.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
	logger, err := config.Build()
	if err != nil {
		return err
	}
	logger.Debug("test", zap.Any("foo", foo{One: "one", Two: "two"}))
	logger.Info("test", zap.Any("foo", foo{One: "one", Two: "two"}))
	logger.Warn("test", zap.Any("foo", foo{One: "one", Two: "two"}))
	logger.Error("test", zap.Any("foo", foo{One: "one", Two: "two"}))
	return nil
}

type foo struct {
	One string
	Two string
}
```

peter-edge:
Can we merge this? I was going to do https://github.com/uber-go/zap/issues/330 but this should be merged first so there's less of an issue.

akshayjshah:
@peter-edge Take a look and let me know what you think. I pulled out some of the string caching and simplified the coloring package a bit.

peter-edge:
LGTM

sectioneight:
Late to the party, but FWIW I disagree with using `isatty` as a general approach to determining whether colorization should be turned on. Specifically I've had to patch a few third-party projects to remove the assumption that just because I'm not attached to a TTY means that the output can't be colorized. For example, in a lot of our build pipelines we `tee` stuff around and the escape codes are totally valid to go to stdout and also when `cat`-ing the file.

Looks like y'all ditched that logic anyway and went with an explicit flag which I like :)

peter-edge:
FORCE_ZAP_COLORS=1 go test aidens/code/...

kc1116:
Super late to the party, but I am trying to get colored output and I just can't seem to. Heres my code to setup the logger 
```

func init() {
	jsonDebugging := zapcore.AddSync(ioutil.Discard)
	jsonErrors := zapcore.AddSync(ioutil.Discard)
	consoleDebugging := zapcore.Lock(os.Stdout)
	consoleErrors := zapcore.Lock(os.Stderr)

	jsonEncoder := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())
	consoleEncoder := zapcore.NewConsoleEncoder(zap.NewDevelopmentEncoderConfig())

	core := zapcore.NewTee(
		zapcore.NewCore(jsonEncoder, jsonErrors, highPriority),
		zapcore.NewCore(consoleEncoder, consoleErrors, highPriority),
		zapcore.NewCore(jsonEncoder, jsonDebugging, lowPriority),
		zapcore.NewCore(consoleEncoder, consoleDebugging, lowPriority),
	)

	var opts []zap.Option
	opts = append(opts, zap.Fields(zap.Int("pid", os.Getpid())))
	opts = append(opts, zap.Fields(zap.String("exe", path.Base(os.Args[0]))))
	Logger = &LoggerWrapper{
		zapLogger: zap.New(core, opts...),
	}
	defer Logger.zapLogger.Sync()
}
```

cup:
https://github.com/uber-go/zap/issues/234
https://github.com/uber-go/zap/issues/489
https://github.com/uber-go/zap/issues/648
https://github.com/uber-go/zap/issues/688

https://github.com/uber-go/zap/pull/307

cup:
https://github.com/uber-go/zap/issues/714

cup:
package main
import "github.com/labstack/gommon/log"
func main() {
   log.EnableColor()
   log.Info("aaaaa bbbbb")
}

cup:
@kc1116 this works for me:

~~~go
package main
import (
   "github.com/mattn/go-colorable"
   "go.uber.org/zap"
   "go.uber.org/zap/zapcore"
)
func main() {
   aa := zap.NewDevelopmentEncoderConfig()
   aa.EncodeLevel = zapcore.CapitalColorLevelEncoder
   bb := zap.New(zapcore.NewCore(
      zapcore.NewConsoleEncoder(aa),
      zapcore.AddSync(colorable.NewColorableStdout()),
      zapcore.DebugLevel,
   ))
   bb.Warn("cc")
}
~~~

Or:

~~~go
package main
import "github.com/labstack/gommon/log"
func main() {
   log.EnableColor()
   log.Info("aaaaa bbbbb")
}
~~~

PR #311, opened 2017-02-17
Skip runtime and some zap frames in stacktraces

### LINKED ISSUE #308: Remove first four calls of stacktrace
When you get a stacktrace, it will always return:

```
go.uber.org/zap.takeStacktrace(0xc42005d000, 0x400, 0x400, 0x0, 0x0, 0xc4200f9901)
	/Users/peteredge/gocode/src/go.uber.org/zap/stacktrace.go:33 +0x64
go.uber.org/zap.Stack(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/peteredge/gocode/src/go.uber.org/zap/field.go:197 +0x7a
go.uber.org/zap.(*Logger).check(0xc420072300, 0x1, 0x12969d2, 0x4, 0xc4200784c0)
	/Users/peteredge/gocode/src/go.uber.org/zap/logger.go:254 +0x212
go.uber.org/zap.(*Logger).Warn(0xc420072300, 0x12969d2, 0x4, 0xc4200784c0, 0x1, 0x1)
	/Users/peteredge/gocode/src/go.uber.org/zap/logger.go:154 +0x44
```

This doesn't have much use to the user, so we should try to remove these. This might be a little tricky, I'll look into it at some point.

peter-edge:
Remove `runtime.goexit`, `runtime.main`, and some zap frames from our stacktraces. Since we're now formatting our own traces, we also get a *big* performance win: nearly 5x faster and half the number of bytes allocated!

Current master:

```
BenchmarkStackField-4             200000              9587 ns/op             960 B/op          2 allocs/op
```

After this PR:

```
BenchmarkStackField-4             500000              2149 ns/op             448 B/op          2 allocs/op
```

Fixes #308.

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -50,7 +50,7 @@ func TestConfig(t *testing.T) {
 			expectRe: "DEBUG\t.*go.uber.org/zap/config_test.go:" + `\d+` + "\tdebug\t" + `{"k": "v", "z": "zz"}` + "\n" +
 				"INFO\t.*go.uber.org/zap/config_test.go:" + `\d+` + "\tinfo\t" + `{"k": "v", "z": "zz"}` + "\n" +
 				"WARN\t.*go.uber.org/zap/config_test.go:" + `\d+` + "\twarn\t" + `{"k": "v", "z": "zz"}` + "\n" +
-				`goroutine \d+ \[running\]:`,
+				`go.uber.org/zap.TestConfig.func1`,
 		},
 	}
 
--- a/config_test.go
+++ b/config_test.go
@@ -50,7 +50,7 @@ func TestConfig(t *testing.T) {
 			expectRe: "DEBUG\t.*go.uber.org/zap/config_test.go:" + `\d+` + "\tdebug\t" + `{"k": "v", "z": "zz"}` + "\n" +
 				"INFO\t.*go.uber.org/zap/config_test.go:" + `\d+` + "\tinfo\t" + `{"k": "v", "z": "zz"}` + "\n" +
 				"WARN\t.*go.uber.org/zap/config_test.go:" + `\d+` + "\twarn\t" + `{"k": "v", "z": "zz"}` + "\n" +
-				`go.uber.org/zap.TestConfig.func1`,
+				`go.uber.org/zap.Stack`,
 		},
 	}
 
@@ -68,19 +68,25 @@ func TestConfig(t *testing.T) {
 			logger, err := tt.cfg.Build(Hooks(hook))
 			require.NoError(t, err, "Unexpected error constructing logger.")
 
-			logger.Debug("debug")
-			logger.Info("info")
-			logger.Warn("warn")
+			withStacktraceIgnorePrefixes([]string{}, func() {
+				logger.Debug("debug")
+				logger.Info("info")
+				logger.Warn("warn")
 
-			byteContents, err := ioutil.ReadAll(temp)
-			require.NoError(t, err, "Couldn't read log contents from temp file.")
-			logs := string(byteContents)
-			assert.Regexp(t, tt.expectRe, logs, "Unexpected log output.")
+				byteContents, err := ioutil.ReadAll(temp)
+				// not doing require so no problem with lock in withStacktraceIgnorePrefixes
+				assert.NoError(t, err, "Couldn't read log contents from temp file.")
+				if err != nil {
+					return
+				}
+				logs := string(byteContents)
+				assert.Regexp(t, tt.expectRe, logs, "Unexpected log output.")
 
-			for i := 0; i < 200; i++ {
-				logger.Info("sampling")
-			}
-			assert.Equal(t, tt.expectN, count.Load(), "Hook called an unexpected number of times.")
+				for i := 0; i < 200; i++ {
+					logger.Info("sampling")
+				}
+				assert.Equal(t, tt.expectN, count.Load(), "Hook called an unexpected number of times.")
+			})
 		})
 	}
 }
--- a/config_test.go
+++ b/config_test.go
@@ -68,17 +68,13 @@ func TestConfig(t *testing.T) {
 			logger, err := tt.cfg.Build(Hooks(hook))
 			require.NoError(t, err, "Unexpected error constructing logger.")
 
-			withStacktraceIgnorePrefixes([]string{}, func() {
+			withNoStacktraceIgnorePrefixes(func() {
 				logger.Debug("debug")
 				logger.Info("info")
 				logger.Warn("warn")
 
 				byteContents, err := ioutil.ReadAll(temp)
-				// not doing require so no problem with lock in withStacktraceIgnorePrefixes
-				assert.NoError(t, err, "Couldn't read log contents from temp file.")
-				if err != nil {
-					return
-				}
+				require.NoError(t, err, "Couldn't read log contents from temp file.")
 				logs := string(byteContents)
 				assert.Regexp(t, tt.expectRe, logs, "Unexpected log output.")
 
--- a/config_test.go
+++ b/config_test.go
@@ -68,21 +68,19 @@ func TestConfig(t *testing.T) {
 			logger, err := tt.cfg.Build(Hooks(hook))
 			require.NoError(t, err, "Unexpected error constructing logger.")
 
-			withNoStacktraceIgnorePrefixes(func() {
-				logger.Debug("debug")
-				logger.Info("info")
-				logger.Warn("warn")
+			logger.Debug("debug")
+			logger.Info("info")
+			logger.Warn("warn")
 
-				byteContents, err := ioutil.ReadAll(temp)
-				require.NoError(t, err, "Couldn't read log contents from temp file.")
-				logs := string(byteContents)
-				assert.Regexp(t, tt.expectRe, logs, "Unexpected log output.")
+			byteContents, err := ioutil.ReadAll(temp)
+			require.NoError(t, err, "Couldn't read log contents from temp file.")
+			logs := string(byteContents)
+			assert.Regexp(t, tt.expectRe, logs, "Unexpected log output.")
 
-				for i := 0; i < 200; i++ {
-					logger.Info("sampling")
-				}
-				assert.Equal(t, tt.expectN, count.Load(), "Hook called an unexpected number of times.")
-			})
+			for i := 0; i < 200; i++ {
+				logger.Info("sampling")
+			}
+			assert.Equal(t, tt.expectN, count.Load(), "Hook called an unexpected number of times.")
 		})
 	}
 }
--- a/field.go
+++ b/field.go
@@ -25,7 +25,6 @@ import (
 	"math"
 	"time"
 
-	"go.uber.org/zap/internal/bufferpool"
 	"go.uber.org/zap/zapcore"
 )
 
@@ -195,20 +194,11 @@ func NamedError(key string, err error) zapcore.Field {
 // extremely expensive (relatively speaking); this function both makes an
 // allocation and takes ~10 microseconds.
 func Stack(key string) zapcore.Field {
-	return stack(key, 4)
-}
-
-// needed for testing
-func stack(key string, skip int) zapcore.Field {
-	// Try to avoid allocating a buffer.
-	buf := bufferpool.Get()
 	// Returning the stacktrace as a string costs an allocation, but saves us
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	field := String(key, takeStacktrace(buf, skip))
-	bufferpool.Put(buf)
-	return field
+	return String(key, takeStacktrace())
 }
 
 // Duration constructs a field with the given key and value. The encoder
--- a/field.go
+++ b/field.go
@@ -195,14 +195,18 @@ func NamedError(key string, err error) zapcore.Field {
 // extremely expensive (relatively speaking); this function both makes an
 // allocation and takes ~10 microseconds.
 func Stack(key string) zapcore.Field {
+	return stack(key, 4)
+}
+
+// needed for testing
+func stack(key string, skip int) zapcore.Field {
 	// Try to avoid allocating a buffer.
 	buf := bufferpool.Get()
-	bs := buf.Bytes()
 	// Returning the stacktrace as a string costs an allocation, but saves us
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	field := String(key, takeStacktrace(bs[:cap(bs)], false))
+	field := String(key, takeStacktrace(buf, skip))
 	bufferpool.Put(buf)
 	return field
 }
--- a/field.go
+++ b/field.go
@@ -191,8 +191,8 @@ func NamedError(key string, err error) zapcore.Field {
 
 // Stack constructs a field that stores a stacktrace of the current goroutine
 // under provided key. Keep in mind that taking a stacktrace is eager and
-// extremely expensive (relatively speaking); this function both makes an
-// allocation and takes ~10 microseconds.
+// expensive (relatively speaking); this function both makes an allocation and
+// takes about two microseconds.
 func Stack(key string) zapcore.Field {
 	// Returning the stacktrace as a string costs an allocation, but saves us
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
--- a/field_test.go
+++ b/field_test.go
@@ -164,7 +164,7 @@ func TestFieldConstructors(t *testing.T) {
 }
 
 func TestStackField(t *testing.T) {
-	withStacktraceIgnorePrefixes([]string{}, func() {
+	withNoStacktraceIgnorePrefixes(func() {
 		f := Stack("stacktrace")
 		assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
 		assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
--- a/field_test.go
+++ b/field_test.go
@@ -164,11 +164,9 @@ func TestFieldConstructors(t *testing.T) {
 }
 
 func TestStackField(t *testing.T) {
-	withNoStacktraceIgnorePrefixes(func() {
-		f := Stack("stacktrace")
-		assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
-		assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
-		assert.Contains(t, f.String, "zap.TestStackField", "Expected stacktrace to contain caller.")
-		assertCanBeReused(t, f)
-	})
+	f := Stack("stacktrace")
+	assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
+	assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
+	assert.Contains(t, f.String, "zap.TestStackField", "Expected stacktrace to contain caller.")
+	assertCanBeReused(t, f)
 }
--- a/field_test.go
+++ b/field_test.go
@@ -164,9 +164,11 @@ func TestFieldConstructors(t *testing.T) {
 }
 
 func TestStackField(t *testing.T) {
-	f := Stack("stacktrace")
-	assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
-	assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
-	assert.Contains(t, f.String, "zap.TestStackField", "Expected stacktrace to contain caller.")
-	assertCanBeReused(t, f)
+	withStacktraceIgnorePrefixes([]string{}, func() {
+		f := Stack("stacktrace")
+		assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
+		assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
+		assert.Contains(t, f.String, "zap.TestStackField", "Expected stacktrace to contain caller.")
+		assertCanBeReused(t, f)
+	})
 }
--- a/field_test.go
+++ b/field_test.go
@@ -164,7 +164,7 @@ func TestFieldConstructors(t *testing.T) {
 }
 
 func TestStackField(t *testing.T) {
-	f := stack("stacktrace", 0)
+	f := Stack("stacktrace")
 	assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
 	assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
 	assert.Contains(t, f.String, "zap.TestStackField", "Expected stacktrace to contain caller.")
--- a/field_test.go
+++ b/field_test.go
@@ -164,7 +164,7 @@ func TestFieldConstructors(t *testing.T) {
 }
 
 func TestStackField(t *testing.T) {
-	f := Stack("stacktrace")
+	f := stack("stacktrace", 0)
 	assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
 	assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
 	assert.Contains(t, f.String, "zap.TestStackField", "Expected stacktrace to contain caller.")
--- a/logger_test.go
+++ b/logger_test.go
@@ -360,23 +360,24 @@ func TestLoggerAddCallerFail(t *testing.T) {
 }
 
 func TestLoggerAddStacktrace(t *testing.T) {
-	assertHasStack := func(t testing.TB, obs observer.LoggedEntry) {
-		assert.Contains(t, obs.Entry.Stack, "zap.TestLoggerAddStacktrace", "Expected to find test function in stacktrace.")
-	}
+	withStacktraceIgnorePrefixes([]string{}, func() {
+		assertHasStack := func(t testing.TB, obs observer.LoggedEntry) {
+			assert.Contains(t, obs.Entry.Stack, "zap.TestLoggerAddStacktrace", "Expected to find test function in stacktrace.")
+		}
+		withLogger(t, DebugLevel, opts(AddStacktrace(InfoLevel)), func(logger *Logger, logs *observer.ObservedLogs) {
+			logger.Debug("")
+			assert.Empty(
+				t,
+				logs.AllUntimed()[0].Entry.Stack,
+				"Unexpected stacktrack at DebugLevel.",
+			)
 
-	withLogger(t, DebugLevel, opts(AddStacktrace(InfoLevel)), func(logger *Logger, logs *observer.ObservedLogs) {
-		logger.Debug("")
-		assert.Empty(
-			t,
-			logs.AllUntimed()[0].Entry.Stack,
-			"Unexpected stacktrack at DebugLevel.",
-		)
+			logger.Info("")
+			assertHasStack(t, logs.AllUntimed()[1])
 
-		logger.Info("")
-		assertHasStack(t, logs.AllUntimed()[1])
-
-		logger.Warn("")
-		assertHasStack(t, logs.AllUntimed()[2])
+			logger.Warn("")
+			assertHasStack(t, logs.AllUntimed()[2])
+		})
 	})
 }
 
--- a/logger_test.go
+++ b/logger_test.go
@@ -360,7 +360,7 @@ func TestLoggerAddCallerFail(t *testing.T) {
 }
 
 func TestLoggerAddStacktrace(t *testing.T) {
-	withStacktraceIgnorePrefixes([]string{}, func() {
+	withNoStacktraceIgnorePrefixes(func() {
 		assertHasStack := func(t testing.TB, obs observer.LoggedEntry) {
 			assert.Contains(t, obs.Entry.Stack, "zap.TestLoggerAddStacktrace", "Expected to find test function in stacktrace.")
 		}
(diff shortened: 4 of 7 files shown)


stacktrace.go
```diff
@@ -20,23 +20,45 @@
...
+		// TODO(pedge): we might want to do a better check than this
+		if name == "runtime.main" {
+			break
+		}
+		// heh
+		if i != 0 {
+			buffer.AppendByte('\n')
+		}
+		file, line := f.FileLine(programCounter - 1)
+		buffer.AppendString(f.Name())
```
akshayjshah: We've already captured this as `name`. Let's re-use that instead of allocating again.

peter-edge: Whoops

stacktrace_test.go
```diff
@@ -22,19 +22,26 @@ package zap
...
 }
+
+func TestRunes(t *testing.T) {
+	// https://golang.org/src/bytes/buffer.go?s=8208:8261#L237
+	// I think this test might not be needed, because this might be checked
+	// implicitly by the fact that we can pass these as bytes to AppendByte
+	assert.True(t, '\n' < utf8.RuneSelf, `You can't cast '\n' to a byte, stop being silly`)
+	assert.True(t, '\t' < utf8.RuneSelf, `You can't cast '\t' to a byte, stop being silly`)
+	assert.True(t, ':' < utf8.RuneSelf, `You can't cast ':' to a byte, stop being silly`)
+}
```
peter-edge: Ya as I said, it's kind of a silly test, but there is something we can do, we could AppendString instead of AppendByte if they weren't part of the ASCII range. Anyways ya I'll remove it

peter-edge:
On the strings.Contains, I actually disagree - I can write a test to cleanly verify the skip works properly. It just seems more proper in practice.

I looked over go-stack and it seems like with how it is implemented, it won't work for us due to performance/allocation reasons. I think we should roll our own for now, and get in touch with Chris Hines to see if we can make some changes to that library so it can work for us, because I rather use his code than maintain our own here.

field_test.go
```diff
@@ -162,7 +162,7 @@ func TestFieldConstructors(t *testing.T) {
 }
 
 func TestStackField(t *testing.T) {
-	f := Stack("stacktrace")
+	f := stack("stacktrace", 0)
```
peter-edge: Actually on second thought - I think we actually have to do the thing you said with strings.Contains, because Stack can be called in different ways, which is the problem here.

peter-edge:
Actually on second thought, you're right about strings.Contains - the reason the TestConfig failed was because we called Stack directly, instead of from a logging function. We should support both.

stacktrace.go
```diff
@@ -20,23 +20,45 @@
...
-		size := 2 * n
-		buf = make([]byte, size)
-		n = runtime.Stack(buf, includeAllGoroutines)
+	programCounters = programCounters[:n]
+
+	for i, programCounter := range programCounters {
+		f := runtime.FuncForPC(programCounter)
+		name := f.Name()
+		// this strips everything below runtime.main
+		// TODO(pedge): we might want to do a better check than this
```
peter-edge: go-stack is doing something more intense, and I don't want to just directly copy his code for now. It also seems more expensive. https://github.com/go-stack/stack/blob/master/stack.go#L330

akshayjshah:
By all means - write a test that verifies that the behavior is correct when users call `Stack` directly, when a `*zap.Logger` auto-adds a stacktrace to an error log, and when a `*zap.SugaredLogger` auto-adds a stacktrace to an error log. In all cases, we want to remove zap's frames.

If we can't make this work consistently, I'd prefer to not do it.

peter-edge:
I think I'm going to do strings.Contains, I think that's a safer way, you're right. Should I add a `buffer.UintptrBuffer` type, or am I being stupid? Really I wish we had a more generic buffer type, but alas, one of the times you miss generics...

peter-edge:
Actually what the hell am I talking about, I can do `[]uintptr` directly I think

akshayjshah:
I think you'll actually want something like 

```go
type stack struct {
  pcs []uintptr
}
```

and then you'll pool `*stack`. If you pool `[]uintptr` directly, you'll have to allocate every time you call `Pool.Put` (because `Put` expects to take `interface{}` and slice headers are values). This is exactly why we have the `buffers.Buffer` type - we can't just pool `[]byte`.

Russ Cox wrote a really [good blog post on interfaces](https://research.swtch.com/interfaces) that dives into interfaces. It explains why zap goes through all this `Field` nonsense - it's purely to keep values on the stack instead of casting them into `interface{}`.

stacktrace.go
```diff
@@ -20,23 +20,45 @@
...
+		if name == "runtime.main" {
```
skipor: Can we find "runtime.main" program counter once, and check equality between program counters, but not between strings?
Also I assume that if "runtime.main" it  is always on same index from end (just assumptions, should be verified), so it's could be better to do smth like 
```golang
if len(programConters) >= mainPcIdx && programCounters[mainPcIdx] == mainPc {
    programCounters = programCounters[:mainPcIdx]
}
```

field.go
```diff
@@ -187,14 +187,18 @@ func Error(err error) zapcore.Field {
...
+func stack(key string, skip int) zapcore.Field {
 	// Try to avoid allocating a buffer.
 	buf := bufferpool.Get()
-	bs := buf.Bytes()
 	// Returning the stacktrace as a string costs an allocation, but saves us
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	field := String(key, takeStacktrace(bs[:cap(bs)], false))
+	field := String(key, takeStacktrace(buf, skip))
```
skipor: I think, string allocation for stack is large enough, and it could be fixed.
What about `BufferType` `FieldType`, with `*bufferpool.Buffer` in `Field.Interface` (don't allocate), that on AddTo calls `ObjectEncoder.AddBinary(bufferBytes)`, and then returns buffer to pool?
But in that case we should document, on `ObjectEccoder.AddBinary` that after `AddBinary` bytes under slice can be modified, so `ObjectEncoder` should not retain reference for it. It's reasonable constraint, `io.Writer` have same.  In zap only *MapObjectEncoder retains references for binary slices.

field.go
```diff
@@ -187,14 +187,18 @@ func Error(err error) zapcore.Field {
 // extremely expensive (relatively speaking); this function both makes an
 // allocation and takes ~10 microseconds.
 func Stack(key string) zapcore.Field {
+	return stack(key, 4)
```
skipor: That is exported function and called outside of package too. 
So `skip` should be 1 by default, but `Logger` should use its `callerSkip` to define caller stack frame. 
Also consider add  exported`StackSkip(key string, skip int) zapcore.Field`, to allow caller skip some frames. It's useful in  recover: skip allows set top stack frame panic caller.

stacktrace.go
```diff
@@ -20,23 +20,45 @@
...
+		}
+		// heh
+		if i != 0 {
+			buffer.AppendByte('\n')
+		}
+		file, line := f.FileLine(programCounter - 1)
+		buffer.AppendString(f.Name())
+		buffer.AppendByte('\n')
+		buffer.AppendByte('\t')
+		buffer.AppendString(file)
```
skipor: What about to trim filename to $GOPATH? It makes stack more compact and readable.
Maybe not now, but in future.

skipor: Also possibly, it's better to make smth like `StackEncoder` and use wrapped and pooled `[]uintptr` as stack representation until pass it to `Encoder`.  That makes possible to choose stack format.
Also it's alternative to get rid of  `string` allocation for stack: `StackEncoder` can directly call `PrimitiveArrayEncoder.AddBinary`, which is not present yet, but should, as for me.
Also, when `zap.Stack` called directly, but message is not logged, that allows to not encode stack at all.

field.go
akshayjshah: `AddBinary` won't work here - it's intended for opaque binary blobs (for example, you fail to unmarshal some proto3 blob and want to log it for later debugging). In JSON, binary is base64-encoded, so it won't be a readable stacktrace.

In general, I really don't want to let users push arbitrary bytes into the encoders. It's too difficult to make it encoding-agnostic.

skipor: Answer in  #324.

peter-edge:
I've uploaded a diff that addresses most of the comments, including using `strings.HasPrefix` instead of skipping callers. Note that tests are failing, and I want to go back to my original statement - I think we should skip callers like we were instead, and internally, use the lowercase `stack(key string, skip int)` function in the logging functions. I could do some hacks (ie adjusting the prefixes that are checked) for tests, but note that with how this currently operates, having tests like we have now are almost impossible to do because any test will have a prefix of `go.uber.org/Zap.` in the name of the functions in the stack.

Short: I'm re-advocating for skipping callers internally, and writing good tests for this. Let me know what you think.

peter-edge:
Also I wish we could @Chris Hines about this.

stacktrace.go
```diff
@@ -20,23 +20,72 @@
...
+	"runtime"
+	"strings"
+	"sync"
+
+	"go.uber.org/zap/internal/bufferpool"
+)
+
+var (
+	defaultUintptrSliceSize = 128
+	uintptrPool             = sync.Pool{
```
prashantv: use `_` prefix here

stacktrace.go
```diff
@@ -20,23 +20,72 @@
...
+import (
+	"runtime"
+	"strings"
+	"sync"
+
+	"go.uber.org/zap/internal/bufferpool"
+)
+
+var (
+	defaultUintptrSliceSize = 128
```
prashantv: Not a big issue since we're dealing with 4 byte values, but this is a huge call stack size. Maybe something like `32` would be more reasonable?

stacktrace.go
```diff
@@ -20,23 +20,72 @@
...
 // successively larger slices until it can capture the whole stack.
-func takeStacktrace(buf []byte, includeAllGoroutines bool) string {
-	if len(buf) == 0 {
-		// We may have been passed a nil byte slice.
-		buf = make([]byte, 1024)
+func takeStacktrace() string {
+	buffer := bufferpool.Get()
+	programCounters := getUintptrSlice()
+	n := runtime.Callers(0, programCounters)
+	size := defaultUintptrSliceSize
```
prashantv: Don't use `size` here, use `len(programCounters)`

stacktrace.go
```diff
@@ -20,23 +20,72 @@
...
+	for n >= size {
```
prashantv: You can remove the `size` variable completely, and remove the `runtime.Callers` duplication
```go
programCounters := getUintptrSlice()
for {
  n := runtime.Callers(0, programCounters)
  if n < len(programCounters) {
    programCounters = programCounters[:n]
    break
  }

  programCounters = make([]uintptr, len(programCounters)*2)
}
```

stacktrace.go
```diff
@@ -20,23 +20,72 @@
...
 
 // takeStacktrace attempts to use the provided byte slice to take a stacktrace.
 // If the provided slice isn't large enough, takeStacktrace will allocate
 // successively larger slices until it can capture the whole stack.
-func takeStacktrace(buf []byte, includeAllGoroutines bool) string {
-	if len(buf) == 0 {
-		// We may have been passed a nil byte slice.
-		buf = make([]byte, 1024)
+func takeStacktrace() string {
+	buffer := bufferpool.Get()
```
prashantv: the cost of `defer` is negligible compared to the cost of this method, so consider `defer bufferpool.Put(buffer)`

stacktrace.go
```diff
@@ -20,23 +20,72 @@
...
-	n := runtime.Stack(buf, includeAllGoroutines)
-	for n >= len(buf) {
-		// Buffer wasn't large enough, allocate a larger one. No need to copy
-		// previous buffer's contents.
-		size := 2 * n
-		buf = make([]byte, size)
-		n = runtime.Stack(buf, includeAllGoroutines)
+	programCounters = programCounters[:n]
+
+	i := 0
```
prashantv: instead of a separate variable `i`, can we check the buffer length instead?

stacktrace_test.go
```diff
@@ -27,14 +27,7 @@ import (
...
-	traceTiny := takeStacktrace(make([]byte, 1), false)
-	for _, trace := range []string{traceNil, traceTiny} {
-		// The top frame should be takeStacktrace.
-		assert.Contains(t, trace, "zap.takeStacktrace", "Stacktrace should contain the takeStacktrace function.")
-		// The stacktrace should also capture its immediate caller.
-		assert.Contains(t, trace, "TestTakeStacktrace", "Stacktrace should contain the test function.")
-	}
+	trace := takeStacktrace()
+	// The stacktrace should also its immediate caller.
+	assert.Contains(t, trace, "TestTakeStacktrace", "Stacktrace should contain the test function.")
```
prashantv: since the test is in the `zap` package, why does it appear in the output of `takeStacktrace`?

peter-edge: This is exactly the problem with this way of doing this, see the last comment, this is broken right now

stacktrace.go
peter-edge: We can't, because of how we skip as of this PR - this variable is used for whether to add a newline or not

stacktrace.go
peter-edge: I think it's fine like this though

stacktrace.go
peter-edge: I don't understand why we're doing this at uber - what is the value?

stacktrace.go
peter-edge: That might be ok, but if we're re-using these, why not just make it larger? There might be some lower-level x86 thing I'm forgetting about but

stacktrace.go
akshayjshah: It makes it much, much clearer when you're using a global, which makes it less likely that you'll assign to it or shadow it by mistake. (Shadowing is fine, but it makes the code a bit more brittle - removing the local declaration won't break the later code, but will likely make it incorrect.)

stacktrace.go
```diff
@@ -20,23 +20,72 @@
...
+	for _, programCounter := range programCounters {
+		f := runtime.FuncForPC(programCounter)
+		name := f.Name()
+		if strings.HasPrefix(name, "runtime.") || strings.HasPrefix(name, "go.uber.org/zap.") {
```
akshayjshah: We should include `zapcore` here too; perhaps removing the trailing period?

akshayjshah:
I'm much more of a fan of this approach (not surprising, since I suggested it). After dealing with a frame-skipping approach to caller annotations (which have to be much faster), it's a *royal* pain. This will be even worse, since there are exported call sites.

One simple approach to testing this code is to make the skip prefix a global variable, and (a) assert that the preset value results in an empty stacktrace (since all frames are internal), and then (b) change the variable to match only one of the zap frames and assert that the others show up. This isn't perfect coverage, but it's enough to make me confident.

peter-edge:
Check it out now @akshayjshah

stacktrace.go
```diff
@@ -20,23 +20,75 @@
...
+var (
+	_stacktraceIgnorePrefixes = []string{
+		"go.uber.org/zap",
+		"runtime.",
+	}
+	_stacktraceUintptrPool = sync.Pool{
+		New: func() interface{} {
+			return make([]uintptr, 64)
+		},
+	}
```
akshayjshah: We shouldn't directly pool `[]uintptr`. Slices are actually values (see "Slices: The slice header" in [this Go Blog post](https://blog.golang.org/slices)), so passing them as `interface{}` allocates. Since `sync.Pool.Put` takes an `interface{}`, putting a plain `[]uintptr` back into the pool allocates, which is exactly what we're trying to avoid.

Instead, try this:

```
type programCounters struct {
  pcs []uintptr
}
```

and pool `*programCounters`.

stacktrace.go
akshayjshah: +1 for deferring here, which you've done. Small nit: can we group the defers and the `Get`s, so that acquiring and releasing the resource are on adjacent lines?

stacktrace.go
```diff
@@ -20,23 +20,75 @@
...
-		// We may have been passed a nil byte slice.
-		buf = make([]byte, 1024)
+func takeStacktrace() string {
+	buffer := bufferpool.Get()
+	programCounters := _stacktraceUintptrPool.Get().([]uintptr)
+	defer bufferpool.Put(buffer)
+	defer _stacktraceUintptrPool.Put(programCounters)
+
+	for {
+		n := runtime.Callers(2, programCounters)
```
akshayjshah: Mind leaving a comment explaining this constant to future us?

stacktrace.go
```diff
@@ -20,23 +20,75 @@
...
 // successively larger slices until it can capture the whole stack.
-func takeStacktrace(buf []byte, includeAllGoroutines bool) string {
-	if len(buf) == 0 {
-		// We may have been passed a nil byte slice.
-		buf = make([]byte, 1024)
+func takeStacktrace() string {
+	buffer := bufferpool.Get()
+	programCounters := _stacktraceUintptrPool.Get().([]uintptr)
+	defer bufferpool.Put(buffer)
+	defer _stacktraceUintptrPool.Put(programCounters)
```
akshayjshah: Should we be resetting the length of this slice (`programCounters = programCounters[:0]`) before putting it back into the pool?

stacktrace_test.go
```diff
@@ -21,20 +21,48 @@
...
+		func() {
+			stacktrace := takeStacktrace()
+			assert.True(t, strings.HasPrefix(stacktrace, "go.uber.org/zap.TestAllStacktrace"),
+				"stacktrace should start with TestAllStacktrace if no prefixes set, but is %s", stacktrace)
+		},
+	)
+}
+
+// What happens with require functions? If they completely break out of the
+// function, will this lock never get unlocked?
```
akshayjshah: Let's use `defer`s here, which will let us continue to use `require` in tests.

logger_test.go
```diff
@@ -360,23 +360,24 @@ func TestLoggerAddCallerFail(t *testing.T) {
...
-		logger.Debug("")
-		assert.Empty(
-			t,
-			logs.AllUntimed()[0].Entry.Stack,
-			"Unexpected stacktrack at DebugLevel.",
-		)
+	withStacktraceIgnorePrefixes([]string{}, func() {
+		assertHasStack := func(t testing.TB, obs observer.LoggedEntry) {
+			assert.Contains(t, obs.Entry.Stack, "zap.TestLoggerAddStacktrace", "Expected to find test function in stacktrace.")
+		}
```
akshayjshah: Does this function need to be in the stacktrace-prefix-clobbering closure?

config_test.go
```diff
@@ -68,19 +68,25 @@ func TestConfig(t *testing.T) {
...
-			byteContents, err := ioutil.ReadAll(temp)
-			require.NoError(t, err, "Couldn't read log contents from temp file.")
-			logs := string(byteContents)
-			assert.Regexp(t, tt.expectRe, logs, "Unexpected log output.")
+				byteContents, err := ioutil.ReadAll(temp)
+				// not doing require so no problem with lock in withStacktraceIgnorePrefixes
+				assert.NoError(t, err, "Couldn't read log contents from temp file.")
+				if err != nil {
+					return
+				}
```
akshayjshah: I think we can use `defer` in the decorator and keep the `require` here.

field.go
```diff
@@ -195,16 +194,11 @@ func NamedError(key string, err error) zapcore.Field {
...
-	buf := bufferpool.Get()
-	bs := buf.Bytes()
 	// Returning the stacktrace as a string costs an allocation, but saves us
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	field := String(key, takeStacktrace(bs[:cap(bs)], false))
-	bufferpool.Put(buf)
-	return field
+	return String(key, takeStacktrace())
```
akshayjshah: Nice, much cleaner.

stacktrace.go
```diff
@@ -20,23 +20,75 @@
...
+)
 
 // takeStacktrace attempts to use the provided byte slice to take a stacktrace.
 // If the provided slice isn't large enough, takeStacktrace will allocate
 // successively larger slices until it can capture the whole stack.
-func takeStacktrace(buf []byte, includeAllGoroutines bool) string {
-	if len(buf) == 0 {
-		// We may have been passed a nil byte slice.
-		buf = make([]byte, 1024)
+func takeStacktrace() string {
```
skipor: Consider add  `skip int` argument here, and pass `callerSkip` for taking stack trace in `Logger`. Skip for making `Stack` `Field` from outside of package is constant.
Skipping "go.uber.org/zap" prefixed frames, by passing correct 'skip' to 'runtime.Callers' seems to have much less overhead, than string compare func names.

stacktrace.go
peter-edge: Done

stacktrace.go
peter-edge: There's no need to do this given how we use programCounters - we just overwrite the data in `runtime.Callers` and iterate up to the number of program counters put into the slice.

stacktrace.go
peter-edge: Done

stacktrace_test.go
peter-edge: Done

stacktrace.go
```diff
@@ -20,23 +20,75 @@
...
-		// previous buffer's contents.
-		size := 2 * n
-		buf = make([]byte, size)
-		n = runtime.Stack(buf, includeAllGoroutines)
+
+	i := 0
+	for _, programCounter := range programCounters {
+		f := runtime.FuncForPC(programCounter)
+		name := f.Name()
+		if shouldIgnoreStacktraceName(name) {
```
skipor: Usually only stack suffix and prefix are trimmed, but here frames in middle of stack could be ignored. That can make stack inconsistent and hard to understand. 
For example, if we pass `Stringer` field to logger, which `String() string` call cause panic, and then take `Stack` in `defer`, we will miss valuable `runtime.panic` frames, and `zap` frames that points to problem inside our stringer. 
Test that illustrates example:
```go
type corruptedStringer struct{}
func (*corruptedStringer) String() string {
	var someNilPtr *string
	return *someNilPtr
}
func TestMiddleStackFramesNotSkipped(t *testing.T) {
	l, err := zap.NewProduction()
	require.NoError(t, err)
	defer func() {
		require.NotNil(t, recover(), "should panic")
		f := zap.Stack("stacktrace")
		enc := zapcore.NewMapObjectEncoder()
		f.AddTo(enc)
		stack := enc.Fields["stacktrace"]
		t.Logf("real stack: %s\n", debug.Stack())
		assert.Contains(t, stack, "panic")
		assert.Contains(t, stack, "zap")
	}()
	l.Warn("", zap.Stringer("should panic inside zap package", &corruptedStringer{}))
}
```
Test run outside of zap repo (allows not to skip test frames) fails with message:
```
test_test.go:32: real stack: goroutine 5 [running]:
    runtime/debug.Stack(0x26b8e0, 0xc420019410, 0x2b9581)
        /usr/local/Cellar/go/1.7.4_2/libexec/src/runtime/debug/stack.go:24 +0x79
    github.com/skipor/playground.TestMiddleStackFramesNotSkipped.func1(0xc420088180)
        /Users/skipor/Dev/Go/src/github.com/skipor/playground/test_test.go:32 +0x251
    panic(0x26cee0, 0xc4200120d0)
        /usr/local/Cellar/go/1.7.4_2/libexec/src/runtime/panic.go:458 +0x243
    github.com/skipor/playground.(*corruptedStringer).String(0x413898, 0x266240, 0x413898)
        /Users/skipor/Dev/Go/src/github.com/skipor/playground/test_test.go:20 +0x2
    go.uber.org/zap/zapcore.Field.AddTo(0x2c0171, 0x1f, 0x17, 0x0, 0x0, 0x0, 0x266240, 0x413898, 0x3df600, 0xc4200ccb20)
        /Users/skipor/Dev/Go/src/go.uber.org/zap/zapcore/field.go:152 +0xca1
    go.uber.org/zap/zapcore.addFields(0x3df600, 0xc4200ccb20, 0xc42000cc40, 0x1, 0x1)
        /Users/skipor/Dev/Go/src/go.uber.org/zap/zapcore/field.go:178 +0x74
    go.uber.org/zap/zapcore.(*jsonEncoder).EncodeEntry(0xc4200cc9e0, 0x1, 0xed04740fa, 0x6396a92, 0x3f7160, 0x0, 0x0, 0x0, 0x0, 0x1, ...)
        /Users/skipor/Dev/Go/src/go.uber.org/zap/zapcore/json_encoder.go:312 +0x346
    go.uber.org/zap/zapcore.(*ioCore).Write(0xc4200193b0, 0x1, 0xed04740fa, 0x6396a92, 0x3f7160, 0x0, 0x0, 0x0, 0x0, 0x1, ...)
        /Users/skipor/Dev/Go/src/go.uber.org/zap/zapcore/core.go:85 +0xa9
    go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc4200a6370, 0xc42000cc40, 0x1, 0x1)
        /Users/skipor/Dev/Go/src/go.uber.org/zap/zapcore/entry.go:170 +0x147
    go.uber.org/zap.(*Logger).Warn(0xc4200107e0, 0x0, 0x0, 0xc42000cc40, 0x1, 0x1)
        /Users/skipor/Dev/Go/src/go.uber.org/zap/logger.go:169 +0x7f
    github.com/skipor/playground.TestMiddleStackFramesNotSkipped(0xc420088180)
        /Users/skipor/Dev/Go/src/github.com/skipor/playground/test_test.go:36 +0x22c
    testing.tRunner(0xc420088180, 0x2dc2b8)
        /usr/local/Cellar/go/1.7.4_2/libexec/src/testing/testing.go:610 +0x81
    created by testing.(*T).Run
        /usr/local/Cellar/go/1.7.4_2/libexec/src/testing/testing.go:646 +0x2ec
		
Error Trace:	
    test_test.go:33
    asm_amd64.s:479
    panic.go:458
    panic.go:62
    sigpanic_unix.go:24
    test_test.go:20
    field.go:152
    field.go:178
    json_encoder.go:312
    core.go:85
    entry.go:170
    logger.go:169
    test_test.go:36
Error:		
    "github.com/skipor/playground.TestMiddleStackFramesNotSkipped.func1
        /Users/skipor/Dev/Go/src/github.com/skipor/playground/test_test.go:28
    github.com/skipor/playground.(*corruptedStringer).String
        /Users/skipor/Dev/Go/src/github.com/skipor/playground/test_test.go:19
    github.com/skipor/playground.TestMiddleStackFramesNotSkipped
        /Users/skipor/Dev/Go/src/github.com/skipor/playground/test_test.go:36
    testing.tRunner
        /usr/local/Cellar/go/1.7.4_2/libexec/src/testing/testing.go:610" does not contain "panic"
    
Error Trace:	
    test_test.go:34
    asm_amd64.s:479
    panic.go:458
    panic.go:62
    sigpanic_unix.go:24
    test_test.go:20
    field.go:152
    field.go:178
    json_encoder.go:312
    core.go:85
    entry.go:170
    logger.go:169
    test_test.go:36
Error:	
    "github.com/skipor/playground.TestMiddleStackFramesNotSkipped.func1
        /Users/skipor/Dev/Go/src/github.com/skipor/playground/test_test.go:28
    github.com/skipor/playground.(*corruptedStringer).String
        /Users/skipor/Dev/Go/src/github.com/skipor/playground/test_test.go:19
    github.com/skipor/playground.TestMiddleStackFramesNotSkipped
        /Users/skipor/Dev/Go/src/github.com/skipor/playground/test_test.go:36
    testing.tRunner
        /usr/local/Cellar/go/1.7.4_2/libexec/src/testing/testing.go:610" does not contain "zap"
```
So, consider only to trim `runtime` stack bottom, and `zap` stack top.

logger_test.go
peter-edge: Yes :)

config_test.go
peter-edge: Done

stacktrace.go
peter-edge: This was discussed earlier in this PR

stacktrace.go
peter-edge: I updated to just trim `runtime.goexit` and `runtime.main`.

peter-edge:
All comments are taken care of @akshayjshah

stacktrace.go
peter-edge: Done

stacktrace.go
skipor: `zap` frames in middle of the stack seems to be valuable too.

peter-edge: We've discussed this a lot, I think I'd like to merge this as is and go from there :)

stacktrace.go
akshayjshah: Yep. In the context of taking a stacktrace, which is more than a microsecond, there's no harm in a few cheap string comparisons. I'd very, very much like to avoid another hard-coded caller skip, since it limits our options for structuring code in zap's internals - either every user-facing API must be the same number of stack frames away from this code, or we need to plumb skip arguments everywhere.

peter-edge:
Ping

stacktrace_test.go
```diff
@@ -21,20 +21,51 @@
...
+
+func withStacktraceIgnorePrefixes(prefixes []string, f func()) {
+	_stacktraceTestLock.Lock()
+	defer _stacktraceTestLock.Unlock()
+	existing := _stacktraceIgnorePrefixes
+	_stacktraceIgnorePrefixes = prefixes
+	defer func() {
+		_stacktraceIgnorePrefixes = existing
+	}()
+	f()
```
akshayjshah: I don't think we need either of these decorators anymore.

PR #313, opened 2017-02-17
Add Pusher and have Core take a Pusher instead of WriteSyncer

peter-edge:
This is a major change, but I think we should have this in before 1.0. This adds the concept of a Pusher, which is basically a Writer that also takes the level of the log message. For some endpoints, this additional data will be necessary.


akshayjshah:
Can you elaborate on the use case here? If it's syslog, which it definitely sounds like, this isn't necessary - we've planned to support this already via the `LevelEnabler` interface and tee'd Facilities.

peter-edge:
It's syslog, but regardless I think we should have this, I would like more power to be encapsulated there.

akshayjshah:
This does push more logic into that layer, which I actively avoided. `WriteSyncer` is a dumb recipient of bytes, and `Encoder` is a dumb producer of bytes. Logic to knit the two together should be a `Facility`.

I'll review this and #314 in more detail in a bit, but I'd prefer to support syslog without this change.

PR #314, opened 2017-02-17
Add syslog pusher implementation

peter-edge:
This adds a syslog implementation. This uses the Pusher concept in https://github.com/uber-go/zap/pull/313.


peter-edge:
I'll fix lint once we get go 1.8 linting in.

PR #315, opened 2017-02-18
RFC: rough functioning syslog encoder and net.Conn sink

supershabam:
Greetings; I'd like to see a syslog encoder and the ability to sync to `/dev/log` over unixgram.

I've implemented a very rough general `syslogEncoder` and `connSyncer`. I wanted to get feedback on this approach before spending too much time.

It seems that building off the encoder and sink interfaces is desired from reading https://github.com/uber-go/zap/pull/313

### DIFF
--- a/zapcore/conn_syncer.go
+++ b/zapcore/conn_syncer.go
@@ -1,0 +1,63 @@
+package zapcore
+
+import (
+	"net"
+	"sync"
+)
+
+var _ WriteSyncer = &connSyncer{}
+
+type connSyncer struct {
+	conn net.Conn
+	dial func() (net.Conn, error)
+	m    sync.Mutex
+}
+
+type ConnSyncerConfig struct {
+	Dial func() (net.Conn, error)
+}
+
+func NewConnSyncer(cfg ConnSyncerConfig) (*connSyncer, error) {
+	conn, err := cfg.Dial()
+	if err != nil {
+		return nil, err
+	}
+	return &connSyncer{
+		conn: conn,
+		dial: cfg.Dial,
+	}, nil
+}
+
+func (cs *connSyncer) Write(p []byte) (n int, err error) {
+	conn := cs.conn
+	if conn != nil {
+		if n, err := conn.Write(p); err == nil {
+			// optimal path; conn already exists and no error writing; no mutex; net.Conn allows concurrent access
+			// each write should be a full message (at least with unix datagram which I'm building for... but may not be true)
+			return n, nil
+		}
+		cs.m.Lock()
+		conn.Close()
+		cs.conn = nil
+		cs.m.Unlock()
+	}
+	// assume we need to build the conn, but we'll do this without the mutex and discard our conn if a concurrent conn is
+	// created.
+	conn, err = cs.dial()
+	if err != nil {
+		return 0, err
+	}
+	cs.m.Lock()
+	if cs.conn != nil {
+		conn.Close()
+		conn = cs.conn
+	} else {
+		cs.conn = conn
+	}
+	cs.m.Unlock()
+	return conn.Write(p)
+}
+
+func (cs *connSyncer) Sync() error {
+	return nil
+}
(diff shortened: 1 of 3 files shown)


peter-edge:
This looks strictly better than what I proposed, but seems like it may put more logic into the encoder than necessary? There's a few other things (caching os.Hostname()) for example, but just my two cents.

zapcore/conn_syncer.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil, err
+	}
+	return &connSyncer{
+		conn: conn,
+		dial: cfg.Dial,
+	}, nil
+}
+
+func (cs *connSyncer) Write(p []byte) (n int, err error) {
+	conn := cs.conn
```
noxiouz: Data race: read without lock. Looks like sync/atomic.Value fits better

akshayjshah:
I will absolutely take a look at this, but won't get to it until later this week. Apologies!

In general (without even looking at the proposed changes), I expect any syslog-related `Core` to be a bit messy. I'm not an expert, but it seems to me that syslog mixes `Encoder` and `WriteSyncer` concerns pretty freely.

zapcore/syslog_encoder.go
```diff
@@ -0,0 +1,274 @@
...
+	Hostname  string
+	PID       int
+	Tag       string
+}
+
+type syslogEncoder struct {
+	cee      bool
+	facility syslog.Priority
+	hostname string // TODO make configurable
+	je       *jsonEncoder
```
jcorbin: This may be better as a value embed, e.g.:
```
type syslogEncoder {
  jsonEncoder
  cee bool
  // ...
}
```

zapcore/syslog_encoder.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func putSyslogEncoder(enc *syslogEncoder) {
+	enc.cee = false
+	enc.facility = 0
+	enc.hostname = ""
+	enc.je = nil
+	enc.pid = 0
+	enc.tag = ""
+	_jsonPool.Put(enc)
```
jcorbin: If we can embed, then we just need to reset all embedded fields here.

Consider doing `*enc = syslogEncoder{}` rather than needing to list all the fields.

zapcore/syslog_encoder.go
```diff
@@ -0,0 +1,274 @@
...
+		enc.pid = os.Getpid()
+	}
+	enc.tag = cfg.Tag
+	if len(enc.tag) == 0 {
+		enc.tag = os.Args[0]
+	}
+	return enc
+}
+
+func (enc *syslogEncoder) AddArray(key string, arr ArrayMarshaler) error {
```
jcorbin: e.g. with an embed, we shouldn't need all these `Add*` and `Append*` methods

zapcore/syslog_encoder.go
```diff
@@ -0,0 +1,274 @@
...
+	clone.je = enc.je.clone()
+	clone.pid = enc.pid
+	clone.hostname = enc.hostname
+	clone.tag = enc.tag
+	return clone
+}
+
+func (enc *syslogEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer, error) {
+	var p syslog.Priority
+	switch ent.Level {
```
jcorbin: I'd mildly prefer to see this switch embodied in something like `func levelToSyslogPriority(Level) syslog.Priority`

peter-edge:
@supershabam is this still something you're interested in? I'd love to see this in :)

supershabam:
@peter-edge sorry; I'm likely not getting back to this soon. Feel free to build on top of this or rework it in your own way if you want some progress you can rely on :)

peter-edge:
Alright, thanks for the response :) I may work on this if I have some free time in the future. @akshayjshah should we close this until then?

akshayjshah:
Closing sounds good - we can always resurrect the branch. Many thanks for the work, @supershabam!

PR #318, opened 2017-02-18
Fix README new logger sample

chapsuk:
https://github.com/uber-go/zap/blob/master/logger.go#L73


akshayjshah:
Oh dear, you're absolutely right.

PR #321, opened 2017-02-19
 AddCallerSkip in RedirectStdLog (#320)

### LINKED ISSUE #320: RedirectStdLog is not adding callerSkip
RedirectStdLog should add correct caller skip, so entries logged through `log.Print` have correct caller.

skipor:
Make entries logged through log.Print have correct caller. Fixes #320.


akshayjshah:
Looks great - I'm going to push a few small changes and then land.

PR #325, opened 2017-02-19
Support non-standard, rich error types

### LINKED ISSUE #303: Support pkg/errors Stacktracer interface
Consider whether we should take on a third-party dependency to support `pkg/errors`. If we really think that this package is garnering widespread adoption, we should support its error type and include both stacktraces and causes in log output.

Thoughts, @prashantv and @jcorbin?

akshayjshah:
Users who take the time (and performance hit) of using Dave Cheney's errors
package should get rich, annotated errors and stacktraces in their logs.
However, that package is only one of a number of similar packages and hasn't yet
committed to a stable API, so we don't want to take a direct dependency on it.

To support these sorts of packages, serialize errors that implement
`fmt.Formatter` differently from those that don't. This is necessarily slower than
serializing standard errors, but users of these packages have already committed
to a more expensive error-handling path.

@sectioneight, this should make UberFX's logs nicer.

Fixes #303.

### DIFF
--- a/zapcore/field.go
+++ b/zapcore/field.go
@@ -151,7 +151,14 @@ func (f Field) AddTo(enc ObjectEncoder) {
 	case StringerType:
 		enc.AddString(f.Key, f.Interface.(fmt.Stringer).String())
 	case ErrorType:
-		enc.AddString(f.Key, f.Interface.(error).Error())
+		val := f.Interface.(error)
+		if fancy, ok := val.(fmt.Formatter); ok {
+			// This is a rich error type, like those produced by
+			// github.com/pkg/errors.
+			enc.AddString(f.Key, fmt.Sprintf("%+v", fancy))
+		} else {
+			enc.AddString(f.Key, val.Error())
+		}
 	case SkipType:
 		break
 	default:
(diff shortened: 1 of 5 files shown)


peter-edge:
LGTM

akshayjshah:
Certainly don't mind doing it that way. I'll update.

zapcore/field.go
```diff
@@ -151,7 +151,13 @@ func (f Field) AddTo(enc ObjectEncoder) {
 	case StringerType:
 		enc.AddString(f.Key, f.Interface.(fmt.Stringer).String())
 	case ErrorType:
-		enc.AddString(f.Key, f.Interface.(error).Error())
+		val := f.Interface.(error)
+		if fancy, ok := val.(fmt.Formatter); ok {
+			// This is a rich error type, like those produced by
+			// github.com/pkg/errors.
```
prashantv: Suggest verifying that the verbose error is actually different from the error message before adding.

It won't cost much extra (since the string length check will probably fail first)

akshayjshah: Sure, done - forgot that you suggested that in the first round.

Issue #319, opened 2017-02-19
Make caller formatting configurable


skipor:
I have problem like #304, but I want truncate file paths to <parent_folder_name/file_name>. Because of really long  my app package path, even truncating to $GOPATH is not enough.
I know that in #201 decided that 
> Support for mutative external hooks has been deemed "not supported", and won't be coming back.

And, as result, there is no way no modify `Entry.Caller` via `Option`
So I found only way to do this: wrap `zapcore.Encoder`, and truncate `Entry.Caller.File` before calling wrapped `Encoder`. But that is really ugly way, and can't be used with beautiful `Config` helpers.

My best idea how to improve usability in this case is add kind of `EncodeCaller` into `EncoderConfig` and use it to encode `Caller`. Also, this is nice way to solve #304.

Other ways to improve usability:
- Add modifying hooks, that are called after `Core` check
  - Not best idea.
- Allow wrap `Encoder` easier.
  - Make `Config` methods exported to add ability for manual `Logger` build.

akshayjshah:
Good point - I hadn't thought about making the caller formatting configurable, but it's a good idea. Since you have a PR open, I'll move discussion there.

akshayjshah:
Fixed in #327.

Issue #304, opened 2017-02-16
Truncate file paths to $GOPATH


akshayjshah:
When adding caller annotations, it's irritating to use the absolute path - the caller annotations become quite long. We should instead report the file's location with $GOPATH stripped off (which should leave files outside $GOPATH unchanged).

skipor:
[go-stack have nice implementation of this ](https://github.com/go-stack/stack/blob/master/stack.go#L344) that can be used as starting point

akshayjshah:
Closing in favor of #319.

PR #327, opened 2017-02-19
Make caller serialization configurable

skipor:
$GOPATH trim logic based on `go-stack/stack` which is under Apache license, so can't be put in repo. So required logic have been moved to another repo (skipor/goenv), and exported as third party. 
That decision is doubtful, so fell free to propose something better.
WARN: skipor/goenv is not pinned in glide.lock. I don't know how to do this without updating and keeping glide.lock hash valid.

### DIFF
--- a/benchmarks/zap_test.go
+++ b/benchmarks/zap_test.go
@@ -69,8 +69,7 @@ var _jane = user{
 
 func newZapLogger(lvl zapcore.Level) *zap.Logger {
 	// use the canned production encoder configuration
-	cfg := zap.NewProductionConfig()
-	enc := zapcore.NewJSONEncoder(cfg.EncoderConfig)
+	enc := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())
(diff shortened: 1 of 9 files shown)


peter-edge:
I don't think copying someone else's code out to your own repo is a nice thing to do, also you just copied the license so this doesn't change anything. We were thinking of using Chris Hines' code in https://github.com/uber-go/zap/pull/311. This code somewhat duplicates that effort as well.

skipor:
Well, I remove `GoPathCallerEncoder` commit with external dependency. 
This PR still fixes #319, and nice start point for #304.

flyingmutant:
Does this PR have a chance to land before 1.0?

skipor:
@peter-edge, @akshayjshah need I fix something, or your will handle this PR yourself?

akshayjshah:
There's certainly a chance that this can land before 1.0; it's not a breaking change, so we can also land it after 1.0 if necessary.

@skipor we can take it from here - I just need to spend the time to review it carefully :)

PR #333, opened 2017-02-23
Further error-logging improvements

### LINKED ISSUE #303: Support pkg/errors Stacktracer interface

akshayjshah:
First, add support for customizing error fields' keys; now that we support rich errors, we can't simply rely on string fields.

Second, use that support in `Any` and `Errors`. This should complete our support for `github.com/pkg/errors`.

Fully resolves #303.


array.go
```diff
@@ -311,7 +316,26 @@ func (errs errArray) MarshalLogArray(arr zapcore.ArrayEncoder) error {
...
+
+type errArrayElem struct {
+	error
+}
+
+func (e *errArrayElem) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	// Re-use the error field's logic, which supports non-standard error types.
+	Error(e.error).AddTo(enc)
+	return nil
+
```
sectioneight: remove extra newline?

akshayjshah: :+1:

array_test.go
```diff
@@ -94,7 +94,11 @@ func TestArrayWrappers(t *testing.T) {
 		{"uint16s", Uint16s("", []uint16{1, 2}), []interface{}{uint16(1), uint16(2)}},
 		{"uint8s", Uint8s("", []uint8{1, 2}), []interface{}{uint8(1), uint8(2)}},
 		{"uintptrs", Uintptrs("", []uintptr{1, 2}), []interface{}{uintptr(1), uintptr(2)}},
-		{"errors", Errors("", []error{nil, errors.New("foo"), nil, errors.New("bar")}), []interface{}{"foo", "bar"}},
+		{
+			"errors",
+			Errors("", []error{nil, errors.New("foo"), nil, errors.New("bar")}),
+			[]interface{}{map[string]interface{}{"error": "foo"}, map[string]interface{}{"error": "bar"}},
```
prashantv: should we add a test which also adds `errorVerbose`?

array.go
```diff
@@ -311,7 +316,25 @@ func (errs errArray) MarshalLogArray(arr zapcore.ArrayEncoder) error {
...
+		_errArrayElemPool.Put(elem)
 	}
 	return nil
 }
+
+type errArrayElem struct {
+	error
+}
+
+func (e *errArrayElem) MarshalLogObject(enc zapcore.ObjectEncoder) error {
```
prashantv: this does mean our output for `[]error` ends up being a list of objects which have a key `error`. I don't have strong opinions on this, but it does seem a little strange.

akshayjshah: Yeah, it's definitely a little odd. I'm honestly not sure what else to do - it seems worse to have arrays that mix strings and objects. Would you prefer that? Or separate arrays of the errors and verbose errors?

array_test.go
akshayjshah: Sure, I can. Those code paths are already covered, but doesn't hurt to do it again.

PR #336, opened 2017-02-24
Update benchmarks

peter-edge:
It's EOD Friday here in Amsterdam, so I thought I'd do something fun :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,3 +69,10 @@ coveralls:
 BENCH ?= .
 bench:
 	@$(foreach pkg,$(PKGS),go test -bench=$(BENCH) -run="^$$" $(BENCH_FLAGS) $(pkg);)
+
+.PHONY: updatebench
+updatebench:
+	@if ! bash --version | grep 'version 4' > /dev/null; then \
+		echo "You need to use bash 4 to run update_benchmarks.sh"; \
+	fi
+	bash -x update_benchmarks.sh
--- a/README.md
+++ b/README.md
@@ -82,32 +82,35 @@ Log a message and 10 fields:
 
 | Library | Time | Bytes Allocated | Objects Allocated |
 | :--- | :---: | :---: | :---: |
-| :zap: zap | 1436 ns/op | 705 B/op | 2 allocs/op |
-| :zap: zap (sugared) | 2436 ns/op | 1931 B/op | 21 allocs/op |
-| logrus | 9393 ns/op | 5783 B/op | 77 allocs/op |
-| go-kit | 6929 ns/op | 3119 B/op | 65 allocs/op |
-| log15 | 25004 ns/op | 5535 B/op | 91 allocs/op |
-| apex/log | 18450 ns/op | 4025 B/op | 64 allocs/op |
+|:zap: zap|       826 ns/op|     706 B/op|       2 allocs/op|
+|:zap: zap (sugared)|      1340 ns/op|    1614 B/op|      20 allocs/op|
+|logrus|      5340 ns/op|    6103 B/op|      78 allocs/op|
+|go-kit|      3363 ns/op|    2898 B/op|      66 allocs/op|
+|log15|     18935 ns/op|    5634 B/op|      93 allocs/op|
+|apex/log|     15290 ns/op|    3835 B/op|      65 allocs/op|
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -37,6 +46,18 @@ logger.Info("Failed to fetch URL.",
 )
 ```
 
+## Development Status: Release Candidate 1
+The current release is `v1.0.0-rc.2`. No further breaking changes are *planned*
```
breerly: If we add this here, we should also add a check to `make version-check` to ensure this doesn't diverge.

Makefile
```diff
@@ -69,3 +69,10 @@ coveralls:
 BENCH ?= .
 bench:
 	@$(foreach pkg,$(PKGS),go test -bench=$(BENCH) -run="^$$" $(BENCH_FLAGS) $(pkg);)
+
+.PHONY: updatebench
+updatebench:
+	@if ! bash --version | grep 'version 4' > /dev/null; then \
+		echo "You need to use bash 4 to run update_benchmarks.sh"; \
```
akshayjshah: Oh dear God. The default on Macs is bash 3.2...

README.md
```diff
@@ -1,5 +1,14 @@
 # :zap: zap [![GoDoc][doc-img]][doc] [![Build Status][ci-img]][ci] [![Coverage Status][cov-img]][cov]
 
+[doc-img]: https://godoc.org/go.uber.org/zap?status.svg
+[doc]: https://godoc.org/go.uber.org/zap
+[ci-img]: https://travis-ci.org/uber-go/zap.svg?branch=master
+[ci]: https://travis-ci.org/uber-go/zap
+[cov-img]: https://coveralls.io/repos/github/uber-go/zap/badge.svg?branch=master
+[cov]: https://coveralls.io/github/uber-go/zap?branch=master
+[benchmarking suite]: https://github.com/uber-go/zap/tree/master/benchmarks
+[glide.lock]: https://github.com/uber-go/zap/blob/master/glide.lock
```
akshayjshah: This is odd to me; I'd prefer to keep the links at the bottom of the doc.

README.md
akshayjshah: This is temporary - once we're out of RC, I'm planning to make this much less specific.

README.md
```diff
@@ -37,6 +46,18 @@ logger.Info("Failed to fetch URL.",
...
+The current release is `v1.0.0-rc.2`. No further breaking changes are *planned*
+unless wider use reveals critical bugs or usability issues, but users who need
+absolute stability should wait for the 1.0.0 release.
+
+<hr>
+Released under the [MIT License](LICENSE.txt).
+
+<sup id="footnote-versions">1</sup> In particular, keep in mind that we may be
+benchmarking against slightly older versions of other libraries. Versions are
+pinned in zap's [glide.lock][] file. [↩](#anchor-versions)
```
akshayjshah: This stuff also needs to be at the bottom of the doc.

README.md
peter-edge: Another PR then I would say haha :) really this can just be a grep in the Makefile

Makefile
peter-edge: Aw man :) Ya I mean I could do the whole thing in golang, I was just messing around. I can make it not-bash-4 too (bash 3 doesn't do associative arrays). Could I do that, and then later on we make this golang?

README.md
peter-edge: It has no effect I think. The whole thing is the `head` call at the end of the script - I could do something like `.readme-template.md` that has a spot I grep for, but I thought I'd just do this for now.

Makefile
akshayjshah: Yep, works for me. As long as it works in bash 3 I'm good.

README.md
akshayjshah: Yeah, it doesn't affect the rendered output - it just makes the raw markdown look weird.

akshayjshah:
I think that the best solution here is a small Go program that generates the README. That's a bit of a pain, though.

In the interim, let's aim a bit lower. Instead of having the script modify the README, let's just have the script write markdown-formatted tables to stdout. Then a maintainer can 

```
bash ./gen-benchmark-tables.sh | pbcopy
```

and manually update the doc. It's a big step forwards from where we are today, but easier to get right than complete automation.

PR #340, opened 2017-02-27
Extend bufferpool test

willhug:
Summary: Adds on to the bufferpool test to write random bytes into
the buffers before putting them back into the sync.Pool

### DIFF
--- a/internal/bufferpool/bufferpool_test.go
+++ b/internal/bufferpool/bufferpool_test.go
@@ -21,6 +21,7 @@
 package bufferpool
 
 import (
+	"math/rand"
 	"sync"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


internal/bufferpool/bufferpool_test.go
```diff
@@ -36,10 +37,23 @@ func TestBuffers(t *testing.T) {
 				buf := Get()
 				assert.Zero(t, buf.Len(), "Expected truncated buffer")
 				assert.NotZero(t, buf.Cap(), "Expected non-zero capacity")
+
+				b := getRandBytes()
```
prashantv: do we really need random data? maybe just write a constant to simplify this
```go
const dummyData = "dummy data"

buf.WriteString(dummyData)
assert.Equal(t, buf.Len(), len(dummyData), "...")
```

internal/bufferpool/bufferpool_test.go
```diff
@@ -36,10 +37,23 @@ func TestBuffers(t *testing.T) {
...
 			}
 			wg.Done()
 		}()
 	}
 	wg.Wait()
 }
+
+func getRandBytes() []byte {
+	b := make([]byte, rand.Intn(5000))
+	rand.Read(b)
```
prashantv: technically read can return an error that is being ignored here. might be cleaner to not use this at all

internal/bufferpool/bufferpool_test.go
```diff
@@ -21,13 +21,14 @@
...
 
 import (
-	"math/rand"
 	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
+const dummyData = "dummy data"
```
akshayjshah: Please underscore-prefix globals.

internal/bufferpool/bufferpool_test.go
```diff
@@ -27,6 +27,8 @@ import (
...
```
prashantv: i'd move the const into the test since it's not used anywhere else

PR #344, opened 2017-03-03
support hijacking other-than-"std" log.logger

flisky:
Sometimes other library requires a log.Logger instance,
but the builtin log doesn't expose the std variable.

### DIFF
--- a/.readme.tmpl
+++ b/.readme.tmpl
@@ -57,40 +57,15 @@ id="anchor-versions">[1](#footnote-versions)</sup>
 
 Log a message and 10 fields:
 
-| Library | Time | Bytes Allocated | Objects Allocated |
-| :--- | :---: | :---: | :---: |
-| :zap: zap | 694 ns/op | 705 B/op | 2 allocs/op |
-| :zap: zap (sugared) | 1174 ns/op | 1613 B/op | 20 allocs/op |
-| logrus | 6824 ns/op | 6100 B/op | 78 allocs/op |
-| go-kit | 3830 ns/op | 2897 B/op | 66 allocs/op |
-| log15 | 16650 ns/op | 5632 B/op | 93 allocs/op |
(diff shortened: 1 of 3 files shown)


global_test.go
```diff
@@ -92,6 +92,18 @@ func TestGlobalsConcurrentUse(t *testing.T) {
...
 
+func TestNewStdLog(t *testing.T) {
+	withLogger(t, DebugLevel, nil, func(l *Logger, logs *observer.ObservedLogs) {
+		std := NewStdLog(l)
+		std.Print("redirected")
+
+		assert.Equal(t, []observer.LoggedEntry{{
+			Entry:   zapcore.Entry{Message: "redirected"},
+			Context: []zapcore.Field{},
+		}}, logs.AllUntimed(), "Unexpected global log output.")
```
akshayjshah: nit: the output here isn't global. Please adjust the error message to clarify :)

global.go
```diff
@@ -67,6 +67,17 @@ func ReplaceGlobals(logger *Logger) func() {
 	return func() { ReplaceGlobals(prev) }
 }
 
+// NewStdLog returns a standard logger which writes to the supplied zap logger at InfoLevel.
+func NewStdLog(l *Logger) *log.Logger {
+	const (
+		stdLogDefaultDepth = 4
+		loggerWriterDepth  = 1
+	)
```
akshayjshah: Now that we're using these constants in two places, let's make them global. Please remember that this project's style underscore-prefixes globals, so this will be 

```
const (
  _stdLogDefaultDepth = 4
  _loggerWriterDepth = 1
)
```

flisky:
Thank you, and please take another look.

akshayjshah:
Hrm, GitHub appears to be interpreting raw git actions oddly. Will re-open in another PR and merge.

PR #345, opened 2017-03-06
config: export DevelopmentEncoderConfig and ProductionEncoderConfig

suyash:
This makes creating loggers with custom write paths easier.

For example

```go
var ws zapcore.WriteSync // probably obtained using zap.Open

core := zapcore.NewCore(
	zapcore.NewConsoleEncoder(zap.ProductionEncoderConfig),
	ws, zapcore.DebugLevel)

logger := zap.New(core)
```

instead of copying the entire `ProductionEncoderConfig` from source.

- [ ] added tests to cover your changes; (Not Applicable)

Thanks for your contribution!

### DIFF
--- a/config.go
+++ b/config.go
@@ -77,6 +77,22 @@ type Config struct {
 	InitialFields map[string]interface{} `json:"initialFields" yaml:"initialFields"`
 }
 
+// NewProductionEncoderConfig returns an opinionated EncoderConfig for
+// production environments.
+func NewProductionEncoderConfig() zapcore.EncoderConfig {
+	return zapcore.EncoderConfig{
+		TimeKey:        "ts",
+		LevelKey:       "level",
+		NameKey:        "logger",
+		CallerKey:      "caller",
+		MessageKey:     "msg",
+		StacktraceKey:  "stacktrace",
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -77,6 +77,20 @@ type Config struct {
 	InitialFields map[string]interface{} `json:"initialFields" yaml:"initialFields"`
 }
 
+// ProductionEncoderConfig is an opinionated EncoderConfig for
+// production environments
+var ProductionEncoderConfig = zapcore.EncoderConfig{
```
peter-edge: Can you make this and the development config `NewProductionEncoderConfig()/NewDevelopmentEncoderConfig()`?  I fear some user will do:

```
ProductionEncoderConfig.TimeKey = "foo"
...
SomeBug()
```

I don't want this to be mutable. It's very unlikely since in general, zap configs will only be done on setup, but with fx for example, if it does something different/access this in some way, then we could run into a bug.

peter-edge:
LGTM but wait for @akshayjshah

akshayjshah:
Looks good. Thanks for the contribution, @suyash!

PR #346, opened 2017-03-06
writer: add CombineWriteSyncers

suyash:
Allowing functions to take interfaces as parameters allows for better
customization.

For example, to have a writer that also counts the number of writes
specifically made to its instance

```go
type MyWriter struct {
	*os.File
	logcount int
}

func (w *MyWriter) Write(data []byte) (int, error) {
	w.logcount++
	return w.File.Write(data)
}

func (w *MyWriter) Sync() error {
	return w.File.Sync()
}

...

func main() {
	f, err := os.Create(...)
	if err != nil {
		log.Fatal(err)
	}
	defer f.Close()

	w := &MyWriter{f, 0}
	ws, err := zap.OpenWriters(os.Stdout, w)
	if err != nil {
		log.Fatal(err)
	}

	core := zapcore.NewCore(
		zapcore.NewConsoleEncoder(zap.NewProductionEncoderConfig()),
		ws, zapcore.DebugLevel)

	logger := zap.New(core)

	...

	fmt.Printf("logged %v times to MyWriter", w.logcount)
}
```

Thanks for your contribution!

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -36,15 +36,14 @@ import (
 // Passing no paths returns a no-op WriteSyncer. The special paths "stdout" and
 // "stderr" are interpreted as os.Stdout and os.Stderr, respectively.
 func Open(paths ...string) (zapcore.WriteSyncer, func(), error) {
-	if len(paths) == 0 {
-		return zapcore.AddSync(ioutil.Discard), func() {}, nil
+	writers, close, err := open(paths)
+
+	if err != nil {
+		return nil, nil, err
 	}
 
-	writers, close, err := open(paths)
-	if len(writers) == 1 {
-		return zapcore.Lock(writers[0]), close, err
-	}
-	return zapcore.Lock(zapcore.NewMultiWriteSyncer(writers...)), close, err
+	writer, err := OpenWriteSyncers(writers...)
+	return writer, close, err
 }
 
 func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
@@ -75,4 +74,13 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
 		}
 	}
 	return writers, close, errs.AsError()
+}
+
+// OpenWriteSyncers combines the passed set of WriteSyncer objects into a locked
+// WriteSyncer. It returns any error encountered.
+func OpenWriteSyncers(writers ...zapcore.WriteSyncer) (zapcore.WriteSyncer, error) {
+	if len(writers) == 0 {
+		return zapcore.AddSync(ioutil.Discard), nil
+	}
+	return zapcore.Lock(zapcore.NewMultiWriteSyncer(writers...)), nil
 }
--- a/writer.go
+++ b/writer.go
@@ -37,13 +37,12 @@ import (
 // "stderr" are interpreted as os.Stdout and os.Stderr, respectively.
 func Open(paths ...string) (zapcore.WriteSyncer, func(), error) {
 	writers, close, err := open(paths)
-
 	if err != nil {
 		return nil, nil, err
 	}
 
-	writer, err := OpenWriteSyncers(writers...)
-	return writer, close, err
+	writer := CombineWriteSyncers(writers...)
+	return writer, close, nil
 }
 
 func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
@@ -76,11 +75,11 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
 	return writers, close, errs.AsError()
 }
 
-// OpenWriteSyncers combines the passed set of WriteSyncer objects into a locked
-// WriteSyncer. It returns any error encountered.
-func OpenWriteSyncers(writers ...zapcore.WriteSyncer) (zapcore.WriteSyncer, error) {
+// CombineWriteSyncers combines the passed set of WriteSyncer objects into a
+// locked WriteSyncer.
+func CombineWriteSyncers(writers ...zapcore.WriteSyncer) zapcore.WriteSyncer {
 	if len(writers) == 0 {
-		return zapcore.AddSync(ioutil.Discard), nil
+		return zapcore.AddSync(ioutil.Discard)
 	}
-	return zapcore.Lock(zapcore.NewMultiWriteSyncer(writers...)), nil
+	return zapcore.Lock(zapcore.NewMultiWriteSyncer(writers...))
 }
--- a/writer_test.go
+++ b/writer_test.go
@@ -81,3 +81,28 @@ func TestOpen(t *testing.T) {
 		assert.Equal(t, tt.filenames, names, "Opened unexpected files given paths %v.", tt.paths)
 	}
 }
+
+type testWriter struct {
+	expected []byte
+	t        testing.TB
+}
+
+func (w *testWriter) Write(actual []byte) (int, error) {
+	assert.Equal(w.t, w.expected, actual, "expected writer to write %v, wrote %v", string(w.expected), string(actual))
+	return 0, nil
+}
+
+func (w *testWriter) Sync() error {
+	return nil
+}
+
+func TestOpenWriteSyncers(t *testing.T) {
+	tw := &testWriter{[]byte("test"), t}
+
+	w, err := OpenWriteSyncers(tw)
+	if err != nil {
+		require.NoError(t, err, "OpenWriters Failed.")
+	}
+
+	w.Write([]byte("test"))
+}
--- a/writer_test.go
+++ b/writer_test.go
@@ -83,13 +83,13 @@ func TestOpen(t *testing.T) {
 }
 
 type testWriter struct {
-	expected []byte
+	expected string
 	t        testing.TB
 }
 
 func (w *testWriter) Write(actual []byte) (int, error) {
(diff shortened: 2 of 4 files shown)


suyash:
cc @peter-edge @akshayjshah

writer.go
```diff
@@ -76,3 +76,33 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
 	}
 	return writers, close, errs.AsError()
 }
+
+// OpenWriters combines the passed io.Writer objects that are compatible with
+// zapcore.WriteSyncer into a locked WriteSyncer. It returns any errors
+// encountered.
+func OpenWriters(writers ...io.Writer) (zapcore.WriteSyncer, error) {
```
peter-edge: I don't think this function should be here - it somewhat defeats the purpose of static typing. Instead, there is a function in this package that can turn a Writer into a WriteSyncer, and then a user can pass the converted Writer to OpenWriteSyncers

writer.go
```diff
@@ -76,3 +76,33 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
...
+		}
+	}
+
+	return OpenWriteSyncers(w...)
+}
+
+// OpenWriteSyncers combines the passed set of WriteSyncer objects into a locked
+// WriteSyncer. It returns any error encountered
+func OpenWriteSyncers(writers ...zapcore.WriteSyncer) (zapcore.WriteSyncer, error) {
+	if len(writers) == 0 {
```
peter-edge: I might just return nil here

writer.go
```diff
@@ -76,3 +76,33 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
...
+		return zapcore.AddSync(ioutil.Discard), nil
+	}
+
+	if len(writers) == 1 {
+		return zapcore.Lock(writers[0]), nil
+	}
+
```
peter-edge: nit: there's a lot of newlines here, maybe just remove them :)

writer.go
akshayjshah: Agreed. In cases where this is needed, I'd much prefer to have users deal with it explicitly.

Please remove this function.

writer.go
akshayjshah: I'd prefer to return a discarder, purely for consistency with `New`.

writer.go
```diff
@@ -76,3 +76,33 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
...
+// OpenWriteSyncers combines the passed set of WriteSyncer objects into a locked
+// WriteSyncer. It returns any error encountered
+func OpenWriteSyncers(writers ...zapcore.WriteSyncer) (zapcore.WriteSyncer, error) {
+	if len(writers) == 0 {
+		return zapcore.AddSync(ioutil.Discard), nil
+	}
+
+	if len(writers) == 1 {
+		return zapcore.Lock(writers[0]), nil
+	}
```
akshayjshah: Let's actually make this improvement to `zapcore.NewMultiWriteSyncer` - if it's passed just one argument, it should return that `WriteSyncer` unchanged. This wrapper can then rely on that behavior.

writer_test.go
```diff
@@ -81,3 +81,39 @@ func TestOpen(t *testing.T) {
 		assert.Equal(t, tt.filenames, names, "Opened unexpected files given paths %v.", tt.paths)
 	}
 }
+
+type testWriter struct {
+	expected []byte
+	t        *testing.T
```
akshayjshah: For future extensibility, let's make this a `testing.TB`.

writer_test.go
```diff
@@ -81,3 +81,39 @@ func TestOpen(t *testing.T) {
...
+	return nil
+}
+
+func TestOpenWriters(t *testing.T) {
+	tw := &testWriter{[]byte("test"), t}
+
+	w, err := OpenWriters(tw)
+	if err != nil {
+		t.Fatalf("OpenWriters failed, error: %v", err)
+	}
```
akshayjshah: Instead, please use the `require` package, which is already included as part of testify.

```
require.NoError(t, err, "OpenWriters failed.")
```

suyash:
@akshayjshah ping

writer.go
```diff
@@ -76,3 +75,12 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
...
 }
+
+// OpenWriteSyncers combines the passed set of WriteSyncer objects into a locked
+// WriteSyncer. It returns any error encountered.
+func OpenWriteSyncers(writers ...zapcore.WriteSyncer) (zapcore.WriteSyncer, error) {
+	if len(writers) == 0 {
+		return zapcore.AddSync(ioutil.Discard), nil
+	}
+	return zapcore.Lock(zapcore.NewMultiWriteSyncer(writers...)), nil
+}
```
akshayjshah: Please remove this function; I'd much rather export just `Open`.

zapcore/write_syncer.go
```diff
@@ -88,6 +88,9 @@ type multiWriteSyncer []WriteSyncer
 // NewMultiWriteSyncer creates a WriteSyncer that duplicates its writes
 // and sync calls, much like io.MultiWriter.
 func NewMultiWriteSyncer(ws ...WriteSyncer) WriteSyncer {
+	if len(ws) == 1 {
+		return ws[0]
+	}
```
akshayjshah: Needs test coverage.

writer.go
suyash: @akshayjshah the whole point of this PR was to get a function in the API that takes interfaces instead of strings as output file paths.

Please see my first comment

zapcore/write_syncer.go
suyash: Added, Currently simply verifying that passing just one argument to NewMultiWriteSyncer simply returns the argument back.

writer.go
akshayjshah: Ack, my fault - I confused this with the previously-present `OpenWriters`.

Since the files/writers are already open, let's rename this to `CombineWriteSyncers` or `ComposeWriteSyncers` instead. This function can't fail, so let's also remove the error return value.

writer_test.go
```diff
@@ -81,3 +81,28 @@ func TestOpen(t *testing.T) {
...
+	return nil
+}
+
+func TestOpenWriteSyncers(t *testing.T) {
+	tw := &testWriter{[]byte("test"), t}
+
+	w, err := OpenWriteSyncers(tw)
+	if err != nil {
+		require.NoError(t, err, "OpenWriters Failed.")
+	}
```
akshayjshah: No need for the conditional.

writer_test.go
```diff
@@ -81,3 +81,28 @@ func TestOpen(t *testing.T) {
...
 	}
 }
+
+type testWriter struct {
+	expected []byte
+	t        testing.TB
+}
+
+func (w *testWriter) Write(actual []byte) (int, error) {
+	assert.Equal(w.t, w.expected, actual, "expected writer to write %v, wrote %v", string(w.expected), string(actual))
```
akshayjshah: The failure message automatically contains the expected and actual values, so we can simplify this to

```
assert.Equal(w.t, w.expected, string(actual), "Unexpected write call.")
```

writer_test.go
```diff
@@ -81,3 +81,28 @@ func TestOpen(t *testing.T) {
 		assert.Equal(t, tt.filenames, names, "Opened unexpected files given paths %v.", tt.paths)
 	}
 }
+
+type testWriter struct {
+	expected []byte
```
akshayjshah: If we make this a string, we'll get better output on test failures and easier initialization.

writer_test.go
```diff
@@ -81,3 +81,28 @@ func TestOpen(t *testing.T) {
...
+	t        testing.TB
+}
+
+func (w *testWriter) Write(actual []byte) (int, error) {
+	assert.Equal(w.t, w.expected, actual, "expected writer to write %v, wrote %v", string(w.expected), string(actual))
+	return 0, nil
```
akshayjshah: Let's fulfill the writer contract and return the correct number of bytes.

zapcore/write_syncer_test.go
```diff
@@ -65,6 +65,12 @@ func TestAddSyncWriter(t *testing.T) {
 	assert.NoError(t, ws.Sync(), "Unexpected error calling a no-op Sync method.")
 }
 
+func TestNewMultiWriteSyncerWorksForSingleWriter(t *testing.T) {
+	w := AddSync(&bytes.Buffer{})
```
akshayjshah: We can use `testutils.Buffer` here instead.

zapcore/write_syncer.go
akshayjshah: ✅  Excellent.

zapcore/write_syncer_test.go
suyash: Done

writer_test.go
suyash: Done

writer_test.go
suyash: returning the length of `actual`

writer_test.go
suyash: Done

writer_test.go
suyash: Done

PR #349, opened 2017-03-07
Do not special case stdout and stderr in Open

peter-edge:
You can close this if you don't want it, but I don't like how stdout and stderr are special cased in Open - you aren't actually opening them. I don't like my code here either, but this is a first pass.


akshayjshah:
I'd prefer to keep `Open` as-is, since we want users to be able to do what `Config` does without too much ceremony. (Eventually, I'd also like to expose more of `Config`'s internals.)

Agreed that we're not really opening stdout and stderr, but I'm okay with that.

PR #350, opened 2017-03-07
Update year from 2016 to 2017 for licenses

peter-edge:
I wanted to start using https://github.com/uber-go/update-license in this repository but this should be done first.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2016 Uber Technologies, Inc.
+Copyright (c) 2017 Uber Technologies, Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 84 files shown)


akshayjshah:
AFAIK, this is simply not the way copyright works. I *think* that the copyright date is intended to capture the date on which the creative work was fixed in tangible form, which was the date we created the file (or the project, depending on how you define the "work" in question). Any book that's been edited and many large, aka legal-approved, software projects (see [this file in the Chromium project](https://chromium.googlesource.com/chromium/src.git/+/31211cb8a35a8d9f6664a0f1f6d1957576e077e3/third_party/WebKit/Source/core/editing/serializers/Serialization.cpp) or [the FreeBSD license](https://www.freebsd.org/copyright/freebsd-license.html)) follow this convention.

I don't particularly care, but I'd like to do whatever legal wants. Let me check with them and we can figure out how to move forward here.

peter-edge:
Note that yarpc just did this a month or so ago heh.

Tbh I don't even think files need license headers,      but I'm not an
expert

akshayjshah:
Per email conversation, we'll need to change our approach here (and probably update `update-license` too).

PR #359, opened 2017-03-09
support concurrent-safe WriteSyncer

flisky:
Follows up #106

### DIFF
--- a/zapcore/write_syncer_test.go
+++ b/zapcore/write_syncer_test.go
@@ -56,6 +56,16 @@ func TestAddSyncWriteSyncer(t *testing.T) {
 	assert.Error(t, ws.Sync(), "Expected to propagate errors from concrete type's Sync method.")
 }
 
+func TestLockedWriteSyncerInterface(t *testing.T) {
+	buf := &bytes.Buffer{}
+	ws := AddSync(buf)
+	var lws WriteSyncer = &lockedWriteSyncer{ws: ws}
+
+	_, ok := lws.(LockedWriteSyncer)
+	require.True(t, ok, "Wrong implementation of the builtin lockedWriteSyncer")
+	require.Equal(t, lws, Lock(lws), "Unnessessary wrapped lock for locked WriteSyncer")
+}
+
 func TestAddSyncWriter(t *testing.T) {
 	// If we pass a plain io.Writer, make sure that we still get a WriteSyncer
 	// with a no-op Sync.
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -34,6 +34,12 @@ type WriteSyncer interface {
 	Sync() error
 }
 
+// LockedWriteSyncer is the interface that groups sync.Locker and WriteSyncer methods.
+type LockedWriteSyncer interface {
+	sync.Locker
+	WriteSyncer
+}
+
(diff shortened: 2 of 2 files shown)


flisky:
Travis errors with "Update failed for go.uber.org/atomic: Cannot detect VCS"

Trigger again.

flisky:
The CI should pass. Please take a review.

suyash:
Any chance you could provide a sample program demonstrating usage for this feature?

flisky:
Actually, I need the lock to do an application-level log rotate, but zap doesn't expose the underlying `lockedWriteSyncer`, and zap will lock the sink in anyway even if the `writesyncer` already does.

akshayjshah:
@flisky I'm sorry, I'm still confused. Can you explain your use case in more detail?

If you're trying to provide your own `WriteSyncer` that's already locked and handles log rotation, this change isn't necessary. Pass your implementation to `zap.New` - this library doesn't do anything by default *unless* you use the config struct.

flisky:
>Pass your implementation to zap.New - this library doesn't do anything by default unless you use the config struct.

I don't think so. It call the `writeSyncer` with an `zapcore.Lock`. In that implementation, it forcely adds an lock, just as #106 said - 

>In future, we may want to expose a way of setting a WriteSyncer that
already supports concurrent calls

This is my way trying to fix this problem. Could you take another check? Thanks!

akshayjshah:
Oh dear, you're right - I was so sure that we'd already removed that. Let's fix this properly and just remove the automatic lock in `NewCore`.

zapcore/write_syncer_test.go
```diff
@@ -56,6 +56,16 @@ func TestAddSyncWriteSyncer(t *testing.T) {
 	assert.Error(t, ws.Sync(), "Expected to propagate errors from concrete type's Sync method.")
 }
 
+func TestLockedWriteSyncerInterface(t *testing.T) {
+	buf := &bytes.Buffer{}
+	ws := AddSync(buf)
+	var lws WriteSyncer = &lockedWriteSyncer{ws: ws}
```
jcorbin: It'd be even better if the test uses a concrete type separately implemented from the internal `lockedWriteSyncer` since, as I understand it, the entire point of this PR is to support external implementations (and avoid double-locking them).

flisky: Yes, half part. It also makes sure the internal `lockedWriteSyncer` is okay.
A separately implementation would be almost same as our default one, so I don't think it necessary.
Would some rewords help? Any thoughts?

flisky:
Could you give me a final okay to get rid of `zapcore.Lock`, @akshayjshah & @jcorbin ? Because it's an incompatible change.

If so, would we keep the `Lock` implementation and just not call it ourselves, or remove it clearly? I'm glad to send another PR:)

akshayjshah:
We shouldn't get rid of `zapcore.Lock`, but we should get rid of the automatic use of it in `zapcore.NewCore`.

seifer:
Hi, what about this PR? It's a very useful. When will it be merged?

flisky:
#369 as you wish, @akshayjshah.

akshayjshah:
@seifer This PR won't be merged, but @flisky was kind enough to open a smaller PR that accomplishes the same thing.

Issue #358, opened 2017-03-08
Support Panic-level logging that doesn't panic


lhecker:
_Jump down to the **last paragraph** for the actual issue._

I'm currently facing the issue that I got a panic-recovery middleware in my server and thus would like errors catched by it to be logged at the `Panic` level. I can't do that right now though since `log.Panic()` also panics in turn which I don't want of course.

This problem could be solved on my end by copying the most essential parts of [`Logger.check()`](https://github.com/uber-go/zap/blob/4e517d38b2138e59475d6468e7a6b626eecdaa66/logger.go#L223-L272) like this:
```go
ce := log.Core().Check(zapcore.Entry{
    LoggerName: log.Name(),
    Time:       time.Now(),
    Level:      zapcore.PanicLevel,
    Message:    "panic recover",
}, nil)

if ce != nil {
    ce.Write(zap.Stack("stack"), zap.Error(err))
}
```

Unfortunately `Logger` does not have a `Name()` method which would make it easily possible to set the `LoggerName` above. Would it be possible to add such a method as well as for the other private `Logger` fields (like `errorOutput`) before v1.0.0 is released?

akshayjshah:
Sure, I'm open to this. We can even do this after 1.0, since loggers are concrete types - adding methods isn't a breaking change.

If you open a PR, I'm happy to review it.

akshayjshah:
This is currently possible, though ugly:

```
if ce := logger.Check(zap.PanicLevel, "message"); ce != nil {
  logger.Core().Write(ce.Entry, []zapcore.Field{...})
}
```

If this is truly common, we can revisit expanding the logger type's API surface.

peter-edge:
I don't think most logging libraries would log at the "panic" level without panicking - this gets into a discussion I had with @prashantv where the claim was that `log.Fatal` is "known" to call `os.Exit(...)`. I rather there be more levels of granularity than have unexpected behavior.

lhecker:
@peter-edge Well my original intention in this issue was to be able to generate `{"level":"panic", ...}` log lines at will and not to have a `logger.Panic()` which does not panic.
Basically I wanted something similar to @akshayjshah's code snippet [above](https://github.com/uber-go/zap/issues/358#issuecomment-285495283), while including [this](https://github.com/uber-go/zap/blob/4e517d38b2138e59475d6468e7a6b626eecdaa66/logger.go#L259-L269) piece of code which is  currently impossible to replicate outside of zap.

[...]

PR #360, opened 2017-03-09
Improve customizability of Logger

lhecker:
In this PR I plan to improve the customizability of the Logger interface etc. as per #358.
When I started adding getter methods for the private fields of Logger I noticed that they would expose a tad bit too much internal details. For instance exposing the `callerSkip` field just didn't seem right to me.

I thus added a `CheckWithAction` method instead solving my original problem differently. It allows "one to explicitly specify the zapcore.CheckWriteAction to take, making it possible to e.g. write a PanicLevel log line without panicking afterwards."

What do you guys think? Do you think I should solve it differently? Or maybe add some actual getters for e.g. the `name` field?
I'm of course open to critique and will try me best to change this PR accordingly. 😊

<hr>

- [ ] added tests to cover your changes;

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -38,9 +38,9 @@ import (
 type Logger struct {
 	core zapcore.Core
 
-	development bool
-	name        string
-	errorOutput zapcore.WriteSyncer
+	dpanicAction zapcore.CheckWriteAction
+	name         string
+	errorOutput  zapcore.WriteSyncer
 
 	addCaller bool
 	addStack  zapcore.LevelEnabler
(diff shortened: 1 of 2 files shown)


lhecker:
I thought the same as well, but unfortunately that's not possible if you aim for a generic setting...
You'd have to replicate [this](https://github.com/uber-go/zap/blob/4e517d38b2138e59475d6468e7a6b626eecdaa66/logger.go#L259-L269) as well, which would require the exposal of _all_ private fields including plain internal details like `callerSkip`.
So what's the solution here? Exposing ugly implementation details or exposing an additional ugly method? I knew that you'll surely dislike this PR with it's `CheckWithAction`, but I'm not sure if it's a wrong PR either, since my original suggestion/solution does not seem to be a viable one.

akshayjshah:
I'm certainly not going to tell you how to log in your application, but I don't want to expose a panic-but-don't-really-panic API from the logger. In all our internal uses, we use error-level logging for the situation you're describing - `logger.Panic` triggers a panic, but once you've recovered, it's just an error.

We don't need to agree about that, though - you should handle log levels in a way that makes sense for your application. By design, `*zap.Logger` and `*zap.SugaredLogger` aren't extension points, they're fairly opinionated loggers. I think they fit many use cases, but your situation might not be one of them.

If you need different behavior, I'd suggest writing your own small wrapper around `zapcore.Core`. The meat of the logger is only ~50 lines of code, and that'll give you full control over behavior.

lhecker:
@akshayjshah As you know calls to e.g. the stdlib can panic - they don't call `logger.Panic()` though which means you have to catch that exception and manually log it, right? So how do you handle logging them? Are you logging those errors at the "error" level or do you simply let your servers crash in such a case?
I'm currently catching all panics in a regular server middleware and was planning to log them at the appropriate "panic" log level, which is not possible with the zap API though.
(There is another problem in that panics thrown by `logger.Panic(...)` are indistinguishable of other calls to `panic(someString)` since zap does not use a custom error type, but that's a problem for another day.)

akshayjshah:
We log all those errors at the `Error` level, since we've already stopped the server from crashing. Once we're no longer crashing, it's not a panic anymore - it's just an error, from which we've already recovered.

lhecker:
@akshayjshah Alright... I'm doing the same right now. I disagree about the reasoning though since originally it _was_ a panic and I feel like just because it was catched as one it's still a far more serious error than... a simple error, since Go follows the "panics are truly exceptional errors"-model.

I don't mind if this PR gets closed (no hard feelings here 😂), but I'd urge you all to rethink wether adding a `zap.Logger` API, which would allow one to log at arbitrary levels without any side effects, really doesn't fit the current API design... 😕

akshayjshah:
Internally, our position is this: if the panic is truly serious, *don't recover*. The application is in an unknown state, and there's no guarantee that continued operation will behave as expected. Instead, crash and let the process manager or scheduler restart the application or container. Well-behaved applications log *no* errors - every error should be treated as a bug, with a patch expected to fix the issue or downgrade the logging.

If you're recovering (e.g., in HTTP middleware), then the panic is just an error - you're catching it and continuing operations, just like you would from any other error.

Again, this is our internal position; we run many Go services, but we're just one company.

jcorbin:
@lhecker the prior thinking, as far as I'm aware, was that the use case of "log at arbitrary levels without any side effects" is supported by reaching into the core, i.e. `(*Logger).Core().Write(Entry{...}, []Field)`.  Would that be a viable approach for your use case?

lhecker:
@jcorbin I think I might stick with what @akshayjshah said from now on and use the "error" level.
The reason the `Logger.Core()` method is not enough is because the `Logger.check()` method is doing some non-trivial things, like setting the name, adding caller and stack information. Those things currently cannot be replicated outside of zap due to the necessary fields being private. And even if those fields were exposed it would still be quite nasty since they expose some real internal implementation details, which would be better off exposed with a `CheckWithAction` method like in this PR.

akshayjshah:
In the short term, you can achieve the effect you're looking for with the current APIs (though it's not super-pretty):

```
if ce := logger.Check(zap.PanicLevel, "message"); ce != nil {
  logger.Core().Write(ce.Entry, []zapcore.Field{...})
}
```

I'll keep the issue you've filed open, and we'll keep thinking on this. It's easy enough to add more API surface area to the logger after 1.0.

PR #363, opened 2017-03-09
Sort benchmarks in readme by time

peter-edge:
This sorts the benchmarks in the readme by time. It also adds a `benchmarkRow` struct that is parsed from the benchmark output to add some semantic representation to a benchmark row, which will make it easier to do such things in the future and/or make this into a more general tool.


peter-edge:
Oh wow...ya I used a go 1.8 specific feature (sort.Slice), need to fix that.

PR #364, opened 2017-03-09
Add APIs to support buffering

### LINKED ISSUE #355: Support offloading I/O to a separate goroutine
The `WriteSyncer` abstraction already supports buffering or asynchronous implementations. However, the logger APIs don't expose a way to flush any buffered messages before shutting down. In a post-1.0 point release, we should add a `Close` method to loggers and an `OnClose` option (to register shutdown hooks). The opinionated `Config` struct can then register hooks to sync all underlying `WriteSyncer`s.

This will have to remain an opt-in setup, since there's no way to guarantee that `Close` is called on all panics and exits (including those not triggered by logger.Panic and logger.Fatal).

akshayjshah:
This is a breaking change that adds a `Sync` method to `zapcore.Core`. The
intention here is to expose `Sync` all the way up to the loggers themselves;
this lets implementations buffer I/O, since the end user can `Sync` at the end
of main.

(Of course, nothing can save users if they directly panic or os.Exit from a
different goroutine.)

This fixes #355.

### DIFF
--- a/internal/observer/observer.go
+++ b/internal/observer/observer.go
@@ -124,6 +124,10 @@ func (o *observer) Write(ent zapcore.Entry, fields []zapcore.Field) error {
 	return o.sink(LoggedEntry{ent, fields})
 }
 
+func (o *observer) Sync() error {
+	return nil
+}
+
 type contextObserver struct {
 	zapcore.LevelEnabler
 	sink    func(LoggedEntry) error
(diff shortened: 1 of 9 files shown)


jcorbin:
Given that we frequently are writing docs like "Sync flushes..." perhaps the better name here would've indeed been `Flush() error` ?

akshayjshah:
`Flush` is definitely better on the loggers, but `Sync` is better on the `WriteSyncer`. Keeping the names the same clearly indicates that they're related, which I weighed more heavily than the name itself.

Issue #354, opened 2017-03-07
Use runtime.Frames in stacktraces


akshayjshah:
As a consequence of Go 1.9's mid-stack inlining, PCs no longer have a 1:1 relationship with the frames we'll want to show in stacktraces. We'll need to switch to using `runtime.Frames` instead.

See https://github.com/golang/go/issues/19426 for some more color.

PR #365, opened 2017-03-09
Use runtime.CallersFrames

peter-edge:
Just doing https://github.com/uber-go/zap/issues/354

### DIFF
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -60,31 +60,29 @@ func takeStacktrace() string {
 	}
 
 	i := 0
-	for _, programCounter := range programCounters.pcs {
-		f := runtime.FuncForPC(programCounter)
-		name := f.Name()
-		if shouldIgnoreStacktraceName(name) {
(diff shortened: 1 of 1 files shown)


prashantv:
Can you post a summary of any differences in the benchmark results from switching to `CallerFrames`? Just curious what kind of performance penalty we're paying (if any).

akshayjshah:
The relevant benchmark here is `BenchmarkStackField`.

peter-edge:
```
# master
BenchmarkStackField-8   	 1000000	      1032 ns/op	     449 B/op	       2 allocs/op
# runtime-frames
BenchmarkStackField-8   	 1000000	      1008 ns/op	     449 B/op	       2 allocs/op
```

no change it seems like

akshayjshah:
The escape analyzer has our back here - even though CallerFrames returns a reference, it doesn't escape and so we're not penalized.

PR #366, opened 2017-03-12
Add ByteString API

### LINKED ISSUE #324: AddBinary(data) result differs from the result of AddString(string(data))
In Go `[]byte` and  `string` is freely converted between. But binary data is encoded as base64 string now. It is very surprising, and seem to be uncommon when `zap.Binary(key, data)` and `zap.String(key, string(zap))` produce different result. 

I think that it is caller responsibility to encode binary data, and it should can choose encoding itself. 
Also, this strange behaviour requires extra allocations for callers that don't want to convert their `[]byte` to base64.

skipor:
This is a breaking change that adds bytestring APIs (to log UTF-8 encoded bytes) to `ObjectEncoder` and `ArrayEncoder`. To actually save allocations along this path, I also benchmarked adding a `[]byte` to the field union; this reduces the allocation count for bytestrings, but increases the cost of adding fields by ~50%.

Fixes #324.

### DIFF
--- a/array.go
+++ b/array.go
@@ -43,6 +43,12 @@ func Bools(key string, bs []bool) zapcore.Field {
 	return Array(key, bools(bs))
 }
 
+// ByteStrings constructs a field that carries a slice of []byte
+// that assumed to be UTF-8 encoded.
+func ByteStrings(key string, bss [][]byte) zapcore.Field {
+	return Array(key, byteStringsArray(bss))
+}
+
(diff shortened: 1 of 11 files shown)


akshayjshah:
This is *great* - nicely done.

I'm going to push a small change (less use of named returns), and I'm going to benchmark whether it's worth adding a `[]byte` to the `Field` union struct.

PR #367, opened 2017-03-12
Remove unused globals

skipor:
The `_bufSize` and `maxUint64` globals were unused, so this PR removes them.


akshayjshah:
Great - thanks for the cleanup, @skipor!

PR #369, opened 2017-03-13
remove WriteSyncer's automatic lock

flisky:
The lock goes up to function call to make changes compatible.

And now, it's user's duty to make `WriteSyncer` concurrent-safe.

Fix up #359.


seifer:
Ok, the same question as in https://github.com/uber-go/zap/pull/359. When will this be merged? :)

akshayjshah:
Thanks for opening a second PR! This goes further than what I'd intended; I'm going to push a commit to back out some of these changes?

writer.go
```diff
@@ -81,5 +81,5 @@ func CombineWriteSyncers(writers ...zapcore.WriteSyncer) zapcore.WriteSyncer {
 	if len(writers) == 0 {
 		return zapcore.AddSync(ioutil.Discard)
 	}
-	return zapcore.Lock(zapcore.NewMultiWriteSyncer(writers...))
```
akshayjshah: I'd prefer to keep this auto-lock, since it lets users easily replicate the behavior of the `Config` struct. In general, I'd like the `zapcore` package to have no defaults and the `zap` package to supply sane defaults.

Anyone who doesn't want this auto-locking can just use `zapcore.NewMultiWriteSyncer`.

options.go
```diff
@@ -63,7 +63,7 @@ func Fields(fs ...zapcore.Field) Option {
 // safe for concurrent use.
```
akshayjshah: Need to update the comment here.

PR #372, opened 2017-03-13
Export zap's internal observing logger

billf:
Simplify `internal/observer` and export it as `zaptest/observer`.

### DIFF
--- a/zapcore/sampler_test.go
+++ b/zapcore/sampler_test.go
@@ -26,20 +26,19 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-
 	"go.uber.org/atomic"
-	"go.uber.org/zap/internal/observer"
 	. "go.uber.org/zap/zapcore"
 	"go.uber.org/zap/zaptest"
+	"go.uber.org/zap/zaptest/observer"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func fakeSampler(lvl LevelEnabler, tick time.Duration, first, thereafter int) (Core, *observer.ObservedLogs) {
-	var logs observer.ObservedLogs
-	core := observer.New(lvl, logs.Add, true)
+	core, logs := observer.New(lvl)
 	core = NewSampler(core, tick, first, thereafter)
-	return core, &logs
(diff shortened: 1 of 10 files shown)


zapcore/tee_test.go
```diff
@@ -137,10 +134,9 @@ func TestTeeEnabled(t *testing.T) {
 }
 
 func TestTeeSync(t *testing.T) {
-	tee := NewTee(
-		observer.New(InfoLevel, nil, false),
-		observer.New(WarnLevel, nil, false),
-	)
+	infoLogger, _ := observer.New(InfoLevel)
+	warnLogger, _ := observer.New(WarnLevel)
+	tee := NewTee(infoLogger, warnLogger)
```
prashantv: do we need real loggers here, can we pass `NopLoggers` or something?

zaptest/observer/observer.go
```diff
@@ -40,13 +40,11 @@ type ObservedLogs struct {
 	logs []LoggedEntry
 }
 
-// Add appends a new observed log to the collection. It always returns nil
-// error, but does so that it can be used as an observer sink function.
-func (o *ObservedLogs) Add(log LoggedEntry) error {
+// Add appends a new observed log to the collection.
+func (o *ObservedLogs) Add(log LoggedEntry) {
```
prashantv: does this need to be exported?

zapcore/sampler_test.go
```diff
@@ -26,20 +26,19 @@ import (
...
-	"github.com/stretchr/testify/require"
-
 	"go.uber.org/atomic"
-	"go.uber.org/zap/internal/observer"
 	. "go.uber.org/zap/zapcore"
 	"go.uber.org/zap/zaptest"
+	"go.uber.org/zap/zaptest/observer"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
akshayjshah: :+1: Thanks for the import reordering.

zapcore/tee_test.go
akshayjshah: The `NopCore` is always disabled, so it's not helpful here.

zaptest/observer/observer.go
akshayjshah: Nope, I'll push a small change to un-export it.

PR #376, opened 2017-03-14
Expose Buffer pools to third-party encoders

akshayjshah:
Zap currently uses an internal package to house its buffer pool, and it returns
all buffers to that pool. This is problematic for two reasons:

1. Third-party encoders can easily corrupt the pool by keeping a
   reference to a buffer.
2. Third-party encoders can't pool their buffers, since zap needs to
   know how to return them to their pools.

This PR introduces a `buffer.Pool` type to solve this problem. Individual
buffers can only be created via a pool, and each buffer keeps a reference to its
pool of origin. This allows zap to continue using a single internal pool, but it
allows third-party encoders to safely pool their buffers too.

### DIFF
--- a/buffer/pool.go
+++ b/buffer/pool.go
@@ -23,26 +23,27 @@ package buffer
 import "sync"
 
 // A Pool is a type-safe wrapper around a sync.Pool.
-type Pool sync.Pool
+type Pool struct {
+	p *sync.Pool
+}
 
 // NewPool constructs a new Pool.
-func NewPool() *Pool {
-	p := sync.Pool{
+func NewPool() Pool {
+	return Pool{p: &sync.Pool{
 		New: func() interface{} {
 			return &Buffer{bs: make([]byte, 0, _size)}
 		},
-	}
-	return (*Pool)(&p)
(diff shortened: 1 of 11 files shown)


buffer/pool.go
```diff
@@ -0,0 +1,48 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package buffer
+
+import "sync"
+
+// A Pool is a type-safe wrapper around a sync.Pool.
+type Pool sync.Pool
```
prashantv: i think this code will be cleaner if we make this a struct with a `sync.Pool` as a field. Retyping is useful for slices and maps since it lets callers create them more easily, but there's no advantage here, and we don't want users to get access to the underlying `sync.Pool`.

akshayjshah: Sure, can do.

PR #386, opened 2017-03-23
zaptest/observer: Support filtering by message or field

### LINKED ISSUE #383: zaptest: Extend ObservableLogs to support filtering
In tests, it's useful to validate that a specific log exists with a specific field (similar to tchannel's [log filtering](https://godoc.org/github.com/uber/tchannel-go/testutils#ChannelOpts.AddLogFilter)).

I'm thinking of adding methods on `ObservedLogs` to filter:
 - filter by message (exact or substring, not sure)
 - filter by exact field match (key and value)
 - filter by level

Imagining something like:
```go
logger, sink := observer.New(...)
[...]
logs := sink.FilterMessage("Warning.").FilterField("serviceName", "foo")
assert.Equal(t, 1, logs.Len(), "Missing expected log in %v", sink)
```

cc @billf

prashantv:
Add `Filter*` functions to filter a set of ObservedLogs and return
a new set of ObservedLogs. This allows chaining of multiple filters
such as `sink.FilterMessage(..).FilterField(..)` for easier use in
tests.

Fixes #383

### DIFF
--- a/zaptest/observer/observer.go
+++ b/zaptest/observer/observer.go
@@ -80,6 +80,39 @@ func (o *ObservedLogs) AllUntimed() []LoggedEntry {
 	return ret
 }
 
+// FilterMessage filters entries to those that have the specified message.
+func (o *ObservedLogs) FilterMessage(msg string) *ObservedLogs {
+	return o.filter(func(e LoggedEntry) bool {
+		return e.Message == msg
+	})
+}
+
+// FilterField filters entries to those that have the specified field.
+func (o *ObservedLogs) FilterField(field zapcore.Field) *ObservedLogs {
+	return o.filter(func(e LoggedEntry) bool {
+		for _, ctxField := range e.Context {
+			if ctxField.Key != field.Key {
+				continue
+			}
+			return ctxField == field
+		}
+		return false
+	})
+}
+
+func (o *ObservedLogs) filter(match func(LoggedEntry) bool) *ObservedLogs {
+	o.mu.RLock()
+	defer o.mu.RUnlock()
+
+	var filtered []LoggedEntry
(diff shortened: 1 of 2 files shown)


zaptest/observer/observer.go
```diff
@@ -80,6 +80,39 @@ func (o *ObservedLogs) AllUntimed() []LoggedEntry {
...
+}
+
+// FilterField filters entries to those that have the specified field.
+func (o *ObservedLogs) FilterField(field zapcore.Field) *ObservedLogs {
+	return o.filter(func(e LoggedEntry) bool {
+		for _, ctxField := range e.Context {
+			if ctxField.Key != field.Key {
+				continue
+			}
+			return ctxField == field
```
akshayjshah: Why not just use `ctxField == field`?

prashantv: `continue` vs `return` -- if we find a field with the right key, then the match is whether the field matches exactly. If the field mismatches on keys, we want to check other fields (since the filter only needs one field to match exactly).

akshayjshah: Sorry, I should be a bit clearer. It's entirely possible to have multiple fields with the same key. Consider

```
logger.With(zap.Int("foo", 1), zap.Namespace("component"), zap.Int("foo", 2))
```

and a filter on `zap.Int("foo", 2)`.

Given that, I don't think we want to explicitly compare keys; instead, I'd expect something more like

```
for _, f := range e.Context {
  if f == field {
    return true
  }
}
return false
```

prashantv: Sure, added a test and supports multiple keys. As soon as any of the fields match with the specified key, the log entry will be returned.

PR #394, opened 2017-03-28
Don't panic when invalid paths are configured

### LINKED ISSUE #390: panic: runtime error: SIGSEGV on non existing directory
Hi,

There is panic (trying to execute nil function) when there is the directory non exists.

```
package main

import (
	"log"

	"go.uber.org/zap"
	"go.uber.org/zap/zapcore"
)

func main() {
	cfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
	cfg.Level.SetLevel(zapcore.InfoLevel)
	cfg.DisableCaller = true
	cfg.OutputPaths = []string{"/tmp/non_existing_dir/file.log"}
	logger, err := cfg.Build() // panic()
	if err != nil {
		log.Fatal(err)
	}
	logger.Info("test message", zap.String("foo", "bar"))
}
```

After run: 

```panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x605933]

goroutine 1 [running]:
go.uber.org/zap.Config.openSinks(0xc420011218, 0x100, 0xc420011230, 0x699642, 0x4, 0x699426, 0x3, 0x6998f4, 0x5, 0x699318, ...)
	/usr/src/go/src/go.uber.org/zap/config.go:220 +0x193
go.uber.org/zap.Config.Build(0xc420011218, 0x100, 0xc420011230, 0x699642, 0x4, 0x699426, 0x3, 0x6998f4, 0x5, 0x699318, ...)
	/usr/src/go/src/go.uber.org/zap/config.go:161 +0xb1
main.main()
```

The problem is in `go.uber.org/zap/config.go#220 and 225-226` in openSinks method - there should be check for nil clouseOut/closeErr or zap.Open should return nopcloser function.

akshayjshah:
If the config struct includes invalid output paths, make sure that we don't try
to execute a nil function pointer.

Fixes #390.

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -93,7 +93,7 @@ func TestConfigWithInvalidPaths(t *testing.T) {
 	}{
 		{"output directory doesn't exist", "/tmp/not-there/foo.log", "stderr"},
 		{"error output directory doesn't exist", "stdout", "/tmp/not-there/foo-errors.log"},
-		{"both output directories don't exist", "/tmp/not-there/foo.log", "/tmp/not-there/foo-errors.log"},
+		{"neither output directory exists", "/tmp/not-there/foo.log", "/tmp/not-there/foo-errors.log"},
 	}
 
 	for _, tt := range tests {
--- a/config_test.go
+++ b/config_test.go
@@ -84,3 +84,25 @@ func TestConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestConfigWithInvalidPaths(t *testing.T) {
+	tests := []struct {
+		desc      string
+		output    string
+		errOutput string
+	}{
+		{"output directory doesn't exist", "/tmp/not-there/foo.log", "stderr"},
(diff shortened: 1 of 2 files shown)


akshayjshah:
Coveralls appears to be hosed somehow; the Coveralls builds themselves are passing, but they're not calling back to GitHub properly to green-light this PR.

If code reviewers think this looks okay, I'm going to force-land (as long as Travis builds continue to pass).

prashantv:
This change looks fine to fix the issue in #390, but it seemed a little odd to have this partial return on error.

On errors in `Open`, I think we should either:
 - Return a `WriteSyncer` with successfully opened files, and a non-nil close function.
 - Return `nil` for both `WriteSyncer` and `close`.

I don't think there's any advantages to returning a `close` without a `WriteSyncer`.

Take a look at #396 which goes the latter route (we still collect all errors during `Open`, but if there's an error, no `WriteSyncer` nor `close` is returned.

PR #396, opened 2017-03-29
Open should only return close on success

### LINKED ISSUE #390: panic: runtime error: SIGSEGV on non existing directory

prashantv:
Config assumes that Open will always return a close function
which causes a panic when Open returns an error, since it doesn't
return a close function.

We can instead clean up the assumption that we return partial values
on error, and instead use a more common pattern:
 - On success, `err == nil` and all other return values are valid
 - On error, `err != nil` and all other return values are zero values

Fixes #390.

### DIFF
--- a/writer_test.go
+++ b/writer_test.go
@@ -66,7 +66,10 @@ func TestOpen(t *testing.T) {
 
 	for _, tt := range tests {
 		wss, cleanup, err := open(tt.paths)
-		defer cleanup()
+		if err == nil {
+			defer cleanup()
+		}
+
 		if tt.error == "" {
 			assert.NoError(t, err, "Unexpected error opening paths %v.", tt.paths)
 		} else {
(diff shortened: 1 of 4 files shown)


writer_test.go
```diff
@@ -82,6 +85,26 @@ func TestOpen(t *testing.T) {
 	}
 }
 
+func TestOpenFails(t *testing.T) {
+	tests := []struct {
+		paths   []string
+		wantErr string
```
akshayjshah: Since we're not using this, can we remove it?

prashantv: 👍

PR #413, opened 2017-04-20
Add MarshalText to AtomicLevel and FilterMessageSnippet to observer

ZymoticB:
My team displays its yaml configs on HTML debug pages and thus requires that AtomicLevel support marshalling.

Additionally, we are currently transitioning from logrus style format logging to zap.SugaredLogger and use a testing strategy where we assert that a log message contains some list of snippets, rather than asserting on the entire formatted log message. This seems generally useful for users of the SugaredLogger.

### DIFF
--- a/level.go
+++ b/level.go
@@ -21,9 +21,13 @@
 package zap
 
 import (
+	"errors"
+
 	"go.uber.org/atomic"
 	"go.uber.org/zap/zapcore"
 )
+
+var errMarshalNilLevel = errors.New("can't marshal a nil *AtomicLevel to text")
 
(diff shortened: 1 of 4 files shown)


ZymoticB:
It was definitely awkward working with the pointer receiver there, thanks for the PR.

I'll break this up.

ZymoticB:
I'm closing this so that the two parts can be merged from sane branch names in my fork.

ZymoticB:
Observer is here GH-415
AtomicLevel will be rebase once GH-414 is finalized

PR #423, opened 2017-05-01
Add method for constructing a logger with custom syncer

GrantSeltzer:
This provides a super easy way to create a logger with a config and custom WriteSyncer's. This way developers using zap can create a WriteSyncer out of whatever implementation of io.Writer they have for their project and create a logger from there.

For example, in a lot of projects developers will create mock writers where values are stored in internal byte slices instead of going to stdout/stderr. This way developers can easily set up unit tests that test to make sure the proper values are being logged. 

Signed-off-by: grantseltzer <grant@capsule8.com>

### DIFF
--- a/config.go
+++ b/config.go
@@ -170,6 +170,23 @@ func (cfg Config) Build(opts ...Option) (*Logger, error) {
 	return log, nil
 }
 
+// BuildWithCustomSyncers constructs a logger from the config, and Options using a custom WriteSyncer
+func (cfg Config) BuildWithCustomSyncers(out zapcore.WriteSyncer, errOut zapcore.WriteSyncer, opts ...Option) (*Logger, error) {
+	enc, err := cfg.buildEncoder()
+	if err != nil {
+		return nil, err
+	}
+
+	log := New(
+		zapcore.NewCore(enc, out, cfg.Level),
+		cfg.buildOptions(errOut)...,
+	)
+	if len(opts) > 0 {
+		log = log.WithOptions(opts...)
+	}
+	return log, nil
+}
+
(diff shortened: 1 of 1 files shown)


akshayjshah:
Thanks for the PR, @GrantSeltzer! I'm going to ponder this one for a bit; I see what you're going for here, but I'm not sure it's worth the extra API.

GrantSeltzer:
@akshayjshah Sure. The code base is pretty vast so if this does something that already exists (that I clearly missed) please let me know. 

Also have you seen anyone accomplish something similar without writing an entire wrapper around zap specific to their project? That's what I had done originally but it felt silly. 

Feel free to give me feedback of stuff you'd like to change, i'd be happy to work with you on it.

akshayjshah:
Sorry for the delay getting back to you, Grant!

I'd prefer not to include this in the config-based logger configuration path. You can construct a logger using any `WriteSyncer` implementation like this:

```
ws := newWriteSyncer()
core := zapcore.NewCore(
  zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig()),
  ws,
  zapcore.DebugLevel,
)
```
Since there are so many different ways to compose the `Core`, `Encoder`, and `WriteSyncer` types, I'd prefer to only support the most common scenarios in configuration. Anything even moderately complex should be in code.

For testing in particular, `go.uber.org/zap/zaptest/observer` provides a useful testing spy.

GrantSeltzer:
@akshayjshah No worries!

That certainly makes sense. Thanks for clearing up how to do that as well! I'll keep an eye out for other ways to contribute.

PR #424, opened 2017-05-04
Make line endings configurable

### LINKED ISSUE #412: Ability to define line endings
So I ran into a problem when I need to delivery logs to a Splunk Cloud instance.
The configuration of this Splunk instance is out of my reach, and by default it uses windows style line endings "\r\n" to separate each item. 
Current version of encoders have the line endings hardcoded to \n. My proposal is to allow configuring the line endings adding a LineSeparator string property the zapcore.EncoderConfig and using that to append in the encoders to the end of the generated log line.

alaczi:
Add `LineEnding` to `EncoderConfig`, which allows Windows users to use non-Unix newlines.

Fixes #412.


akshayjshah:
@alaczi This is great - thanks for the contribution! The test coverage is great.

I'm going to push a commit that changes `LineSeparator` to `LineEnding`, which is the term I see used more widely. Other than that small change, this is ready to merge :tada:

alaczi:
Cheers!

PR #425, opened 2017-05-04
Keep time location in zap.Time

hnakamur:
The `zap.Time()` did convert the time value to integer using `time.UnixNano()`.
https://github.com/uber-go/zap/blob/7641ffe6315b3cb6274a020e6db8e87db4d46bc0/field.go#L178
This means we lost the location information.
So `EncodeEntry()` for `JSONEncoder` and `ISO8601TimeEncoder` returned the time string with the local timezone, not the original location specified in the time value passed to `zap.Time()`.

I add a test case to reproduce this problem at zapcore/json_encoder_entry_test.go.
Notice this test case fails only for environments whose timezone is not UTC.

This pull request fixes this problem with setting the time value to the `Interface` field of `Field` instead of `Integer` field.

To maintain backward compatibility, I modified the `TimeType` case in `func (f Field) AddTo(enc ObjectEncoder)` to use `Interface` field when it is set (not nil) and use `Integer` field otherwise.

Could you review this pull request?
Thanks!


zapcore/json_encoder_entry_test.go
```diff
@@ -0,0 +1,79 @@
...
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
```
akshayjshah: Good catch... :(

Thanks for adding the license.

field.go
```diff
@@ -175,7 +175,8 @@ func Stringer(key string, val fmt.Stringer) zapcore.Field {
 // Time constructs a zapcore.Field with the given key and value. The encoder
 // controls how the time is serialized.
 func Time(key string, val time.Time) zapcore.Field {
-	return zapcore.Field{Key: key, Type: zapcore.TimeType, Integer: val.UnixNano()}
+	// Use Interface instead of Integer here to keep location in time.
+	return zapcore.Field{Key: key, Type: zapcore.TimeType, Interface: val}
```
akshayjshah: Unfortunately, this makes every call to `Time` allocate.

hnakamur:
Thanks for your comments.
I need to preserve the original time zone.
I think times in UTC should be printed with timezone 'Z' and times in JST+9 should be printed with timezone '+0900'.
The current zap prints timezone '+0900' for times in UTC too.

field.go
hnakamur: I made another try: https://github.com/uber-go/zap/pull/425/commits/384c2f1fd8cda0b7df2a8a1762831400b74f5377
This time, we don't have extra allocate.

hnakamur:
Thanks for merging!

PR #429, opened 2017-05-05
Add changelog entry for v1.4.0

akshayjshah:
Add a changelog entry for the 1.4.0 minor release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## v1.4.0 (05 May 2017)
+
+This release adds the ability to customize the line endings used in zap's log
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## v1.4.0 (05 May 2017)
+
+This release adds the ability to customize the line endings used in zap's log
+output. It is fully backward-compatible.
```
breerly: nit - Probably dont need to mention it's backwards-compatible, that's implied by the fact that 'dis is a minor.

akshayjshah: I like to harp on that, since so many projects don't follow semver properly :)

PR #435, opened 2017-05-18
Accept capital strings in Level.UnmarshalText

jcorbin:
Makes it easier to parse zap's own log output (especially since capital level strings is our internal standard).

### DIFF
--- a/zapcore/level.go
+++ b/zapcore/level.go
@@ -117,19 +117,19 @@ func (l *Level) UnmarshalText(text []byte) error {
 		return errUnmarshalNilLevel
 	}
 	switch string(text) {
-	case "debug":
+	case "debug", "DEBUG":
 		*l = DebugLevel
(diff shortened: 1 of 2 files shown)


jcorbin:
- I'd prefer to avoid introducing an alloc as I'm using this on the hot path (external down sampler) and not just as part of startup config parse
- while understand the impulse to "Parse Them All! ™", the all-caps version actually is more privileged since it's one of the configured formats that zap itself will produce
- my real intention with this pr was: "if we can write it (and we do by default internally), we should parse it"

akshayjshah:
@jcorbin What do you think of pulling out the switch into a separate function and making two attempts at parsing - once with the original string, and once with the lowercased version if the first pass fails? That keeps the success case fast, but successfully parses a wider array of reasonably well-formed inputs.

PR #436, opened 2017-05-19
Add zerolog benchmark, fix go-kit bench and update README

rs:


peter-edge:
Tests are broken because of a missing license header in `zerolog_test.go`.

Just putting it out there - why not improve zap to fix any performance gaps instead of having more golang logger fragmentation? I have my own logging lib too github.com/peter-edge/lion-go but I'm moving to put my work towards this. There's probably some issue but just wondering.

rs:
Thanks for the heads up, I will fix the header.

About making zap faster, I’m afraid it would not be possible without changing the API in a non backward compatible way. Zerolog can be so fast because of its API is not using variadic arguments and because its output format is only JSON so it can incrementally build the JSON even in a single buffer with no intermediary form.

Zap is more versatile but it comes at a cost. When you need this flexibility it is a great choice. If all you need is fast JSON logging, you might want to consider zerolog.

peter-edge:
@akshayjshah

peter-edge:
I'd say this would be better to merge if there was such a pros/cons list advertised on zerolog, as opposed to the sensationalist headline on https://www.reddit.com/r/golang/comments/6c9k7n/zerolog_is_now_faster_than_zap, but just my two (friendly) cents :) haha

rs:
I plead guilty on the clickbait title :)

rs:
Thank you for you feedback @akshayjshah. Can you elaborate on "zerolog actually allocates more frequently"?

akshayjshah:
As measured by these same benchmarks, zerolog allocates at least as often as zap, and sometimes more often.

rs:
The benchmark shows that because one of the field types in the "fake fields" is allocating more on zerolog than on zap. For usual field types, there is no allocation at all in either case.

akshayjshah:
Yep, I'm aware. I'm just pointing out that the performance difference observed isn't because zerolog is allocating fewer objects on the heap.

peter-edge:
@rs https://github.com/uber-go/zap/pull/483

Issue #449, opened 2017-06-12
Lack of api to log to non-file target


xmxiaoq:
**#450** 

What if our log targets are **NOT** named files, say io.Writer?
Unfortunately, there's no exported function to do this easily.
#
In old version, it's easy:
```go
var out io.Writer
logger := zap.New(
		zap.NewJSONEncoder(),
		zap.DebugLevel,
		zap.Output(zap.Tee(zap.AddSync(os.Stdout), zap.AddSync(out))),
	)
```

akshayjshah:
Sorry for the long delay, @xmxiaoq - this is already possible with the exported APIs.

```
var out zapcore.WriteSyncer
// use zapcore.NewJSONEncoder, or implement your own encoder
var encoder zapcore.Encoder
core := zapcore.NewCore(
  encoder,
  zapcore.NewMultiWriteSyncer(out, os.Stdout),
  zapcore.DebugLevel,
)
logger := zap.New(core)
```

I'm very, very late writing good documentation for zap, so these APIs aren't super-obvious. Think of `zap.Config` as just one possible way to drive the `zapcore` APIs from configuration - there are many others that might be better for your use case. It's only a little code, so I expect that many users (especially in companies that need a standardized logging config format) to write their own configuration wrappers.

PR #450, opened 2017-06-12
Add function to customize output targets

xmxiaoq:
What if our log targets are NOT named files, say io.Writer?
Currently, there's no exported function to do this easily.
Add a SinkFunction, hook into the log build pipeline, it's backward compatibility:
```go
// SinkFunc customize output targets, for example log to io.Writer, not just files named by OutputPaths.
type SinkFunc func(Config) (zapcore.WriteSyncer, zapcore.WriteSyncer, error)
type Config struct {

  // ...

  SinkFunction SinkFunc `json:"-" yaml:"-"`
}

func (cfg Config) openSinks() (zapcore.WriteSyncer, zapcore.WriteSyncer, error) {
	// customize output?
	if cfg.SinkFunction != nil {
		return cfg.SinkFunction(cfg)
	}

	// No customize, as usual  ...
}
```

Usage:
```go
var output bytes.Buffer
var errOutput bytes.Buffer
cfg := zap.NewDevelopmentConfig()
// Customize output target
cfg.SinkFunction = func(cfg zap.Config) (zapcore.WriteSyncer, zapcore.WriteSyncer, error) {
	return zap.CombineWriteSyncers(os.Stdout, zapcore.AddSync(&output)), zap.CombineWriteSyncers(os.Stderr, zapcore.AddSync(&errOutput)), nil
}
logger, _ := cfg.Build()
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -37,6 +37,9 @@ type SamplingConfig struct {
 	Thereafter int `json:"thereafter" yaml:"thereafter"`
 }
 
+// SinkFunc customize output targets, for example log to io.Writer, not just files named by OutputPaths.
+type SinkFunc func(Config) (zapcore.WriteSyncer, zapcore.WriteSyncer, error)
+
 // Config offers a declarative way to construct a logger.
 //
 // It doesn't do anything that can't be done with New, Options, and the various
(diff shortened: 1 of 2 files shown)


akshayjshah:
Addressed in #449 - this already possible with exported APIs, it's just not possible through the `Config` struct. The built-in `Config` struct isn't intended to address all use cases - it's only a little code, and I expect that users who don't log to files (or standard out/err) will want to write their own configuration wrappers.

PR #455, opened 2017-06-27
Use go.uber.org/multierr

abhinav:
This drops zap's custom multierror implementation in favor of
go.uber.org/multierr.

Performance should be the same because we're relying on `Append`, which
short circuits in the no-error case.

---

Relies on #456


abhinav:
Ah, derp. `glide up` broke tests.

PR #460, opened 2017-06-30
errors: Support multierr and pkg/errors' causer

abhinav:
This changes how we log errors to break down multierr and pkg/errors
errors if possible.

If an error is a multierr error, its individual items will be listed
under `${key}Causes` as objects. Roughly,

    "error": "foo; bar; baz",
    "errorVerbose":
        `the following errors occurred:
          -  foo
          -  bar
          -  baz`,
    "errorCauses": [
        {"error": "foo"},
        {"error": "bar"},
        {"error": "baz"},
    ]

Similarly, if an error is a pkg/errors causer, the cause will be listed
under `${key}Causes` by itself.

    "error": "foo: bar",
    "errorVerbose": "foo: bar\n[stack trace]"
    "errorCauses": [
        {
            "error": "bar",
            "errorVerbose": "bar\n[stack trace]",
        },
    ]

See: uber-go/multierr#6

### DIFF
--- a/zapcore/error.go
+++ b/zapcore/error.go
@@ -1,0 +1,124 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zapcore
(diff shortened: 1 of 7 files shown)


zapcore/error.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (e *errArrayElem) MarshalLogArray(arr ArrayEncoder) error {
+	return arr.AppendObject(e)
+}
+
+func (e *errArrayElem) MarshalLogObject(enc ObjectEncoder) error {
+	return encodeError("error", e.err, enc)
+}
+
+func (e *errArrayElem) Release() {
```
akshayjshah: Elsewhere in zap, I believe that we're calling these methods `Free`. Would be nice to be consistent here.

akshayjshah:
@prashantv That's a good point, and one that I didn't stop to consider fully before we landed this. Appreciate the vigilance.

I mulled this over a while last night, and I'm increasingly dissatisfied with this whole approach to fancy errors - it feels odd and un-idiomatic to me. The `error` interface is just `Error() string`, and I'd really like to respect that and log *all* errors as a single string. Would we be globally happier if we took a simpler approach?

1. For most errors, log `err.Error()`.
2. For errors that implement `fmt.Formatter`, log `fmt.Sprintf("+v", err)`. Users of `pkg/errors`, for example, usually want this verbose output, and they can always get terse output by explicitly logging `err.Error()`.

If we don't like `multierr`'s multi-line logging in logs, perhaps that's a sign that it's just a bad idea. In all non-test contexts, I'd much prefer the output of my program to be line-oriented.

prashantv:
I'm not convinced that we should do something special for how `pkg/errors` works, but not for `multierr`. The main issue is that the interfaces we're dealing with don't really specify how they're used. E.g., even if a type implements `Formatter`, there's no guarantee it'll have unique output.

The simplest option is to have  `zap.Error` output `error: err.Error()`. It's unsurprising and doesn't automatically add any extra fields, nor does it bloat the output if a `pkg/errors` error is used. The issue with forcing users to call `err.Error()` is:

 * The callsite changes from a simple `zap.Error(err)` to `zap.String("error", err.Error())`
 * If the `err` was somehow `nil`, it'll panic

Perhaps we should have a separate `ErrorVerbose` that does extra checks, but in that I'd want to detect `multierr` explicitly, and only log `error` and `errorCauses` for it.

ansel1:
or have a configurable error encoder, as is done with a very other types? I
gotta say, we've been using zap.Error() with our own error class which
implements Formatter, and we built our own multiline-friendly encoder, and
it's great for development to see readable stack traces for errors.

akshayjshah:
Let's discuss on uber-go/multierr#23 - I made the mistake of kicking this discussion off in two places, and it's a bit confusing.

PR #465, opened 2017-07-03
Added support for providing custom logger name encoders

pavius:
Users can now provide a `LoggerNameEncoder` to override encoding of the logger name, which currently defaults to simply outputting the string.

Two `LoggerNameEncoder`s are provided - `FullLoggerNameEncoder`, the default, which dumps the string as before to maintain backwards compatibility and `CapitalLoggerNameEncoder` which is there mostly to give the tests another option to use. 

Quite a bit of the PR is slight formatting to accommodate the longish name `LoggerNameEncoder`. I considered calling it `NameEncoder` seeing how `LoggerName`'s key name is referred to as `Name`, but opted to stick to the member name.


zapcore/console_encoder.go
```diff
@@ -79,8 +79,8 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer,
 	if c.LevelKey != "" && c.EncodeLevel != nil {
 		c.EncodeLevel(ent.Level, arr)
 	}
-	if ent.LoggerName != "" && c.NameKey != "" {
-		arr.AppendString(ent.LoggerName)
+	if ent.LoggerName != "" && c.NameKey != "" && c.EncodeLoggerName != nil {
+		c.EncodeLoggerName(ent.LoggerName, arr)
```
akshayjshah: Unfortunately, this isn't backward-compatible. If `c.EncodeLoggerName` is nil, we should use `FullLoggerNameEncoder`.

zapcore/encoder.go
```diff
@@ -196,6 +197,31 @@ func (e *CallerEncoder) UnmarshalText(text []byte) error {
...
+
+// FullLoggerNameEncoder serializes a logger name as is
+func FullLoggerNameEncoder(loggerName string, enc PrimitiveArrayEncoder) {
+	enc.AppendString(loggerName)
+}
+
+// CapitalLoggerNameEncoder serializes a logger name in all caps
+func CapitalLoggerNameEncoder(loggerName string, enc PrimitiveArrayEncoder) {
+	enc.AppendString(strings.ToUpper(loggerName))
+}
```
akshayjshah: Rather than exporting this, let's put it into the test code. That'll give unit tests a way to exercise this functionality without adding this unusual option to the public API.

zapcore/encoder_test.go
```diff
@@ -360,70 +393,93 @@ func TestEncoderConfiguration(t *testing.T) {
...
+				TimeKey:          "T",
+				MessageKey:       "M",
+				NameKey:          "N",
+				CallerKey:        "C",
+				StacktraceKey:    "S",
+				EncodeTime:       base.EncodeTime,
+				EncodeDuration:   base.EncodeDuration,
+				EncodeLevel:      base.EncodeLevel,
+				EncodeCaller:     base.EncodeCaller,
+				EncodeLoggerName: base.EncodeLoggerName,
```
akshayjshah: We should only need to add two tests here: one that uses the capitalized encoder, and one that uses a no-op encoder. The other tests should pass as-is, and keeping them that way ensures that we don't break backward-compatibility later.

zapcore/json_encoder.go
```diff
@@ -302,7 +302,13 @@ func (enc *jsonEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer,
 	}
 	if ent.LoggerName != "" && final.NameKey != "" {
 		final.addKey(final.NameKey)
-		final.AppendString(ent.LoggerName)
+		cur := final.buf.Len()
+		final.EncodeLoggerName(ent.LoggerName, final)
```
akshayjshah: Same here - if the user-supplied logger name encoder is nil, we should fall back to the current behavior.

zapcore/encoder.go
```diff
@@ -211,10 +237,11 @@ type EncoderConfig struct {
...
 	// seconds since epoch, while others may prefer ISO8601 strings.
-	EncodeLevel    LevelEncoder    `json:"levelEncoder" yaml:"levelEncoder"`
-	EncodeTime     TimeEncoder     `json:"timeEncoder" yaml:"timeEncoder"`
-	EncodeDuration DurationEncoder `json:"durationEncoder" yaml:"durationEncoder"`
-	EncodeCaller   CallerEncoder   `json:"callerEncoder" yaml:"callerEncoder"`
+	EncodeLevel      LevelEncoder      `json:"levelEncoder" yaml:"levelEncoder"`
+	EncodeTime       TimeEncoder       `json:"timeEncoder" yaml:"timeEncoder"`
+	EncodeDuration   DurationEncoder   `json:"durationEncoder" yaml:"durationEncoder"`
+	EncodeCaller     CallerEncoder     `json:"callerEncoder" yaml:"callerEncoder"`
+	EncodeLoggerName LoggerNameEncoder `json:"nameEncoder" yaml:"nameEncoder"`
```
akshayjshah: Let's document that `EncoderLoggerName` is optional, and that the fallback behavior is to include the name as-is.

pavius:
Thanks for the comments @akshayjshah. Was too busy following suite to think about backwards compatibility. Addressed your comments and renamed to `EncodeName` (and consequently `FullLoggerNameEncoder` became `FullNameEncoder`).

Issue #139, opened 2016-09-17
Clean up documentation


akshayjshah:
Before releasing v1.0.0, we should clean up the documentation (both the README and the GoDoc). The current README is a bit text-heavy and doesn't show off enough code.

magicalbanana:
What about doing the documentation using `mkdocs` or `readthedocs`?

sectioneight:
In addition, should we pull in some of our fabulous tech writers? e.g. @beklapko

akshayjshah:
Yeah, let's definitely recruit a ringer :)

@magicalbanana, I'd prefer to stick to a README and GoDoc until they're proven insufficient. The README isn't optional, since the project's on GitHub, and Go packages should always have complete GoDoc. If those aren't enough, we can explore a more general-purpose option like Github Pages, ReadTheDocs, or mkdocs.

moitias:
Even a simple Contributing.MD would be nice, as the first indication of testing/linting practices I came up with a recent PR were displayed when making the PR (checklist that was automatically added to the PR). 

That checklist also seems to link to a different CLA (on Google Docs) than the one CLAssistant requires in the PR, which is confusing (even if they do work together?)

akshayjshah:
Both good points. I'll update the PR template to mention the bot (instead of providing a direct link), and I'll make sure to add a contributing guide with the rest of the docs.

akshayjshah:
Done! The PR template predated the CLA bot - thanks for reminding me that they pointed to different docs.

akshayjshah:
The lion's share of this is done (finally!).

PR #466, opened 2017-07-04
Add documentation and examples for zap

akshayjshah:
Add a bunch of testable examples and documentation for the main `zap`
package. As we get closer to launching fx and pushing zap adoption
internally, documentation will get more and more important.

We'll still need to improve the documentation of the other packages, but
this addresses the lion's share of #139.


doc.go
```diff
@@ -18,5 +18,92 @@
...
+//    "attempt", 3,
+//    "backoff", time.Second,
+//  )
+//  sugar.Printf("failed to fetch URL: %s", "http://example.com")
+//
+// In the rare contexts where every microsecond and every allocation matter,
+// use the Logger. It's even faster than the SugaredLogger and allocates far
+// less, but it only supports strongly-typed, structured logging.
+//  logger := zap.NewExample()
+//  defer logger.Sync()
```
willhug: Can you add an explanation of the logger.Sync somewhere in here?  It's used a lot but I'm not sure when it should be called (i.e. after I grab a logger, after I log a line, etc).

example_test.go
```diff
@@ -0,0 +1,324 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
willhug: should this be 2017?

doc.go
```diff
@@ -18,5 +18,92 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-// Package zap provides fast, structured, leveled logging in Go.
+// Package zap provides fast, structured, leveled logging.
```
willhug: do we have a good link that can explain leveled logging?  Wouldn't be surprised if new engineers are not as familiar with the nuances.

example_test.go
```diff
@@ -0,0 +1,324 @@
...
+	// zapcore.Cores, then tee the four cores together.
+	core := zapcore.NewTee(
+		zapcore.NewCore(kafkaEncoder, topicErrors, highPriority),
+		zapcore.NewCore(consoleEncoder, consoleErrors, highPriority),
+		zapcore.NewCore(kafkaEncoder, topicDebugging, lowPriority),
+		zapcore.NewCore(consoleEncoder, consoleDebugging, lowPriority),
+	)
+
+	// From a zapcore.Core, it's easy to construct a Logger.
+	logger := zap.New(core)
```
willhug: should I always call some form of `zap.New` before I make a log, or should I store a logger on a struct and reuse it.  And if we're in the reuse case, when should I call `logger.Sync`?

example_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+	logger.Info("info logging enabled")
+
+	atom.SetLevel(zap.ErrorLevel)
+	logger.Info("info logging disabled")
+	// Output:
+	// {"level":"info","msg":"info logging enabled"}
+}
+
+func ExampleAtomicLevel_config() {
```
willhug: should `config` be capitalized?

logger.go
```diff
@@ -203,19 +229,21 @@ func (log *Logger) Panic(msg string, fields ...zapcore.Field) {
...
+// disabled.
 func (log *Logger) Fatal(msg string, fields ...zapcore.Field) {
 	if ce := log.check(FatalLevel, msg); ce != nil {
 		ce.Write(fields...)
 	}
 }
 
-// Sync flushes any buffered log entries.
+// Sync calls the underlying Core's Sync method, flushing any buffered log
+// entries. Applications should take care to call Sync before exiting.
```
willhug: ahh, so an `OnStop` functionality

options.go
```diff
@@ -22,7 +22,7 @@ package zap
 
 import "go.uber.org/zap/zapcore"
 
-// Option is used to set options for the logger.
+// An Option configures a Logger.
```
willhug: I though gofmt comments needed to start with the name of the interface.

example_test.go
akshayjshah: Nope.

logger.go
akshayjshah: Yep, will clarify in the top-level examples.

options.go
akshayjshah: Nope, preceding articles are allowed.

example_test.go
akshayjshah: No. It should actually be `2016-2017`, but I'm punting that whole thing until I can write a tool to handle it properly. 

I'll forward you the whole email chain with legal.

doc.go
akshayjshah: Will do.

doc.go
akshayjshah: Errrr...wow, I hadn't thought of that. I don't know of one.

If this comes up in an issue or something, I'll dig up a link and add an FAQ entry.

PR #467, opened 2017-07-04
Clean up README

akshayjshah:
Fix some small formatting and hyphenation errors in the readme, and
correct the places where we're using "library" instead of "package."

### DIFF
--- a/.readme.tmpl
+++ b/.readme.tmpl
@@ -11,15 +11,15 @@ Note that zap only supports the two most recent minor versions of Go.
 ## Quick Start
 
 In contexts where performance is nice, but not critical, use the
-`SugaredLogger`. It's 4-10x faster than than other structured logging libraries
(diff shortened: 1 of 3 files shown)


.readme.tmpl
```diff
@@ -11,30 +11,31 @@ Note that zap only supports the two most recent minor versions of Go.
...
+`SugaredLogger`. It's 4-10x faster than than other structured logging
+packages and includes both structured and `printf`-style APIs.
 
 ```go
 defer logger.Sync() // flushes buffer, if any
 sugar := logger.Sugar()
-sugar.Infow("Failed to fetch URL.",
-  // Structured context as loosely-typed key-value pairs.
+sugar.Infow("failed to fetch URL",
```
akshayjshah: @peter-edge After the team vote, I concede - you were right, let's just lower-case log messages.

Sic transit gloria mundi.

PR #469, opened 2017-07-04
Add an FAQ

akshayjshah:
Add an `FAQ.md` to the main repository. This keeps the GoDoc focused and
is lighter-weight than a separate HTML documentation site (GH pages,
ReadTheDocs, or similar).

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,0 +1,135 @@
+# Frequently Asked Questions
+
+## Design
+
+### Why spend so much effort on logger performance?
+
+Of couse, most applications won't notice the impact of a slow logger: they
+already take tens or hundreds of milliseconds for each operation, so an extra
+millisecond doesn't matter.
+
+On the other hand, why *not* make structured logging fast? The `SugaredLogger`
+isn't any harder to use than other logging packages, and the `Logger` makes
+structured logging possible in performance-sensitive contexts. Across a fleet
+of Go microservices, making each application even slightly more efficient adds
+up quickly.
+
+### Why aren't `Logger` and `SugaredLogger` interfaces?
+
+Unlike most widely used interfaces, both `Logger` and `SugaredLogger`
+interfaces would include *many* methods. As [Rob Pike points
+out][go-proverbs], "The bigger the interface, the weaker the abstraction."
+Interfaces are also rigid &mdash; *any* change requires releasing a new major
+version, since it breaks all third-party implementations.
+
+Making the `Logger` and `SugaredLogger` concrete types doesn't sacrifice much
+abstraction, and it lets us add methods without introducing breaking changes.
+Your applications can and should define and depend upon an interface that
+includes just the methods you use.
+
+### Why sample application logs?
+
+Applications often experience runs of errors, either because of a bug or
+because of a misbehaving user. Logging errors is usually a good idea, but it
+can sometimes make a bad situation worse: not only is your application coping
+with an increased number of errors, it's also spending extra CPU cycles and
+I/O logging those errors. In the worst case, each call to your logger is
+contending over the same mutex, limiting throughput when you need it most.
+
+Sampling fixes this problem by dropping repetitive log entries. Under normal
+conditions, your application writes out every entry. When the similar entries
+are logged hundreds or thousands of times each second, though, zap begins
+dropping duplicates to preserve throughput.
+
+### Why do the structured logging APIs take a message in addition to fields?
+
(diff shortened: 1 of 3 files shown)


FAQ.md
```diff
@@ -0,0 +1,135 @@
...
+Unlike most widely used interfaces, both `Logger` and `SugaredLogger`
+interfaces would include *many* methods. As [Rob Pike points
+out][go-proverbs], "The bigger the interface, the weaker the abstraction."
+Interfaces are also rigid &mdash; *any* change requires releasing a new major
+version, since it breaks all third-party implementations.
+
+Making the `Logger` and `SugaredLogger` concrete types doesn't sacrifice much
+abstraction, and it lets us add methods without introducing breaking changes.
+Your applications can and should define and depend upon an interface that
+includes just the methods you use.
```
prashantv: Is it worth mention that the underlying `Core` is an interface, so the logger can still be stubbed out?

FAQ.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Why sample application logs?
+
+Applications often experience runs of errors, either because of a bug or
+because of a misbehaving user. Logging errors is usually a good idea, but it
+can sometimes make a bad situation worse: not only is your application coping
+with an increased number of errors, it's also spending extra CPU cycles and
+I/O logging those errors. In the worst case, each call to your logger is
+contending over the same mutex, limiting throughput when you need it most.
```
prashantv: Rather than calling out a mutex, I think it's better to say something like "since writes are typically serialized, [...]". Since the cost of the mutex is actually pretty negligible, it's that we are blocked till the writes are done which ends up serializing all log calls

FAQ.md
```diff
@@ -0,0 +1,135 @@
...
+
+Sampling fixes this problem by dropping repetitive log entries. Under normal
+conditions, your application writes out every entry. When the similar entries
+are logged hundreds or thousands of times each second, though, zap begins
+dropping duplicates to preserve throughput.
+
+### Why do the structured logging APIs take a message in addition to fields?
+
+Zap's sampling algorithm uses the message to identify duplicate entries. In
```
prashantv: While messages are useful for sampling, I think having a primary "field" that gives a brief description is pretty useful. I don't think the reason we have message is just for sampling, can we come up with a higher level reason for it?

FAQ.md
```diff
@@ -0,0 +1,135 @@
...
+logs will never be lost, but it eliminates a common error.
+
+See the discussion in uber-go/zap#207 for more details.
+
+### What's `DPanic`?
+
+`DPanic` stands for "panic in development." In development, it logs at
+`PanicLevel`; otherwise, it logs at `ErrorLevel`. It's designed to make it
+easier to catch errors that are theoretically possible, but shouldn't ever
+happen in practice.
```
prashantv: Can we add a little blurb about how there are errors/situations that should never happen, but if they do, we want to notice it straight away in development, but in production, we don't want to crash the whole app.

FAQ.md
```diff
@@ -0,0 +1,135 @@
...
+
+If you've ever written code like this, you need `DPanic`:
+
+```go
+if err != nil {
+  logger.Print("shouldn't ever get here: %v", err)
```
prashantv: i think this example might be better with a `panic`?

akshayjshah: Sure, sounds good.

FAQ.md
akshayjshah: I'd prefer to leave that out for now. Most of the issues, emails, and chat conversations I've had about this are basically about the "accept interfaces, return structs" rule.

I still need to add docs for `zapcore`; once I do that, I'll add an FAQ entry about extending zap.

FAQ.md
akshayjshah: I wasn't trying to suggest that the locking and unlocking is expensive, but that contention over the mutex serializes operations - your suggestion makes that much clearer.

FAQ.md
akshayjshah: Sure, I'll point that out too. Had some discussion about this with Chris Hines and Peter Bourgon on a Google doc a while ago, which I'll try to find again.

FAQ.md
akshayjshah: Sure, sounds good - your edit below makes this a bit clearer.

akshayjshah:
Thanks for the suggestions! I think I've addressed all the comments.

PR #470, opened 2017-07-06
Less verbose output for multierr

akshayjshah:
Following the discussions on uber-go/zap#460, uber-go/multierr#6, and
(most recently) uber-go/multierr#23, reduce log verbosity for
`multierr`. This is fully backward-compatible with the last released
version of zap.

The small changes introduced here do two things:

1. First, we either report `errorCauses` or `errorVerbose`, but not
   both.
2. Second, we prefer `errorCauses` to `errorVerbose`.

I think that this addresses our top-level wants without breaking any
interfaces or removing behavior we've already shipped.

In a future minor, we can add an `ErrorEncoder` interface that the
JSON encoder and console encoder implement, and make the error field
attempt an upcast into that type. That would let the user supply their
own error encoder (much like they supply their own time and duration
encoders now). Even if we do that, though, I suspect that we'll want to
preserve the behavior here as the default.

### DIFF
--- a/zapcore/error.go
+++ b/zapcore/error.go
@@ -46,6 +46,10 @@ func encodeError(key string, err error, enc ObjectEncoder) error {
 	basic := err.Error()
 	enc.AddString(key, basic)
 
+	if group, ok := err.(errorGroup); ok {
+		return enc.AddArray(key+"Causes", errArray(group.Errors()))
+	}
+
 	if fancy, ok := err.(fmt.Formatter); ok {
 		verbose := fmt.Sprintf("%+v", fancy)
 		if verbose != basic {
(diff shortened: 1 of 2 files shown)


zapcore/error.go
```diff
@@ -46,6 +46,10 @@ func encodeError(key string, err error, enc ObjectEncoder) error {
 	basic := err.Error()
 	enc.AddString(key, basic)
 
+	if group, ok := err.(errorGroup); ok {
+		return enc.AddArray(key+"Causes", errArray(group.Errors()))
+	}
+
 	if fancy, ok := err.(fmt.Formatter); ok {
 		verbose := fmt.Sprintf("%+v", fancy)
```
abhinav: nit: use type switch now that it's one or the other?

```
switch e := err.(type) {
case errorGroup:
  ...
case fmt.Formatter:
  ...
}
```

akshayjshah: Oh...yeah, that totally makes sense.

akshayjshah:
Yep, outlined a plan for making this configurable in the PR description. I'm not planning to put cycles into that until we need that customizability or someone else files an issue, though.

PR #473, opened 2017-07-08
Small documentation formatting fixes

akshayjshah:
GoDoc.org renders docs a little differently from my local `godoc`. This
PR fixes some whitespace and breaks up some long lines so that they look
better.

It also removes backticks from a last-minute documentation addition I
made (Markdown habits die hard).

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -75,17 +75,18 @@ func Example_basicConfiguration() {
 	// See the documentation for Config and zapcore.EncoderConfig for all the
 	// available options.
 	rawJSON := []byte(`{
-		"level": "debug",
-		"encoding": "json",
-		"outputPaths": ["stdout", "/tmp/logs"],
-		"errorOutputPaths": ["stderr"],
-		"initialFields": {"foo": "bar"},
-		"encoderConfig": {
-			"messageKey": "message",
-			"levelKey": "level",
-			"levelEncoder": "lowercase"
-		}
+	  "level": "debug",
+	  "encoding": "json",
+	  "outputPaths": ["stdout", "/tmp/logs"],
+	  "errorOutputPaths": ["stderr"],
+	  "initialFields": {"foo": "bar"},
+	  "encoderConfig": {
+	    "messageKey": "message",
+	    "levelKey": "level",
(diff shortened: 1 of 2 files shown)


example_test.go
```diff
@@ -75,17 +75,18 @@ func Example_basicConfiguration() {
...
+	  "encoding": "json",
+	  "outputPaths": ["stdout", "/tmp/logs"],
+	  "errorOutputPaths": ["stderr"],
+	  "initialFields": {"foo": "bar"},
+	  "encoderConfig": {
+	    "messageKey": "message",
+	    "levelKey": "level",
+	    "levelEncoder": "lowercase"
+	  }
 	}`)
```
akshayjshah: This is a little hard to grok in the review view, but it formats the JSON to be indented by a tab (to line up with `rawJSON`) and then two spaces per level of JSON indentation. GoDoc.org displays tabs as eight spaces in docs (I think), which makes this JSON look ridiculous.

example_test.go
```diff
@@ -124,11 +125,13 @@ func Example_advancedConfiguration() {
...
 	// mutex with zapcore.Lock.)
-	topicDebugging, topicErrors := zapcore.AddSync(ioutil.Discard), zapcore.AddSync(ioutil.Discard)
+	topicDebugging := zapcore.AddSync(ioutil.Discard)
+	topicErrors := zapcore.AddSync(ioutil.Discard)
 
 	// High-priority output should also go to standard error, and low-priority
 	// output should also go to standard out.
-	consoleDebugging, consoleErrors := zapcore.Lock(os.Stdout), zapcore.Lock(os.Stderr)
+	consoleDebugging := zapcore.Lock(os.Stdout)
+	consoleErrors := zapcore.Lock(os.Stderr)
```
akshayjshah: Unlike my local `godoc`, GoDoc.org uses a fixed-width element for code. These lines are cute, but they make the docs scroll horizontally there.

example_test.go
prashantv: Thanks for the explanation, I was a little confused initially -- then figured it was just the contents in the backticks that now use spaces. As long as `gofmt` is happy, looks good.

PR #475, opened 2017-07-09
Add filtering cores

### LINKED ISSUE #453: How to implement a filtering middleware core?
Hi,

I implemented a core that sends log entries to Sentry. But I want to skip entries containing particular fields from being sent to Sentry. For that reason I tried to implement a core that would wrap the Sentry core and simply skip writing the entry when certain field is present.

The thing is, I don't see how the filtering core can be done in a generic way since calling `Check` on the filtering core should obviously call `Check` on the core being wrapped. But that will add the underlying core into the checked entry, causing it to always contain the log entry. In other words, this does not work:

https://github.com/tchap/zapext/blob/a5d992351555de581c6d36e2829287a6d1c2f16a/middleware/core_filtering.go

The problem seems to be that fields are not available in `Check` so I cannot run the filter function without really calling the underlying core's `Check` first.

Any ideas how to do this? Am I missing something obvious?

tchap:
Closes #453 .

`NewFieldFilteringCore` can be used to wrap a core with a middleware core
that uses the given filter function to filter fields before passing them
to the core being wrapped.

This PR is not complete, I intend to add `NewEntryFilteringCore` as well.

What is important, though, that the test is not passing, for rather unknown reasons.

### DIFF
--- a/filtering_cores.go
+++ b/filtering_cores.go
@@ -1,0 +1,63 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


akshayjshah:
Ah, I see the problem here: you also need to implement the `Check` method from `zapcore.Core`. Unfortunately, I don't see a clean, 100% correct way for you to do this.

To make your test pass, you could add a method like this:

```go
func (core *fieldFilteringCore) Check(ent zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
    if ce := core.Core.Check(ent, ce); ce == nil {
        return nil
    }
    return ce.AddCore(ent, core)
}
```

However, this won't behave correctly if the wrapped core (`next` in your constructor) uses `CheckedEntry.Should`. None of the cores that come with zap do this, but there may be other implementations that do. Since I don't see a clean, fully correct way for us to implement this feature, I don't want to merge it directly into zap.

Don't let that stop you, though! You can take this code, with the snippet above, and use it in your own code (or release it as your own zap extension package). It works correctly with all the cores that are *currently* part of zap, and as long as you're careful to not wrap a core that uses `Should`, you'll be fine.

PR #479, opened 2017-07-13
Add AddClassFunction option

nplanel:
Configures the Logger to annotate each message with the caller class and
function name.

for example :

http/server.go:127 (*Server).ListenAndServe

Signed-off-by: Nicolas PLANEL <nplanel@redhat.com>

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -45,8 +45,9 @@ type Logger struct {
 	name        string
 	errorOutput zapcore.WriteSyncer
 
-	addCaller bool
-	addStack  zapcore.LevelEnabler
+	addCaller        bool
+	addClassFunction bool
+	addStack         zapcore.LevelEnabler
 
 	callerSkip int
 }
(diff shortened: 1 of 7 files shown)


akshayjshah:
Thanks for the PR, @nplanel - I appreciate the work you put into making sure that all the new code is tested!

I don't think that this particular enhancement needs to be integrated into zap, though. It doesn't require access to any package internals, and can easily live outside of zap (either in a standalone extension package or with the code that uses it).

Here's a self-contained `CallerEncoder` that accomplishes the same thing:

```go
func ShortCallerWithClassFunctionEncoder(caller EntryCaller, enc PrimitiveArrayEncoder) {
  path := caller.TrimmedPath()
  if f := runtime.FuncForPC(caller.PC); f != nil {
    name := f.Name()
    i := strings.LastIndex(name, "/")
    j := strings.Index(name[i+1:], ".")
    path += " " + name[i+j+2:]
  }
  enc.AppendString(path)
}
```

You can pass that `CallerEncoder` in an `EncoderConfig` just like zap's built-ins.

nplanel:
Ok thanks @akshayjshah 
I was seeing that more like a general feature/option for uber-zap

PR #484, opened 2017-07-26
Fix zerolog benchmark

rs:
- Configure zerolog to match zap
- Update zerolog and use new Object/Array functions
- Add a new bench scenario with only primitive field types
- Add some missing Check bench with zerolog

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Log a message and 10 fields:
 | :--- | :---: | :---: |
 | :zap: zap | 3131 ns/op | 5 allocs/op |
 | :zap: zap (sugared) | 4173 ns/op | 21 allocs/op |
-| zerolog | 16154 ns/op | 90 allocs/op |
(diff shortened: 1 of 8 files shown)


rs:
Any chance to merge this?

akshayjshah:
Probably! Unfortunately, I won't have a chance to look at this in detail until later this week.

Right off the bat, I'm skeptical of the sub-100ns benchmarks. That's on par with direct usage of `os.Stdout.Write`, and suggests to me that you're running the benchmarks without timestamping zerolog's output. Can you please make sure that you're fully rebased on master?

rs:
Yes, I am rebased on the master. The reason you got such a difference in the master is because by default timestamps are formatted as string using RFC3339 in zerolog, as opposed to simple UNIX timestamps in production's zap configuration.

Direct usage of `Write` is fine as the write syscall is thread safe.

Issue #439, opened 2017-06-01
Make level configurable for NewStdLog


heyimalex:
For things like `net/http#Server.ErrorLog`, it might make sense for all log lines to be at the error level. Since we can add fields and/or name the logger it's not a huge deal to pick these lines out of stream, but it kind of reduces the utility of levels as a catch all this-is-something-that-should-be-seen-by-a-human field.

Something to consider? Understand if it's too minor of a change, honestly having no issues working around it today.

akshayjshah:
No opposition from me, but I probably won't get to this for a while.

delicb:
I needed this for exact case that was mentioned (`net/http#Server.ErrorLog`). However, it turned out that I needed `io.Writer` that will produce logs as well and zap does not expose writer that it uses, so I wrote it myself and I am willing to create pull request. Would something like this be accepted (ignore hardcoded values for `AddCallerSkip` for now)?

```go
// logWriter is wrapper around Logger that implements io.Writer interface.
type logWriter struct {
	logFunc func(msg string, fields ...zapcore.Field)
}

// Write implements io.Writer for logWriter.
// It will never return error.
func (l *logWriter) Write(p []byte) (n int, err error) {
	l.logFunc(string(p))
	return len(p), nil
}

// ToWriter creates and returns io.Writer that logs all messages written to in
// by using provided Logger and log level.
// If unsupported log level is provided as parameter there will be not errors,
// but Warn level is used.
func ToWriter(logger *zap.Logger, level zapcore.Level, options ...zap.Option) io.Writer {
	if logger == nil {
		return ioutil.Discard
	}
	var logFunc func(msg string, fields ...zapcore.Field)

	// since we are introducing new level of indirection, if provided
	// implementation is zap.Logger, we know how to indicate to it to skip
	// this level when reporting origin or the log message.
	logger = logger.WithOptions(zap.AddCallerSkip(2)).WithOptions(options...)

	switch level {
	case zapcore.DebugLevel:
		logFunc = logger.Debug
	case zapcore.InfoLevel:
		logFunc = logger.Info
	case zapcore.WarnLevel:
		logFunc = logger.Warn
	case zapcore.ErrorLevel:
		logFunc = logger.Error
	case zapcore.PanicLevel:
		logFunc = logger.Panic
	case zapcore.DPanicLevel:
		logFunc = logger.DPanic
	case zapcore.FatalLevel:
		logFunc = logger.Fatal
	default:
		logFunc = logger.Warn
	}
[...]

PR #487, opened 2017-08-02
Make level configurable for NewStdLog. (#439)

delicb:
With this change new function is added called `NewStdLogAt`. It has same
behaviour as NewStdLog, but it allows providing log level to use by
returned `*log.Logger`.

Couple of things to note:
- it would be much more straight forward to implement this if there is function `*zap.Logger.Log(Level, string, …zapcore.Fields)` in addition to `Debug`, `Info`, etc. I understand if you do not want to introduce it, but if you are opened to it, I can add it and adapt implementation.
- `loggerWriter` is changed to accept function to use, so that decision which function should be used is elsewhere. Existing code is compatible, since default was `Info` and that is passed. This can be other way around - `loggerWriter` can have both logger and level and decide which function to call during `Write` execution, but I do not really like idea of executing switch case every time `Write` is called.
- Tests are refactored a bit, so that common parts among new tests and existing one `TestNewStdLog` can be reused. It might be possible to generalize this to table-driven test, but DPanic, Panic and Fatal levels have special handling, so it might be a bit long. I prefer it the way I wrote it, but I am opened to changing it to table driven test if that is preferred.

### DIFF
--- a/global.go
+++ b/global.go
@@ -22,14 +22,17 @@ package zap
 
 import (
 	"bytes"
+	"fmt"
 	"log"
 	"os"
 	"sync"
+
+	"go.uber.org/zap/zapcore"
 )
 
 const (
 	_stdLogDefaultDepth = 2
-	_loggerWriterDepth  = 1
+	_loggerWriterDepth  = 2
 )
 
 var (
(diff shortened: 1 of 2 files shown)


akshayjshah:
@delicb Sorry for the long delay reviewing this! Changes and test coverage look great. I'm going to push a commit with a few small nits (e.g., `assert.Nil(t, err)` is better as `assert.NoError(t, err)`) and then merge this.

Thank you for the contribution!

delicb:
@akshayjshah Ah, sure, `NoError` is better. I actually found out that it exists few days ago (after I created this pull request). I can fix that, if you would like.

Thank you for accepting PR.

akshayjshah:
🤦‍♂️ I completely forgot about this - I'm so sorry, @delicb! I'll merge this today.

djui:
@akshayjshah Was there a specific reason to not additionally make this change (log level configurable) for `redirectStdLog`, or just to stay focus on the PRs description? Any plans on adding it?

akshayjshah:
Nope, you're welcome to open a PR with `RedirectStdLogAt`. The existing function can't start taking extra parameters without breaking backward compatibiliy.

PR #490, opened 2017-08-15
Add ContextMap method to LoggedEntry which returns fields as a map

prashantv:
Tests often care about specific fields, or want fields without worrying
about order, or marshalling objects etc. It makes more sense for tests
to compare a fields `map[string]interface{}` rather than `[]zap.Field`

### DIFF
--- a/zaptest/observer/fields.go
+++ b/zaptest/observer/fields.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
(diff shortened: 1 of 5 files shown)


zaptest/observer/fields.go
```diff
@@ -0,0 +1,32 @@
...
+import "go.uber.org/zap/zapcore"
+
+// FieldsMap takes a list of fields and converts it to a map.
+func FieldsMap(fields []zapcore.Field) map[string]interface{} {
+	encoder := zapcore.NewMapObjectEncoder()
+	for _, f := range fields {
+		f.AddTo(encoder)
+	}
+	return encoder.Fields
+}
```
akshayjshah: Love the implementation. Would this be more ergonomic as a `ContextMap` method on `LoggedEntry`?

prashantv: I'd prefer to keep this separate from `LoggedEntry` since it doesn't have the message or timestamp.

I'm open to `ContextMap`, however any underlying nested objects would use `map[string]interface{}`. If we're OK with that, I can make this a new `ContextMap` type.

akshayjshah: Clarified offline - since the `[]zapcore.Field` usually comes from `LoggedEntry.Context`, we're going to start by exposing this as `LoggedEntry.ContextMap()`.

prashantv: Misunderstood the comment, moved this method to be a function on `LoggedEntry`

PR #504, opened 2017-09-22
MapEncoder: Store strings in map when AddByteString is used

### LINKED ISSUE #503: MapEncoder.AddByteString should add the value as a string
The `AddByteString` was added as a performance optimization to avoid allocations from `[]byte` to `string` when logging, when the final value is a string:
https://github.com/uber-go/zap/pull/366

Currently the MapEncoder stores this as a `[]byte`, which is incorrect. We should store the value as a `string`

prashantv:
Fixes #503.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v1.7.1 (25 Sep 2017)
+
+Bugfixes:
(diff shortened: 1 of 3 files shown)


akshayjshah:
If you also add a changelog entry, we can just cut a patch as soon as this lands.

Issue #96, opened 2016-07-05
How can zap support syslog


lysu:
rt~

akshayjshah:
Eventually, yes - supporting syslog will require both a syslog encoder (similar to zap's current JSON encoder) and a `WriteSyncer`.

We're definitely not going to get to this until after a 1.0 release, though.

skypacer210:
Hi @akshayjshah 
Dose it mean that zap can't write log to file? If zap support write to file, how to use it? Thanks a lot.

prashantv:
@skypacer210 This issue is not related to writing to a file, but specifically to the [syslog](https://en.wikipedia.org/wiki/Syslog) interface.

You can write to a file using zap, take a look at [`zap.Output` option](https://godoc.org/github.com/uber-go/zap#Output). The `os.File` type satisfies this interface so you can open a file and pass it in as the output.

I've filed #100 so we have an example to point to for file logging.

skypacer210:
@akshayjshah Thanks a lot!

skypacer210:
@akshayjshah Do zap support logging with rotation?

prashantv:
@skypacer210 Please look at existing documentation and issues, there is a separate issue about this: https://github.com/uber-go/zap/issues/98. As I mentioned earlier, this issue is not related to writing to a file, but about the syslog interface, so let's move the discussion of logging to file away from this issue.

skypacer210:
@prashantv Got it, thanks.

bclermont:
I actually implemented a nearly completed Syslog friendly encoder (mostly just a slightly modified JSON encoder) and a syslog writer.

But I can't finish it because `zap.Encoder.Encoder` interface `WriteEntry(io.Writer, string, Level, time.Time) error` function first argument `io.Writer` is not enough.

The problem is: syslog writer need to know the `zap.Level` and not just the `[]byte` of the encoded message. Syslog must specify priority and level on each new log entry.

To achieve this I suggest create the new following interface:

```
type LogWriter interface {
		Write([]byte, Level) (int, error)
}
```

and use it instead of `io.Writer` in `WriteEntry`.

For writers that don't need the level such to use `os.Stdout` as a `LogWriter`, we can use that wrapper:

```
type wrappedWriter struct {
	w io.Wrapper
}

func (w *wrappedWriter) Write(p []byte, _ Level) (int, error) {
	return w.Write(p)
}

func WrapIgnoreLevel(w io.Wrapper) LogWriter {
[...]

PR #506, opened 2017-09-27
syslog writer

avidak:
This commit allows zap to write log in syslog writer

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -22,7 +22,9 @@ package zap
 
 import (
 	"io/ioutil"
+	"log/syslog"
 	"os"
+	p "path"
 
 	"go.uber.org/zap/zapcore"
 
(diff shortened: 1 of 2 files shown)


akshayjshah:
Thank you for the PR, @avidak! I probably won't have a chance to review it until next week.

jeffbean:
This PR is useful for a project I'm working on. 

Just a friendly ping.

lafolle:
We're also waiting for this to merge.

Just another friendly ping :)

akshayjshah:
Thank you for the effort you put into this PR, @avidak!

I've added a comment to the discussion on #96, which I won't repeat in full here. The short version is that the zap maintainers (including me) don't understand syslog well enough to support this code, so we'd prefer to keep it in a separate extension package that's owned by someone with more expertise.

I'm going to close this PR - let's continue the discussion #96 instead.

PR #522, opened 2017-10-30
DRY up standard logger redirection code

akshayjshah:
Remove some duplication from the standard logger redirection code.

/cc @djui - thanks again for contributing this code, just tagging you here in case you're interested.

### DIFF
--- a/global.go
+++ b/global.go
@@ -31,8 +31,10 @@ import (
 )
 
 const (
-	_stdLogDefaultDepth = 2
-	_loggerWriterDepth  = 2
+	_stdLogDefaultDepth      = 2
+	_loggerWriterDepth       = 2
+	_programmerErrorTemplate = "You've found a bug in zap! Please file a bug at " +
+		"https://github.com/uber-go/zap/issues/new and reference this error: %v"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


global.go
```diff
@@ -113,58 +98,64 @@ func NewStdLogAt(l *Logger, level zapcore.Level) (*log.Logger, error) {
...
+	}
+	log.SetOutput(&loggerWriter{logFunc})
+	return func() {
+		log.SetFlags(flags)
+		log.SetPrefix(prefix)
+		log.SetOutput(os.Stderr)
+	}, nil
+}
+
+func levelToFunc(logger *Logger, lvl zapcore.Level) (f func(string, ...zapcore.Field), err error) {
```
akshayjshah: This felt like a place where named return values simplified the code; happy to remove them if any reviewers disagree.

global.go
```diff
@@ -113,58 +98,64 @@ func NewStdLogAt(l *Logger, level zapcore.Level) (*log.Logger, error) {
...
-	log.SetOutput(&loggerWriter{logFunc})
-	return func() {
-		log.SetFlags(flags)
-		log.SetPrefix(prefix)
-		log.SetOutput(os.Stderr)
+	f, err := redirectStdLogAt(l, InfoLevel)
+	if err != nil {
+		// Can't get here, since passing InfoLevel to redirectStdLogAt always
+		// works.
+		panic(err)
```
abhinav: optional: if this happens, we should ask the user to file a bug report (similar to the dig panics)

global.go
prashantv: not strong objections, but I don't think it provides a huge amount of value over
```
return logger.Debug, nil
```

it's slightly less verbose, but doesn't really affect clarity otherwise, so i'd prefer to avoid.

akshayjshah: Fair enough, I'm not strongly attached to the named results.

global.go
akshayjshah: Can do.

PR #523, opened 2017-10-30
Break dependency between zap and zaptest

akshayjshah:
In order to introduce new testing helpers into the `zaptest` package, where users will expect to find them, we need to break the dependency between `zap` and `zaptest`. (Unfortunately, we introduced `zaptest` before Go supported internal packages and squatted on the best user-facing name.)

This PR can be reviewed commit-by-commit if necessary. It makes a wordy, but logically simple series of changes: we copy the `zaptest` package to `internal/ztest`, change the rest of `zap` to use the new internal helpers, and finally convert `zaptest` to a shim around the internal helpers. In Go 1.9, we use type aliases to also clean up the documentation (aliases don't have their methods documented); for older Go versions, we leave a copy of the original code and a test to make sure that the two versions don't drift out of sync.

This should make #518 much nicer. Since it begins using type aliases, it depends on #524.

### DIFF
--- a/zaptest/timeout.go
+++ b/zaptest/timeout.go
@@ -21,31 +21,25 @@
 package zaptest
 
 import (
-	"log"
-	"os"
-	"strconv"
 	"time"
+
+	"go.uber.org/zap/internal/ztest"
 )
 
-var _timeoutScale = 1.0
-
 // Timeout scales the provided duration by $TEST_TIMEOUT_SCALE.
+//
+// Deprecated: This function is intended for internal testing and shouldn't be
+// used outside zap itself. It was introduced before Go supported internal
+// packages.
 func Timeout(base time.Duration) time.Duration {
-	return time.Duration(float64(base) * _timeoutScale)
+	return ztest.Timeout(base)
 }
 
 // Sleep scales the sleep duration by $TEST_TIMEOUT_SCALE.
+//
+// Deprecated: This function is intended for internal testing and shouldn't be
+// used outside zap itself. It was introduced before Go supported internal
+// packages.
 func Sleep(base time.Duration) {
-	time.Sleep(Timeout(base))
-}
-
-func init() {
-	if v := os.Getenv("TEST_TIMEOUT_SCALE"); v != "" {
-		fv, err := strconv.ParseFloat(v, 64)
-		if err != nil {
-			panic(err)
-		}
-		_timeoutScale = fv
-		log.Printf("Scaling timeouts by %vx.\n", _timeoutScale)
-	}
+	ztest.Sleep(base)
 }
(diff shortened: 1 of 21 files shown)


zaptest/timeout.go
```diff
@@ -21,31 +21,17 @@
...
-	"os"
-	"strconv"
 	"time"
-)
 
-var _timeoutScale = 1.0
+	"go.uber.org/zap/internal/ztest"
+)
 
 // Timeout scales the provided duration by $TEST_TIMEOUT_SCALE.
```
prashantv: can you mark this as deprecated? I don't think these functions are intended for end-users right? This should purely be for internal testing

zaptest/timeout.go
```diff
@@ -21,31 +21,17 @@
...
 func Timeout(base time.Duration) time.Duration {
-	return time.Duration(float64(base) * _timeoutScale)
+	return ztest.Timeout(base)
 }
 
 // Sleep scales the sleep duration by $TEST_TIMEOUT_SCALE.
```
prashantv: same as above, should this be deprecated?

zaptest/writer_go_19.go
```diff
@@ -0,0 +1,45 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// +build go1.9
```
prashantv: optional: should we call out why we need 1.9 (e.g., that we're using type aliases?)

zaptest/writer_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func readCode(t testing.TB, fname string) []string {
+	f, err := os.Open(fname)
+	require.NoError(t, err, "Failed to read %s.", fname)
```
prashantv: `defer f.Close()` after the `require.NoError`

zaptest/timeout.go
akshayjshah: Yep, can do.

zaptest/timeout.go
akshayjshah: Yep.

zaptest/writer_go_19.go
akshayjshah: Sure, sounds good!

zaptest/writer_test.go
akshayjshah: 🤦‍♂️ Good catch.

PR #525, opened 2017-11-03
stacktrace: skip frames

### LINKED ISSUE #512: CallerSkip does not apply to stacktraces captured by zap
Currently, logging wrappers use `CallerSkip` to avoid the logging wrapper showing up as the caller. However, this is not used by stacktraces, so we end up with the logging wrappers showing up in stacktraces.

@akshayjshah: I'm thinking we should respect caller skip in stacktraces as well, any thoughts?

boz:
* Honor StackSkip option when including stacktrace.
 * Replace package-based stack pruning - the same behavior is achieved by
   skipping frames appropriately.

 fixes #512

### DIFF
--- a/field.go
+++ b/field.go
@@ -188,7 +188,14 @@ func Stack(key string) zapcore.Field {
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	return String(key, takeStacktrace())
+	return stackField(key, 1)
+}
+
+// stackField returns a stacktrace field with the first skip frames ignored.
+//
+// stackField("foo",0) is equivalent to Stack("foo").
+func stackField(key string, skip int) zapcore.Field {
+	return String(key, takeStacktrace(skip+1))
 }
 
 // Duration constructs a field with the given key and value. The encoder
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -22,13 +22,10 @@ package zap
 
 import (
 	"runtime"
-	"strings"
 	"sync"
 
 	"go.uber.org/zap/internal/bufferpool"
 )
-
-const _zapPackage = "go.uber.org/zap"
 
 var (
 	_stacktracePool = sync.Pool{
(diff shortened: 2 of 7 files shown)


boz:
Hi @prashantv, @akshayjshah.  Any thoughts on this?

Thanks!

field.go
```diff
@@ -184,11 +184,19 @@ func Time(key string, val time.Time) zapcore.Field {
...
 // takes about two microseconds.
 func Stack(key string) zapcore.Field {
+	return String(key, takeStacktrace(1))
+}
+
+// StackSkip returns a stacktrace field like Stack and also allows for skipping
+// stack frames via the skip argument.
+//
+// StackSkip("foo",0) is equivalent to Stack("foo").
+func StackSkip(key string, skip int) zapcore.Field {
```
prashantv: I'm not sure if we should expose this right now -- I'd prefer to keep just `Stack` for now and think about whether we really need the skip version separately.

stacktrace.go
```diff
@@ -22,38 +22,31 @@ package zap
...
+func takeStacktrace(skip int) string {
 	buffer := bufferpool.Get()
 	defer buffer.Free()
 	programCounters := _stacktracePool.Get().(*programCounters)
 	defer _stacktracePool.Put(programCounters)
 
 	var numFrames int
+	skip += 2
 	for {
 		// Skip the call to runtime.Counters and takeStacktrace so that the
```
prashantv: can you move this comment to where we do `+= 2`

stacktrace_ext_test.go
```diff
@@ -37,22 +32,23 @@ import (
...
 		logger.Error("test log")
 		logger.Sugar().Error("sugar test log")
 
-		require.Contains(t, out.String(), "TestStacktraceFiltersZapLog", "Should not strip out non-zap import")
-		verifyNoZap(t, out.String())
+		frames := strings.Split(out.String(), "\n")
+
+		assert.True(t, len(frames) > 6)
+
+		// The logging function should be the first frame ...
```
prashantv: instead of testing partially, can we just test the entire result of `frames`?

stacktrace.go
```diff
@@ -63,19 +56,12 @@ func takeStacktrace() string {
 	}
 
 	i := 0
-	skipZapFrames := true // skip all consecutive zap frames at the beginning.
```
prashantv: This behaviour was added on purpose, any reason why we're deleting this logic?

prashantv:
Thanks for the contribution @boz 

Can you provide more context about why we're removing the `zap` stacktrace pruning?

boz:
Hi @prashantv 

As far as I can tell, accounting for frames correctly with `skip` and `zap.StackSkip()` achieves the same thing that the name-based frame skipping did.  I might be missing something, tho.

stacktrace_ext_test.go
boz: Sure, added.  I was worried it'd be too brittle.

akshayjshah:
Today, the offset for determining callers is separate from the stack-pruning behavior. That means that it's entirely possible to use a logger that *never* includes the calling function, but does occasionally log stacktraces. In that case, the caller skip likely wouldn't be configured, but stacktraces would still prune zap's internals.

Unfortunately, that means that this change isn't backward-compatible and can't ship as-is :/

In retrospect, I regret the offset-based skip mechanism we used to determine the calling function. It's useful, but has turned out to be even more fiddly and error-prone than I expected.

akshayjshah:
This has been open for a long time, and I'd prefer not to introduce even small backward-incompatible changes if they're not absolutely necessary. For housekeeping, I'm going to close this PR - @boz, happy to discuss further if you want to re-open it.

PR #527, opened 2017-11-07
Added type method for plfag.Value interface

### LINKED ISSUE #526: Add Type() method to zapcore.Level to satisfy pflag's Value interface
The `zapcore.Level` (as of v1.7.1) satisfies the stdlib's [`flag.Value`](https://golang.org/pkg/flag/#Value) interface, so that [`zap.LevelFlag`](https://godoc.org/go.uber.org/zap#LevelFlag) can be used. 
```go
type Value interface {
        String() string
        Set(string) error
}
```

For lots of projects, the pflag library (by [spf13](https://github.com/spf13/pflag)) is used instead of the default flag lib. In order to have a pflag that takes a `zapcore.Level`, one addition `Type() string` method must be added to satisfy the [`Value` interface](https://godoc.org/github.com/spf13/pflag#Value) :
```go
type Value interface {
    String() string
    Set(string) error
    Type() string
}
```

I'd like to add a `Type() string` method to `zapcore.Level` to satisfy the `plfag.Value` interface.

micahhausler:
Closes #526

### DIFF
--- a/zapcore/level.go
+++ b/zapcore/level.go
@@ -155,6 +155,11 @@ func (l *Level) Get() interface{} {
 	return *l
 }
 
+// Type gets the type for the pflag.Value interface
+func (l *Level) Type() string {
+	return "level"
+}
+
 // Enabled returns true if the given level is at or above this level.
 func (l Level) Enabled(lvl Level) bool {
 	return lvl >= l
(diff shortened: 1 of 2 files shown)


akshayjshah:
Discussed on #526 - I'd much prefer that you support `pflag` with a small wrapper in a separate project. I'm happy to merge an addition to zap's FAQ that sends `pflag` users to that project.

PR #528, opened 2017-11-09
Fixed NPE in Stringer()

### LINKED ISSUE #495: logger.Info(zap.Any("",*url.URL)) should not panic if *url.URL is nil
code:
```
var a *url.URL
l, _ := zap.NewProduction()
l.Info("msg", zap.Any("url", a))
```
result:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1232c70]

goroutine 25 [running]:
testing.tRunner.func1(0xc4200745b0)
	/usr/local/Cellar/go/1.8.3/libexec/src/testing/testing.go:622 +0x29d
panic(0x13e4ec0, 0x16b1430)
	/usr/local/Cellar/go/1.8.3/libexec/src/runtime/panic.go:489 +0x2cf
net/url.(*URL).String(0x0, 0x1430a80, 0x0)
	/usr/local/Cellar/go/1.8.3/libexec/src/net/url/url.go:726 +0x40
code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore.Field.AddTo(0x145b2e5, 0x3, 0x18, 0x0, 0x0, 0x0, 0x1430a80, 0x0, 0x1690c80, 0xc420175e40)
	/Users/yutongpei/Uber/Code/gocode/src/code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore/field.go:163 +0xd52
code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore.addFields(0x1690c80, 0xc420175e40, 0xc420015000, 0x1, 0x1)
	/Users/yutongpei/Uber/Code/gocode/src/code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore/field.go:199 +0x71
code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore.(*jsonEncoder).EncodeEntry(0xc420175d20, 0x0, 0xed1315168, 0xd98249d, 0x16bbae0, 0x0, 0x0, 0x145b2af, 0x3, 0x1, ...)
	/Users/yutongpei/Uber/Code/gocode/src/code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore/json_encoder.go:339 +0x209
code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore.(*ioCore).Write(0xc42001de60, 0x0, 0xed1315168, 0xd98249d, 0x16bbae0, 0x0, 0x0, 0x145b2af, 0x3, 0x1, ...)
	/Users/yutongpei/Uber/Code/gocode/src/code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore/core.go:86 +0xa9
code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc42008c2c0, 0xc420015000, 0x1, 0x1)
	/Users/yutongpei/Uber/Code/gocode/src/code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/zapcore/entry.go:215 +0x116
code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap.(*Logger).Info(0xc4200183c0, 0x145b2af, 0x3, 0xc420015000, 0x1, 0x1)
	/Users/yutongpei/Uber/Code/gocode/src/code.uber.internal/engsec/auth-common.git/vendor/go.uber.org/zap/logger.go:187 +0x7f
```

the problem is zap found out url has a `String()` function and directly used it. Unfortunately, net/url doesn't perform nil ptr check in `String()` function.

I don't think zap user should perform the nil ptr check, it is very normal to log a nil value.
I think ideally the `String()` implementor should do the nil ptr check, but that is hardly enforced for every struct implemented `String()`
So maybe zap can do the nil ptr check. The easy way is print "\<nil\>" if input object is nil. The harder way is call `String()` first(`String()` may handle nil case already) and trap the panic if there is one.

selslack:
Fixes #495

### DIFF
--- a/field.go
+++ b/field.go
@@ -23,6 +23,7 @@ package zap
 import (
 	"fmt"
 	"math"
+	"reflect"
 	"time"
 
 	"go.uber.org/zap/zapcore"
@@ -170,6 +171,9 @@ func Namespace(key string) zapcore.Field {
 // Stringer constructs a field with the given key and the output of the value's
 // String method. The Stringer's String method is called lazily.
 func Stringer(key string, val fmt.Stringer) zapcore.Field {
+	if val == nil || reflect.ValueOf(val).IsNil() {
+		return Skip()
+	}
 	return zapcore.Field{Key: key, Type: zapcore.StringerType, Interface: val}
 }
 
--- a/field.go
+++ b/field.go
@@ -170,6 +170,7 @@ func Namespace(key string) zapcore.Field {
 
 // Stringer constructs a field with the given key and the output of the value's
 // String method. The Stringer's String method is called lazily.
+// If passed a nil value, the field is a no-op.
 func Stringer(key string, val fmt.Stringer) zapcore.Field {
 	if val == nil || reflect.ValueOf(val).IsNil() {
 		return Skip()
--- a/field.go
+++ b/field.go
@@ -172,9 +172,19 @@ func Namespace(key string) zapcore.Field {
 // String method. The Stringer's String method is called lazily.
 // If passed a nil value, the field is a no-op.
 func Stringer(key string, val fmt.Stringer) zapcore.Field {
-	if val == nil || reflect.ValueOf(val).IsNil() {
+	if val == nil {
 		return Skip()
 	}
+
+	r := reflect.ValueOf(val)
+
+	switch r.Type().Kind() {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.Interface, reflect.Slice:
+		if r.IsNil() {
+			return Skip()
+		}
+	}
+
 	return zapcore.Field{Key: key, Type: zapcore.StringerType, Interface: val}
 }
 
--- a/field_test.go
+++ b/field_test.go
@@ -22,6 +22,7 @@ package zap
 
 import (
 	"net"
+	"net/url"
 	"sync"
 	"testing"
 	"time"
@@ -63,6 +64,10 @@ func TestFieldConstructors(t *testing.T) {
 	name := username("phil")
 	ints := []int{5, 6}
 
+	// For Stringer() test
+	var nUrl *url.URL
+	nUrl = nil
+
 	tests := []struct {
 		name   string
 		field  zapcore.Field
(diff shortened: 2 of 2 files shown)


field.go
```diff
@@ -169,7 +170,21 @@ func Namespace(key string) zapcore.Field {
 
 // Stringer constructs a field with the given key and the output of the value's
 // String method. The Stringer's String method is called lazily.
+// If passed a nil value, the field is a no-op.
 func Stringer(key string, val fmt.Stringer) zapcore.Field {
+	if val == nil {
+		return Skip()
+	}
+
+	r := reflect.ValueOf(val)
```
prashantv: I have a few concerns with this approach:
 - we're adding a reflect call when creating a field, which could have a performance impact (we should measure the impact)
 - we're changing the behaviour of `Stringer`. It's possible for a type to implement a `String` method that works for `nil`:
https://play.golang.org/p/EUNoR7Sz9Q

Now, we're changing the behaviour where it'll `Skip` the field.

This is also bad for `nil` slices for example, where we might want to output a empty sice rather than skipping the field.

selslack: 1) It's 288 ns/op vs 316 ns/op on my laptop, additional datapoints are welcome.

2) Yea, I'm aware of that possibility. Simplest way of maintaining compatibility is to introduce something like `StringerNilSafe` field or bump the library version. If you have any ideas -- they are welcome.

I find the behavior of `Error` simple and useful: if argument is nil it's not rendered.

As for slices: there could be scenario where I want to distinguish nil-slice vs empty slice in logs.

akshayjshah: I don't mind adding the nil check above; it's cheap and always correct.

I'd prefer not to have the reflection-based checks below. If we get a non-nil interface with a nil value that also dereferences the receiver in `String`, it's completely reasonable to panic - in my book, that's definitely programmer error.

akshayjshah:
😶  Closing due to inactivity. 😶 

This is just housekeeping - leave a comment and re-open if necessary!

PR #538, opened 2017-12-19
Implementing grpclog.LoggerV2 compatible logger

### LINKED ISSUE #534: zapgrpc upgrade for grpclog.LoggerV2
The latest gRPC releases support an expanded logger interface (https://github.com/grpc/grpc-go/blob/master/grpclog/loggerv2.go). It'd be great if zap's gRPC support was updated accordingly.

Thanks.

frankgreco:
Currently, `go.uber.org/zap/zapgrpc.Logger` only provides an implementation for `grpclog.Logger`. With the release of `grpclog.LoggerV2`, an updated implementation is needed so that `go.uber.org/zap/zapgrpc.Logger` implements both `grpclog.Logger` and `grpclog.LoggerV2`.

Fixes #534

### DIFF
--- a/zapgrpc/zapgrpc.go
+++ b/zapgrpc/zapgrpc.go
@@ -21,7 +21,10 @@
 // Package zapgrpc provides a logger that is compatible with grpclog.
 package zapgrpc // import "go.uber.org/zap/zapgrpc"
 
-import "go.uber.org/zap"
+import (
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
 
 // An Option overrides a Logger's default configuration.
 type Option interface {
@@ -48,11 +51,17 @@ func WithDebug() Option {
 // By default, Loggers print at zap's InfoLevel.
 func NewLogger(l *zap.Logger, options ...Option) *Logger {
 	logger := &Logger{
-		log:    l.Sugar(),
-		fatal:  (*zap.SugaredLogger).Fatal,
-		fatalf: (*zap.SugaredLogger).Fatalf,
-		print:  (*zap.SugaredLogger).Info,
-		printf: (*zap.SugaredLogger).Infof,
+		log:      l.Sugar(),
+		info:     (*zap.SugaredLogger).Info,
+		infof:    (*zap.SugaredLogger).Infof,
+		warning:  (*zap.SugaredLogger).Warn,
+		warningf: (*zap.SugaredLogger).Warnf,
+		err:      (*zap.SugaredLogger).Error,
+		errf:     (*zap.SugaredLogger).Errorf,
+		fatal:    (*zap.SugaredLogger).Fatal,
+		fatalf:   (*zap.SugaredLogger).Fatalf,
+		print:    (*zap.SugaredLogger).Info,
+		printf:   (*zap.SugaredLogger).Infof,
 	}
 	for _, option := range options {
 		option.apply(logger)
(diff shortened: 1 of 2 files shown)


zapgrpc/zapgrpc.go
```diff
@@ -48,38 +51,95 @@ func WithDebug() Option {
 // By default, Loggers print at zap's InfoLevel.
 func NewLogger(l *zap.Logger, options ...Option) *Logger {
 	logger := &Logger{
-		log:    l.Sugar(),
-		fatal:  (*zap.SugaredLogger).Fatal,
-		fatalf: (*zap.SugaredLogger).Fatalf,
-		print:  (*zap.SugaredLogger).Info,
-		printf: (*zap.SugaredLogger).Infof,
+		log:      l.Sugar(),
+		info:     (*zap.SugaredLogger).Info,
```
prashantv: is there a reason why we have to keep function pointers for each level, instead of just doing `l.sugar.Info(..)` etc?

frankgreco: @prashantv no reason. I was just trying to maintain the programming style that was previously present.

peter-edge:
Also see https://github.com/grpc/grpc-go/blob/master/grpclog/loggerv2.go.

It's not hard to implement, but it should be done in a thread-safe manner if you make V modifiable. glog does this.

zapgrpc/zapgrpc.go
peter-edge: @prashantv this is for the `WithDebug` function, which changes the zap function used for the `Print` functions.

prashantv: Ah, I see. In that case, aren't the only functions we need to do this for `print` and `printf`? The rest should just be normal function calls on the Logger.

peter-edge: That would work too, either/or, I was just being consistent, especially “back in the day” before this PR when there were only a few functions :-)

peter-edge:
@frankgreco any progresss on this?

kazegusuri:
I fixed the V implementation in my repo. Can I overtake this PR?

frankgreco:
Yea! Sorry I haven’t had time to complete this. I’d be awesome if you could provide the V implementation!

kazegusuri:
Thanks. I sent PR based on your commit, so please sign CLA.

AlekSi:
@frankgreco please sign CLA. That should take only a few minutes.

frankgreco:
@AlekSi 

It seems that I have already signed the CLA.
screen shot 2018-08-29 at 8 43 39 am

You can verify from the username in that image and the username from my commits in this PR that they are the same.

Maybe we should get a repo admin involved?

AlekSi:
Let's continue at #613.

frankgreco:
@AlekSi I think I fixed it :) I did an interactive rebase an modified the author/email for this commit to an email that was on my account (I don't have access to verify the original email as I no longer work for that company). As you will see, the CLA has been signed in this PR. Now, the reason it isn't signed in #613 is because that is off of @kazegusuri branch which does not have my modified commit. 

@kazegusuri can you update your branch so that you are using my modified commit? Then, we should all be good. 

Sorry that this was such a mess :(

ash2k:
Would be great to get this merged! @frankgreco Perhaps you could squash all commits and use the right email to commit it?

PR #548, opened 2018-01-10
Export Config.BuildEncoder

jcorbin:
Otherwise it's pretty challenging to build new cores when all you have ahold of is either a `*zap.Config` or a `*zap.Logger`.

### DIFF
--- a/config.go
+++ b/config.go
@@ -164,7 +164,7 @@ func NewDevelopmentConfig() Config {
 
 // Build constructs a logger from the Config and Options.
 func (cfg Config) Build(opts ...Option) (*Logger, error) {
-	enc, err := cfg.buildEncoder()
+	enc, err := cfg.BuildEncoder()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


jcorbin:
Right, okay so use case first for a second: I want to create a "session logger", that both writes to global logs (at whatever level that's enabled for), and also writes to a session-specific file, potentially at a more verbose level.

Currently the shortest path to get there is to make a logger into a tee the normal logger core, and into a newly/dynamically created, session-specific core.

I'd prefer to not force users to duplicate config for my widget, just to basically say "logs encoded like the rest of the application". In fact, it wouldn't even currently be possible to re-use just the encoder to config to support such a case.

akshayjshah:
I'd prefer not to simply expose `BuildEncoder`; if we end up needing to export something like this, it should be a method of `zapcore.EncoderConfig`. That will likely require moving the encoder constructor registry.

However, I'm not sure that this is required. If your users have already provided a `zap.Config` for the main (non-session-specific) logger, I think you should be able to accomplish your goals through the top-level config - you can copy it, drop the log level, and change the output file path, call `Build`, yank out the resulting `zapcore.Core`, and tee it into the main logger. Users should need to see the duplicate config. Would that work?

akshayjshah:
I'm going to assume that we've resolved this, since we can accomplish these goals through copying the config struct.

Re-open if necessary :)

PR #555, opened 2018-02-04
Use t.Run for some encoder tests

jkanywhere:
Wrap each encoder test case with t.Run for improved test output and
ability to selectively run specific sub tests.

### DIFF
--- a/zapcore/json_encoder_impl_test.go
+++ b/zapcore/json_encoder_impl_test.go
@@ -223,7 +223,9 @@ func TestJSONEncoderObjectFields(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		assertOutput(t, tt.desc, tt.expected, tt.f)
+		t.Run(tt.desc, func(t *testing.T) {
+			assertOutput(t, tt.desc, tt.expected, tt.f)
+		})
 	}
 }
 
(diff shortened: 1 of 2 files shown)


zapcore/json_encoder_impl_test.go
```diff
@@ -223,7 +223,9 @@ func TestJSONEncoderObjectFields(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		assertOutput(t, tt.desc, tt.expected, tt.f)
+		t.Run(tt.desc, func(t *testing.T) {
+			assertOutput(t, tt.desc, tt.expected, tt.f)
```
prashantv: do we need to pass `tt.desc` through and log them again on failures? it seems like just `t.Run(tt.desc, ...)` is enough.

jkanywhere: We don't need to pass `tt.desc`, I removed it.

jkanywhere:
Is there anything additional I need to do for this? I am by no means in a hurry, and I don't want to be the blocker.

akshayjshah:
Nope, you've done everything you need to @jkanywhere! I've just been really slow handling the backlog of issues and PRs for zap.

PR #565, opened 2018-03-26
zaptest: Expand TestingT interface

abhinav:
This expands the TestingT interface to a much larger subset of
`testing.TB`.

The following were left out:

- `Error` and `Log` were left out in favor of `Errorf` and `Logf`
- `Fatal*` methods were left out in favor of `Errorf` followed by
  `FailNow`
- `Skip*` methods were left out because our test logger shouldn't be
  skipping tests
- `Helper` was left out because not all supported verisons of Go have
  that

I'm putting this in a separate PR so that we can discuss every included/omitted
method outside #518.

CC @akshayjshah @prashantv

### DIFF
--- a/zaptest/testingt.go
+++ b/zaptest/testingt.go
@@ -29,6 +29,15 @@ type TestingT interface {
 	// Logs the given message and marks the test as failed.
 	Errorf(string, ...interface{})
 
+	// Marks the test as failed.
+	Fail()
+
+	// Returns true if the test has been marked as failed.
+	Failed() bool
+
+	// Returns the name of the test.
+	Name() string
+
 	// Marks the test as failed and stops execution of that test.
 	FailNow()
 }


zaptest/testingt.go
```diff
@@ -29,6 +29,15 @@ type TestingT interface {
 	// Logs the given message and marks the test as failed.
 	Errorf(string, ...interface{})
 
+	// Marks the test as failed.
+	Fail()
```
prashantv: do we need `Fail` when we already have `Logf` and `Errorf`?

abhinav: Fail marks the test as having failed without logging a message.
`Errorf("")` will still emit an empty string.

We don't necessarily need this but I can see its value if you want to log a
multi-line complex message before you mark the test as failed.

prashantv: I'm OK either way, if we have use-cases where we want to log a message separately to using `Errorf`, then we can keep it. If we haven't done this in our other projects, then perhaps we don't need it.

abhinav: How we mark tests as failed in normal tests in other projects isn't
that important here. This interface controls which methods are
available to our testing utility. If we decide to add more features to
the test logger in the future, it'll be difficult to request more of
the methods from `testing.TB`.

The intent here is to get the minimal standalone functionality of
testing.TB. I wouldn't mind dropping Errorf in favor of just Logf and
Fail but we want Errorf for testify compatibility.

FWIW, we *are* using Fail in the logger returned by zaptest.NewLogger:
We use it when the SyncWriter needs to mark the test as failed if
anything gets written to ErrorOutput. (#566)

abhinav:
Ping :)

PR #566, opened 2018-03-26
zaptest/NewLogger: fail test on internal errors

abhinav:
Zap's errorOutput is used exclusively to log internal errors. If this
ever gets used in a test, something catastrophic happened and we should
fail the test.

This changes the logger returned by zaptest.NewLogger to implement this
behavior by passing a `WriteSyncer` to `zap.ErrorOutput` that marks the
test as failed upon being used.

To test this, we set up a test logger and replace its core with one that
will always fail to write.

### DIFF
--- a/zaptest/logger_test.go
+++ b/zaptest/logger_test.go
@@ -29,12 +29,16 @@ import (
 	"testing"
 
 	"go.uber.org/zap"
+	"go.uber.org/zap/internal/ztest"
+	"go.uber.org/zap/zapcore"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestTestLogger(t *testing.T) {
 	ts := newTestLogSpy(t)
+	defer ts.AssertPassed()
+
 	log := NewLogger(ts)
 
 	log.Info("received work order")
@@ -57,6 +61,8 @@ func TestTestLogger(t *testing.T) {
 
 func TestTestLoggerSupportsLevels(t *testing.T) {
 	ts := newTestLogSpy(t)
+	defer ts.AssertPassed()
+
 	log := NewLogger(ts, Level(zap.WarnLevel))
 
 	log.Info("received work order")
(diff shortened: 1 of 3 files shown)


zaptest/logger_test.go
```diff
@@ -105,5 +153,9 @@ func (t *testLogSpy) Logf(format string, args ...interface{}) {
 }
 
 func (t *testLogSpy) AssertMessages(msgs ...string) {
-	assert.Equal(t, msgs, t.Messages)
+	assert.Equal(t.TB, msgs, t.Messages, "logged messages did not match")
+}
+
+func (t *testLogSpy) AssertFailed(v bool) {
```
prashantv: Can we make this an unexported method and have methods `AssertFailed()` and `AssertPassed()` rather than a boolean argument?

zaptest/logger_test.go
```diff
@@ -74,18 +80,61 @@ func TestTestLoggerSupportsLevels(t *testing.T) {
...
+		assert.Regexp(t, `write error: failed`, ts.Messages[0])
+	}
+}
+
 // testLogSpy is a testing.TB that captures logged messages.
 type testLogSpy struct {
 	testing.TB
 
-	mu       sync.Mutex
+	mu       sync.RWMutex
```
prashantv: optional/can follow up:
this spy is only used internally, and I don't think any of the tests are concurrent, so we might be able to just remove the mutex.

zaptest/logger_test.go
abhinav: Makes sense

zaptest/logger_test.go
abhinav: The spy is thread safe only because `testing.TB` is supposed to be thread-
safe, but yeah, I'll drop it.

abhinav: https://github.com/uber-go/zap/pull/567

Issue #553, opened 2018-01-27
Ability to have custom sinks


dimroc:
Hi all,
We need finer grained control of our console output. This led me to do some investigation around `OutputPaths` and `Open` in `writer.go`:

https://github.com/uber-go/zap/blob/master/writer.go#L60

```golang
	for _, path := range paths {
		switch path {
		case "stdout":
			writers = append(writers, os.Stdout)
			// Don't close standard out.
			continue
		case "stderr":
                        ...
```

After perusing some other PRs and issues, I noticed that there's growing consensus around the need for "syslog", which ended up with this draft adding `case "syslog"` to the switch statement.

https://github.com/uber-go/zap/pull/506/files#diff-dc03dbc53655fac7a4c8d16b1b4144c1R70

Rather than adding string after string in the case statement, we could introduce an interface that returns an `zapcore.WriteSyncer` and a `close` method. That way we could register new sinks like syslog or my customized `Stdout`, and not increase the number of special cases.

Quick outside-in proposal:

```golang
func (cfg Config) BuildWithSinkFactories(sinkMakers map[string]SinkFactory, opts ...Option) (*Logger, error) {
   ...
}

func (cfg Config) Build(opts ...Option) (*Logger, error) {
    BuildWithSinkFactories(DefaultSinkFactories, opts)
}

...

func Open(sinkMakers map[string]SinkFactory, ...) {
        sinks := make([]Sink, 0, len(paths))
	for _, path := range paths {
            factory, ok := sinkMakers[path]
            if ok {
                // sink{writer, closer}
                sinks = append(writers, sink)
            }
            ...
```

Any other ideas or approaches around this would be great.

---

There could be a discussion around me finding a place outside of zap to customize my stdout, but for this issue, I would like to focus on the viability of customizing sinks in `writer.go` or extending zap.

etsangsplk:
@akshayjshah or other reviewers
What is the status on this? And on the associated PR?

akshayjshah:
landed, cutting a release shortly. Sorry for the long delay!

PR #572, opened 2018-04-10
Ability to build logger with custom Sink

dimroc:
Introduce the ability to pass in `Sink`s enabling one to customize a logger's `Writer` (technically `WriteSyncer`) and `Closer`.

* Completes issue https://github.com/uber-go/zap/issues/553
* Allows clean up of https://github.com/uber-go/zap/pull/506/files#diff-dc03dbc53655fac7a4c8d16b1b4144c1R85 by allowing one to place `syslog` logic like `syslog.Dial` outside of `writer.go#Open` (https://github.com/uber-go/zap/pull/506)
* Enables the creation of future sinks without the need to tamper with `writer.go#Open`

Sample usage:

```go
func CreateLogger() *zap.Logger {
	config := zap.NewProductionConfig()
	file := path.Join("/tmp", "log.jsonl")

        RegisterSink("customsink",  func() (Sink, error) { return CustomSink{}, nil })
	config.OutputPaths = []string{"stdout", "customsink", file}
	zl, err := config.Build()

	if err != nil {
		log.Fatal(err)
	}
	return zl
}

// Complies with type Sink interface (zapcore.WriteSyncer, io.Closer)
type CustomSink struct{}

func (CustomSink) Sync() error  { return nil }
func (CustomSink) Close() error { return nil }

func (CustomSink) Write(p []byte) (int, error) {
	log.Print(string(p)) // send to syslog, or do anything
	return len(p), nil
}
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -22,7 +22,6 @@ package zap
 
 import (
 	"io"
-	"io/ioutil"
 	"os"
 	"sort"
 	"time"
@@ -167,19 +166,20 @@ func NewDevelopmentConfig() Config {
 
 // Build constructs a logger from the Config and Options.
 func (cfg Config) Build(opts ...Option) (*Logger, error) {
-	return cfg.BuildWithSinks(DefaultSinkFactories(), opts...)
+	return cfg.BuildWithSinks(DefaultSinks(), opts...)
 }
 
-// BuildWithSinks uses the map of sink factories to constrct a logger from the
+// BuildWithSinks uses the map of sinks to construct a logger from the
 // passed options. For example, the option "stdout" will construct a logger
-// that routes output to the stdout sink.
-func (cfg Config) BuildWithSinks(sf map[string]SinkFactory, opts ...Option) (*Logger, error) {
+// that routes output to the stdout sink, but "/filepath" will fallback
+// to writing to file.
+func (cfg Config) BuildWithSinks(sm map[string]Sink, opts ...Option) (*Logger, error) {
 	enc, err := cfg.buildEncoder()
 	if err != nil {
 		return nil, err
 	}
 
-	sink, errSink, err := cfg.openSinks(sf)
+	sink, errSink, err := cfg.openSinks(sm)
 	if err != nil {
 		return nil, err
 	}
@@ -235,12 +235,12 @@ func (cfg Config) buildOptions(errSink zapcore.WriteSyncer) []Option {
 	return opts
 }
 
-func (cfg Config) openSinks(sf map[string]SinkFactory) (zapcore.WriteSyncer, zapcore.WriteSyncer, error) {
-	sink, closeOut, err := Open(sf, cfg.OutputPaths...)
+func (cfg Config) openSinks(sm map[string]Sink) (zapcore.WriteSyncer, zapcore.WriteSyncer, error) {
+	sink, closeOut, err := OpenWithSinks(sm, cfg.OutputPaths...)
 	if err != nil {
 		return nil, nil, err
 	}
-	errSink, _, err := Open(sf, cfg.ErrorOutputPaths...)
+	errSink, _, err := OpenWithSinks(sm, cfg.ErrorOutputPaths...)
 	if err != nil {
 		closeOut()
 		return nil, nil, err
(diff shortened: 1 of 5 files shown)


dimroc:
The core of the change can be found here https://github.com/uber-go/zap/pull/572/files#diff-dc03dbc53655fac7a4c8d16b1b4144c1L59

config.go
```diff
@@ -164,12 +166,20 @@ func NewDevelopmentConfig() Config {
...
 // Build constructs a logger from the Config and Options.
 func (cfg Config) Build(opts ...Option) (*Logger, error) {
+	return cfg.BuildWithSinks(DefaultSinks(), opts...)
+}
+
+// BuildWithSinks uses the map of sinks to construct a logger from the
+// passed options. For example, the option "stdout" will construct a logger
+// that routes output to the stdout sink, but "/filepath" will fallback
+// to writing to file.
+func (cfg Config) BuildWithSinks(sm map[string]Sink, opts ...Option) (*Logger, error) {
```
akshayjshah: This is a nice idea - I'd like to make this portion of zap more extensible. I also like that this approach avoids adding global state.

However, this works very differently from extending encoders: for that, we use a global registry and a top-level `RegisterEncoder` function. For consistency, I'd actually prefer to add another global registry for sinks. (Generally, I much prefer dependency injection, but I think consistency is more important here.)

@prashant and @abhinav, what do you think?

prashantv: I think I'd prefer consistency as well here, and have a global registry for both encoders and sinks.

If we want users to be able to specify the exact registries, we'll likely want to add a single method that accepts both registries rather than a method for each.

dimroc: Thanks for the feedback @akshayjshah @prashantv. I never saw `RegisterEncoder` but after seeing it, your requests make complete sense. I have since added a commit with `RegisterSink` to mimc that functionality. If @abhinav feels strongly about reverting back to the dependency injection style, no worries, we can just revert one commit.

writer.go
```diff
@@ -49,28 +50,26 @@ func Open(paths ...string) (zapcore.WriteSyncer, func(), error) {
...
 		}
 	}
 	for _, path := range paths {
-		switch path {
-		case "stdout":
-			writers = append(writers, os.Stdout)
-			// Don't close standard out.
+		if sink, err := NewSink(path); err == nil {
+			writers = append(writers, sink)
+			closers = append(closers, sink)
```
dimroc: I kept the two lists, `writers` and `closers`, because I didn't want to be too aggressive with the refactor and change the return value of 
`func open(paths []string) ([]zapcore.WriteSyncer, func(), error)`
to
`func open(paths []string) ([]Sink, func(), error)`

It would have rippled out to affect a few other funcs and we would have needed an array of `[]zapcore.WriteSyncer` for `CombineWriteSyncers` anyways.

etsangsplk:
@akshayjshah 
What's the decision??

rbastic:
I'd also like to see this merged. Any update?

se3000:
Would love to see this merged, it'd be very helpful for us. What's left to be done?

akshayjshah:
This is great. I'm going to push a commit tomorrow to un-export a few of the helper structs, but the core of this is excellent - thanks for the work and the refactoring, @dimroc.

To everyone waiting on this, thanks for your patience. Work and fatherhood distracted me for the past few months.

Expect a release with this functionality by the end of the week.

akshayjshah:
Merged! I'm going to open PRs to prep a release tomorrow. We use this package extensively internally; for my own sanity and family life, I'm not going to cut a release on a Friday just in case something goes horribly wrong. I'll tag a new relase first thing monday morning.

sink.go
```diff
@@ -0,0 +1,94 @@
...
+	_sinkMutex     sync.RWMutex
+	_sinkFactories map[string]func() (Sink, error)
+)
+
+func init() {
+	resetSinkRegistry()
+}
+
+func resetSinkRegistry() {
+	_sinkMutex.Lock()
```
prashantv: Should we have this mutex? It seems like we'd be masking a race where loggers might fail to build because the scheme isn't recognized by the sink?

sink.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Sink defines the interface to write to and close logger destinations.
+type Sink interface {
+	zapcore.WriteSyncer
+	io.Closer
+}
+
+// RegisterSink adds a Sink at the given key so it can be referenced
+// in config OutputPaths.
+func RegisterSink(key string, sinkFactory func() (Sink, error)) error {
```
prashantv: This will work well for static sinks, but I think we might need something more dynamic (e.g., a prefix match, or a URI scheme match). E.g., even for something like syslog, we'd probably need more info like port. I'd prefer to have less APIs exposed, so one that is more flexible might be better

sink.go
```diff
@@ -0,0 +1,94 @@
...
+	_sinkMutex.Lock()
+	defer _sinkMutex.Unlock()
```
prashantv: optional nit: blank after `defer` to separate it from the rest of the logic (same for `newSink`)

sink.go
akshayjshah: Certainly possible. We lock around the encoder registry; IMO, consistency with that API is important here.

sink.go
akshayjshah: :+1: will do.

sink.go
akshayjshah: That's a good point - I hadn't thought about this use case.

I'd prefer to avoid a URI-based scheme here, since the current special-cased sinks don't use `file://`. Prefix match would would, as long as we ensure that we don't get any conflicting rules.

akshayjshah: Actually, URIs with special cases for `stdout` and `stderr` might be best :thinking:

akshayjshah: https://github.com/uber-go/zap/pull/606

DGollings:
I'm either the first or the least intelligent person to attempt to use this custom sink stuff. Took me quite some time and digging through issues/code to figure this out.  To anyone wondering why the example doesn't work, it has changed slightly due to #606:

```
func CreateLogger() *zap.Logger {
	config := zap.NewProductionConfig()
	file := path.Join("/tmp", "log.jsonl")

        RegisterSink("customsink://",  func() (Sink, error) { return CustomSink{}, nil })
	config.OutputPaths = []string{"stdout", "customsink", file}
	zl, err := config.Build()

	if err != nil {
		log.Fatal(err)
	}
	return zl
}

// Complies with type Sink interface (zapcore.WriteSyncer, io.Closer)
type CustomSink struct{}

func (CustomSink) Sync() error  { return nil }
func (CustomSink) Close() error { return nil }

func (CustomSink) Write(p []byte) (int, error) {
	log.Print(string(p)) // send to syslog, or do anything
	return len(p), nil
}
```

`RegisterSink("customsink",  func() (Sink, error) { return CustomSink{}, nil })`
should be
`RegisterSink("customsink://",  func() (Sink, error) { return CustomSink{}, nil })`

PR #573, opened 2018-04-11
Add Go import comments.

functionary:
This pull request was made with an automated tool.

The suggested change adds an import comment to a file in each package
to ensure that the package is always imported using its canonical
import path. With this comment in place, the go tool guarantees
that the package can be imported only from this path.

For more information, see https://golang.org/s/go14customimport,
in particular the "Proposal" section.


akshayjshah:
Can't accept this without a signed CLA.

PR #600, opened 2018-06-18
Add Quoted Field Support

jcorbin:
This adds a path to fast sub-encoded data (i.e. logging a field whose value is quoted JSON) without forcing the log site to either:
- drive a stripped down copy of a zapcore json encoder targeted at a temporary byte buffer
- fallback to using `buf, _ := json.Marshal(thing); zap.ByteString("thing", buf)`

I added an example test first, for PoC demonstration, and minimal testing.

However this PR, at present, is just a start, things still TODO:
- [ ] implement memory encoder "quoting" (whatever that even means) so that tests can have parity
- [ ] consider whether we want to fail encoding when quoting is not supported, or instead fallback to a JSON string (see TOOD note inline below)
- [ ] write more (unit) tests, if we can agree to add this feature

Note how I've gone the optional interface upgrade route with `QuotedObjectEncoder` ; if this feature gets accepted, we should add a point the 2.x wish list to fold this in as an `ObjectEncoder` requirement.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -59,10 +59,32 @@ func Example_presets() {
 		zap.Int("attempt", 3),
 		zap.Duration("backoff", time.Second),
 	)
+
+	dat := struct {
+		Foo string            `json:"foo"`
+		Bar float64           `json:"bar"`
+		Baz map[string]string `json:"baz"`
+	}{
+		Foo: "now is the time for string-\"ception\"!",
+		Bar: 3.14,
+		Baz: map[string]string{
+			"such": "octopus 🐙", // TODO why doesn't this encode as \u0001f419
+		},
+	}
+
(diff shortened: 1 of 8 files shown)


akshayjshah:
@jcorbin Will https://github.com/thriftrw/thriftrw-go/pull/366 resolve the performance issues we're seeing internally?

jcorbin:
Well, that won't really affect any case of "the user really wanted to log a quoted sub-document (whatever that means for a given encoding)"; they're really separate topics, although found adjacently in at least one instance.

jcorbin:
Leaving for another day is fine.

akshayjshah:
Ok - let's circle back if this continues to come up.

PR #602, opened 2018-06-18
Pool the buffer and encoder used for generic JSON reflection

jcorbin:


zapcore/json_encoder_test.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			buf, err := enc.EncodeEntry(tt.ent, tt.fields)
+			if assert.NoError(t, err, "Unexpected JSON encoding error.") {
+				assert.JSONEq(t, tt.expected, buf.String(), "Incorrect encoded JSON entry.")
+			}
+			buf.Free()
+		})
+	}
+}
```
akshayjshah: Similar tests already exist in the top-level zap package; would we be better off changing the Makefile to pass `coverpkg=all` instead of adding a new test suite?

jcorbin:
Hmm, that `-coverpkg=all` trick didn't get me a greenlight from the bot; should I revert back to the new test?

zapcore/json_encoder_test.go
akshayjshah: For future reviewers: this turned into a yak shave, will fix it later.

PR #603, opened 2018-06-26
Simplify coverage calculation

akshayjshah:
We only use the current `cover.sh` script on Travis, where we'd like coverage
tracked. We don't need to calculate coverage twice; in fact, we'd prefer to
calculate coverage only on the most recent supported Go release. Now that Go
1.10 supports coverage reports across multiple packages, we can remove
`cover.sh` entirely and amend the Makefile a bit. Along the way, we enable
cross-package coverage calculation.


akshayjshah:
Ugh, this isn't going to help either.

PR #605, opened 2018-06-28
Prepare v1.9.0

akshayjshah:
Includes some perf optimization when using `zap.Reflect` and some exposed API for user-supplied log sinks.


prashantv:
Left some minor comments related to the new API (https://godoc.org/go.uber.org/zap#RegisterSink).

If we're fine with that API, then this lgtm.

akshayjshah:
Believe I've addressed feedback re: user-defined log sinks.

Issue #609, opened 2018-07-17
zaptest: Add possibility to pass zap.Option's 


iaroslav-ciupin:
## Synopsis

`zaptest.NewLogger` gives you possibility to set logging level using `zaptest.Level(enab zapcore.LevelEnabler)`. It would be great to optionally add caller to test logger using something like: 
```go
logger := zaptest.NewLogger(t, zaptest.WrapOptions(zap.AddLogger(), ...))
```
This should result in logging caller like so:
```go
...
42: log.Info("received work order") 
43: // logs "INFO    zaptest/logger_test.go:42    received work order"
```

Pull request with suggested change: https://github.com/uber-go/zap/pull/610

iaroslav-ciupin:
Hi @akshayjshah ! Can you please have a look at this proposal(and PR with corresponding implementation) ? Thanks in advance.

PR #610, opened 2018-07-17
Add zaptest logger option to wrap zap.Option's

iaroslav-ciupin:
## Synopsis

This PR adds feature suggested in https://github.com/uber-go/zap/issues/609


iaroslav-ciupin:
@akshayjshah Thank you for your reply I've adjusted PR and proposal accordingly. Please have a look ;)

PR #614, opened 2018-07-30
MapObjectEncoder: Empty arrays should not be nil

mh-park:
This change fixes a bug in MapObjectEncoder.AddArray where an empty slice would log nil instead of an empty array. Logging an empty array makes the MapObjectEncoder's behavior correctly reflect the JSON encoder's.


zapcore/memory_encoder.go
```diff
@@ -45,6 +45,9 @@ func NewMapObjectEncoder() *MapObjectEncoder {
 func (m *MapObjectEncoder) AddArray(key string, v ArrayMarshaler) error {
 	arr := &sliceArrayEncoder{}
 	err := v.MarshalLogArray(arr)
+	if arr.elems == nil {
+		arr.elems = []interface{}{}
```
abhinav: I would make this the default value of `arr.elems` since it'll be appended to anyway.

```
arr := &sliceArrayEncoder{elems: make([]interface{})}
```

zapcore/memory_encoder.go
```diff
@@ -43,7 +43,7 @@ func NewMapObjectEncoder() *MapObjectEncoder {
 
 // AddArray implements ObjectEncoder.
 func (m *MapObjectEncoder) AddArray(key string, v ArrayMarshaler) error {
-	arr := &sliceArrayEncoder{}
+	arr := &sliceArrayEncoder{elems: make([]interface{}, 0)}
```
abhinav: 0 unnecessary in `make(..)` because it's the default.

mh-park: I initially tried it without the 0 and `zapcore/memory_encoder.go:46:39: missing len argument to make([]interface {})` popped up

abhinav: Derp, you're right.

Issue #621, opened 2018-08-13
Windows paths not supported by default file sink


Apeopex:
When using absolute Windows paths, url.Parse does not retreieve the "file" scheme.

`panic: couldn't open sink "C:\\Users\\...\\AppData\\Roaming\\...\\info.log.json": no sink found for scheme "c"`

Appending, "file:///" (https://github.com/golang/go/issues/6027#issuecomment-66083310) does not work as expected. (Ref: https://github.com/golang/go/issues/13276 )

https://github.com/uber-go/zap/blob/ff33455a0e382e8a81d14dd7c922020b6b5e7982/sink.go#L99

Ref: https://github.com/golang/go/issues/13276#issuecomment-157074191
Changed by PR: #606 

Temporary workaround:

```
func newWinFileSink(u *url.URL) (zap.Sink, error) {
    // Remove leading slash left by url.Parse()
    return os.OpenFile(u.Path[1:], os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0644)
}
zap.RegisterSink("winfile", newWinFileSink)
zap.Config{
    OutputPaths: []string{"stdout", "winfile:///" + filepath.Join(GlobalConfigDir.Path, "info.log.json")},
}
```

wenzhihong2003:
yes, in windows `file:///d:/a/wen88.log` is not supported. Because, in this scene, url.Path is `/d:/a/wen88.log`, in windows you can't create file start with `/`,  so in windows, you will strip the `/`.

wenzhihong2003:
I have pull a request in https://github.com/uber-go/zap/pull/624

prashantv:
The reason `file:///d:/a/wen88.log` doesn't work is that the path is `/d:/a/...`, can you replace the 3 `/` with 2 `/`, e.g., `file://d:/a/wen88.log`?

Apeopex:
I did try that and ran into the same issue listed here: https://github.com/golang/go/issues/13276

wenzhihong2003:
@prashantv , if use `file://d:/a/wen88.log` , it will paic with message 
```
couldn't open sink "file://d:/a/wen88.log": file URLs must leave host empty or use localhost: got file://d:/a/wen88.log
```

prashantv:
That makes sense, thanks. It looks like it has to be 3 slashes, and is covered here:
https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/

[...]

PR #624, opened 2018-08-20
fix in windows file path can't support

wenzhihong2003:
fix bug #621

### DIFF
--- a/sink.go
+++ b/sink.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"runtime"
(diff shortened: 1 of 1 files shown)


selslack:
@prashantv 

Then it'll fail with error: `can't parse "file://C:\\Users\\..." as a URL: parse file://C:\Use
rs\...: invalid character "\\" in host name`. See https://github.com/uber-go/zap/issues/621#issuecomment-415310700 as well as https://github.com/golang/go/issues/13276.


Fix from #621 works fine.

PR #628, opened 2018-08-28
Fix misleading tables

deanveloper:
The tables on the main page currently _look_ like they are sorted by `ns/op`, but are instead sorted by `ns/op but put zap on top no matter what`, which is a bit misleading.

This pull request sorts the tables by `ns/op`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,9 @@ Log a message with a logger that already has 10 fields of context:
 
 | Package | Time | Objects Allocated |
 | :--- | :---: | :---: |
+| zerolog | 321 ns/op | 0 allocs/op |
 | :zap: zap | 380 ns/op | 0 allocs/op |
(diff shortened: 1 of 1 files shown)


jcorbin:
So performance comparison tables like this frequently put the "primary subject under comparison" on top, and then frame everything relative to that reference. However that usually also goes along with focusing primarily on relative (percent) difference, and such header is usually marked more clearly to make the frame of reference obvious.

I'm not sure what the right answer is here, and will of course defer to @akshayjshah.

akshayjshah:
First, a low-level issue: the readme in this project is mechanically generated (since it's really irritating to hand-update markdown each time we run the benchmarks), so making this change requires changing `internal/readme/readme.go`.

More broadly, I'd prefer to keep zap on top the top of these tables. As Josh mentions, this is common - the primary subject remains on top. Adding percent-wise comparisons would be great, though.

PR #629, opened 2018-08-31
Added zapcore.TimeEncoderOfLayout

tmshn:
I added `zapcore.TimeEncoderOfFormat` to make it easy to build `TimeEncoder` of custom format.

You can use it in code:

```go
config := zap.NewDevelopmentConfig()
config.EncoderConfig.EncodeTime = zapcore.TimeEncoderOfLayout("06/01/02 03:04pm")
```

or, with JSON:

```json
{
  "encoderConfig": {
    "timeEncoder": {
       "layout": "06/01/02 03:04pm"
     }
  }
}
```

or, with YAML:
```yaml
encoderConfig:
  timeEncoder:
    layout: '06/01/02 03:04pm'
```


zapcore/encoder.go
```diff
@@ -151,15 +152,25 @@ func RFC3339NanoTimeEncoder(t time.Time, enc PrimitiveArrayEncoder) {
 	encodeTimeLayout(t, time.RFC3339Nano, enc)
 }
 
+// TimeEncoderOfFormat returns TimeEncoder which serializes a time.Time using
+// given format.
+func TimeEncoderOfFormat(format string) TimeEncoder {
+	return func(t time.Time, enc PrimitiveArrayEncoder) {
```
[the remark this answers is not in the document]
tmshn: ✅ 7de11fd

tmshn:
Actually I'm wondering which word to use, "format" or "layout"...

zapcore/encoder.go
```diff
@@ -170,6 +181,8 @@ func (e *TimeEncoder) UnmarshalText(text []byte) error {
 		*e = EpochMillisTimeEncoder
 	case "nanos":
 		*e = EpochNanosTimeEncoder
+	case "format=":
+		*e = TimeEncoderOfFormat(vals[1])
```
abhinav, line 185: I see the concern about the `=` splitting thing. This could become a problem
if we want more parameters in the future, or if escaping parameters becomes
necessary.

Perhaps we can accept objects for when the format is specified:

```
timeEncoder: rfc3339
# or
timeEncoder:
  format: "06/01/02 03:04pm"
```

We can accomplish this by implementing [`yaml.Unmarshaler`][2] to decode into
an object, and if that fails, to fall back to UnmarshalText.

  [1]: https://golang.org/pkg/encoding/json/#Unmarshaler
  [2]: https://godoc.org/gopkg.in/yaml.v2#Unmarshaler

Something like the following should suffice:

    func (e *TimeEncoder) UnmarshalYAML(unmarshal func(interface{}) error) error {
      # timeEncoder:
      #   format: "..."
      var o struct {
        Format string `yaml:"format" json:"format"`
      }
      if err := unmarshal(&o); err == nil {
        *e = TimeEncoderOfFormat(o.Format)
        return nil
      }
      
      # timeEncoder: "..."
      var s string
      if err := unmarshal(&s); err != nil {
        return err
      }
      return e.UnmarshalText([]byte(s))
    }

We can even reuse the `UnmarshalYAML` method for JSON support:

    func (e *TimeEncoder) UnmarshalJSON(bs []byte) error {
      return e.UnmarshalYAML(func(v interface{}) error {
        return json.Unmarshal(bs, v)
      })
    }

prashantv:
> Actually I'm wondering which word to use, "format" or "layout"...

Good question, since Go tends to use the word layout for the argument, it makes sense to use that in the YAML config.

zapcore/encoder.go
tmshn: Seems more expressive!
Fixed at 0c345e2

Also, modified tests so that it explicitly uses yaml.Unmarshal (5f2bb49)

tmshn:
Don't know why codecov fails...

zapcore/encoder.go
```diff
@@ -176,6 +185,35 @@ func (e *TimeEncoder) UnmarshalText(text []byte) error {
...
+		Layout string `json:"layout" yaml:"layout"`
+	}
+	if err := unmarshal(&o); err == nil {
+		*e = TimeEncoderOfLayout(o.Layout)
+		return nil
+	}
+
+	var s string
+	if err := unmarshal(&s); err != nil {
+		return err
```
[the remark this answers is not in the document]
tmshn: Will fix!
BTW, the diff coverage should be 100% to get merged?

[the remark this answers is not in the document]
tmshn: ✅ d907620

> the diff coverage should be 100% to get merged?

Oh I noticed that this "target" value might be the required boarder line, right?

codecov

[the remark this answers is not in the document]
prashantv: Yes, the target is the value above which the coverage check passes.

zapcore/encoder_test.go
```diff
@@ -21,10 +21,13 @@
 package zapcore_test
 
 import (
+	"encoding/json"
 	"strings"
 	"testing"
 	"time"
 
+	"gopkg.in/yaml.v2"
+
```
prashantv, line 30: nit: delete this blank line between third party imports

abhinav, line 30:
```suggestion
```

zapcore/encoder_test.go
```diff
@@ -523,29 +526,65 @@ func TestLevelEncoders(t *testing.T) {
...
+		"timeEncoder: [1, 2, 3]", // wrong type
+		"timeEncoder: {foo:bar",  // broken yaml
+	}
+	for _, tt := range tests {
+		cfg := EncoderConfig{}
+		assert.Error(t, yaml.Unmarshal([]byte(tt), &cfg), "Expected unmarshaling %q to become error, but not.", tt)
+	}
+}
+
+func TestTimeEncodersParseFromJson(t *testing.T) {
```
abhinav, line 569: s/Json/JSON
```suggestion
func TestTimeEncodersParseFromJSON(t *testing.T) {
```

zapcore/encoder_test.go
tmshn: ✅ c6de7e9c783b51707bc7962284be5f08d86883f6

tmshn: ✅ c6de7e9c783b51707bc7962284be5f08d86883f6

PR #633, opened 2018-09-07
Allow thereafter in sampler to be zero

gunnihinn:
When creating a new Sampler, there is nothing stopping us from passing
in 0 as a value for thereafter. One might expect this to be legal -- as
the documentation doesn't say thereafter should be greater than zero and
no error is returned when passing in this configuration -- and to mean
"drop all messages after the first n". What actually happens is a

    panic: runtime error: integer divide by zero

when we try to log a message.

This patch avoids the division by zero and enforces the behavior
described above.

This fixes issue 632.

### DIFF
--- a/zapcore/sampler.go
+++ b/zapcore/sampler.go
@@ -127,7 +127,7 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
 
 	counter := s.counts.get(ent.Level, ent.Message)
 	n := counter.IncCheckReset(ent.Time, s.tick)
-	if n > s.first && (n-s.first)%s.thereafter != 0 {
+	if n > s.first && (s.thereafter == 0 || (n-s.first)%s.thereafter != 0) {
 		return ce
 	}
 	return s.Core.Check(ent, ce)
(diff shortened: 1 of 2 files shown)


prashantv (commented):
Thanks for the contribution @gunnihinn

Rather than add more logic to the sampler's counting logic, we could simplify this (and make the caller's code more efficient) by detecting this early, and returning the passed in core.

E.g., `NewSampler` could check if `thereafter == 0`, and if so, return the passed in `core` directly, since using `0` means we're disabling sampling.

gunnihinn:
> since using 0 means we're disabling sampling.

Did I miss where we can choose to log only the first `first` messages?

prashantv (approved):
Ah, I misunderstood the change -- I thought we were changing `thereafter: 0` to mean that we'd log everything. However, that's the `thereafter == 1` case, where we could do the optimization (which doesn't need to be a part o this change).

akshayjshah (approved):
This looks great - thank you, @gunnihinn!

justinhwang:
Superceded by #1033

Issue #635, opened 2018-09-12
Implement gokit/log's Logger interface in the SugaredLogger


alex-laties:
I'd like to extend the [`SugaredLogger`](https://godoc.org/go.uber.org/zap#SugaredLogger) to implement [go-kit/log](https://github.com/go-kit/kit/tree/master/log)'s [`log.Logger`](https://godoc.org/github.com/go-kit/kit/log#Logger) interface.

This would make `zap` easier to integrate with codebases that already use [`go-kit/log`](https://github.com/go-kit/kit/tree/master/log) and simplify migrating to `zap` down the road.

[`SugaredLogger`](https://godoc.org/go.uber.org/zap#SugaredLogger) seems like the correct place to implement this as there is already an expectation around loosely typed, but structured, logging.

alex-laties:
https://github.com/alex-laties/gokitzap does the thing

PR #636, opened 2018-09-12
implement gokit logger interface to SugaredLogger

alex-laties:
Implements the proposal in #635

### DIFF
--- a/sugar.go
+++ b/sugar.go
@@ -92,6 +92,14 @@ func (s *SugaredLogger) With(args ...interface{}) *SugaredLogger {
 	return &SugaredLogger{base: s.base.With(s.sweetenFields(args)...)}
 }
 
+// Log implements the gokit/log.Logger interface.
+// Log acts almost identically to Infow, except the first argument is _not_ a message.
+// Log will always return nil, as there is no error condition to check currently
+func (s *SugaredLogger) Log(keysAndValues ...interface{}) error {
+	s.log(InfoLevel, "", nil, keysAndValues)
+	return nil
+}
+
 // Debug uses fmt.Sprint to construct and log a message.
 func (s *SugaredLogger) Debug(args ...interface{}) {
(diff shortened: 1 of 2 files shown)


alex-laties:
um... I've signed the CLA... how do I get that CLAssistant bot to post that here?

abhinav:
I see a couple issues with putting this directly on SugaredLogger:

- Log is a very generic name. For people using Zap directly, it adds no value.
  If this adapter was provided by Zap, it would probably need to be a function
  similar `ToGoKit(*Logger) *GoKitLogger`.
- The empty message is a bit of a problem. I imagine codebases have some
  convention around which field in their go-kit `Log` is the message, i.e.
  `log.Log("msg", "did something", [other keys])`. Ideally, the Zap-based
  implementation of `go-kit/log.Logger` is aware of this key and pulls the
  key-value pair out of `keyvals` if found. Further, because the key probably
  varies between projects, this would need to be configurable. This is quite a
  lot of logic that definitely does not belong on the SugaredLogger itself,
  and possibly not in Zap at all.

My recommendation is to create a separate standalone library out of this
functionality. That avoids zap or go-kit taking dependencies on each other and
lets you version that library separately.

alex-laties:
@abhinav both are very sound points.

Regarding the message field in go-kit... there really is no standard, so `msg, message`, and any other variation of letters may be valid. Really depends on the org. I absolutely agree that the logic to extract and uplift those things should be outside of the `SugaredLogger`.

I think you're recommendation to create a bridging library is a good one.
I'll try it out.

Thanks!

bjm88:
@alex-laties hi - did you ever do that?  could share the link?

we having this compile issue

	*zap.SugaredLogger does not implement logger.Interface (missing Fatalln method)

just want to pass the logger to another library

alex-laties:
@bjm88 yep! https://github.com/alex-laties/gokitzap.

Issue #645, opened 2018-10-20
log-writing blocks forever if `EncodeTime` is unset


Alex-liutao:
if i'm not set the EncodeTime explicitly when initialize my logger, I will get blocked on any log-writing function, such as `log.Info()`

my use case is: 

```go
        logFilePath := "/path/to/log/file"
	rawJSON := []byte(`{
	  "level": "debug",
	  "encoding": "json",
	  "outputPaths": ["stdout", "` + logFilePath + `"],
	  "errorOutputPaths": ["stderr", "` + logFilePath + `"],
	  "encoderConfig": {
	    "messageKey": "message",
	    "levelKey": "level",
	    "levelEncoder": "lowercase"
	  }
	}`)

	var cfg zap.Config
	if err := json.Unmarshal(rawJSON, &cfg); err != nil {
		panic(err)
	}

	cfg.EncoderConfig.TimeKey = "ts" // I set "TimeKey" but don't set "EncodeTime"
	// cfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder

	logger, err := cfg.Build()
	if err != nil {
		panic(err)
	}
	defer logger.Sync()
``` 

I did some debug and looks like it blocks at `jsonEncoder::EncodeTime()` function.

the callstack is as following:
```shell
go.uber.org/zap/zapcore.(*jsonEncoder).AppendTime at json_encoder.go:256
go.uber.org/zap/zapcore.(*jsonEncoder).AddTime at json_encoder.go:170
go.uber.org/zap/zapcore.(*jsonEncoder).EncodeEntry at json_encoder.go:323
go.uber.org/zap/zapcore.(*ioCore).Write at core.go:86
go.uber.org/zap/zapcore.(*CheckedEntry).Write at entry.go:215
go.uber.org/zap.(*Logger).Info at logger.go:187
```

Alex-liutao:
is this expected behavior？ if it's not, I think I can fix this one

Alex-liutao:
I think is the framework I'm using that blocks.
The proper description of this issue is that it will case a `nil pointer dereference` panic when user set the `TimeKey` and did not set the `EncodeTime`.

kevin-ip-incorta:
@Alex-liutao Adding "timeKey" and "timeEncoder" to the rawJSON works for me:

```
rawJSON := []byte(`{
	  "level": "debug",
	  "encoding": "json",
	  "outputPaths": ["stdout", "` + logFilePath + `"],
[...]

PR #646, opened 2018-10-20
skip timestamp if EncodeTime is nil (#645)

Alex-liutao:
This PR fix the issue #645 
Program will panic when user set the `TimeKey` and did not set `EncodeTime` explicitly.
I've noticed that in `consoleEncoder` it will skip timestamp if EncodeTime is nil, so I add this into `jsonEncoder` to ensure behavioral consistency.


abhinav:
Superseded by #791

PR #664, opened 2018-12-31
Only set entry time when it will be logged

pjvds:
This pull request optimizes the Check scenario. It only sets the entry time after the decision has been made to log the entry or not. This improves performance in a disabled scenario by 63.55%.

```
benchmark                                           old ns/op     new ns/op     delta
BenchmarkDisabledAccumulatedContext/Zap.Check-4     20.3          7.40          -63.55%
```

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -262,7 +262,6 @@ func (log *Logger) check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
 	// log message will actually be written somewhere.
 	ent := zapcore.Entry{
 		LoggerName: log.name,
-		Time:       time.Now(),
 		Level:      lvl,
 		Message:    msg,
 	}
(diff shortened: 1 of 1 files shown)


fgrosse:
We also noticed that recording the time in high throughput scenarios was a bottleneck, basically because we had a lot of debug messages. However I'm afraid you cannot merge this change without breaking compatibility because cores such as the sampling core rely on the time information in order to check if the log entry should be printed or not. Additionally there might be other hooked cores which work in a similar way.

What we ended up doing was to wrap the debug log calls in a function which only checks the level and aborts early if possible. Of course that means that we cannot use any sampling based logging there.

pjvds:
Is there a reason why we don't remove the sampling state to the sampling core itself, or make it not time depending but just a percentage of the logs with a atomic counter? This would mean you only pay the overhead for sampling support if the sampling core is enabled.

Issue #388, opened 2017-03-23
2.0 Wishlist


akshayjshah:
This is a catch-all issue for our 2.0 wish list. If this list gets long and compelling enough, we should consider cutting a new major version.

akshayjshah:
## Un-export Field internals

The original plan (see #13) with fields was to keep the implementation details completely private, and the only way to use a field would be to `AddTo(Encoder)`.

It seems like `FieldType` is exported only to let the zap package expose field constructors, but I think what we should unexport all implementation details and instead:

* Expose the same field constructors in the zapcore package
* The zap package just wraps the underlying zapcore functions, and doesn't need access to the internals of a field.

This gives us more flexibility to change Field implementation details in future.

akshayjshah:
## Unify the `zap` and `zapcore` packages

The intention here was to separate user-facing APIs from APIs relevant only to extension authors. Unfortunately, the division has only made things more confusing - in particular, separating the many `zapcore.Field` constructors from the type definition makes the GoDoc output quite daunting.

akshayjshah:
# Make field order configurable in the console encoder

Not everyone likes the order in the current console encoder. Making this configurable would be nice.

prashantv:
Copied from https://github.com/uber-go/zap/issues/387

# Provide `String` method on Field

If a user tries to print fields, the output is pretty hard to understand, e.g.,
```
fmt.Println(zap.Error(errors.New("test")))
```

results in:
```
{error 25 0  test}
```

We should probably include a `String` method that returns easier to understand output.

Since the current field implementation exports the internals (https://github.com/uber-go/zap/issues/387), we can't add a `String` method until we unexport the internals.

tchap:
### Change zapcore.Core interface to split checking and adding core to the entry

I have been constantly having difficulties wrapping `zapcore.Core` to add what I needed.

Imagine I want to wrap a core to add filtering based on log entry. This would have to be done in `Write` since only there you know all the fields, which is what I needed. But checking whether to write an entry or not in `Write` is too late, because the entry has already been added to the wrapped core in `Check`, because it not only checks, it also adds the core to the entry.

[...]

PR #669, opened 2019-01-22
Add forgotten AppendBinary to ArrayEncoder

NWilson:
For some reason, ObjectEncoder has AddBinary and AddByteString methods, but ArrayEncoder only has AppendByteString. This PR adds the missing (forgotten?) AppendBinary method.

I realise this will have to wait until zap v2, since it's a breaking API change.

### DIFF
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -308,7 +308,8 @@ type ArrayEncoder interface {
 type PrimitiveArrayEncoder interface {
 	// Built-in types.
 	AppendBool(bool)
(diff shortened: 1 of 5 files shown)


NWilson:
Ping - could you create a "v2-unstable" branch (or similar) where API breaking changes can be merged that will go into the v2 release, whenever it is? Otherwise without a branch, any future improvements can't be developed.

prashantv:
We've been tracking improvements in https://github.com/uber-go/zap/issues/388

I'd suggest leaving a comment linking to this PR in that task as we're not currently maintaining a v2 branch.

PR #671, opened 2019-01-25
Skip low level message to avoid `clock_gettime`

Laisky:
Skip lower level logs to avoid too much`time.Now()`,  especially each `time.Now()` will invoke `clock_gettime`. This costs a lot of time even though there is no log match the level to print.

log

This fix will significant reduce the number of syscall and improve performance.

On my test (set log level as `INFO`, then invoke `logger.Debug("xxx")` in benchmark ):

srceen_shot 2019-01-25 at 14 28 31


fgrosse:
There was some discussion around the topic in #664 . The issue with the change is that this would break compatibility for sample based logging cores. This is however on the wishlist for 2.0 (#388)

PR #674, opened 2019-01-30
Fix putting nil to fmt.Stringer interface

fpuc:


fpuc:
closing in favour of https://github.com/uber-go/zap/pull/675/

PR #675, opened 2019-01-30
Don't panic when a given field value is nil

joa:
Ideally a logging framework doesn't cause a panic for code like this:

```go
foo, err := getFoo()

if err != nil {
   log.Error("Couldn't acquire foo", zap.Error(err))
}

// foo could be nil here
log = log.With(zap.Stringer("foo", foo))

bar, err := getBar()

if err != nil {
  // panics
  log.Error("Couldn't acquire bar", zap.Error(err))
}
```

This is probably also fixing #495 and #528

### DIFF
--- a/zapcore/field.go
+++ b/zapcore/field.go
@@ -160,7 +160,7 @@ func (f Field) AddTo(enc ObjectEncoder) {
 	case NamespaceType:
 		enc.OpenNamespace(f.Key)
 	case StringerType:
-		enc.AddString(f.Key, f.Interface.(fmt.Stringer).String())
+		err = encodeStringer(f.Key, f.Interface, enc)
 	case ErrorType:
 		encodeError(f.Key, f.Interface.(error), enc)
 	case SkipType:
@@ -198,4 +198,15 @@ func addFields(enc ObjectEncoder, fields []Field) {
 	for i := range fields {
 		fields[i].AddTo(enc)
 	}
+}
+
+func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) (err error) {
+	defer func() {
+		if v := recover(); v != nil {
+			err = fmt.Errorf("PANIC=%v", v)
+		}
+	}()
+
+	enc.AddString(key, stringer.(fmt.Stringer).String())
+	return
 }
--- a/zapcore/field_test.go
+++ b/zapcore/field_test.go
@@ -24,13 +24,12 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"net/url"
 	"testing"
 	"time"
 
-	"go.uber.org/zap"
-
 	"github.com/stretchr/testify/assert"
-
+	"go.uber.org/zap"
 	. "go.uber.org/zap/zapcore"
 )
 
@@ -58,6 +57,28 @@ func (u users) MarshalLogArray(enc ArrayEncoder) error {
 	return nil
 }
 
+type obj struct {
+	kind int
+}
+
+func (o *obj) String() string {
+	if o == nil {
+		return "nil obj"
+	}
+
+	if o.kind == 1 {
+		panic("panic with string")
+	} else if o.kind == 2 {
+		panic(errors.New("panic with error"))
+	} else if o.kind == 3 {
+		// panic with an arbitrary object that causes a panic itself
+		// when being converted to a string
+		panic((*url.URL)(nil))
+	}
+
+	return "obj"
+}
+
 func TestUnknownFieldType(t *testing.T) {
 	unknown := Field{Key: "k", String: "foo"}
 	assert.Equal(t, UnknownType, unknown.Type, "Expected zero value of FieldType to be UnknownType.")
@@ -67,19 +88,27 @@ func TestUnknownFieldType(t *testing.T) {
 }
 
 func TestFieldAddingError(t *testing.T) {
+	var empty interface{}
 	tests := []struct {
-		t    FieldType
-		want interface{}
+		t     FieldType
+		iface interface{}
+		want  interface{}
+		err   string
 	}{
-		{ArrayMarshalerType, []interface{}{}},
-		{ObjectMarshalerType, map[string]interface{}{}},
+		{t: ArrayMarshalerType, iface: users(-1), want: []interface{}{}, err: "too few users"},
+		{t: ObjectMarshalerType, iface: users(-1), want: map[string]interface{}{}, err: "too few users"},
+		{t: StringerType, iface: obj{}, want: empty, err: "PANIC=interface conversion: zapcore_test.obj is not fmt.Stringer: missing method String"},
+		{t: StringerType, iface: &obj{1}, want: empty, err: "PANIC=panic with string"},
+		{t: StringerType, iface: &obj{2}, want: empty, err: "PANIC=panic with error"},
+		{t: StringerType, iface: &obj{3}, want: empty, err: "PANIC=<nil>"},
+		{t: StringerType, iface: (*url.URL)(nil), want: empty, err: "PANIC=runtime error: invalid memory address or nil pointer dereference"},
 	}
 	for _, tt := range tests {
-		f := Field{Key: "k", Interface: users(-1), Type: tt.t}
+		f := Field{Key: "k", Interface: tt.iface, Type: tt.t}
 		enc := NewMapObjectEncoder()
 		assert.NotPanics(t, func() { f.AddTo(enc) }, "Unexpected panic when adding fields returns an error.")
 		assert.Equal(t, tt.want, enc.Fields["k"], "On error, expected zero value in field.Key.")
-		assert.Equal(t, "too few users", enc.Fields["kError"], "Expected error message in log context.")
+		assert.Equal(t, tt.err, enc.Fields["kError"], "Expected error message in log context.")
 	}
 }
 
@@ -116,6 +145,8 @@ func TestFields(t *testing.T) {
 		{t: ReflectType, iface: users(2), want: users(2)},
 		{t: NamespaceType, want: map[string]interface{}{}},
 		{t: StringerType, iface: users(2), want: "2 users"},
+		{t: StringerType, iface: &obj{}, want: "obj"},
+		{t: StringerType, iface: (*obj)(nil), want: "nil obj"},
 		{t: SkipType, want: interface{}(nil)},
 	}
 


prashantv:
Thanks for the contribution @joa.

Your solution works when dealing with functions that return interfaces, and explicitly return `nil`. However, it doesn't work when the return value has a type, but the value is `nil`, as mentioned in https://github.com/uber-go/zap/issues/495#issuecomment-324805952

With your PR merged in, I wrote a simple test:
```go
func getURL() (*url.URL, error) {
  return nil, nil
}

url, err := getURL()
if err != nil {
  logger.Error("got error", zap.Error(err))
}

logger.Info("got url", zap.Stringer("url", url))
```

This still causes a panic:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x5d3550]

(truncated)

panic(0x74d580, 0x9bc7d0)
        /home/prashant/bin/go1.10.4/src/runtime/panic.go:502 +0x229
net/url.(*URL).String(0x0, 0x790a00, 0x0)
        /home/prashant/bin/go1.10.4/src/net/url/url.go:742 +0x40
go.uber.org/zap/zapcore.Field.AddTo(0x7b73a5, 0x3, 0x18, 0x0, 0x0, 0x0, 0x790a00, 0x0, 0x81ed20, 0xc42009d170)
        /home/prashant/go/src/go.uber.org/zap/zapcore/field.go:172 +0xc10
```

See related questions:
https://golang.org/doc/faq#nil_error
https://stackoverflow.com/questions/38070015/check-if-interface-value-is-nil-in-go-without-using-reflect/38070426

joa:
Good catch. I expanded the test case and added the additional nil check using reflection. 

I was unsuccessful with other variants. I hacked a bit around `unsafe.Pointer` but that's also not promising. Each `(type, value)` tuple is a new instance and you can't compare the heap address for anything.

prashantv:
Unfortunately reflection comes with a performance penalty, and the only alternative is `unsafe`, which we strongly discourage.

It's also not correct to assume that `nil` always means that the type cannot be logged. E.g., it's completely reasonable to have a type which has a `String()` method that handles the `nil` case.

joa:
@prashantv The commit does not use reflection to test for the nil pointer in side the non-nil `(type, ptr)` tuple. Do you have a particular reason why you'd advise against unsafe in this instance? There are unit tests in place that cover both the positive and negative cases.  Of course the compatibility promise isn't given when using unsafe but that's why the tests exist.

I understand the argument around a valid `String()` implementation for the nil case. In fact it is best practice to provide meaningful nil values and have these cases covered. However as with `url.URL` we see this isn't even consistent in within the go lib itself. 

Of course both cases could be supported with `recover` and accepting a potential panic. I guess however I'd rather have a different layer catch these (e.g. a middleware in a server). I guess there is nothing left in this MR to improve directly. Feel free to close if you think the risks outweigh the benefits.

joa:
I see from your comment in https://github.com/uber-go/zap/issues/495#issuecomment-326355875 that fmt is acutally doing this. Maybe that'd be an acceptable solution.

joa:
With defer/recover
```
BenchmarkAddingFields/Zap-8                                  	 1000000	      2212 ns/op	    2214 B/op	      15 allocs/op
BenchmarkAddingFields/Zap.Check-8                            	 1000000	      2221 ns/op	    2214 B/op	      15 allocs/op
BenchmarkAddingFields/Zap.CheckSampled-8                     	 5000000	       312 ns/op	     277 B/op	       1 allocs/op
BenchmarkAddingFields/Zap.Sugar-8                            	  500000	      2762 ns/op	    3010 B/op	      21 allocs/op
BenchmarkAddingFields/apex/log-8                             	   50000	     33792 ns/op	    6694 B/op	     115 allocs/op
BenchmarkAddingFields/go-kit/kit/log-8                       	  200000	      8049 ns/op	    5641 B/op	     105 allocs/op
BenchmarkAddingFields/inconshreveable/log15-8                	   50000	     30263 ns/op	    6625 B/op	     122 allocs/op
BenchmarkAddingFields/sirupsen/logrus-8                      	  200000	      9931 ns/op	    9082 B/op	     122 allocs/op
BenchmarkAddingFields/go.pedge.io/lion-8                     	  200000	      8117 ns/op	    8761 B/op	     105 allocs/op
BenchmarkAddingFields/rs/zerolog-8                           	  200000	      6691 ns/op	    4725 B/op	      76 allocs/op
BenchmarkAddingFields/rs/zerolog.Check-8                     	  200000	      6646 ns/op	    4728 B/op	      76 allocs/op
```

Without defer/recover:
```
BenchmarkAddingFields/Zap-8                                  	 1000000	      1614 ns/op	     770 B/op	       5 allocs/op
BenchmarkAddingFields/Zap.Check-8                            	 1000000	      1614 ns/op	     770 B/op	       5 allocs/op
BenchmarkAddingFields/Zap.CheckSampled-8                     	10000000	       225 ns/op	      91 B/op	       0 allocs/op
BenchmarkAddingFields/Zap.Sugar-8                            	 1000000	      1986 ns/op	    1565 B/op	      11 allocs/op
BenchmarkAddingFields/apex/log-8                             	   50000	     33770 ns/op	    6694 B/op	     115 allocs/op
BenchmarkAddingFields/go-kit/kit/log-8                       	  200000	      8100 ns/op	    5640 B/op	     105 allocs/op
BenchmarkAddingFields/inconshreveable/log15-8                	   50000	     30802 ns/op	    6623 B/op	     122 allocs/op
BenchmarkAddingFields/sirupsen/logrus-8                      	  200000	     10017 ns/op	    9082 B/op	     122 allocs/op
BenchmarkAddingFields/go.pedge.io/lion-8                     	  200000	      8220 ns/op	    8760 B/op	     105 allocs/op
BenchmarkAddingFields/rs/zerolog-8                           	  200000	      6620 ns/op	    4728 B/op	      76 allocs/op
BenchmarkAddingFields/rs/zerolog.Check-8                     	  200000	      6662 ns/op	    4725 B/op	      76 allocs/op
```

joa:
Implemented recover within defer and expanded test cases as discussed (e.g. valid nil impl for Stringer).

prashantv:
Thanks for adding the benchmarks @joa and some responses to your questions:

> @prashantv The commit does not use reflection to test for the nil pointer in side the non-nil (type, ptr) tuple. Do you have a particular reason why you'd advise against unsafe in this instance? There are unit tests in place that cover both the positive and negative cases. Of course the compatibility promise isn't given when using unsafe but that's why the tests exist.

From the unsafe package documentation:
> Packages that import unsafe may be non-portable and are not protected by the Go 1 compatibility guidelines.

Even if the test covers it, a client may upgrade to a new Go version, and zap may stop working in that version, and will require a fix + library upgrades. I'd much rather avoid the concerns that crop up with unsafe.

> Of course both cases could be supported with recover and accepting a potential panic. I guess however I'd rather have a different layer catch these (e.g. a middleware in a server). I guess there is nothing left in this MR to improve directly. Feel free to close if you think the risks outweigh the benefits.

Yep, the `recover` approach works well, but also has a measurable performance impact. Given that this could be better handled elsewhere, I'm not sure the benefits outweigh the cost.

However, I do want to consider all the options, since this has come up multiple times. We could:
 * Do what you had originally: only check if `value == nil`, which works if you return a nil type/value interface, but not otherwise.
 * (In zap or externally) Add a `SafeStringer` function which will wrap a stringer, and add panic handling. Field constructors can be done outside of zap, and it would probably look like:

```go
type safeStringer struct{
  s fmt.Stringer
}

func (ss safeStringer) String() {
  defer func() {
    // recover and return "nil"
  }()
  return ss.s.String()
}

func SafeStringer(s fmt.Stringer) fmt.Stringer {
  return safeStringer{s}
}
```

joa:
Agree.

There is another option. If you're just interested in the `Stringer` case the performance impact could be reduce to just that. And I think it would be the best solution since we don't expect panic in any of the other cases (e.g. `IntField`) and should not pay the `recover` tax there and doesn't introduce a new `SafeXXX` kind of field. Because why would I not want to use the  variant one?

`ArrayMarshaler` and `ObjectMarshaler` are defined by Zap. It would be okay IMHO to make the  requirement that a meaningful `nil` for implementation must exist.

Note that these benchmark numbers are to be taken with a grain of salt. `BenchmarkAddingFields` doesn't use `zap.Stringer`. This just shows you that the performance impact is zero when not using `zap.Stringer`.

```
BenchmarkAddingFields/Zap-8                                      1000000              1699 ns/op             770 B/op          5 allocs/op
BenchmarkAddingFields/Zap.Check-8                                1000000              1705 ns/op             770 B/op          5 allocs/op
BenchmarkAddingFields/Zap.CheckSampled-8                         5000000               236 ns/op              91 B/op          0 allocs/op
BenchmarkAddingFields/Zap.Sugar-8                                1000000              2058 ns/op            1565 B/op         11 allocs/op
BenchmarkAddingFields/apex/log-8                                   50000             34314 ns/op            6694 B/op        115 allocs/op
BenchmarkAddingFields/go-kit/kit/log-8                            200000              8087 ns/op            5640 B/op        105 allocs/op
BenchmarkAddingFields/inconshreveable/log15-8                      50000             30024 ns/op            6625 B/op        122 allocs/op
BenchmarkAddingFields/sirupsen/logrus-8                           200000              9941 ns/op            9082 B/op        122 allocs/op
BenchmarkAddingFields/go.pedge.io/lion-8                          200000              8358 ns/op            8758 B/op        105 allocs/op
BenchmarkAddingFields/rs/zerolog-8                                200000              6914 ns/op            4725 B/op         76 allocs/op
BenchmarkAddingFields/rs/zerolog.Check-8                          200000              7137 ns/op            4724 B/op         76 allocs/op
```

prashantv:
Only doing `defer` for the `Stringer` case is a reasonable option, but what is the performance impact to logging `Stringer` fields when using `defer`?

If it's negligible, then it definitely makes sense to do by default. Otherwise, the "safe" field constructors could provide an alternative between safety and performance.

joa:
@prashantv: Sorry for the late reply but here are the benchmarks

### Benchmark using `defer` for `fmt.Stringer`
```
BenchmarkAddingFields/Zap-8                                      1000000              1991 ns/op            1004 B/op          8 allocs/op
BenchmarkAddingFields/Zap.Check-8                                1000000              1887 ns/op            1004 B/op          8 allocs/op
BenchmarkAddingFields/Zap.CheckSampled-8                         5000000               267 ns/op             122 B/op          0 allocs/op
BenchmarkAddingFields/Zap.Sugar-8                                1000000              2475 ns/op            1865 B/op         14 allocs/op
BenchmarkAddingFields/apex/log-8                                   50000             34326 ns/op            6691 B/op        115 allocs/op
BenchmarkAddingFields/go-kit/kit/log-8                            200000              8357 ns/op            5855 B/op        110 allocs/op
BenchmarkAddingFields/inconshreveable/log15-8                      50000             32977 ns/op            8114 B/op        128 allocs/op
BenchmarkAddingFields/sirupsen/logrus-8                           200000             10069 ns/op            9088 B/op        122 allocs/op
BenchmarkAddingFields/go.pedge.io/lion-8                          200000              9393 ns/op            8688 B/op        105 allocs/op
BenchmarkAddingFields/rs/zerolog-8                                200000              7106 ns/op            4729 B/op         76 allocs/op
BenchmarkAddingFields/rs/zerolog.Check-8                          200000              7171 ns/op            4725 B/op         76 allocs/op
```

### Baseline
```
BenchmarkAddingFields/Zap-8                                      1000000              1927 ns/op             875 B/op          7 allocs/op
BenchmarkAddingFields/Zap.Check-8                                1000000              1813 ns/op             875 B/op          7 allocs/op
BenchmarkAddingFields/Zap.CheckSampled-8                         5000000               258 ns/op             105 B/op          0 allocs/op
BenchmarkAddingFields/Zap.Sugar-8                                1000000              2316 ns/op            1736 B/op         13 allocs/op
BenchmarkAddingFields/apex/log-8                                   30000             35845 ns/op            6695 B/op        115 allocs/op
BenchmarkAddingFields/go-kit/kit/log-8                            200000              8246 ns/op            5856 B/op        110 allocs/op
BenchmarkAddingFields/inconshreveable/log15-8                      50000             33425 ns/op            8110 B/op        128 allocs/op
BenchmarkAddingFields/sirupsen/logrus-8                           200000              9958 ns/op            9088 B/op        122 allocs/op
BenchmarkAddingFields/go.pedge.io/lion-8                          200000              8169 ns/op            8696 B/op        105 allocs/op
BenchmarkAddingFields/rs/zerolog-8                                200000              7984 ns/op            4724 B/op         76 allocs/op
BenchmarkAddingFields/rs/zerolog.Check-8                          200000              7372 ns/op            4724 B/op         76 allocs/op
```

I applied the following changes to the benchmark functions:

```diff
diff --git a/benchmarks/zap_test.go b/benchmarks/zap_test.go
index 92f7120..aa503bc 100644
--- a/benchmarks/zap_test.go
+++ b/benchmarks/zap_test.go
@@ -23,6 +23,7 @@ package benchmarks
 import (
        "errors"
        "fmt"
+       "net/url"
        "time"
 
        "go.uber.org/multierr"
@@ -37,6 +38,7 @@ var (
        _messages   = fakeMessages(1000)
        _tenInts    = []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 0}
        _tenStrings = []string{"a", "b", "c", "d", "e", "f", "g", "h", "i", "j"}
+       _oneStringer, _ = url.Parse("http://localhost:1234")
        _tenTimes   = []time.Time{
                time.Unix(0, 0),
                time.Unix(1, 0),
@@ -130,6 +132,7 @@ func fakeFields() []zap.Field {
                zap.Ints("ints", _tenInts),
                zap.String("string", _tenStrings[0]),
                zap.Strings("strings", _tenStrings),
+               zap.Stringer("stringer", _oneStringer),
                zap.Time("time", _tenTimes[0]),
                zap.Times("times", _tenTimes),
                zap.Object("user1", _oneUser),
@@ -145,6 +148,7 @@ func fakeSugarFields() []interface{} {
                "ints", _tenInts,
                "string", _tenStrings[0],
                "strings", _tenStrings,
+               "stringer", _oneStringer,
                "time", _tenTimes[0],
                "times", _tenTimes,
                "user1", _oneUser,
@@ -162,6 +166,7 @@ func fakeFmtArgs() []interface{} {
                _tenInts,
                _tenStrings[0],
                _tenStrings,
+               _oneStringer,
                _tenTimes[0],
                _tenTimes,
                _oneUser,
```

In conclusion: the performance impact when using `defer` with `fmt.Stringer` is neglectable and the number of additional allocations (one) scales linear with the amount of `zap.Stringer` fields.

prashantv:
Thanks @joa. To better understand the performance impact, I wrote a benchmark which:
 * Uses a production logger config, writing to `/dev/null`
 * Created a custom stringer type which doesn't need to encode anything

E.g.,
```go
type customString string

func (cs customString) String() string {
        return string(cs)
}

func BenchmarkStringer(b *testing.B) {
        cfg.OutputPaths = []string{"/dev/null"}
        l, err := cfg.Build()
        require.NoError(b, err, "failed to build logger")

        cs := customString("encode")

        b.Run("1 field", func(b *testing.B) {
                for i := 0; i < b.N; i++ {
                        l.Info("test",
                                zap.Stringer("cs1", cs),
                        )
                }
        })
        b.Run("4 fields", func(b *testing.B) {
                for i := 0; i < b.N; i++ {
                        l.Info("test",
                                zap.Stringer("cs1", cs),
                                zap.Stringer("cs2", cs),
                                zap.Stringer("cs3", cs),
                                zap.Stringer("cs4", cs),
                        )
                }
        })
}
```

To get a closer look at the impact to just the stringer performance, I wrote a benchmark which only logs stringer fields,
```
        b.Run("1 field", func(b *testing.B) {
                for i := 0; i < b.N; i++ {
                        l.Info("test",
                                zap.Stringer("cs1", cs),
                        )
                }
        })
        b.Run("4 fields", func(b *testing.B) {
                for i := 0; i < b.N; i++ {
                        l.Info("test",
                                zap.Stringer("cs1", cs),
                                zap.Stringer("cs2", cs),
                                zap.Stringer("cs3", cs),
                                zap.Stringer("cs4", cs),
                        )
                }
        })
```

I then compared the difference with and without this change on Go 1.12,

Without this change (no `defer`):
```
BenchmarkStringer/1_field-4             20000000               203 ns/op              82 B/op          2 allocs/op
BenchmarkStringer/4_fields-4            20000000               373 ns/op             322 B/op          5 allocs/op
```

With this change (with `defer`):
```
BenchmarkStringer/1_field-4             20000000               207 ns/op              83 B/op          2 allocs/op
BenchmarkStringer/4_fields-4            10000000               377 ns/op             327 B/op          5 allocs/op
```

There's a very slight increase in the number of bytes allocated, but the actual number of allocations remains the same, and the performance is the same (the difference looks like noise).

zapcore/field.go
```diff
@@ -160,6 +160,23 @@ func (f Field) AddTo(enc ObjectEncoder) {
 	case NamespaceType:
 		enc.OpenNamespace(f.Key)
 	case StringerType:
+		defer func() {
```
prashantv: Given the function is pretty large and this is a large block, it'd be better to move the logic into a separate `encodeStringer` function

zapcore/field.go
```diff
@@ -160,6 +160,23 @@ func (f Field) AddTo(enc ObjectEncoder) {
...
+			if r := recover(); r != nil {
+				var msg string
+
+				switch v := r.(type) {
```
prashantv: should this behaviour match `fmt`?

zapcore/field_test.go
```diff
@@ -58,6 +57,27 @@ func (u users) MarshalLogArray(enc ArrayEncoder) error {
 	return nil
 }
 
+type obj struct {
+	kind int
+}
+
+func (o *obj) String() string {
```
prashantv: let's add a custom object type which mimics production use-cases: we have a `String` method which accesses some field, but it causes a panic because the object is nil.

zapcore/field.go
joa: - `fmt` panics when printing the panic value panics. I added a test case for this and currently the code would print out `<unknown>` 
- `fmt` prints `<nil>` when the receiver is `nil` which is something we could do as well
- `fmt` prints something like `%s!(PANIC=some additional text)` which is specific to the `fmt` package IMHO but it might make sense to adopt the `PANIC=` syntax

zapcore/field_test.go
joa: There is the following additional test in line 103:
`{t: StringerType, iface: (*url.URL)(nil), want: empty, err: "runtime error: invalid memory address or nil pointer dereference"},`

Is that sufficient for you? The current obj struct makes testing some of the edge cases a lot simpler but I am happy to change that.

zapcore/field.go
joa: Done.  We call `encodeStringer` for `StringerType` which contains a call to `catchPanic`.

zapcore/field.go
```diff
@@ -199,3 +199,25 @@ func addFields(enc ObjectEncoder, fields []Field) {
...
+
+func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) {
+	defer catchPanic(key, enc)
+	enc.AddString(key, stringer.(fmt.Stringer).String())
+}
+
+func catchPanic(key string, enc ObjectEncoder) {
+	if err := recover(); err != nil {
+		var msg string
+
```
abhinav: minor style nit:

```suggestion
```

zapcore/field_test.go
abhinav: That case seems sufficient; it exercises the `fmt.Stringer` with a nil value case.

zapcore/field.go
```diff
@@ -160,7 +160,7 @@ func (f Field) AddTo(enc ObjectEncoder) {
 	case NamespaceType:
 		enc.OpenNamespace(f.Key)
 	case StringerType:
-		enc.AddString(f.Key, f.Interface.(fmt.Stringer).String())
+		encodeStringer(f.Key, f.Interface, enc)
```
prashantv: rather than duplicating the logic of how errors are encoded, this should probably be `err = encodeStringer(..)` and the panic handling should return an error

zapcore/field.go
```diff
@@ -199,3 +199,25 @@ func addFields(enc ObjectEncoder, fields []Field) {
 		fields[i].AddTo(enc)
 	}
 }
+
+func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) {
+	defer catchPanic(key, enc)
```
prashantv: ok with inlining `catchPanic`, especially since we'll need to modify the returned error here (if we make `encodeStringer` return `error`:

```go
func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) (retError error) {
  defer func() 
    if err := recover(); err != nil {
      // massage error
      retError = err
    }
  }()

  enc.AddString(key, stringer.(fmt.Stringer).String())
```

zapcore/field.go
```diff
@@ -199,3 +199,25 @@ func addFields(enc ObjectEncoder, fields []Field) {
...
+	if err := recover(); err != nil {
+		var msg string
+
+		switch v := err.(type) {
+		case error:
+			msg = v.Error()
+		case string:
+			msg = v
+		default:
+			msg = "<unknown>"
```
prashantv: optional way to reduce some code,
```go
msg := "<unknown>"
switch v := err.(type) {
case error:
   msg = v.Error()
case string:
  msg = v
}
```

zapcore/field_test.go
prashantv: yep makes sense, thanks

zapcore/field.go
```diff
@@ -199,3 +199,25 @@ func addFields(enc ObjectEncoder, fields []Field) {
...
+func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) {
+	defer catchPanic(key, enc)
+	enc.AddString(key, stringer.(fmt.Stringer).String())
+}
+
+func catchPanic(key string, enc ObjectEncoder) {
+	if err := recover(); err != nil {
+		var msg string
+
+		switch v := err.(type) {
```
prashantv: can't reply to the conversation about matching `fmt`, but I like the idea of `PANIC=` to make it explicit that this was from a panic (rather than a standard returned error).

I like their use of `<nil>` as well, though that seems like it needs string matching, which I'm not a fan of. @abhinav thoughts?

abhinav: The `PANIC=` prefix sounds nice. It helps differentiate between `panic("foo")`
and `return errors.New("foo")` from an error-able encode call.

I'm not a fan of the `nil` check. `PANIC=runtime error: invalid memory address or nil pointer dereference` is enough information for the user to know what was nil.

joa: `fmt` performs a `nil` check via reflection at this stage. Given that this code path should be rather cold it is something to consider.

zapcore/field.go
joa: I added the `PANIC=` prefix and removed the case of an indefinitely reentrant panic as I think it is overly constructed. 

This simplifies the code and ensures panic values are printed the same way `fmt` behaves. 

However as we remove the `<unknown>` special case, printing the panic could causes a panic itself and that would be potentially unhandled.

zapcore/field.go
joa:
```go
if v := recover(); v != nil {
  err = errors.New(fmt.Sprintf("PANIC=%v", v))
}
```

This has a nice simplicity to it.

joa:
Thank you for the review @prashantv and @abhinav. This MR became much more concise thanks to your feedback.

joa:
Might want to revisit in Go 1.13: https://go-review.googlesource.com/c/go/+/171758/

PR #678, opened 2019-02-05
Add LazyMessage function as message argument in debug methods

hudlngc:
# Add LazyMessage Function as Message Argument in Debug Methods

Logging debug messages is a common way to print out diagnostic information in log files during development and testing phases. However, too many debug-level logs may impact system performance in production especially when debug messages have to be generated before being passed to log functions. Examples of such use cases are but not limited to:

* Printing out string representation of a byte array in a request body to find out what is actually in the request
* Printing out a struct in a custom (mostly a calculated form) string form instead of Go's default implementation.

This change introduces `LazyMessage` function as an argument to be passed to Debug functions. `LazyMessage` is a simple no-arg function that returns a string, exactly the same signature as `stringer` does. When a debug function is called, the logging level is checked before `LazyMessage` function is evaluated, which eliminates unnecessary debug message generation if debug is disabled in the current log level (often in production).

Both `zap.Logger` and `zap.SugaredLogger` are changed with new debug functions added. Mild modifications to existing internal functions and structures are made to avoid code redundancy and assure correct level of call stack depth (caller skip values).

Unit tests are adjusted with newly added debug function calls in various test cases.

Benchmark tests with `BenchmarkDebuglWithoutFields` are added in scenario_bench_test.go. The benchmark test runs all the new debug functions compared to existing debug functions with current log level set to `info`. Here is the result of a benchmark run:

```
Running tool: /usr/local/bin/go test -benchmem -run=^$ go.uber.org/zap/benchmarks -bench ^(BenchmarkDebuglWithoutFields)$

goos: darwin
goarch: amd64
pkg: go.uber.org/zap/benchmarks
BenchmarkDebuglWithoutFields/Zap.DebugOnInfo-12         	10000000	       164 ns/op	     529 B/op	       5 allocs/op
BenchmarkDebuglWithoutFields/Zap.DebuglOnInfo-12        	100000000	        21.9 ns/op	      16 B/op	       1 allocs/op
BenchmarkDebuglWithoutFields/Zap.DebugOnInfoWithBytesToString-12         	20000000	        93.2 ns/op	     416 B/op	       1 allocs/op
BenchmarkDebuglWithoutFields/Zap.DebuglOnInfoWithBytesToString-12        	100000000	        22.6 ns/op	      16 B/op	       1 allocs/op
BenchmarkDebuglWithoutFields/Zap.SugarDebugOnInfo-12                     	10000000	       188 ns/op	     545 B/op	       6 allocs/op
BenchmarkDebuglWithoutFields/Zap.SugarDebugflOnInfo-12                   	200000000	         9.55 ns/op	      16 B/op	       1 allocs/op
BenchmarkDebuglWithoutFields/Zap.DebugOnInfoWithBytesToString#01-12      	10000000	       123 ns/op	     432 B/op	       2 allocs/op
BenchmarkDebuglWithoutFields/Zap.DebuglOnInfoWithBytesToString#01-12     	100000000	        10.1 ns/op	      16 B/op	       1 allocs/op
PASS
ok  	go.uber.org/zap/benchmarks	15.610s
Success: Benchmarks passed.
```

This idea comes from Log4j2's `logger.debug()` method: https://logging.apache.org/log4j/2.x/log4j-api/apidocs/index.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ _testmain.go
 *.pprof
 *.out
 *.log
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ _testmain.go
 *.out
 *.log
 .vscode/
+*.swp
--- a/benchmarks/scenario_bench_test.go
+++ b/benchmarks/scenario_bench_test.go
@@ -21,9 +21,12 @@
 package benchmarks
 
 import (
+	"fmt"
 	"io/ioutil"
 	"log"
 	"testing"
+
+	"go.uber.org/zap/zapcore"
 
 	"go.uber.org/zap"
 )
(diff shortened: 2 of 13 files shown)


yurishkuro:
how is that different from using `zap.Stringer()` field? Why does the message itself need to be dynamic, are you doing string concatenation when preparing it?

prashantv:
Thanks for the contribution @hudlngc 

Zap is intended for structured logging, where the message is typically a constant string. If you'd like to log a dynamic string as a field, and if your type doesn't implement `fmt.Stringer`, then you can add a custom type in your application code to pass an anonymous function in as a `fmt.String`:

```go
type StringerFunc func() string

func (f StringerFunc) String() string {
  return f()
}
```

You can then pass in any anonymous function:
```go
logger.Debug("message", zap.Stringer("obj", StringerFunc(func() string { return "test" })))
```

hudlngc:
Thanks for the explanation. I actually just realized it recently and
started to use that trick in my code. I should’ve noticed that earlier but
as a newbie of Go, I still have a lot to learn. Thanks again for this.
Please feel free to abandon the PR. Maybe you could add this in README for
other people to refrence.

Thank you
Dongliang

hudlngc:
By the way sorry for not notifying you earlier for the finding.

prashantv:
No worries, glad you found a solution!

PR #684, opened 2019-03-03
zap.BuildForTest allows injection of an output WriteSyncer 

keynan:
I wanted to test some context sensitive custom fields. However there does not currently exist a method to test the output of a logger constructed from the handy New[Development|Production]Logger.

This adds a public function that accepts the WriteSyncers directly so that buffers can be supplied and then reviewed after a message is logged.

### DIFF
--- a/config.go
+++ b/config.go
@@ -164,12 +164,16 @@ func NewDevelopmentConfig() Config {
 
 // Build constructs a logger from the Config and Options.
 func (cfg Config) Build(opts ...Option) (*Logger, error) {
-	enc, err := cfg.buildEncoder()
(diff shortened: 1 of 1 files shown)


prashantv:
Thanks for the contribution @keynan.

Have you seen the [`observer` subpackage](https://godoc.org/go.uber.org/zap/zaptest/observer)? It lets you create a logger for unit tests that capture the message and fields in-memory. That's what we typically use in tests. If you must use the config struct, then you should create a temporary file using `ioutil.TempFile` and pass that path in to the `OutputPaths`.

I don't think we want to add a `Build` method that ignores some of the fields in the configuration struct given the alternatives.

PR #685, opened 2019-03-04
Move addFields to ObjectEncoder  to let logger#With be extendible

lysu:
1. What did you do?

we write a [TextEncoder](https://github.com/pingcap/log/blob/master/zap_text_encoder.go#L124) to let log be formated as

```
[2019/03/04 23:00:05.943 +08:00] [ERROR] [log_test.go:209] [common=comval] [k=v]
```

but found that `With` method cannot add `[ ]`

```
newLogger := logger.With(zap.String("c", "c1"))
newLogger.Error("", zap.String("k", "v"))
```

will got 

```
[2019/03/04 23:00:05.943 +08:00] [ERROR] [log_test.go:209],c=c1 [k=v]
```

2. What's the problem

found that `addFields` at

https://github.com/uber-go/zap/blob/master/zapcore/field.go#L197

maybe better move to  `objectEncoder` at

https://github.com/uber-go/zap/blob/master/zapcore/encoder.go#L247

so transfer the responsibility of how to print multiple field  to objectEncoder, so other people can customize it, eg. we can add `[` `]` sourrounding each field even them add via [logger#With](https://github.com/uber-go/zap/blob/master/logger.go#L159)

3. What is changed

move addFields to ObjectEncoder  and let current implements it.

ref https://github.com/pingcap/log/issues/3, https://github.com/pingcap/tidb/pull/9548

### DIFF
--- a/zapcore/console_encoder.go
+++ b/zapcore/console_encoder.go
@@ -128,7 +128,7 @@ func (c consoleEncoder) writeContext(line *buffer.Buffer, extra []Field) {
 	context := c.jsonEncoder.Clone().(*jsonEncoder)
 	defer context.buf.Free()
 
-	addFields(context, extra)
+	context.AddFields(extra)
 	context.closeOpenNamespaces()
 	if context.buf.Len() == 0 {
 		return
--- a/zapcore/core.go
+++ b/zapcore/core.go
@@ -71,7 +71,7 @@ type ioCore struct {
 
 func (c *ioCore) With(fields []Field) Core {
 	clone := c.clone()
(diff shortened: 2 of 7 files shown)


lysu:
@akshayjshah PTAL thx~

prashantv:
Thanks for the contribution @lysu.

Unfortunately this change is not backwards compatible, since you're modifying the interface, so we won't be able to accept this change in zap 1.0.

lysu:
@prashantv  we will take another temp way to solve this problem in current and wait for the 2.x, thx :D

PR #689, opened 2019-03-19
Adding reportingSampler

pavel-shpilev:
A version of sampler that generates reporting log entries on number of records it sampled.

### DIFF
--- a/zapcore/counter.go
+++ b/zapcore/counter.go
@@ -1,0 +1,85 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zapcore
+
+import (
+	"time"
+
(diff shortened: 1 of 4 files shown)


jcorbin:
Meta: this would be easier to review if you had left the counters in `sampler.go`; as it is, you've increased the cognitive load for reviewers, part of why I've been "leaving it for someone else". As it is, I'll need to clone this branch to compare the old/new code directly, rather than jumping back and forth here in github.

prashantv:
@pavel-shpilev Thank you for the contribution. I have only taken a high level look, but I have some questions:

 * As Josh mentioned, moving counters out adds some noise. It may be better to either: leave it in the original location with a `TODO` to move it out, or do a prefactoring to move it out. Changes that just rearrange code without making any functional changes are easy to review and can reduce the load on further reviews.
 * It's not clear to me why this needs to be a completely separate sampler, or whether it can reuse the same sampler with some additional logic.
 * The reporting is fixed, would callers want to customize the reporting message/level? Perhaps we should just call a function with parameters used for sampling, and have the function return a `Entry`. We can provide a default.

pavel-shpilev:
Thank you for the feed back.
To answer your points:
- **Counters** At the first approach I thought I'd need a special version for the counters, and therefore I moved it into its own file. I soon figured out that wasn't the case. Nevertheless I leaved it there because it felt atomic enough to be its own thing. I can make it a separate diff or I can move it out. You, guys, are more familiar with the codebase, so I will let you call what's the best course of action here.
- **Separate sampler**  The idea was to avoid surprises in form of a new behavior for existing zap users. If you feel like it can be part of the regular sampler, I'm happy to do that.
- **Configurable reporting** Valid points, will do.

prashantv:
> Separate sampler The idea was to avoid surprises in form of a new behavior for existing zap users. If you feel like it can be part of the regular sampler, I'm happy to do that.

Totally agree that we shouldn't change the behaviour, but I think we can have the additional logic be optional in the existing sampler. It can still be disabled by default, but I think we can reuse most of the additional logic.

> Counters At the first approach I thought I'd need a special version for the counters, and therefore I moved it into its own file. I soon figured out that wasn't the case. Nevertheless, I left it there because it felt atomic enough to be its own thing. I can make it a separate diff or I can move it out. You, guys, are more familiar with the codebase, so I will let you call what's the best course of action here.

Agree that it makes sense to move out, but I'd defer till we make a decision on whether the existing sampler can be shared. If we can share the sampler, less of a benefit in moving it out, and we can do so separately as a follow-up if required.

pavel-shpilev:
> Totally agree that we shouldn't change the behaviour, but I think we can have the additional logic be optional in the existing sampler. It can still be disabled by default, but I think we can reuse most of the additional logic.

I'm looking into making this part of existing sampler. Making it optional and configurable seems quite tricky without changing `NewSampler()` signature, which I try to avoid. What are the cons of having it as a separate sampler?

prashantv:
> I'm looking into making this part of existing sampler. Making it optional and configurable seems quite tricky without changing NewSampler() signature, which I try to avoid.

Since we can't change `NewSampler` without breaking backwards compatibility, I think we should add a new constructor that takes a struct with the parameters (sampling frequency, whether to log after sampling, etc)

> What are the cons of having it as a separate sampler?

Code duplication and ongoing maintenance. These samplers are almost exactly the same code, which means that any fixes or changes to the logic will need to be applied to both samplers increasing the long-term maintenance cost.

pavel-shpilev:
Revised PR: https://github.com/uber-go/zap/pull/695
Closing this one.

PR #695, opened 2019-04-10
Adding reporting functionality to sampler

pavel-shpilev:


zapcore/sampler.go
```diff
@@ -62,33 +63,76 @@ func fnv32a(s string) uint32 {
...
+// sampled entries. reportingLevel must be enabled on the core.
+func NewReportingSampler(
+	core Core,
+	tick time.Duration,
+	first, thereafter int,
+	reportingLevel Level,
+	reportingLoggerName, reportingMessage string,
+) Core {
+	if !core.Enabled(reportingLevel) {
+		panic("Core must have at least info logging level.")
```
jcorbin: Hardcoded "info"; use fmt.Sprintf and interpolate the actual value of reportingLevel

config.go
```diff
@@ -205,6 +219,18 @@ func (cfg Config) buildOptions(errSink zapcore.WriteSyncer) []Option {
...
 	if cfg.Sampling != nil {
 		opts = append(opts, WrapCore(func(core zapcore.Core) zapcore.Core {
+			if cfg.Sampling.Reporting != nil && cfg.Sampling.Reporting.Enabled {
+				loggerName := cfg.Sampling.Reporting.LoggerName
+				if loggerName == "" {
+					loggerName = "zapcore"
+				}
+				level := cfg.Sampling.Reporting.Level.Level()
+				message := cfg.Sampling.Reporting.Message
+				if message == "" {
```
jcorbin: Push this zero default down into the `NewReportingSampler` constructor; likewise for the loggerName above.

zapcore/sampler.go
```diff
@@ -126,9 +178,35 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
...
+		Message:    s.reporting.message,
+	}
+	fields := []Field{
+		{Key: "original_level", Type: StringType, String: ent.Level.String()},
+		{Key: "original_message", Type: StringType, String: ent.Message},
+		{Key: "count", Type: Int64Type, Integer: int64(sampleCount)},
+	}
+	err := s.reporting.core.Write(entry, fields)
+	if err != nil {
+		fmt.Printf("%v write error: %v\n", time.Now(), err)
```
jcorbin: This is what `Logger.errorOutput` is for, but probably this `report` method should just return an error, and be used in existing error handling paths... The fact that `With` above doesn't support error propagation, seems a sign that we shouldn't be writing output during a `With`, but instead defer reporting until actual attempt to log time.

zapcore/sampler_test.go
```diff
@@ -35,12 +35,30 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	repLogName = "zaptet"
```
jcorbin: nit: misspelling of "zaptest" ?

zapcore/sampler.go
```diff
@@ -126,9 +178,35 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
 	}
 
 	counter := s.counts.get(ent.Level, ent.Message)
-	n := counter.IncCheckReset(ent.Time, s.tick)
+	n, prevN := counter.IncCheckReset(ent.Time, s.tick)
 	if n > s.first && (n-s.first)%s.thereafter != 0 {
 		return ce
 	}
+	if s.reporting.enabled && prevN > 0 {
+		s.report(ent, prevN)
```
jcorbin: `With` seems to be the wrong time to be doing this reporting, better if it could be right at checked entry write, which would then support proper error propagation (as noted below).

jcorbin (changes requested):

zapcore/sampler.go
pavel-shpilev: This one is called by `Check()` which doesn't support error propagation either. I thought printing would be better than just ignoring.

zapcore/sampler.go
pavel-shpilev: Again, this is `Check()` not `With()`, and this is where we check counters value. So it got to be here.

zapcore/sampler.go
```diff
@@ -126,9 +193,36 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
...
+		Time:       time.Now(),
+		Level:      s.reporting.level,
+		Message:    s.reporting.message,
+	}
+	fields := []Field{
+		{Key: "original_level", Type: StringType, String: ent.Level.String()},
+		{Key: "original_message", Type: StringType, String: ent.Message},
+		{Key: "count", Type: Int64Type, Integer: int64(sampleCount)},
+	}
+	err := s.reporting.core.Write(entry, fields)
```
jcorbin: nit: just `return s.reporting.core.Write(...)`; you could even inline the fields in the log site; clearer to read if you use field constructors, also less fragile if we ever iterate on field particulars.

pavel-shpilev: Updated return.
To use field constructors here would require importing zap, which would introduce circular reference.
Not sure what you mean by 'inline the fields in the log site'. If you're talking about construction like:
```
reportingCore := s.reporting.core.With(fields)
reportingCore.Write()
```
I don't think it makes it any simpler or neater.

config.go
```diff
@@ -27,14 +27,28 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+// ReportingSamplingConfig offers way to configure reporting of sampling
+// events.
+//
+// With sampling there is always added ambiguity. A record is missing from the
```
prashantv: I think a better name may be `SamplingReportingConfig`

"This avoids the ambiguity of whether a record is missing because it was sampled, or whether it never took palce"

config.go
```diff
@@ -27,14 +27,28 @@ import (
...
+// logs because it was sampled or because it never took place? Sampling reports
+// intend to compensate for that. See zapcore.NewReportingSampler for details.
+type ReportingSamplingConfig struct {
+	Enabled    bool        `json:"enabled" yaml:"enabled"`
+	LoggerName string      `json:"loggerName" yaml:"loggerName"`
+	Level      AtomicLevel `json:"level" yaml:"level"`
```
prashantv: why this an `AtomicLevel`?

config.go
```diff
@@ -27,14 +27,28 @@ import (
...
+	Message    string      `json:"message" yaml:"message"`
```
prashantv: I'm not sure whether we want to add a set of fields to represent a log message, or whether we should only provide a simple function hook that returns a log message. otherwise, this may end up expanding significantly to cater to every use-case

zapcore/sampler.go
```diff
@@ -62,33 +69,85 @@ func fnv32a(s string) uint32 {
...
-		// We raced with another goroutine trying to reset, and it also reset
-		// the counter to 1, so we need to reincrement the counter.
-		return c.counter.Inc()
+		// We raced with another goroutine trying to reset, and it reset the
+		// counter to 1, so we just increment.
+		return c.counter.Inc(), 0
 	}
+	preResetVal = c.counter.Load()
+	currentVal = 1
+	c.counter.Store(currentVal)
```
prashantv: i think this change can now "lose" increments

previously, we assumed that the `resetAt` was changed *after* we set the counter value. now it's set after, so we could have:

g1 changes `resetAt`
g2 tries to change `resetAt`, fails, and then increments counter (let's say to 5)
g1 now stores `1` in the counter

we've lost g2's increment.

zapcore/sampler.go
```diff
@@ -62,33 +69,85 @@ func fnv32a(s string) uint32 {
...
+// subset of your logs. The difference with the regular sampling core is that it
+// reports the number of sampled entries for the same logging level and message.
+// Reporting is in form of a separate log entry per every tick when there are
+// sampled entries. reportingLevel must be enabled on the core.
+func NewReportingSampler(
+	core Core,
+	tick time.Duration,
+	first, thereafter int,
+	reportingLevel Level,
+	reportingLoggerName, reportingMessage string,
```
prashantv: we shouldn't take these as function args, as any additional fields we may want now require a breaking API change.

I think the implementation should be more generic rather than trying to describe every single field. E.g., maybe a function that returns a log message, or a function that can add a specific log message to the core.

zapcore/sampler.go
```diff
@@ -126,9 +193,32 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
 	}
 
 	counter := s.counts.get(ent.Level, ent.Message)
-	n := counter.IncCheckReset(ent.Time, s.tick)
+	n, prevN := counter.IncCheckReset(ent.Time, s.tick)
 	if n > s.first && (n-s.first)%s.thereafter != 0 {
 		return ce
 	}
+	if s.reporting.enabled && prevN > 0 {
```
prashantv: rather than complicating the logic here, what if we just ran a function (similar to a callback) to report, and we passed in parameters to it?

separate the responsibility of how to form the message/log away from here.

pavel-shpilev:
@jcorbin Can I have another go please?

config_test.go
```diff
@@ -80,6 +109,23 @@ func TestConfig(t *testing.T) {
...
+			if tt.expectSamplingN > 0 {
+				ztest.Sleep(time.Second)
+				logger.Info("sampling")
+				byteContents, err = ioutil.ReadAll(temp)
+				require.NoError(t, err, "Couldn't read log contents from temp file.")
+				logs = string(byteContents)
+				logSlice := strings.Split(logs, "\n")
+				samplingRep := logSlice[len(logSlice)-3]
+				require.Regexpf(t, tt.expectSamplingRe, samplingRep, "Unexpected sampling report output.")
+				re := regexp.MustCompile(`[\d]+`)
```
jcorbin: nit: regexp simpler as just `\d+`
nit: prefer to pre-compile regexps at package level, ala
```
var numberPattern = regexp.MustCompile(`\d+`)
```

zapcore/sampler.go
```diff
@@ -38,6 +47,37 @@ type counter struct {
 
 type counters [_numLevels][_countersPerLevel]counter
 
+// Report defines a function that generates and writes sampling activity report
+// log entries.
+type Report func(core Core, ent Entry, n int64) error
```
jcorbin: In my view, any exported `type T func()` being added, especially to a post-1.0 stable project, should immediately be re-written as an interface:
```
type Ter interface {
  thing()
}
```
with maybe a convenience
```
type tFunc func()
func (tf tFunc) thing() { tf() }
```

zapcore/sampler.go
```diff
@@ -38,6 +47,37 @@ type counter struct {
...
+// SamplingReport is an example and the default implementation of Report type.
+// Refer to it when you want to customise reporting behaviour and message
+// composition.
+func SamplingReport(core Core, ent Entry, n int64) error {
+	if n <= 0 {
+		return nil
+	}
+	if !core.Enabled(ReportingLogLevel) {
+		return fmt.Errorf(
+			"reporting core must have at least %s logging level",
```
jcorbin: Don't use `%s` and `x.String()` directly, just `fmt.Sprintf("%v", x)` (by any other alias, errorf in this case)

zapcore/sampler.go
```diff
@@ -62,33 +102,42 @@ func fnv32a(s string) uint32 {
...
+	c.counter.Store(currentVal)
 
 	newResetAfter := tn + tick.Nanoseconds()
 	if !c.resetAt.CAS(resetAfter, newResetAfter) {
 		// We raced with another goroutine trying to reset, and it also reset
 		// the counter to 1, so we need to reincrement the counter.
-		return c.counter.Inc()
+		return c.counter.Inc(), 0
 	}
+	return
```
jcorbin: Naked returns are generally frowned on, prefer the explicit `return currentVal, preResetVal`

> Naked returns are okay if the function is a handful of lines. Once it's a medium sized function, be explicit with your return values. 

https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

zapcore/sampler.go
```diff
@@ -126,9 +183,47 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
...
 	counter := s.counts.get(ent.Level, ent.Message)
-	n := counter.IncCheckReset(ent.Time, s.tick)
+	n, prevN := counter.IncCheckReset(ent.Time, s.tick)
 	if n > s.first && (n-s.first)%s.thereafter != 0 {
 		return ce
 	}
+	if s.reporting.enabled {
+		err := s.reporting.report(s.reporting.core, ent, s.getSampleCount(prevN))
+		if err != nil {
+			fmt.Printf("%v sampling report write error: %v\n", time.Now(), err)
```
jcorbin: This is still writing directly to stdout, you need to at least get a hold of `zap.Logger.errorOutput` and write to it instead (especially if we cannot propagate the error).

nit: always prefer tighter `if err := f(); err != nil {}` scoping when possible.

zapcore/sampler.go
```diff
@@ -62,28 +71,68 @@ func fnv32a(s string) uint32 {
...
-		return c.counter.Inc()
+		return c.counter.Inc(), 0
 	}
+	return currentVal, preResetVal
+
+}
+
+// Reporter defines report function that generates and writes sampling activity
+// report log entries.
+type Reporter interface {
```
jcorbin: this almost certainly doesn't need to be exported (its only method is unexported, so it can't be implemented outside the zapcore package anyhow; which is the right thing to do, since we really don't want to grow zap's supported surface area if possible)

zapcore/sampler.go
```diff
@@ -126,9 +183,48 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
...
 
 	counter := s.counts.get(ent.Level, ent.Message)
-	n := counter.IncCheckReset(ent.Time, s.tick)
+	n, prevN := counter.IncCheckReset(ent.Time, s.tick)
 	if n > s.first && (n-s.first)%s.thereafter != 0 {
 		return ce
 	}
+	if err := s.Reporter.report(ent, s.getSampleCount(prevN)); err != nil {
+		if ce.ErrorOutput != nil {
+			fmt.Fprintf(ce.ErrorOutput, "%v sampling report write error: %v\n", time.Now(), err)
```
jcorbin: you need to sync it after write: `ce.ErrorOutput.Sync()`

zapcore/sampler.go
pavel-shpilev: The idea is to allow users write their own Reporter if/when there's a need to customize sampling report entry

zapcore/sampler.go
```diff
@@ -29,6 +30,14 @@ import (
 const (
 	_numLevels        = _maxLevel - _minLevel + 1
 	_countersPerLevel = 4096
+
+	// ReportingLoggerName is the default name of logger that generates sampling
```
prashantv: can we avoid exporting these (as they become part of our API that we have to support forever, which means we can't move them into a subpackage in future, etc)

instead, let's document it in the sampler config?

zapcore/sampler.go
prashantv: since we don't export this functionality today, let's unexport till needed

config_test.go
```diff
@@ -80,6 +111,22 @@ func TestConfig(t *testing.T) {
 			for i := 0; i < 200; i++ {
 				logger.Info("sampling")
 			}
+
+			// Rolling over tick to produce sampling report.
+			if tt.expectSamplingN > 0 {
+				ztest.Sleep(time.Second)
```
prashantv: this is a long sleep which will slow down the test execution. In general, use of sleep is bad for tests (slow and also cause flakiness). I think we can't avoid it for the sampler tests currently, but the other tests are doing milliseconds rather than seconds to avoid a large impact to the test suite runtime.

zapcore/sampler.go
```diff
@@ -126,9 +183,49 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
 	}
 
 	counter := s.counts.get(ent.Level, ent.Message)
-	n := counter.IncCheckReset(ent.Time, s.tick)
+	n, prevN := counter.IncCheckReset(ent.Time, s.tick)
 	if n > s.first && (n-s.first)%s.thereafter != 0 {
 		return ce
 	}
+	if err := s.Reporter.Report(ent, s.getSampleCount(prevN)); err != nil {
```
prashantv: not sure if we can make a change here, but it's a little unexpected that `Check` does a write here.

should we instead add hooks into `CheckedEntry` that trigger a function to run on `Write`?

prashantv (commented):
I only just realized that the sampler reporting only happens if the user logs the same level + message again in the future. However, this isn't guaranteed and it may happen a lot later, in which case it may not help much. The advantage of doing it at that point is we can give an accurate number of logs that were dropped.

If the goal is to just be aware that sampling occurred, I think logging as soon as any sampling happens is actually more beneficial. If the goal is to know exactly how much sampling occurred, then the current approach is more useful.

We probably should make sure we can support both in future, but for now, what is the goal?

zapcore/sampler.go
jcorbin: Hmm, but wouldn't that only write during the next unsampled log entry, which could be much later as you point out? Maybe that's okay? Again it all comes down to: what actually is the goal here.

At any rate, `Check` doing a write is mildly less surprising than `With` doing a write as was the initial version.

prashantv: If we wanted to be aware that sampling was happening (e.g., log as soon as we know we'll sample), then I think we could make it work by either:

 * Having the log that gets to the sampling threshold also log "all future entries will be sampled". The con is that we'll log this even if there were no other sampled logs. It could get noisy for low sampling rates (like 1 a second) even if the event only happens once.
 * Having the first log that would be sampled returning a checked entry, but the write doesn't write the original log, but a "dropping log due to sample". This might be surprising since the entry that is written isn't the entry that was checked.

There could also be a slightly different approach, where the original log entry has an added field like "sampleLimitHit" or something, which makes the first option less intrusive.

PR #697, opened 2019-04-11
Support configurable delimiter for console encoder

lixingwang:
It would be nice to have a configurable delimiter for console encoder, tab by default.

For our cases, we prefer to have space as element delimiter.

A custom console delimiter can be set through the `consoleSeparator` encoder configuration.


jcorbin:
Not getting into a deep/full review at this point, but directionally: we probably don't want to add new ways of configuring zap (e.g. environment variable conventions like this). Instead this should just be directly added to the console encoder config, and then integration with environment can be done by the application.

lixingwang:
@jcorbin Agree! env control should done at end-user application. I have made delimiter in encode config, please have a look. thanks.

lixingwang:
@jcorbin or anyone, could you help to take a look this PR please?

zapcore/console_encoder.go
```diff
@@ -22,10 +22,9 @@ package zapcore
 
 import (
 	"fmt"
-	"sync"
-
 	"go.uber.org/zap/buffer"
 	"go.uber.org/zap/internal/bufferpool"
+	"sync"
```
jcorbin: nit: un-churn this import move

zapcore/encoder.go
```diff
@@ -239,6 +239,8 @@ type EncoderConfig struct {
 	// Unlike the other primitive type encoders, EncodeName is optional. The
 	// zero value falls back to FullNameEncoder.
 	EncodeName NameEncoder `json:"nameEncoder" yaml:"nameEncoder"`
+	//More on console, the delimiter of each element
```
jcorbin: - let's name it `ConsoleFoo` then rather than generically `ElementFoo`
- doc string needs to be formatted / written better, as in `// Configures the field separator used by the console encoder.` (note: leading space, final period, and phrased as a complete sentence)
- nit: in my view, this should be called a "separator" rather than a "delimiter", so `ConsoleSeparator` instead of `ElementDelimiter`
- corollary: I care less about s/delimiter/separator/ in the internal code paths above, more so in this exported struct; but probably best for to do so for internal consistency

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,56 @@
...
+// THE SOFTWARE.
+package zapcore_test
+
+import (
+	"github.com/stretchr/testify/assert"
+	. "go.uber.org/zap/zapcore"
+	"testing"
+)
+
+func TestSetConsoleElementDelimiter(t *testing.T) {
```
jcorbin: - consider making this a https://github.com/golang/go/wiki/TableDrivenTests
- add explicit coverage for the default path (delimiter unset / set to 0, default to tab)

zapcore/console_encoder.go
lixingwang: Thanks @jcorbin Updated

zapcore/console_encoder_test.go
lixingwang: @jcorbin  Done

zapcore/encoder.go
lixingwang: @jcorbin  Thanks for details instruction,  I made the changes, please have another look.

lixingwang:
@jcorbin  Thanks for your details review comments,  I updated code, Please have another look

lixingwang:
@jcorbin  Any chance to have another look?

lixingwang:
@abhinav  @prashantv  @jcorbin Could you please have a look?

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,87 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package zapcore_test
+
+import (
+	"github.com/stretchr/testify/assert"
+	. "go.uber.org/zap/zapcore"
+	"testing"
```
abhinav: Would you mind running goimports over this. The imports should be in two
groups: standard library, everything else.

zapcore/console_encoder.go
```diff
@@ -56,6 +56,10 @@ type consoleEncoder struct {
 // encoder configuration, it will omit any element whose key is set to the empty
 // string.
 func NewConsoleEncoder(cfg EncoderConfig) Encoder {
+	if cfg.ConsoleSeparator == 0 {
+		//Default delimiter '\t'
```
abhinav: nit:

```suggestion
		// Default delimiter '\t'
```

zapcore/encoder.go
```diff
@@ -239,6 +239,8 @@ type EncoderConfig struct {
 	// Unlike the other primitive type encoders, EncodeName is optional. The
 	// zero value falls back to FullNameEncoder.
 	EncodeName NameEncoder `json:"nameEncoder" yaml:"nameEncoder"`
+	// Configures the field separator used by the console encoder.
```
abhinav: Let's mention the default.

```suggestion
	// Configures the field separator used by the console encoder. Defaults
	// to tab.
```

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,87 @@
...
+		Level:      InfoLevel,
+		Message:    `hello`,
+		Time:       _epoch,
+		Stack:      "fake-stack",
+		Caller:     EntryCaller{Defined: true, File: "foo.go", Line: 42},
+	}
+)
+
+func TestConsoleSeparator(t *testing.T) {
+
```
abhinav:
```suggestion
```

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,87 @@
...
+	tests := []struct {
+		desc            string
+		cfg             EncoderConfig
```
abhinav: I think this can be simplified slightly by changing this field to a `separator byte`.

```
tests := []struct{
	desc        string
	separator   byte
	wantConsole string
}{
	...
}
```

Omit the field from the struct for the default test case and set up the
configuration for the encoder in the test itself.

```
cfg := testEncoderConfig()
cfg.ConsoleSeparator = tt.separator
```

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+		{
+			desc:            "dash console separator",
+			cfg:             encoderTestEncoderConfig('-'),
+			expectedConsole: "0-info-main-foo.go:42-hello\nfake-stack\n",
+		},
+	}
+
+	for i, tt := range tests {
+		console := NewConsoleEncoder(tt.cfg)
```
abhinav: We should use subtests here for better error messages.

```
for _, tt := range tests {
	t.Run(tt.desc, func(t *testing.T) {
		...
	})
}
```


Then the individual `assert` statements don't need to include the test case
description or index.

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,87 @@
...
+		entry := testEntry
+		consoleOut, consoleErr := console.EncodeEntry(entry, nil)
```
abhinav: The error should be called just `err` unless there's a need to maintain a
reference to an existing `err` that's already in scope.

vosst:
@lixingwang @abhinav Seems like this PR has stalled. Happy to pick up the change unless either of you has objections/plans to complete it :smile:

lixingwang:
@vosst  thanks for reminding.  I am totally lost this PR.  let me update it in the coming days.

lixingwang:
Hi, @abhinav and @vosst please approve.

vosst:
> Hi, @abhinav and @vosst please approve.

Nice, thank you :smile: @prashantv @jcorbin @abhinav probably worth taking another look.

zapcore/encoder.go
```diff
@@ -329,6 +329,9 @@ type EncoderConfig struct {
 	// Unlike the other primitive type encoders, EncodeName is optional. The
 	// zero value falls back to FullNameEncoder.
 	EncodeName NameEncoder `json:"nameEncoder" yaml:"nameEncoder"`
+	// Configures the field separator used by the console encoder. Defaults
+	// to tab.
+	ConsoleSeparator byte `json:"consoleSeparator" yaml:"consoleSeparator"`
```
prashantv, line 334: Should we use a string type for the separator so it's more flexible in the future? A `byte` is efficient, but will have issues if there's some non-ASCII character that someone would like to use.

cc @abhinav what do you think?

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,89 @@
...
+		wantConsole string
+	}{
+		{
+			desc:        "space console separator",
+			separator:   ' ',
+			wantConsole: "0 info main foo.go:42 hello\nfake-stack\n",
+		},
+		{
+			desc: "default console separator",
+			separator: '	',
```
prashantv, line 53: is this the same as below (just with a literal tab vs `\t`)? If so, can we just keep the `\t`, since it's only really testing the difference in how go handles literals

zapcore/console_encoder.go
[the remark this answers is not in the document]
lixingwang: I did go import for this one not sure why it gets moved.  let me revert that. thx

zapcore/encoder.go
lixingwang: Agree, As of today, I think the string makes more sense for the future.  @prashantv @abhinav  let me know if I should go ahead to change to the string.

abhinav: Yeah, a string separator makes sense.

lixingwang: @abhinav @prashantv  done please have another look

zapcore/console_encoder.go
```diff
@@ -56,6 +56,10 @@ type consoleEncoder struct {
 // encoder configuration, it will omit any element whose key is set to the empty
 // string.
 func NewConsoleEncoder(cfg EncoderConfig) Encoder {
+	if len(cfg.ConsoleSeparator) <= 0 {
+		// Default delimiter '\t'
```
prashantv, line 60:
```suggestion
		// Use a default delimiter of '\t' for backwards compatibility
```

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,89 @@
...
+			wantConsole: "0\tinfo\tmain\tfoo.go:42\thello\nfake-stack\n",
+		},
+		{
+			desc:        "tag console separator",
+			separator:   "\t",
+			wantConsole: "0\tinfo\tmain\tfoo.go:42\thello\nfake-stack\n",
+		},
+		{
+			desc:        "dash console separator",
+			separator:   "-",
```
prashantv, line 63: can we use a multi-line string in a test,  maybe this one?

lixingwang: @prashantv I don't think it is necessary, test cases in encoder_test.go do the same. we should keep consistent. Let me know otherwise.

zapcore/console_encoder.go
```diff
@@ -56,6 +56,10 @@ type consoleEncoder struct {
 // encoder configuration, it will omit any element whose key is set to the empty
 // string.
 func NewConsoleEncoder(cfg EncoderConfig) Encoder {
+	if len(cfg.ConsoleSeparator) <= 0 {
```
abhinav, line 59: minor: len can never be `< 0`.

```suggestion
	if len(cfg.ConsoleSeparator) == 0 {
```

zapcore/console_encoder_test.go
abhinav: I don't feel strongly about this. I think multi-line strings here would aid in readability.

If we don't want to mess up the indentation, an option is something like the
following:

```
func multiline(lines ...string) string {
  return strings.Join(append(lines, ""), "\n")
}
```

Used like so:

```
wantConsole: multiline(
  "0\tinfo\tmain\tfoo.go:42\thello",
  "fake-stack",
),
```

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,89 @@
...
+			wantConsole: "0-info-main-foo.go:42-hello\nfake-stack\n",
+		},
+	}
+
+	for _, tt := range tests {
+		console := NewConsoleEncoder(encoderTestEncoderConfig(tt.separator))
+		t.Run(tt.desc, func(t *testing.T) {
+			entry := testEntry
+			consoleOut, err := console.EncodeEntry(entry, nil)
+			assert.Nil(t, err)
```
abhinav, line 73: We should use `assert.NoError` here, and if it is non-nil, skip the following assert.Equal.

```suggestion
			if !assert.NoError(t, err) {
				return
			}
```

zapcore/console_encoder_test.go
```diff
@@ -0,0 +1,89 @@
...
+			assert.Equal(
+				t,
+				tt.wantConsole,
+				consoleOut.String(),
+				"Unexpected console output: expected to %v.", tt.desc,
```
abhinav, line 78: minor: since tt.desc is part of the test name, we don't need it in the error message again.
```suggestion
				"Unexpected console output",
```

zapcore/console_encoder_test.go
lixingwang: Let keep what it is now.  Everyone should know what `\n` means here :)

prashantv: Sorry, I misspoke, what I actually meant was a multi-character string, so instead of just `-`, if we used `--`.

That would verify that we're not just adding the first character of the separator. Sorry for the confusion!

lixingwang: Make sense.  Done.   @prashantv  Please just merge it if you are ok.

PR #707, opened 2019-04-29
Add Go 1.12 for Travis

prashantv:
Makefile rules have to change for 1.12 support (can't use go tool vet).

Drops 1.10 as we only support current and previous release.


.travis.yml
```diff
@@ -3,6 +3,7 @@ sudo: false
 go:
   - 1.10.x
```
abhinav: we can probably drop 1.10 

```suggestion
```

PR #710, opened 2019-05-01
remove dep and bench for lion

lvjing2:
Current, the lion had been down for a long time, https://github.com/peter-edge/lion-go/issues/5
And I suggest to remote bench for this.


prashantv:
Done as part of #720

PR #720, opened 2019-06-05
Re-run benchmarks and add % over zap measurement

glibsm:
Benchmarks were last run 2y ago, so it's time to see how the
lay of the land has changed.

Just baseline wise it appears that my laptop is about 50%
faster than the hardware that the initial benchmarks were
ran on, but it doesn't really change the relative difference
in the loggers.

Zerolog appears to have closed the distance quite a bit,
while others are still an order of magnitude slower than
zap.


jcorbin:
For kicks, I pulled your branch to try to reproduce results / give you another modern comparison point, but I get:
```
[ERROR] Update failed for go.pedge.io/lion: The Remote does not match the VCS endpoint
[ERROR] Failed to install: The Remote does not match the VCS endpoint
```

How do feel about switching to dep while we're at it?

abhinav:
According to #710, the repo's been down a while. I'm okay removing that benchmark.

glibsm:
I did see #710, but it seems like my local glide worked just fine. I'll drop lion I guess? 🤷‍♂ 

> How do feel about switching to dep while we're at it?

I'm not sure that dep alone would help here, as it seems to be an upstream issue. I'm not terribly opposed to switching, but probably not part of this PR.

Also remember that dep doesn't do the whole test imports thing so in order to keep the dependency tree clean, I think booting out the zap benchmarks into a separate repo may be desired.

glibsm:
Since there is so much variation in baseline hardware (my personal mbp 13 from a few years ago is like 3x slower than the work laptop in these benches) I added a relative column which should provide better visibility into how the loggers actually stack up.

As mentioned earlier zerolog appears to perform better in certain benches, but otherwise zap blows everything else away. Might be worthwhile to look into why and how zerolog is faster, perhaps some lessons there.

image

jcorbin:
+1 relative column; your newly-reran resuts are >= 2x faster than my aging mid 2015 (escape key toting) mbp

glibsm:
> +1 relative column; your newly-reran resuts are >= 2x faster than my aging mid 2015 (escape key toting) mbp

This is on 2018 15".

I'm happy to re-run like dockerized with limited resources or something, but I'm not sure the baseline numbers really matter as much as the relative column.

glide.yaml
```diff
@@ -22,12 +22,11 @@ testImport:
 - package: github.com/mattn/goveralls
 - package: github.com/pborman/uuid
 - package: github.com/pkg/errors
-- package: go.pedge.io/lion
 - package: github.com/rs/zerolog
 - package: golang.org/x/tools
   subpackages:
   - cover
-- package: github.com/golang/lint
+- package: golang.org/x/lint
```
abhinav: Thanks for fixing this!

internal/readme/readme.go
```diff
@@ -142,12 +152,21 @@ func getBenchmarkRow(input []string, benchmarkName string, libraryName string) (
...
 	}
-	return &benchmarkRow{
-		libraryNameToMarkdownName[libraryName],
-		duration,
-		allocatedBytes,
-		allocatedObjects,
-	}, nil
+	r := &benchmarkRow{
+		Name: libraryNameToMarkdownName[libraryName],
+
```
abhinav: nit: delete empty line

```suggestion
```

prashantv:
Adding relative is a great idea, and definitely more useful. I tried to reproduce your numbers, but my relative numbers are a little different, especially logrus and log15 which show 3000% vs 1500% in mine.

I wonder if we have some sort of concurrency that's causing this? I noticed the benchmark used anywhere between 200 and 400% on my machine.

-----
Log a message and 10 fields:

| Package | Time | Time % to zap | Objects Allocated |
| :------ | :--: | :-----------: | :---------------: |
| :zap: zap | 2121 ns/op | +0% | 5 allocs/op
| :zap: zap (sugared) | 2503 ns/op | +18% | 11 allocs/op
| zerolog | 6434 ns/op | +203% | 76 allocs/op
| go-kit | 9245 ns/op | +336% | 105 allocs/op
| apex/log | 31056 ns/op | +1364% | 115 allocs/op
| logrus | 32328 ns/op | +1424% | 125 allocs/op
| log15 | 34025 ns/op | +1504% | 122 allocs/op

Log a message with a logger that already has 10 fields of context:

| Package | Time | Time % to zap | Objects Allocated |
| :------ | :--: | :-----------: | :---------------: |
| :zap: zap | 215 ns/op | +0% | 0 allocs/op
| :zap: zap (sugared) | 293 ns/op | +36% | 2 allocs/op
| zerolog | 144 ns/op | -33% | 0 allocs/op
| go-kit | 7933 ns/op | +3590% | 103 allocs/op
| log15 | 20268 ns/op | +9327% | 73 allocs/op
| apex/log | 29399 ns/op | +13574% | 104 allocs/op
| logrus | 31134 ns/op | +14381% | 113 allocs/op

Log a static string, without any context or `printf`-style templating:

| Package | Time | Time % to zap | Objects Allocated |
| :------ | :--: | :-----------: | :---------------: |
| :zap: zap | 231 ns/op | +0% | 0 allocs/op
| :zap: zap (sugared) | 274 ns/op | +19% | 2 allocs/op
| zerolog | 134 ns/op | -42% | 0 allocs/op
| go-kit | 362 ns/op | +57% | 11 allocs/op
| standard library | 379 ns/op | +64% | 2 allocs/op
| apex/log | 1807 ns/op | +682% | 10 allocs/op
| logrus | 3045 ns/op | +1218% | 24 allocs/op
| log15 | 4650 ns/op | +1913% | 23 allocs/op

glibsm:
Looks like our baselines are a bit off. I'm not sure why.

If we really want more accuracy into this more control over the hardware is required. We can have travis-ci settle the score :D Just turn this into a test that prints out stdout or something.

Alternatively we can also just ignore if it's 15x or 30x slower... bottom line is it's much slower :)

glibsm:
Alternatively some more sophistication like https://github.com/aclements/perflock to make sure that the CPU is not distracted by other things running in background, especially with all the modern electron gubbins that we may have open.

PR #723, opened 2019-06-19
Delete lion benchmarks

pedgeio:
I believe this benchmark is effectively responsible for the majority of traffic to go.pedge.io - go.pedge.io/lion is deprecated and shouldn't be used anyways, and this doesn't add anything.

Instead of redoing all the benchmarks with `make updatereadme`, I just manually deleted the lion benchmarks - this has the same effect as `make updatereadme` except it does not update the other benchmarks (I verified this).


prashantv:
I think #720 already does this, so we should merge that, then this won't be needed.

pedgeio:
OK great - the sooner you can merge that the better, I'd like to reduce traffic to this website

prashantv:
Just merged #720 :)

pedgeio:
Thanks!

PR #733, opened 2019-08-01
Make encoding binary configurable in EncoderConfig

### LINKED ISSUE #730: Change bytes encoding in JSON Encoder
By default, `jsonEncoder` encodes bytes using base64. In my case, it would be much more useful if bytes are encoded using hex encoding.

I tried to build a wrapper around `jsonEncoder`:
```go
type jsonHexEncoder struct {
	zapcore.Encoder
}

func NewJSONHexEncoder(cfg zapcore.EncoderConfig) zapcore.Encoder {
	jsonEncoder := zapcore.NewJSONEncoder(cfg)
	return &jsonHexEncoder{
		Encoder: jsonEncoder,
	}
}

func (enc *jsonHexEncoder) AddBinary(key string, val []byte) {
	enc.AddString(key, "0x"+hex.EncodeToString(val))
}
``` 

The problem is that not only `Clone()` needs to be overridden (this is easy) but also `EncodeEntry()` which is not really possible without copying a huge chunk of code.

I see two options:
* create a new field type `zap.BinaryHex`, 
* configure bytes encoding in the `NewJSONEncoder()` constructor, or, for backward compatibility, provide a new one `NewJSONHexEncoder()` or something similar.

I would like to hear your opinion about this. Maybe there is a simpler solution I haven't noticed so far.

adambabik:
This change allows to configure how binary fields are serialized by `Encoder`. It is backward compatible and defaults to base64 encoding.

Closes #730

### DIFF
--- a/config.go
+++ b/config.go
@@ -102,6 +102,7 @@ func NewProductionEncoderConfig() zapcore.EncoderConfig {
 		EncodeLevel:    zapcore.LowercaseLevelEncoder,
 		EncodeTime:     zapcore.EpochTimeEncoder,
 		EncodeDuration: zapcore.SecondsDurationEncoder,
+		EncodeBinary:   zapcore.Base64BinaryEncoder,
 		EncodeCaller:   zapcore.ShortCallerEncoder,
 	}
 }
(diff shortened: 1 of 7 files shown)


prashantv (commented):
This change cannot be accepted as-is as it modifies the `ArrayEncoder` interface which makes it backwards incompatible.

PR #734, opened 2019-08-06
feat:allow set logger name by option function

detailyang:
Hello.

Now zap can set logger name by `logger.Named(xx)` only and wish can provide a better way like AddCallerSkip API Style to set name:)


prashantv:
Thanks for the contribution @detailyang.

Can you provide more context on where you expect to use this?

We intentionally designed the name/field APIs such that you can only add context, and it's not possible to drop context. For example, if I pass in a logger named "sync.flush" to some library call, I don't want it to replace that name with something that it chooses, it should only be appended to.

detailyang:
Hello @prashantv 

My scenario is initialized the logger name which is empty default.  Current, what I do is as the following:

````golang
	buf := &bytes.Buffer{}
	encoder := zapcore.NewConsoleEncoder(NewDevelopmentEncoderConfig())
	core := zapcore.NewCore(encoder, zapcore.AddSync(buf), zapcore.DebugLevel)
       	opts := []zap.Option{
		zap.AddCaller(),
		zap.AddCallerSkip(1),
		zap.AddStacktrace(zap.FatalLevel),
	}
	logger := zap.New(core, opts...)
        logger = logger.Named("xxx")
````

Well, `logger.Named` is enough but the same as zap.AddXXXAPI style

tsingson:
> Hello @prashantv
> 
> My scenario is initialized the logger name which is empty default. Current, what I do is as the following:
> 
> ```go
> 	buf := &bytes.Buffer{}
> 	encoder := zapcore.NewConsoleEncoder(NewDevelopmentEncoderConfig())
> 	core := zapcore.NewCore(encoder, zapcore.AddSync(buf), zapcore.DebugLevel)
>        	opts := []zap.Option{
> 		zap.AddCaller(),
> 		zap.AddCallerSkip(1),
> 		zap.AddStacktrace(zap.FatalLevel),
> 	}
> 	logger := zap.New(core, opts...)
>         logger = logger.Named("xxx")
> ```
> 
> Well, `logger.Named` is enough but not the same as zap.AddXXXAPI style

as you said,  logger.named is enough.

Issue #659, opened 2018-12-14
Add support for RFC3339Nano as another zapcore.TimeEncoder


nattawitc:
zapcore package has implemented 4 types of TimeEncoder (Epoch, EpochMillis, EpochNanos, and ISO8601). Is it possible to add RFC3339 and RFC3339Nano or any other standard format in go's time package. I've seen this implemented in #276 , but I couldn't find it anymore in the current version.

Another solution is to implement a custom TimeEncoder ourself
`func rfc3339TimeEncoder(t time.Time, enc zapcore.PrimitiveArrayEncoder) {
  enc.AppendString(t.Format(time.RFC3339Nano))
}`
but I'm not sure if it will cause any further issues as TimeEncoder.UnmarshalText does not support custom one.

PR #736, opened 2019-08-18
Add RFC3339, RFC3339Nano and update docs to include 'nano'

uhthomas:
I was going to open an issue but #659 is already open.

Go's time package suggests that 'RFC3339 should be preferred for new protocols' so it would be great to have RFC3339 as a standard time format in zap.


uhthomas:
@prashantv Absolutely! I've added some unit tests :)

uhthomas:
Awesome, thanks for the approval @prashantv! What's required to get this merged?

prashantv:
Nothing else needed, I'll merge now :)

Issue #747, opened 2019-10-21
Introduce TRACE log level


mr-kashif:

jcorbin:
Without much detail, I can't conclude that this isn't a dupe of #680; please comment there with any distinctive needs not already covered, or re-open this one if it's truly different.

Issue #680, opened 2019-02-11
Support custom log levels/general logging.


NathanZook:
My company wants to implement a trace level log.  Currently, this is problematic because zap.Logger.check is private.

Digging in, it would seem that a full implementation would extract https://github.com/uber-go/zap/blob/master/logger.go#L273-L282 into a method call.  Of course, the natural target of the call is probably the levels themselves, which would be a problem.  Making it a method on *zap.Logger would do, I think.

Ideally, we would probably want to change the zapcore.Level constants to variables, and implement a zap.Logger#Log method that takes a zapcore.Level as its second argument.

I know of one major user of Go that implements TRACE1, TRACE2, and TRACE3...

iorlas:
+1 for custom levels. Maybe custom levels not argent right now, but at least adding Trace level would be soooo goood.

lysu:
+1 to add trace level

kolkov:
Yes, it’s very good if there are more levels.

mpfaff:
Is anything ever going to be done about this? At a minimum, a trace log level is a must, but support for custom levels would be the ideal solution.

prashantv:
It's not clear whether this can be done in a fully backwards compatible way (as existing cores and encoders may break). For example, the sampler panics when a negative level is passed in (#713). While the sampler can be fixed because it's part of the zap repository, there may be other implementations that would not support new levels.

We are open to PRs that implement custom levels in a backwards compatible way (ideally with an analysis of existing implementations of zap cores/encoders to ensure that it is not breaking). Unfortunately we can't accept any backwards incompatible changes to zap 1.0

tv42:
Stackdriver has a few more levels on the severe end:

```
    // Alert means a person must take an action immediately.
    Alert = Severity(logtypepb.LogSeverity_ALERT)
    // Emergency means one or more systems are unusable.
    Emergency = Severity(logtypepb.LogSeverity_EMERGENCY)
```

https://godoc.org/cloud.google.com/go/logging#Severity

alxn:
> Stackdriver has a few more levels on the severe end:
> 
> ```
>     // Alert means a person must take an action immediately.
>     Alert = Severity(logtypepb.LogSeverity_ALERT)
>     // Emergency means one or more systems are unusable.
>     Emergency = Severity(logtypepb.LogSeverity_EMERGENCY)
> ```
> 
> https://godoc.org/cloud.google.com/go/logging#Severity

[...]

PR #749, opened 2019-10-24
Add New Trace Logging Level

LeviMatus:
A few people have expressed interest in having a TraceLevel added to Zap (Some mention of this in #680 and again in #747). This PR addresses that interest by introducing this logging level.

### DIFF
--- a/level.go
+++ b/level.go
@@ -26,6 +26,10 @@ import (
 )
 
 const (
+	// TraceLevel logs are especially granular log messages that are used in development and debugging
+	// situations in which tracing of function inputs at outputs are necessary. Almost always disabled in production
+	// due to performance hits associated with very voluminous logs.
+	TraceLevel = zapcore.TraceLevel
 	// DebugLevel logs are typically voluminous, and are usually disabled in
 	// production.
 	DebugLevel = zapcore.DebugLevel
--- a/level_test.go
+++ b/level_test.go
@@ -35,6 +35,7 @@ func TestLevelEnablerFunc(t *testing.T) {
 		level   zapcore.Level
 		enabled bool
 	}{
+		{TraceLevel, false},
 		{DebugLevel, false},
 		{InfoLevel, true},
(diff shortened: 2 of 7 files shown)


prashantv:
I'm not sure that we can add a level to zap 1.0 and retain backwards compatibility. Existing cores, hooks, etc do not know about the trace level and may not handle it properly. For example, here's an adapter library that checks for known levels:

https://github.com/uber-common/bark/blob/master/zbark/zapify.go#L64

That library at least handles it will by returning an error, but there may be other cases where it panics.

I don't think we can add any new levels unless we do a zap 2.0.

LeviMatus:
Thanks for the code reference there to check against. Having external wrappers is something I had not thought of and is a great point. I see now how this could make introducing new levels more difficult.

Moreover, I realized that I forgot to add Trace methods to `sugar.go`, so the PR can't even be considered complete at this time.

novabyte:
@LeviMatus You've done great work on this pull request! It's sorely needed within Zap even if we cannot specify our own set of severity levels for output.

@prashantv Is there any kind of timeline for when a set of changes like this would be acceptible to the release constraints of Zap?

prashantv:
> @prashantv Is there any kind of timeline for when a set of changes like this would be acceptible to the release constraints of Zap?

We don't currently have a timeline for zap 2.0, but the improvements we'd like to make in a 2.0 are tracked here: #388.

We don't currently have a concrete timeline for a 2.0 though.

se3000:
Any update on this? We would like to use it.

Antonboom:
What's the bottom line? Wait for zap v2?

prashantv:
We definitely can't add it in 1.0. It's possible for 2.0, but that's not in the roadmap. At that point, we'd want to consider the cost/benefit (it's added cognitive overhead to add another level, so we should ensure there's a significant benefit).

Closing for 1.0.

PR #742, opened 2019-09-25
modules: include go module information

cep21:

PR #751, opened 2019-10-29
Add support for Go modules

### LINKED ISSUE #639: Go Modules
Hello,
Are you plan to support Go Modules with go.mod file?

Thx in adv

abhinav:
This switches Zap to using Go modules instead of Glide for managing
dependencies, while keeping the glide.yaml around to provide
constraints for legacy ecosystems.

The change ended up being more complicated than a `go mod init` because
Zap has a few tests that verify its behavior inside `GOPATH` and a
vendor/ directory. Those tests were passing fine in CI for #742 because
of the `go_import_path` directive in the Travis configuration which
places it under `$GOPATH/src/go.uber.org/zap`. This isn't necessary
with Go modules but I've left it in-place for now.

Note that to keep the direct dependencies of Zap small, the benchmarks/
directory was turned into its own Go module with a `replace` clause
pointing to the local copy of Zap. We will not tag a release of the
benchmarks/ module.

Supersedes #742
Resolves #639

---

Note to reviewers: Each commit is meant to be reviewed individually.
The change should not be squashed when merged.


stacktrace_ext_test.go
```diff
@@ -84,33 +85,35 @@ func TestStacktraceFiltersZapMarshal(t *testing.T) {
...
+			setupSymlink(t, dep.Dir, filepath.Join(vendorDir, dep.ImportPath))
 		}
 
-		// Now run the above test which ensures we filter out zap stacktraces, but this time
-		// zap is in a vendor
+		// Now run the above test which ensures we filter out zap
+		// stacktraces, but this time zap is in a vendor
 		cmd := exec.Command("go", "test", "-v", "-run", "TestStacktraceFiltersZap")
 		cmd.Dir = testDir
+		cmd.Env = envWith(t, "GO111MODULE=off")
```
prashantv: i think this could be simplified to:
```
cmd.Env = append(os.Environ(), "GO11MODULE=off")
```

since duplicate entries causes the last value to be used;
>     // If Env contains duplicate environment keys, only the last
>     // value in the slice for each duplicate key is used.
https://golang.org/pkg/os/exec/#Cmd

stacktrace_ext_test.go
```diff
@@ -162,3 +165,64 @@ func setupSymlink(t *testing.T, src, dst string) {
...
+	ImportPath string `json:"Path"` // import path of the dependency
+	Dir        string `json:"Dir"`  // location on disk
+}
+
+// Downloads all dependencies for the current Go module and reports their
+// module paths and locations on disk.
+func downloadDependencies(t *testing.T) []dependency {
+	cmd := exec.Command("go", "mod", "download", "-json")
+
+	stdout, err := cmd.StdoutPipe()
```
prashantv: is streaming worth it here, suggest simplifying to:
```
stdout, err := cmd.Output()
require.NoError(t, err, ...)

dec := json.NewDecoder(bytes.NewReader(stdout))
```

Makefile
```diff
@@ -1,74 +1,67 @@
...
+	-o -name '*.go' -print | cut -b3-)
 
-# The linting tools evolve with each Go version, so run them only on the latest
-# stable release.
+# The linting tools evolve with each Go version, so run them only on the
+# latest stable release.
 GO_VERSION := $(shell go version | cut -d " " -f 3)
 GO_MINOR_VERSION := $(word 2,$(subst ., ,$(GO_VERSION)))
-LINTABLE_MINOR_VERSIONS := 12
+LINTABLE_MINOR_VERSIONS := 13
```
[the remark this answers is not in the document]
abhinav: Reasonable. That's what we do in the other projects. Switching.

Makefile
```diff
@@ -1,74 +1,67 @@
...
 	@echo "Skipping linters on" $(GO_VERSION)
 endif
 
+$(GOLINT):
+	go install golang.org/x/lint/golint
+
 .PHONY: test
 test:
-	go test -race $(PKGS)
+	go test -race ./...
```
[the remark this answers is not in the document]
abhinav: Adding `foreach` for `make test` but not for `cover` because then we'll have to deal with cross-module coverage merging. Adding a comment that we track coverage only for the main module.

stacktrace_ext_test.go
abhinav: 🤦‍♀

stacktrace_ext_test.go
abhinav: Yeah, streaming is necessary here. The output contains a series of newline-delimited JSON objects.

Per `go help mod download`,

```
By default, download reports errors to standard error but is otherwise silent.
The -json flag causes download to print a sequence of JSON objects
to standard output, describing each downloaded module (or failure),
corresponding to this Go struct:

    type Module struct {
        Path     string // module path
        Version  string // module version
        Error    string // error loading module
        Info     string // absolute path to cached .info file
        GoMod    string // absolute path to cached .mod file
        Zip      string // absolute path to cached .zip file
        Dir      string // absolute path to cached source root directory
        Sum      string // checksum for path, version (as in go.sum)
        GoModSum string // checksum for go.mod (as in go.sum)
    }
```

abhinav:
Updated based on comments.

stacktrace_ext_test.go
prashantv: We need to use `json.NewDecoder` since the output has multiple objects, but I don't think we need to use a pipe -- we can get all the output at once and then use `json.NewDecoder`

abhinav: Ah, I see what you mean. Switching.

PR #754, opened 2019-10-31
Add LogWithLevel method to logger

tmshn:
## This PR introduces...

`LogWithLevel(lvl, msg, fields...)` method, which logs messages at specified log level.

This can be seen as shorthand for (but more appropriate caller depth to [`check()`](https://github.com/uber-go/zap/blob/a6015e13fab9b744d96085308ce4e8f11bad1996/logger.go#L256-L259))

```go
if ce := logger.Check(lvl, msg); ce != nil {
	ce.Write(fields...)
}
```

## Use-cases

When writing a logging middleware library for a http server, I want to let users to define log level using req/res while hiding logging implementation behind the library.

Here's an example of defining log levels using response status (for Echo).

```go
// User specify log level
func LogLevelFromStatus(ctx echo.Context) zapcore.Level {
	status := ctx.Response().Status
	if status >= 500 {
		return zapcore.ErrorLevel
	}
	return zapcore.InfoLevel
}
```

With this definition, you can log like

```go
logger.LogWithLevel(LogLevelFromStatus(ctx), "request completed", fields...)
```

- - -

In addition to such usecases, this opens a door to implementing custom log level as discussed in #680 

## Discussion point

- Naming...some logging library name this much simpler like `Log` (e.g.: [logrus](https://github.com/sirupsen/logrus/blob/67a7fdcf741f4d5cee82cb9800994ccfd4393ad0/logger.go#L189))

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -237,6 +237,18 @@ func (log *Logger) Fatal(msg string, fields ...Field) {
 	}
 }
 
+// LogWithLevel logs a message at specified level. The message includes any
+// fields passed at the log site, as well as any fields accumulated on the
+// logger.
+//
+// The logger then might panic or call os.Exit(1) depending on the level
+// (see DPanic, Panic or Fatal for more details).
+func (log *Logger) LogWithLevel(lvl zapcore.Level, msg string, fields ...Field) {
+	if ce := log.check(lvl, msg); ce != nil {
+		ce.Write(fields...)
+	}
+}
+
 // Sync calls the underlying Core's Sync method, flushing any buffered log
 // entries. Applications should take care to call Sync before exiting.
 func (log *Logger) Sync() error {
--- a/logger_test.go
+++ b/logger_test.go
@@ -120,6 +120,7 @@ func TestLoggerLogPanic(t *testing.T) {
 		expected string
 	}{
 		{func(logger *Logger) { logger.Check(PanicLevel, "bar").Write() }, true, "bar"},
+		{func(logger *Logger) { logger.LogWithLevel(PanicLevel, "baz") }, true, "baz"},
 		{func(logger *Logger) { logger.Panic("baz") }, true, "baz"},
 	} {
 		withLogger(t, DebugLevel, nil, func(logger *Logger, logs *observer.ObservedLogs) {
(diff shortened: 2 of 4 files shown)


prashantv:
> This can be seen as shorthand for (but more appropriate caller depth to [`check()`](https://github.com/uber-go/zap/blob/a6015e13fab9b744d96085308ce4e8f11bad1996/logger.go#L256-L259))
> 
> ```go
> if ce := logger.Check(lvl, msg); ce != nil {
> 	ce.Write(fields...)
> }
> ```

Can you expand a little more on the "appropriate caller depth" point, as I'm not following exactly why this is required given that `Check` can do the same thing. Is this about API ergonomics (it's already possible, but this simplifies it), or about a functional difference?

tmshn:
@prashantv Sorry to be such vague, it's just about the API ergonomics as you said, and it's quite tiny point as you can ignore.

To avoid introducing more noice in the thread, I will clarify the detailed point in the folded section:

<details>

Currently I'm writing wrapper function like this (the code is simplified)

```go
func (l *WrapperOfLogger) LogWithLevel(lvl zapcore.Level, msg string, fields ...Field) {
	if ce := l.logger.Check(lvl, msg); ce != nil {
		ce.Write(fields...)
	}
}

func main() {
	zapLogger, _ := zap.NewProduction()
	l := NewWrapperOfLogger(zapLogger)
	l.LogWithLevel(zapcore.Info, "hello")
}
```

In this case, call stack looks like

```
zap.Logger#check
↑
zap.Logger#Check
↑
WrapperOfLogger#LogWithLevel
↑
main
```

As the caller info in `check` skips 2, it starts from `WrapperOfLogger#LogWithLevel`

https://github.com/uber-go/zap/blob/a6015e13fab9b744d96085308ce4e8f11bad1996/logger.go#L257-L259

On the other hand, with `LogWithLevel` directly implemented on the `zap.Logger`, I call `LogWithLevel` like this

```go
func main() {
	l, _ := zap.NewProduction()
	l.LogWithLevel(zapcore.Info, "hello")
}
```

and call stack become

```
zap.Logger#check
↑
zap.Logger#LogWithLevel
↑
main
```

In this case, caller info starts from `main`, which is bit more preferable.


Of course, I can achieve same thing by writing `Check`+`Write` pattern in main, , so it's only about "API ergonomics".

```go
func main() {
	l, _ := zap.NewProduction()
	if ce := l.Check(zapcore.Info, "hello"); ce != nil {
		ce.Write()
	}
}
```

</details>

prashantv:
Thanks for the detailed explanation @tmshn. I think I have a solution for your `WrapperOfLogger` type, you can use [`WithOptions`](https://godoc.org/go.uber.org/zap#Logger.WithOptions) and use the [`AddCallerSkip`](https://godoc.org/go.uber.org/zap#AddCallerSkip) option to increase the caller skip, so that when callers use your `LogWithLevel` function, the caller will skip your wrapped method.

```go
func NewWrapperOfLogger(logger *zap.Logger) *WrapperOfLogger {
  return &WrapperOfLogger{logger.WithOptions(zap.AddCallerSkip(1))}
}
```

Since this can be solved in the wrapper, I don't think we should expand the API surface of the `*Logger` right now, but instead encourage those callers to use `Check` or use a wrapper like yours above.

PR #758, opened 2019-11-07
Allow representations of nillable primitive field values

### LINKED ISSUE #753: Pointer to T Field constructors.
Zap has support for constructing Field from primitive types but doesn't have constructors for pointers to the primitive types (`*bool`, `*int`, etc). These types are encountered often enough that constructing through Reflect is not ideal.

Let's discuss this feature.

jbizzle:
Fixes #753 by exporting util functions to handle nillable primitives. The version here uses `Reflect()` under the hood to encode `nil`, and makes corresponding changes in `json_encoder.go` to efficiently handle encoding `nil` field values.

### DIFF
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -44,6 +44,11 @@ func (b *Buffer) AppendString(s string) {
 	b.bs = append(b.bs, s...)
 }
 
+// AppendBytes appends a byte slice to the Buffer.
+func (b *Buffer) AppendBytes(bs []byte) {
+	b.bs = append(b.bs, bs...)
+}
+
 // AppendInt appends an integer to the underlying buffer (assuming base 10).
 func (b *Buffer) AppendInt(i int64) {
 	b.bs = strconv.AppendInt(b.bs, i, 10)
@@ -94,7 +99,7 @@ func (b *Buffer) Reset() {
 
 // Write implements io.Writer.
 func (b *Buffer) Write(bs []byte) (int, error) {
-	b.bs = append(b.bs, bs...)
+	b.AppendBytes(bs)
 	return len(bs), nil
 }
 
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -44,11 +44,6 @@ func (b *Buffer) AppendString(s string) {
 	b.bs = append(b.bs, s...)
 }
 
-// AppendBytes appends a byte slice to the Buffer.
-func (b *Buffer) AppendBytes(bs []byte) {
-	b.bs = append(b.bs, bs...)
-}
-
 // AppendInt appends an integer to the underlying buffer (assuming base 10).
 func (b *Buffer) AppendInt(i int64) {
 	b.bs = strconv.AppendInt(b.bs, i, 10)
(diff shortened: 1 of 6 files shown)


buffer/buffer.go
```diff
@@ -44,6 +44,11 @@ func (b *Buffer) AppendString(s string) {
 	b.bs = append(b.bs, s...)
 }
 
+// AppendBytes appends a byte slice to the Buffer.
+func (b *Buffer) AppendBytes(bs []byte) {
+	b.bs = append(b.bs, bs...)
+}
```
abhinav: Why create this method instead of just using Write?

Since this package and type is public, we'd prefer to not introduce new
methods if possible.

zapcore/encoder.go
```diff
@@ -294,8 +294,8 @@ type ObjectEncoder interface {
 	AddUint8(key string, value uint8)
 	AddUintptr(key string, value uintptr)
 
-	// AddReflected uses reflection to serialize arbitrary objects, so it's slow
-	// and allocation-heavy.
+	// AddReflected uses reflection to serialize arbitrary objects, so it can be
+	// slow and allocation-heavy.
```
abhinav: heh nice detail

zapcore/json_encoder.go
```diff
@@ -145,16 +145,28 @@ func (enc *jsonEncoder) resetReflectBuf() {
...
+			return err
+		}
+		enc.reflectBuf.TrimNewline()
+		valueBytes = enc.reflectBuf.Bytes()
 	}
-	enc.reflectBuf.TrimNewline()
 	enc.addKey(key)
-	_, err = enc.buf.Write(enc.reflectBuf.Bytes())
-	return err
+	enc.buf.AppendBytes(valueBytes)
```
abhinav: Use of Write here should be fine

buffer/buffer.go
jbizzle: Just figured it helped avoid checking for errors via `Write()` which were known-impossible, but not worth disrupting the larger picture just for that (and in a previous, unsubmitted version, I was calling it from more than one site, so it was a mild ergonomic improvement).

zapcore/json_encoder.go
```diff
@@ -145,15 +145,27 @@ func (enc *jsonEncoder) resetReflectBuf() {
...
+
 func (enc *jsonEncoder) AddReflected(key string, obj interface{}) error {
-	enc.resetReflectBuf()
-	err := enc.reflectEnc.Encode(obj)
-	if err != nil {
-		return err
+	// Only invoke the standard JSON encoder if there is actually something to
+	// encode; otherwise write JSON null literal directly. For now, we only
+	// apply this optimization here, but not in AppendReflected. We could apply
+	// it there as well if needed (some discussion in
```
abhinav: Since we're treating this as an optimization, I'm in favor of doin it to AppendReflected too.
It already supports `nil` entries inside slices. https://play.golang.org/p/bZb6t58qhpy

prashantv: +1

PR #760, opened 2019-11-11
feat: add option `HooksWithFields` 

Laisky:
add new Option `HooksWithFields`.

It's similar with `Hooks`, but let hooks can access fields.

fully backward compatible.


Laisky:
@prashantv Thank you for your explaining. But It seems hard to load internal fields from Core, so I add a new `fields` in Core, is this OK?

If it's OK, I'll write test cases later.

prashantv:
> @prashantv Thank you for your explaining. But It seems hard to load internal fields from Core, so I add a new fields in Core, is this OK?

Since `Core` is an interface, adding a new method to it is not backwards compatible. Any existing core implementations will no longer satisfy the interface with the new method, so we cannot make this change in zap 1.0.

Changing cores to track all fields also has a performance penalty, so we wouldn't want to pay this cost in all cases -- we should instead only pay the cost when the user uses a hook that requires access to fields.

Instead of adding a new method to the `Core` interface, you could store the fields as part of the `hookWithFields` core. That avoids any backwards incompatible changes, while limiting the performance penalty of storing all fields to users of the hook.

Laisky:
@prashantv It seems that there's no way to get `Core.enc` without change `Core` interface, so hooks cannot get included fields that stored in `Core.enc`.

Do you have any idea that how to get included fields in hooks? or we have to give up included fields in hooks.

prashantv:
It's not possible to get the fields out of existing cores, however, the wrapped core can store fields.

However, this will be limited to only access fields added after the wrapped core is added.

Laisky:
@prashantv I updated the code, it can achieve this:

```go
logger := zap.NewExample(zap.HooksWithFields(func(e zapcore.Entry, fs []zapcore.Field) error {
    fmt.Println(len(fs)) 
    return nil
}))

logger.With(zap.String("key", "val")).Info("test")
```

output:

```
1
```

logger_test.go
```diff
@@ -401,6 +410,15 @@ func TestLoggerHooks(t *testing.T) {
 	assert.Equal(t, int64(2), seen.Load(), "Hook saw an unexpected number of logs.")
 }
 
+func TestLoggerHooksWithFields(t *testing.T) {
+	hook, seen := makeCountingHookWithFields()
```
prashantv, line 414: The counting hooks doesn't validate the fields, which would be useful since that is the main difference between `TestLoggerHooks` and this test. Given that `countingHookWithFields` is only used once, it may be better to inline that function here, and validate the entry + fields

options.go
```diff
@@ -54,6 +54,13 @@ func Hooks(hooks ...func(zapcore.Entry) error) Option {
 	})
 }
 
+// HooksWithFields similiar with Hooks, but hook funcs can access fields
```
prashantv, line 57:
```suggestion
// HooksWithFields is similar to Hooks, but allows access to fields in addition to the entry.
```

we should also mention include documentation that the fields are read-only (fields can't be changed or added), as well as note that there's a performance impact as it requires all fields to be stored separately (in addition to the encoded form)

zapcore/hook.go
```diff
@@ -66,3 +81,32 @@ func (h *hooked) Write(ent Entry, _ []Field) error {
...
+		return downstream.AddCore(ent, h)
+	}
+	return ce
+}
+
+func (h *hookedWithFields) With(fields []Field) Core {
+	return &hookedWithFields{
+		Core:   h.Core.With(fields),
+		funcs:  h.funcs,
+		fields: append(h.fields[:len(h.fields):len(h.fields)], fields...),
```
prashantv, line 99: Couple of requests:
 * Can you move the creation of the new `fields` up and add comments
 * Can you add a unit test that verifies that the behaviour works correctly. Ideally, a test that fails with `append(h.fields, fields...)` but passes with the correct copy behaviour
* For performance reasons, I think using `append` could end up creating a slice larger than necessary.

See example: https://play.golang.org/p/MBNMAMzOsBs

The resulting slice may have `len < cap`, but this slice is never appended to (it's immutable after creation) -- we only ever make a full copy and add new fields.

It would be more efficient to do something like:
```go
copyFields := make([]Field, len(h.fields) + len(fields))
copy(copyFields, h.fields)
copy(copyFields[len(h.fields):], fields)
```

that way the capacity == length

zapcore/hook_test.go
```diff
@@ -71,3 +71,47 @@ func TestHooks(t *testing.T) {
...
+		f := func(e Entry, fs []Field) error {
+			called++
+			assert.Equal(t, ent, e, "Hook called with unexpected Entry.")
+			return nil
+		}
+
+		h := RegisterHooksWithFields(fac, f)
+		if ce := h.With([]Field{intField}).With([]Field{intField2}).Check(ent, nil); ce != nil {
+			ce.Write()
+		}
```
prashantv, line 102: can you verify that the hook is only called after `Write` is called

zapcore/hook_test.go
```diff
@@ -71,3 +71,47 @@ func TestHooks(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		fac, logs := observer.New(tt.coreLevel)
+		intField := makeInt64Field("foo", 42)
+		intField2 := makeInt64Field("bar", 67)
+		ent := Entry{Message: "bar", Level: tt.entryLevel}
+
+		var called int
+		f := func(e Entry, fs []Field) error {
```
prashantv, line 93: we're not verifying fields, we should verify that the fields match the logger that's used (ideally with different loggers that have different fields attached)

titus12:
Laisky, Please submit

brenu:
I guess I can make these requested changes starting from where Laisky stopped, and open it as a different PR. What do you guys think?

prashantv:
> I guess I can make these requested changes starting from where Laisky stopped, and open it as a different PR. What do you guys think?

Since this change hasn't been updated in a few months, that makes sense to me. Please credit @Laisky in your new PR as well (assuming it's based on this PR).

zapcore/hook.go
brenu: Don't know if I'm getting it right, so I'm gonna ask before commiting. Is the request to implement a separate method for creating new `fields` and then add a unit test for that?

prashantv: No need to create a separate method for appending the fields, just a separate test to verify that the original core's fields backing array is not shared by the new resulting fields. 

E.g., if this code was just `append(h.fields, fields...)`, then the backing fields array could be shared, and so a test should detect this misuse.

PR #764, opened 2019-11-21
export OpenSinks/BuildEncoder/BuildOptions

zhiqiangxu:
When trying to implement a logger that will separate info and error logs based on `zap.NewTee`, I found that these three methods are very necessary:

https://github.com/zhiqiangxu/util/blob/master/logger/zap.go

Users will have to duplicate such code everywhere if these three methods are not exported.

### DIFF
--- a/config.go
+++ b/config.go
@@ -176,7 +176,7 @@ func (cfg Config) Build(opts ...Option) (*Logger, error) {
 
 	log := New(
 		zapcore.NewCore(enc, sink, cfg.Level),
-		cfg.buildOptions(errSink)...,
+		cfg.BuildOptions(errSink)...,
 	)
 	if len(opts) > 0 {
 		log = log.WithOptions(opts...)
@@ -184,7 +184,8 @@ func (cfg Config) Build(opts ...Option) (*Logger, error) {
 	return log, nil
 }
 
-func (cfg Config) buildOptions(errSink zapcore.WriteSyncer) []Option {
+// BuildOptions for building options from cfg
+func (cfg Config) BuildOptions(errSink zapcore.WriteSyncer) []Option {
 	opts := []Option{ErrorOutput(errSink)}
 
 	if cfg.Development {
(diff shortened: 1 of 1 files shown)


prashantv:
Can you provide more context on how these functions being exported would help?

We try to keep the API surface minimal, as any exposed API has to be maintained going forward.

zhiqiangxu:
It is helpful in the same way as it is in the zap source tree, but it is not exported, so I have to duplicate the code.

prashantv:
I understand that exporting reduces duplication for your use-case, but it also increase the API surface that we have to maintain, which is an ongoing cost for the maintainers, so I want to see if there's alternate ways to support your use-case.

For example, I think you can achieve this by:
 * Create logger1 with Info level, and the options you want
 * Create logger2 with Error level, the options aren't relevant, as you'll just want the underlying core, accessible via `logger2.Core()`
 * Use `logger1.WithOptions(WrapCore(...))` and pass a function that will only pass through Info logs to the underlying core, and send Error or higher logs to logger2's core.

zhiqiangxu:
> I understand that exporting reduces duplication for your use-case, but it also increase the API surface that we have to maintain, which is an ongoing cost for the maintainers, so I want to see if there's alternate ways to support your use-case.
> 
> For example, I think you can achieve this by:
> 
> * Create logger1 with Info level, and the options you want
> * Create logger2 with Error level, the options aren't relevant, as you'll just want the underlying core, accessible via `logger2.Core()`
> * Use `logger1.WithOptions(WrapCore(...))` and pass a function that will only pass through Info logs to the underlying core, and send Error or higher logs to logger2's core.

That may also work, but a little too much work for such a simple requirement..

prashantv:
> That may also work, but a little too much work for such a simple requirement..

I see, it seems like it would be less code than copy+pasting.

Unfortunately, I'm not comfortable with exporting the APIs as it's not clear which part of the `Config` is used to build sinks, build an encoder, build a core, and build options for a logger. We don't document which field is used for which component, and if these are exported, then that would become an important part of the API, and we'd have to maintain backwards compatibility.

PR #771, opened 2019-12-10
logger: Check level before creating entry

### LINKED ISSUE #770: Calling Debug or Check on a Production Logger is slow
First of all: thank you for such a great library!
Unfortunately I have some performance issues when calling Check or Debug on a production logger. I am unsure if I am using the library wrong. I am using the following code:

```go
package main

import (
	"os"
	"testing"

	log "github.com/sirupsen/logrus"
	"go.uber.org/zap"
)

func BenchmarkZap(b *testing.B) {
	defer logger.Sync()

	for n := 0; n < b.N; n++ {
		logger.Debug("Test")
	}
}

func BenchmarkZapCheck(b *testing.B) {
	defer logger.Sync()

	for n := 0; n < b.N; n++ {
		if ce := logger.Check(zap.DebugLevel, "Test"); ce != nil {
			ce.Write()
		}
	}
}

func BenchmarkLogrus(b *testing.B) {
	log.SetFormatter(&log.JSONFormatter{})
	log.SetOutput(os.Stdout)
	log.SetLevel(log.WarnLevel)

	for n := 0; n < b.N; n++ {
		log.Debug("Test")
	}
}
```
And I get the following results:
```
goos: linux
goarch: amd64
pkg: bachelorarbeit
BenchmarkZap-4        	15698260	        74.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkZapCheck-4   	17619888	        68.1 ns/op	       0 B/op	       0 allocs/op
BenchmarkLogrus-4     	341667280	         3.54 ns/op	       0 B/op	       0 allocs/op
```
Is this performance expected?

prashantv:
Calling `time.Now()` and creating an entry is unnecessary
if the underlying core has the specified level disabled.

To reduce the cost of logs at disabled levels, skip entry
creation if the log level is disabled in the core.

This special logic does not apply to DPanic or higher logs as
they may need to panic/exit even if the entry does not cause
any log to be emitted.

On my machine, disabled debugging logs are 6x (~60ns to ~10ns)
based on the example in the issue.

benchcmp:
```
benchmark                                                 old ns/op     new ns/op     delta
BenchmarkDisabledWithoutFields/Zap-12                     8.42          1.59          -81.12%
BenchmarkDisabledWithoutFields/Zap.Check-12               8.01          1.32          -83.52%
BenchmarkDisabledWithoutFields/Zap.Sugar-12               12.4          11.4          -8.06%
BenchmarkDisabledWithoutFields/Zap.SugarFormatting-12     117           102           -12.82%
```

Fixes #770.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -258,6 +258,12 @@ func (log *Logger) check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
 	// (e.g., Check, Info, Fatal).
 	const callerSkipOffset = 2
 
+	// Check the level first to reduce the cost of disabled log calls.
+	// Since Panic and higher may exit, we skip the optimization for those levels.
+	if lvl < zapcore.DPanicLevel && !log.core.Enabled(lvl) {
+		return nil
+	}
+
 	// Create basic checked entry thru the core; this will be non-nil if the
 	// log message will actually be written somewhere.
 	ent := zapcore.Entry{


alxn:
new test?

prashantv:
> new test?

Since this is an optimization, it doesn't have any impact to the functional tests, just a performance improvement.

While it is possible to write a test to check for this (we can verify at the core which methods are called), it seems like we're breaking the abstraction.

abhinav:
Perhaps not a test, but perhaps we should add a benchmark for messages being discarded if we don't already have one.

prashantv:
We do have existing benchmarks for this,
```
benchmark                                                 old ns/op     new ns/op     delta
BenchmarkDisabledWithoutFields/Zap-12                     8.42          1.59          -81.12%
BenchmarkDisabledWithoutFields/Zap.Check-12               8.01          1.32          -83.52%
BenchmarkDisabledWithoutFields/Zap.Sugar-12               12.4          11.4          -8.06%
BenchmarkDisabledWithoutFields/Zap.SugarFormatting-12     117           102           -12.82%
```

Will update the summary as well.

PR #773, opened 2019-12-14
add millisecond duration encoder

caibirdme:
Since network requests are always in milliseconds level, if procTime shows in seconds, it may look like `0.003`, to many zeros. While if the procTime shows in nanoseconds, it may look like `12312312323`, too big to identify. So it'd be better to support milliseconds encoder, something like `30.5`, pretty intuisive.

### DIFF
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -168,6 +168,12 @@ func NanosDurationEncoder(d time.Duration, enc PrimitiveArrayEncoder) {
 	enc.AppendInt64(int64(d))
 }
 
+// MillisDurationEncoder serializes a time.Duration to an integer number of
+// milliseconds elapsed.
+func MillisDurationEncoder(d time.Duration, enc PrimitiveArrayEncoder) {
+	enc.AppendInt64(d.Nanoseconds() / 1e6)
+}
+
 // StringDurationEncoder serializes a time.Duration using its built-in String
 // method.
 func StringDurationEncoder(d time.Duration, enc PrimitiveArrayEncoder) {
@@ -183,6 +189,8 @@ func (e *DurationEncoder) UnmarshalText(text []byte) error {
 		*e = StringDurationEncoder
 	case "nanos":
 		*e = NanosDurationEncoder
+	case "ms":
+		*e = MillisDurationEncoder
 	default:
 		*e = SecondsDurationEncoder
 	}
--- a/zapcore/encoder_test.go
+++ b/zapcore/encoder_test.go
@@ -566,6 +566,7 @@ func TestDurationEncoders(t *testing.T) {
 	}{
 		{"string", "1.0000005s"},
 		{"nanos", int64(1000000500)},
+		{"ms", int64(1000)},
 		{"", 1.0000005},
 		{"something-random", 1.0000005},
 	}


caibirdme:
> Thanks for the contribution @caibirdme
> 
> This change looks good to me, just needs some tests. Can you please add tests for this change?

Ok, no problem! I'll add those testcases soon

caibirdme:
@prashantv I'v added the test case

zapcore/encoder.go
```diff
@@ -168,6 +168,12 @@ func NanosDurationEncoder(d time.Duration, enc PrimitiveArrayEncoder) {
 	enc.AppendInt64(int64(d))
 }
 
+// MilliSecondsDurationEncoder serializes a time.Duration to an integer number of
+// milliseconds elapsed.
+func MilliSecondsDurationEncoder(d time.Duration, enc PrimitiveArrayEncoder) {
```
prashantv: Should this be called `MillisDurationEncoder` to match the naming of the other methods

caibirdme:
@prashantv Well, I understand that, but if we just use float64, something like `31.0000005` may not that much good. What about defining a constructor to control the precision?
```
func NewMillisDurationEncoder(precision int) func MilliSecondsDurationEncoder(d time.Duration, enc PrimitiveArrayEncoder)
```

prashantv:
> What about defining a constructor to control the precision?

Constructor will help if you manually create the encoder, but it won't be as easy to use when unmarshalling a configuration from a YAML file.

For your use-case, do you want to control the precision, or would you prefer to use int vs float?

caibirdme:
> > What about defining a constructor to control the precision?
> 
> Constructor will help if you manually create the encoder, but it won't be as easy to use when unmarshalling a configuration from a YAML file.
> 
> For your use-case, do you want to control the precision, or would you prefer to use int vs float?

Actually, I don't care about the precision, what I want is just how many milliseconds the operation costs. 30 or 30.05 are both ok for me, but 30.00005 smells bad. But as the test case shows, all the other encoder could show how nanoseconds it costs(1.0000005s 1.0000005 1000000500)，I don't know if something like 1.02 will bring some inconsistencies. But really, I think float with just two decimal places is better. When people choose ms encoder, nobody cares about the nanoseconds

zapcore/encoder.go
prashantv: Following up on this, to match `NanosDurationEncoder`, this should be `MillisDurationEncoder`

caibirdme: > NanosDurationEncoder

Done

zapcore/encoder.go
```diff
@@ -168,6 +168,12 @@ func NanosDurationEncoder(d time.Duration, enc PrimitiveArrayEncoder) {
 	enc.AppendInt64(int64(d))
 }
 
+// MillisDurationEncoder serializes a time.Duration to an integer number of
+// milliseconds elapsed.
+func MillisDurationEncoder(d time.Duration, enc PrimitiveArrayEncoder) {
+	enc.AppendInt64(d.Nanoseconds() / 1e6)
```
prashantv:
```suggestion
	enc.AppendInt64(d.Milliseconds())
```

prashantv:
The CLA bot is blocked on user liudeen -- if that's an old username you no longer use, can you squash your commits into one owned by the new user

zapcore/encoder.go
caibirdme: the d.Milliseconds() is a new API only for go1.13+

caibirdme: https://github.com/golang/go/issues/28564

prashantv: Good point, we can migrate to it once we drop support for 1.12. This is good for now, thanks.

Issue #581, opened 2018-04-16
provide an easy way to derive a logger that logs at a different level


rogpeppe:
When logging HTTP requests, we've found it useful to be able to increase the logging threshold for some requests at particular paths (for example, the trivial endpoint that the Nagios check uses) to avoid overwhelming the logs with noise.

It's possible to do this oneself with something like the below code, but this isn't particularly obvious and probably isn't that efficient, so perhaps it might be worth adding a `WithLevel` method to zap.Logger.

```go
// in the code
logger = logger.WithOptions(wrapCoreWithLevel(zap.WarnLevel))

func wrapCoreWithLevel(level zapcore.Level) zap.Option {
	return zap.WrapCore(func(core zapcore.Core) zapcore.Core {
		return &coreWithLevel{
			Core:  core,
			level: level,
		}
	})
}

type coreWithLevel struct {
	zapcore.Core
	level zapcore.Level
}

func (c *coreWithLevel) Enabled(level zapcore.Level) bool {
	return c.level.Enabled(level) && c.Core.Enabled(level)
}

func (c *coreWithLevel) Check(e zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
	if !c.level.Enabled(e.Level) {
		return ce
	}
	return c.Core.Check(e, ce)
}

etsangsplk:
@rogpeppe Would you mind explaining how your code works, because zap is not easy to customize.
Am thinking about doing something similar to you. 
Child logger can change their logging level and their descendants level at will without affecting its parent.

rogpeppe:
@etsangsplk It works by creating a wrapper type (coreWithLevel) for the zapcore.Core interface that reimplements the Enabled and Check methods to choose a higher logging level. There's an existing option that allows replacing the zapcore.Core associated with a given logger, so we use that and the coreWithLevel type to provide a function that can raise the level associated with a given logger.

I'm not sure if that will help much - if in doubt, read the docs and/or browse through the code.

FWIW we're using the above technique in the WithLevel function [here](https://godoc.org/github.com/juju/zaputil/zapctx#WithLevel).

NathanZook:
I wonder if WithOption might be what you want here. https://godoc.org/go.uber.org/zap#Logger.WithOptions

rogpeppe:
@NathanZook The code in the issue above does use WithOptions. Have you got a particular way of using WithOptions to raise the logging threshold that's simpler than that?

[...]

PR #775, opened 2019-12-23
Add option to increase the level of a logger

### LINKED ISSUE #774: Change LogLevel without creating a new core
I'd like to create a new subLogger which only logs messages at the `WarnLevel` instead of the `InfoLevel` (which is desired for the rest of the application). I read the other issues about doing so, but they implied creating a new zapcore.

I do not want to create a new zapcore for my sublogger, because my logger's core has hooks which should still be used for the sublogger. So is there any way to create a sublogger which reuses my existent zapcore but only emits log messages at the WarnLevel+?

```go
stripeLogger := logger.
	With(zap.String("source", "stripe")).
	WithOptions(zap.WrapCore(/** Help **/))
stripe.Logger = zap.NewStdLog(stripeLogger)
```

prashantv:
We've had a few requests on how to change the level of a logger. With
the current APIs, it's not possible to reduce the log level (e.g., go
from Info to Debug), but it is possible to increase the level using
`WrapCore` with a custom filtering core.

Since it's a common request, I think we should add an option to make
this easy to use.

I want to make sure "increase" is part of the API to avoid confusion on
whether it can reduce the log level.

Fixes #774.

### DIFF
--- a/increase_level_test.go
+++ b/increase_level_test.go
@@ -1,0 +1,81 @@
+// Copyright (c) 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zap
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+func newLoggedEntry(level zapcore.Level, msg string) observer.LoggedEntry {
+	return observer.LoggedEntry{
+		Entry:   zapcore.Entry{Level: level, Message: msg},
+		Context: []zapcore.Field{},
+	}
+}
+
+func TestIncreaseLevelTryDecrease(t *testing.T) {
+	errorOut := &bytes.Buffer{}
+	opts := []Option{
+		ErrorOutput(zapcore.AddSync(errorOut)),
+	}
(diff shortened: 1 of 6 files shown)


weeco:
I think this is related to https://github.com/uber-go/zap/issues/581 , but when I got the author correctly he wishes an option to change the loglevel on a per message basis rather than just having an easy way to create a new child logger which prints at a different log level.

zapcore/level_filter.go
```diff
@@ -0,0 +1,45 @@
...
+
+type levelFilterCore struct {
+	Core
+	level LevelEnabler
+}
+
+// NewLevelCore creates a core that can be used to increase the level of an existing
+// Core. It cannot be used to decrease the logging level, as it acts as a filter
+// before calling the underlying core.
+func NewLevelCore(core Core, level LevelEnabler) Core {
```
abhinav: I suspect this needs a similar name that implies increasing levels only.

prashantv:
> Do we have any way to guard against attempts to lower the level?
> 
> I'm concerned that despite naming, this will be very easy to misuse.

We can't protect against it, but we can access the `errorOutput` for the logger, and write a message to it if we detect that someone tries to lower the log level. I don't think adding errors to the API will be very ergonomic. We could do a `DPanic` but that might be a little surprising. Any other ideas?

abhinav:
Logging to errorOutput seems reasonable. I agree that having this error out
would be less ergonomic.

prashantv:
Updated with more explicit name + error if the level is decreased. Could definitely improve the message so looking for suggestions

zapcore/increase_level.go
```diff
@@ -0,0 +1,54 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zapcore
+
+import "fmt"
+
+type levelFilterCore struct {
+	Core
+	level LevelEnabler
```
abhinav: nit: empty line between embedded type and fields
```suggestion

	level LevelEnabler
```

zapcore/increase_level_test.go
```diff
@@ -0,0 +1,102 @@
...
+	. "go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+func TestIncreaseLevel(t *testing.T) {
+	tests := []struct {
+		coreLevel     Level
+		increaseLevel Level
+		wantErr       bool
+		wantLogLevels []Level
```
abhinav: Unused field
```suggestion
```

logger_test.go
```diff
@@ -392,6 +392,21 @@ func TestLoggerReplaceCore(t *testing.T) {
 	})
 }
 
+func TestLoggerIncreaseLevel(t *testing.T) {
+	withLogger(t, DebugLevel, opts(IncreaseLevel(WarnLevel)), func(logger *Logger, logs *observer.ObservedLogs) {
+		logger.Info("logger.Info")
+		logger.Warn("logger.Warn")
+		logger.Error("logger.Error")
+		require.Equal(t, 2, logs.Len(), "Expected no-op core to write no logs.")
```
abhinav: the error message is wrong

increase_level_test.go
```diff
@@ -0,0 +1,81 @@
...
+		debugLogger.Debug("ignored debug log")
+		debugLogger.Warn("increase level warn log")
+		debugLogger.Error("increase level error log")
+
+		assert.Equal(t, []observer.LoggedEntry{
+			newLoggedEntry(WarnLevel, "original warn log"),
+			newLoggedEntry(WarnLevel, "increase level warn log"),
+			newLoggedEntry(ErrorLevel, "increase level error log"),
+		}, logs.AllUntimed(), "unexpected logs")
+		assert.Equal(t, "failed to IncreaseLevel: invalid increase level, as info is allowed by increased level, but not by existing core", errorOut.String(), "unexpected error output")
```
abhinav: Can we tweak the warning a bit? This reads weird because `info`, the level mixes into the warning message. Maybe we should quote it and/or say "[..] as logs with level info are allowed by increased level but not by existing level."

PR #776, opened 2019-12-31
Return options generated from Config for inspection.

kyegupov:
Adds an extra public method that returns `[]Option` generated from the `Config`.

Rationale: our project uses zap's config but also performs additional manipulations on the generated options. For that, we need to access Options generated from the config. However, the current code layout doesn't allow that. Hence, a new intermediary-but-public method.

Please suggest alternative solutions if this is not acceptable.

### DIFF
--- a/config.go
+++ b/config.go
@@ -164,24 +164,32 @@ func NewDevelopmentConfig() Config {
 
 // Build constructs a logger from the Config and Options.
 func (cfg Config) Build(opts ...Option) (*Logger, error) {
+	_, log, err := cfg.BuildOptionsAndLogger()
+	if len(opts) > 0 {
+		log = log.WithOptions(opts...)
+	}
+	return log, err
+}
+
+// BuildOptionsAndLogger generates Options from the Config, builds a logger
+// and returns both Options and the logger.
+func (cfg Config) BuildOptionsAndLogger() ([]Option, *Logger, error) {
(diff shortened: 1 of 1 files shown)


prashantv:
There was a similar request in #764, and I think I have the same apprehension. Is this API surface easy to understand for a consumer? Which configuration struct fields result in encoder changes vs logger options? Is that an API that we want to maintain going forward.

Can you build the logger from the configuration struct and use `WithOptions` to modify the options?

prashantv:
Closing as there hasn't been a response in a couple of months.

kyegupov:
@prashantv Our use case is:
* create a logger1 from configuration
* create a new logger2 using options from logger1, possibly extended

For that, we need to transplant options generated by a config.

The current API is quite strange: options are "write-only". You can add new options to a logger using WithOptions, but you cannot read the options currently applied.

Could you suggest an approach we could use for our case?

prashantv:
> The current API is quite strange: options are "write-only". You can add new options to a logger using WithOptions, but you cannot read the options currently applied.

Yes, this is typical for functional options in Go libraries.

Instead of converting the configuration to `[]Option` and then passing those around, why not pass the `Configuration` around, and when logger 2 needs additional options, then pass additional options to [`Build`](https://pkg.go.dev/go.uber.org/zap?tab=doc#Config.Build).

PR #782, opened 2020-02-08
Add buffering wrapper around WriteSyncer

### LINKED ISSUE #663: what about implement a buffer writer syncer?
Support sync through a buffer writer, zap should be faster...

Just like golang/glog does.
https://github.com/golang/glog/blob/master/glog.go#L841

Glad to send a PR if you guys are interested :-)

Simple Demo:
```
type bufferWriterSync struct {
	buf *bufio.Writer
}

func Buffer(ws WriteSyncer, size int) WriteSyncer {
        bw := &bufferWriterSync{
		buf: bufio.NewWriterSize(ws, size),
	}

        ws = zapcore.Lock(bw)
        return ws
}

// Sync syncs data to file
func (w bufferWriterSync) Sync() error {
	return w.buf.Flush()
}

// Write writes data to buffer
func (w bufferWriterSync) Write(p []byte) (int, error) {
	return w.buf.Write(p)
}
```

hnlq715:
Fixes #663 

This change adds `Buffer`, which wraps a `WriteSyncer` with buffering. This use `bufio.Writer` to buffer in-memory, and flush changes periodically (or when the buffer is full). The `Sync` method forces an immediate flush.

This can improve performance by amortizing any fixed overheads of the underlying `WriteSyncer`.

bechmark result
```
Running tool: /usr/local/opt/go/libexec/bin/go test -benchmem -run=^$ go.uber.org/zap/zapcore -bench ^(BenchmarkMultiWriteSyncer|BenchmarkWriteSyncer)$

goos: darwin
goarch: amd64
pkg: go.uber.org/zap/zapcore
BenchmarkMultiWriteSyncer/2_discarder-4         	41213204	        66.5 ns/op	      16 B/op	       1 allocs/op
BenchmarkMultiWriteSyncer/4_discarder-4         	19545279	        72.0 ns/op	      16 B/op	       1 allocs/op
BenchmarkMultiWriteSyncer/4_discarder_with_buffer-4         	 8759560	       143 ns/op	      16 B/op	       1 allocs/op
BenchmarkWriteSyncer/write_file_with_no_buffer-4            	  160742	      9367 ns/op	      16 B/op	       1 allocs/op
BenchmarkWriteSyncer/write_file_with_buffer-4               	 7559112	       198 ns/op	      16 B/op	       1 allocs/op
PASS
ok  	go.uber.org/zap/zapcore	12.013s
Success: Benchmarks passed.
```

### DIFF
--- a/zapcore/core.go
+++ b/zapcore/core.go
@@ -42,6 +42,8 @@ type Core interface {
 	Write(Entry, []Field) error
 	// Sync flushes buffered logs (if any).
 	Sync() error
+	// Close cleans all WriterSync resources (if any).
+	Close() error
 }
 
 type nopCore struct{}
@@ -53,6 +55,7 @@ func (n nopCore) With([]Field) Core                           { return n }
 func (nopCore) Check(_ Entry, ce *CheckedEntry) *CheckedEntry { return ce }
 func (nopCore) Write(Entry, []Field) error                    { return nil }
 func (nopCore) Sync() error                                   { return nil }
+func (nopCore) Close() error                                  { return nil }
 
 // NewCore creates a Core that writes logs to a WriteSyncer.
 func NewCore(enc Encoder, ws WriteSyncer, enab LevelEnabler) Core {
@@ -102,6 +105,10 @@ func (c *ioCore) Write(ent Entry, fields []Field) error {
 
 func (c *ioCore) Sync() error {
 	return c.out.Sync()
+}
+
+func (c *ioCore) Close() error {
+	return c.out.Close()
 }
 
 func (c *ioCore) clone() *ioCore {
--- a/zapcore/core.go
+++ b/zapcore/core.go
@@ -42,8 +42,6 @@ type Core interface {
 	Write(Entry, []Field) error
 	// Sync flushes buffered logs (if any).
 	Sync() error
-	// Close cleans all WriterSync resources (if any).
-	Close() error
 }
 
 type nopCore struct{}
@@ -55,7 +53,6 @@ func (n nopCore) With([]Field) Core                           { return n }
 func (nopCore) Check(_ Entry, ce *CheckedEntry) *CheckedEntry { return ce }
 func (nopCore) Write(Entry, []Field) error                    { return nil }
 func (nopCore) Sync() error                                   { return nil }
-func (nopCore) Close() error                                  { return nil }
 
 // NewCore creates a Core that writes logs to a WriteSyncer.
 func NewCore(enc Encoder, ws WriteSyncer, enab LevelEnabler) Core {
@@ -105,10 +102,6 @@ func (c *ioCore) Write(ent Entry, fields []Field) error {
 
 func (c *ioCore) Sync() error {
 	return c.out.Sync()
-}
-
-func (c *ioCore) Close() error {
-	return c.out.Close()
 }
 
 func (c *ioCore) clone() *ioCore {
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -127,7 +127,7 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteS
 	})
 
 	// flush buffer every interval
-	// we do not need exit this goroutine explicitly
+	// we only exit this goroutine after closefunc called explicitly
 	go func() {
 		select {
 		case <-ticker.C:
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -80,6 +80,7 @@ func (s *lockedWriteSyncer) Sync() error {
 
 type bufferWriterSyncer struct {
 	sync.Mutex
+
 	bufferWriter *bufio.Writer
 	ticker       *time.Ticker
 }
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -80,6 +80,7 @@ func (s *lockedWriteSyncer) Sync() error {
 
 type bufferWriterSyncer struct {
 	bufferWriter *bufio.Writer
+	ticker       *time.Ticker
 }
 
 // defaultBufferSize sizes the buffer associated with each WriterSync.
@@ -117,16 +118,19 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteS
 		flushInterval = defaultFlushInterval
 	}
 
+	ticker := time.NewTicker(flushInterval)
+
 	// bufio is not goroutine safe, so add lock writer here
 	ws = Lock(&bufferWriterSyncer{
 		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
+		ticker:       ticker,
 	})
 
 	// flush buffer every interval
 	// we do not need exit this goroutine explicitly
 	go func() {
 		select {
-		case <-time.NewTicker(flushInterval).C:
+		case <-ticker.C:
 			// the background goroutine just keep syncing
 			// until the close func is called.
 			_ = ws.Sync()
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -21,8 +21,10 @@
 package zapcore
 
 import (
+	"bufio"
 	"io"
 	"sync"
+	"time"
 
 	"go.uber.org/multierr"
 )
@@ -73,6 +75,53 @@ func (s *lockedWriteSyncer) Sync() error {
 	err := s.ws.Sync()
 	s.Unlock()
 	return err
+}
+
+type bufferWriterSyncer struct {
+	bufferWriter *bufio.Writer
+	ws           WriteSyncer
+}
+
+// bufferSize sizes the buffer associated with each log file. It's large
+// so that log records can accumulate without the logging thread blocking
+// on disk I/O. The flushDaemon will block instead.
+const bufferSize = 256 * 1024
+
+// flushInterval means the default flush interval
+const flushInterval = 30 * time.Second
+
+// Buffer wraps a WriteSyncer in a buffer to improve performance,
+// which is implemented in https://github.com/golang/glog.
+func Buffer(ws WriteSyncer) WriteSyncer {
+	if _, ok := ws.(*bufferWriterSyncer); ok {
+		// no need to layer on another buffer
+		return ws
+	}
+
+	// bufio is not goroutine safe, so add lock writer here
+	ws = Lock(&bufferWriterSyncer{
+		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
+	})
+
+	// flush buffer every interval
+	// we do not need exit this goroutine explicitly
+	go func() {
+		for range time.NewTicker(flushInterval).C {
+			if err := ws.Sync(); err != nil {
+				return
+			}
+		}
+	}()
+
+	return ws
+}
+
+func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
+	return s.bufferWriter.Write(bs)
+}
+
+func (s *bufferWriterSyncer) Sync() error {
+	return s.bufferWriter.Flush()
 }
 
 type writerWrapper struct {
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -79,8 +79,8 @@ func (s *lockedWriteSyncer) Sync() error {
 
 type bufferWriterSyncer struct {
 	sync.Mutex
-	stop chan struct{}
 
+	stop         chan struct{}
 	bufferWriter *bufio.Writer
 	ticker       *time.Ticker
 }
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -34,7 +34,6 @@ import (
 // that *os.File (and thus, os.Stderr and os.Stdout) implement WriteSyncer.
 type WriteSyncer interface {
 	io.Writer
-	Close() error
 	Sync() error
 }
 
@@ -79,17 +78,9 @@ func (s *lockedWriteSyncer) Sync() error {
 	return err
 }
 
-func (s *lockedWriteSyncer) Close() error {
-	s.Lock()
-	err := s.ws.Close()
-	s.Unlock()
-	return err
-}
-
 type bufferWriterSyncer struct {
 	ws           WriteSyncer
 	bufferWriter *bufio.Writer
-	cancel       context.CancelFunc
 }
 
 // defaultBufferSize sizes the buffer associated with each WriterSync.
@@ -98,14 +89,24 @@ const defaultBufferSize = 256 * 1024
 // defaultFlushInterval means the default flush interval
 const defaultFlushInterval = 30 * time.Second
 
+// CancelFunc should be called when the caller exits to clean up buffers.
+type CancelFunc func() error
+
 // Buffer wraps a WriteSyncer in a buffer to improve performance,
 // if bufferSize = 0, we set it to defaultBufferSize
 // if flushInterval = 0, we set it to defaultFlushInterval
-func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) WriteSyncer {
+func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteSyncer, CancelFunc) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	cancelfunc := func() error {
+		cancel()
+		return ws.Sync()
+	}
+
 	if lws, ok := ws.(*lockedWriteSyncer); ok {
 		if _, ok := lws.ws.(*bufferWriterSyncer); ok {
 			// no need to layer on another buffer
-			return ws
+			return ws, cancelfunc
 		}
 	}
 
@@ -117,12 +118,9 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) WriteSy
 		flushInterval = defaultFlushInterval
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
-
 	// bufio is not goroutine safe, so add lock writer here
 	ws = Lock(&bufferWriterSyncer{
 		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
-		cancel:       cancel,
 	})
 
 	// flush buffer every interval
@@ -138,7 +136,7 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) WriteSy
 		}
 	}()
 
-	return ws
+	return ws, cancelfunc
 }
 
 func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
@@ -159,11 +157,6 @@ func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
 
 func (s *bufferWriterSyncer) Sync() error {
 	return s.bufferWriter.Flush()
-}
-
-func (s *bufferWriterSyncer) Close() error {
-	s.cancel()
-	return s.Sync()
 }
 
 type writerWrapper struct {
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -111,11 +111,6 @@ const (
 // The buffer size defaults to 256 kB if set to zero.
 // The flush interval defaults to 30 seconds if set to zero.
 func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (_ WriteSyncer, close func() error) {
-	if _, ok := ws.(*bufferWriterSyncer); ok {
-		// no need to layer on another buffer
-		return ws, func() error { return nil }
-	}
-
 	if bufferSize == 0 {
 		bufferSize = _defaultBufferSize
 	}
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -99,17 +99,12 @@ type CloseFunc func() error
 // if bufferSize = 0, we set it to defaultBufferSize
 // if flushInterval = 0, we set it to defaultFlushInterval
 func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteSyncer, CloseFunc) {
-	ctx, cancel := context.WithCancel(context.Background())
-
-	closefunc := func() error {
-		cancel()
-		return ws.Sync()
-	}
-
 	if _, ok := ws.(*bufferWriterSyncer); ok {
 		// no need to layer on another buffer
-		return ws, closefunc
+		return ws, func() error { return nil }
 	}
+
+	ctx, cancel := context.WithCancel(context.Background())
 
 	if bufferSize == 0 {
 		bufferSize = defaultBufferSize
@@ -140,6 +135,11 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteS
 			}
 		}
 	}()
+
+	closefunc := func() error {
+		cancel()
+		return ws.Sync()
+	}
 
 	return ws, closefunc
 }
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -91,6 +91,8 @@ const defaultBufferSize = 256 * 1024
 const defaultFlushInterval = 30 * time.Second
 
 // Buffer wraps a WriteSyncer in a buffer to improve performance,
+// if bufferSize=0, we set it to defaultBufferSize
+// if flushInterval=0, we set it to defaultFlushInterval
 func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) WriteSyncer {
 	if lws, ok := ws.(*lockedWriteSyncer); ok {
 		if _, ok := lws.ws.(*bufferWriterSyncer); ok {
@@ -126,6 +128,19 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) WriteSy
 }
 
 func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
+
+	// there are some logic internal for bufio.Writer here:
+	// 1. when the buffer is not enough, log would be written to disk directly
+	// 2. when the buffer is enough, log would not be flushed until the buffer is filled up
+	// this would lead to log spliting, which is not acceptable for log collector
+	// so we need to flush bufferWriter before writing the log into bufferWriter
+	if len(bs) > s.bufferWriter.Available() && s.bufferWriter.Buffered() > 0 {
+		err := s.bufferWriter.Flush()
+		if err != nil {
+			return 0, err
+		}
+	}
+
 	return s.bufferWriter.Write(bs)
 }
 
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -84,11 +84,13 @@ type bufferWriterSyncer struct {
 	ticker       *time.Ticker
 }
 
-// defaultBufferSize sizes the buffer associated with each WriterSync.
-const defaultBufferSize = 256 * 1024
+const (
+	// defaultBufferSize sizes the buffer associated with each WriterSync.
+	defaultBufferSize = 256 * 1024
 
-// defaultFlushInterval means the default flush interval
-const defaultFlushInterval = 30 * time.Second
+	// defaultFlushInterval means the default flush interval
+	defaultFlushInterval = 30 * time.Second
+)
 
 // CloseFunc should be called when the caller exits to clean up buffers.
 type CloseFunc func() error
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -183,6 +183,7 @@ func (s *bufferWriterSyncer) flushLoop() {
 // close closes the buffer, cleans up background goroutines, and flushes
 // remaining, unwritten data.
 func (s *bufferWriterSyncer) close() error {
+	s.ticker.Stop()
 	close(s.stop)
 	return s.Sync()
 }
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -127,7 +127,7 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteS
 	})
 
 	// flush buffer every interval
-	// we only exit this goroutine after closefunc called explicitly
+	// we do not need to exit this goroutine until closefunc called explicitly
 	go func() {
 		select {
 		case <-ticker.C:
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -22,7 +22,6 @@ package zapcore
 
 import (
 	"bufio"
-	"context"
 	"io"
 	"sync"
 	"time"
@@ -80,6 +79,7 @@ func (s *lockedWriteSyncer) Sync() error {
 
 type bufferWriterSyncer struct {
 	sync.Mutex
+	stop chan struct{}
 
 	bufferWriter *bufio.Writer
 	ticker       *time.Ticker
@@ -96,7 +96,7 @@ const (
 // CloseFunc should be called when the caller exits to clean up buffers.
 type CloseFunc func() error
 
-// Buffer wraps a WriteSyncer in a buffer to improve performance,
+// Buffer wraps a WriteSyncer in a buffer to improve performance
 // if bufferSize = 0, we set it to defaultBufferSize
 // if flushInterval = 0, we set it to defaultFlushInterval
 func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteSyncer, CloseFunc) {
@@ -104,8 +104,6 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteS
 		// no need to layer on another buffer
 		return ws, func() error { return nil }
 	}
-
-	ctx, cancel := context.WithCancel(context.Background())
 
 	if bufferSize == 0 {
 		bufferSize = defaultBufferSize
@@ -117,7 +115,8 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteS
 
 	ticker := time.NewTicker(flushInterval)
 
-	ws = &bufferWriterSyncer{
+	bws := &bufferWriterSyncer{
+		stop:         make(chan struct{}),
 		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
 		ticker:       ticker,
 	}
@@ -130,21 +129,24 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteS
 			case <-ticker.C:
 				// the background goroutine just keep syncing
 				// until the close func is called.
-				_ = ws.Sync()
-			case <-ctx.Done():
+				_ = bws.Sync()
+			case <-bws.stop:
 				return
 			}
 		}
 	}()
 
 	closefunc := func() error {
-		cancel()
-		return ws.Sync()
+		bws.stop <- struct{}{}
+
+		return bws.Sync()
 	}
 
-	return ws, closefunc
+	return bws, closefunc
 }
 
+// Write writes log data into buffer syncer directly, multiple Write calls will be batched,
+// and log data will be flushed to disk when the buffer is full or periodically.
 func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
 	// bufio is not goroutine safe, so add lock writer here
 	s.Lock()
@@ -164,6 +166,7 @@ func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
 	return s.bufferWriter.Write(bs)
 }
 
+// Sync flushes buffered log data into disk directly.
 func (s *bufferWriterSyncer) Sync() error {
 	// bufio is not goroutine safe, so add lock writer here
 	s.Lock()
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -138,7 +138,6 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (_ Writ
 // Write writes log data into buffer syncer directly, multiple Write calls will be batched,
 // and log data will be flushed to disk when the buffer is full or periodically.
 func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
-	// bufio is not goroutine safe, so add lock writer here
 	s.Lock()
 	defer s.Unlock()
 
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -146,10 +146,7 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (_ Writ
 		}
 	}()
 
-	return bws, func() error {
-		bws.stop <- struct{}{}
-		return bws.Sync()
-	}
+	return bws, bws.close
 }
 
 // Write writes log data into buffer syncer directly, multiple Write calls will be batched,
@@ -180,6 +177,13 @@ func (s *bufferWriterSyncer) Sync() error {
 	defer s.Unlock()
 
 	return s.bufferWriter.Flush()
+}
+
+// Close closes the buffer, cleans up background goroutines, and flushes
+// remaining, unwritten data.
+func (s *bufferWriterSyncer) close() error {
+	s.stop <- struct{}{}
+	return s.Sync()
 }
 
 type writerWrapper struct {
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -167,10 +167,6 @@ func (w writerWrapper) Sync() error {
 	return nil
 }
 
-func (w writerWrapper) Close() error {
-	return nil
-}
-
 type multiWriteSyncer []WriteSyncer
 
 // NewMultiWriteSyncer creates a WriteSyncer that duplicates its writes
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -153,8 +153,7 @@ func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
 	// this would lead to log spliting, which is not acceptable for log collector
 	// so we need to flush bufferWriter before writing the data into bufferWriter
 	if len(bs) > s.bufferWriter.Available() && s.bufferWriter.Buffered() > 0 {
-		err := s.bufferWriter.Flush()
-		if err != nil {
+		if err := s.bufferWriter.Flush(); err != nil {
 			return 0, err
 		}
 	}
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -124,12 +124,10 @@ func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (_ Writ
 		flushInterval = _defaultFlushInterval
 	}
 
-	ticker := time.NewTicker(flushInterval)
-
 	bws := &bufferWriterSyncer{
 		stop:         make(chan struct{}),
 		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
-		ticker:       ticker,
+		ticker:       time.NewTicker(flushInterval),
 	}
 
 	go bws.flushLoop()
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -89,16 +89,16 @@ const defaultBufferSize = 256 * 1024
 // defaultFlushInterval means the default flush interval
 const defaultFlushInterval = 30 * time.Second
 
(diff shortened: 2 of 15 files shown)


zapcore/write_syncer.go
```diff
@@ -75,6 +77,53 @@ func (s *lockedWriteSyncer) Sync() error {
...
 }
 
+type bufferWriterSyncer struct {
+	bufferWriter *bufio.Writer
+	ws           WriteSyncer
+}
+
+// bufferSize sizes the buffer associated with each log file. It's large
+// so that log records can accumulate without the logging thread blocking
+// on disk I/O. The flushDaemon will block instead.
```
prashantv: The logging thread can block on disk I/O if the buffer fills up, so I don't think this comment is accurate.

Repro:
https://play.golang.org/p/CB5qRcz84nW

If you change the limit from 1000 to 100, then there's no output (the prints are buffered), but eventually once the buffer is full, it calls the underlying `Write` method

zapcore/write_syncer.go
```diff
@@ -75,6 +77,53 @@ func (s *lockedWriteSyncer) Sync() error {
...
+
+	// bufio is not goroutine safe, so add lock writer here
+	ws = Lock(&bufferWriterSyncer{
+		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
+	})
+
+	// flush buffer every interval
+	// we do not need exit this goroutine explicitly
+	go func() {
+		for range time.NewTicker(flushInterval).C {
```
prashantv: Can we make this configurable?

zapcore/write_syncer.go
```diff
@@ -75,6 +77,53 @@ func (s *lockedWriteSyncer) Sync() error {
...
+			if err := ws.Sync(); err != nil {
```
prashantv: This is racy, simple repro:
```go
package main

import (
        "bufio"
        "fmt"
        "os"
)

func main() {
        w := bufio.NewWriter(os.Stdout)
        fmt.Fprintf(w, "hello world\n")
        go w.Flush()
        fmt.Fprintf(w, "hello 2\n")
}
```

Running this with `-race` causes data races.

hnlq715: We get Lock for this WriteSync, so there's no race actually :-)
```
	// bufio is not goroutine safe, so add lock writer here
	ws = Lock(&bufferWriterSyncer{
		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
	})

	// flush buffer every interval
	// we do not need exit this goroutine explicitly
	go func() {
		for range time.NewTicker(flushInterval).C {
			if err := ws.Sync(); err != nil {
				return
			}
		}
	}()
```

zapcore/write_syncer.go
hnlq715: Yep, we set a large bufferSize to avoid blocking on disk I/O, however when flushing or buffer fills up, it will be blocked.

zapcore/write_syncer.go
hnlq715: Yep, I'm wondering if we need add Options to do this stuff? 

Like
```golang
type Option interface {
	apply(*Logger)
}

// optionFunc wraps a func so it satisfies the Option interface.
type optionFunc func(*Logger)

func (f optionFunc) apply(log *Logger) {
	f(log)
}

// WrapCore wraps or replaces the Logger's underlying zapcore.Core.
func WrapCore(f func(zapcore.Core) zapcore.Core) Option {
	return optionFunc(func(log *Logger) {
		log.core = f(log.core)
	})
}
```

hnlq715: Rethink this, I simply add two arguments

zapcore/write_syncer.go
```diff
@@ -75,6 +77,62 @@ func (s *lockedWriteSyncer) Sync() error {
...
+
+// defaultBufferSize sizes the buffer associated with each log file. It's large
+// so that log records can accumulate without the logging thread blocking
+// on disk I/O untill buffer fills up. The flushDaemon will block instead.
+const defaultBufferSize = 256 * 1024
+
+// defaultFlushInterval means the default flush interval
+const defaultFlushInterval = 30 * time.Second
+
+// Buffer wraps a WriteSyncer in a buffer to improve performance,
```
prashantv: Add comments about periodic flushing here, and mention that if 0 is passed in, what the defaults used are (can just reference the constant names)

zapcore/write_syncer.go
prashantv: My bad, I missed the `Lock`, thanks!

zapcore/write_syncer_test.go
```diff
@@ -65,6 +66,18 @@ func TestAddSyncWriter(t *testing.T) {
...
 }
 
+func TestBufferWriter(t *testing.T) {
+	// If we pass a plain io.Writer, make sure that we still get a WriteSyncer
+	// with a no-op Sync.
+	buf := &bytes.Buffer{}
+	ws := Buffer(Buffer(AddSync(buf), 0, time.Millisecond), 0, 0)
+	requireWriteWorks(t, ws)
+	assert.Equal(t, "", buf.String(), "Unexpected log calling a no-op Write method.")
+	time.Sleep(2 * time.Millisecond)
```
prashantv: tests with `Sleep` tend to be flaky, please write the test avoiding `Sleep`.

One option is that you can pass in a custom `Writer` that signals when its' `Write` method is called

zapcore/write_syncer.go
```diff
@@ -75,6 +77,62 @@ func (s *lockedWriteSyncer) Sync() error {
...
+
+	return ws
+}
+
+func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
+	return s.bufferWriter.Write(bs)
+}
+
+func (s *bufferWriterSyncer) Sync() error {
+	return s.bufferWriter.Flush()
```
prashantv: I think we also need to call `Sync` on the underlying writer? E.g., when `Fatal` is used, we `Sync()` to make sure the log is written out before returning. If the underlying writer is some network call, then the current implementation's `Flush` will call `Write` on the underlying connection but not flush, so the log may not be emitted.

zapcore/write_syncer_test.go
hnlq715: Fixed

zapcore/write_syncer.go
hnlq715: done

prashantv: Looking at the changes to `Write`, it looks like we're trying to guarantee that a single `Write` is not split up -- but we may want to mention here that multiple `Write` calls will be batched. Just in case there's some write syncers which assume that a `Write` only contains a single log entry, which isn't compatible with buffering.

zapcore/write_syncer.go
```diff
@@ -75,6 +77,75 @@ func (s *lockedWriteSyncer) Sync() error {
...
+		flushInterval = defaultFlushInterval
+	}
+
+	// bufio is not goroutine safe, so add lock writer here
+	ws = Lock(&bufferWriterSyncer{
+		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
+	})
+
+	// flush buffer every interval
+	// we do not need exit this goroutine explicitly
```
prashantv: I think we should add a `Close` method to this interface and allow the `Close` to stop goroutines. Otherwise, we'll leak a goroutine which will affect tests that use something like goleak: https://github.com/uber-go/goleak

We enable goleak for many of our internal tests to verify there's no leaks, and usage of `Buffer` would cause a leak that can't be stopped.

The `Close` method should:
 - do one last flush
 - stop the flush goroutine

zapcore/write_syncer.go
```diff
@@ -75,6 +77,75 @@ func (s *lockedWriteSyncer) Sync() error {
...
+				return
+			}
+		}
+	}()
+
+	return ws
+}
+
+func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
+
```
prashantv: nit: delete unnecessary blank line

zapcore/write_syncer.go
```diff
@@ -75,6 +77,75 @@ func (s *lockedWriteSyncer) Sync() error {
...
+	// there are some logic internal for bufio.Writer here:
+	// 1. when the buffer is enough, data would not be flushed.
```
prashantv: Thanks for adding these notes, it seems like we may be coupled to the implementation details of the `bufio.Writer`, I'm not sure if the documentation guarantees that the implementation will behave in the same way in future.

Would it be better to just implement our own simple buffer that provides the guarantees that we need?

hnlq715: Thanks for you advise. However, this code should not be changed frequently in the future.
And I think this logic is simple enough, we do not need to implement another stuff for this.

zapcore/write_syncer.go
hnlq715: I agree, and this change may lead to add `Close` method for zap.
Please check the latest commit, thanks.

zapcore/write_syncer.go
hnlq715: fixed

zapcore/core.go
```diff
@@ -42,6 +42,8 @@ type Core interface {
 	Write(Entry, []Field) error
 	// Sync flushes buffered logs (if any).
 	Sync() error
+	// Close cleans all WriterSync resources (if any).
+	Close() error
```
prashantv: we can't change the interface here, as it would be a backwards incompatible change -- existing core implementations without `Close()` will break if we make this change.

we can either:
 - return a separate exported type from `Buffer` that implements `WriteSyncer` but also adds `Close()` and document that the caller must call `Close()` when they're done
 - return `(WriteSyncer, io.Closer, error)` from the call to buffer

We could also return just `WriteSyncer` and document that there's a close method if the returned value is up-casted, but I'd prefer to have the type-system call out the list of methods explicitly.

hnlq715: Good idea, I prefered this one like context:
```
buffer, cancel := Buffer(...)
defer cancel()
```

hnlq715: updated @prashantv

zapcore/write_syncer.go
```diff
@@ -75,6 +78,87 @@ func (s *lockedWriteSyncer) Sync() error {
...
+type bufferWriterSyncer struct {
+	ws           WriteSyncer
+	bufferWriter *bufio.Writer
+}
+
+// defaultBufferSize sizes the buffer associated with each WriterSync.
+const defaultBufferSize = 256 * 1024
+
+// defaultFlushInterval means the default flush interval
+const defaultFlushInterval = 30 * time.Second
```
prashantv, line 90: let's group these
```suggestion
const (
	// defaultBufferSize sizes the buffer associated with each WriterSync.
	defaultBufferSize = 256 * 1024

	// defaultFlushInterval means the default flush interval
	 defaultFlushInterval = 30 * time.Second
)
```

zapcore/write_syncer.go
```diff
@@ -75,6 +78,87 @@ func (s *lockedWriteSyncer) Sync() error {
...
+
+// CancelFunc should be called when the caller exits to clean up buffers.
+type CancelFunc func() error
+
+// Buffer wraps a WriteSyncer in a buffer to improve performance,
+// if bufferSize = 0, we set it to defaultBufferSize
+// if flushInterval = 0, we set it to defaultFlushInterval
+func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (WriteSyncer, CancelFunc) {
```
prashantv, line 98: instead of `CancelFunc`, `io.Closer` may be a better interface,
https://golang.org/pkg/io/#Closer

zapcore/write_syncer.go
```diff
@@ -75,6 +78,87 @@ func (s *lockedWriteSyncer) Sync() error {
...
+	ctx, cancel := context.WithCancel(context.Background())
+
+	cancelfunc := func() error {
+		cancel()
```
prashantv, line 102: after `cancel`, can we wait for the background goroutine to end?

zapcore/write_syncer.go
```diff
@@ -75,6 +78,87 @@ func (s *lockedWriteSyncer) Sync() error {
...
+		return ws.Sync()
```
prashantv, line 103: we should collect errors that may have caused the background goroutine to have ended already too

zapcore/write_syncer.go
prashantv: I think the above is still pending.

We should also call out the behaviour on errors in `Sync`

notedit:
any update?

michelaquino:
Hi, any update about that?

fevin:
any update?

hnlq715:
I'm coming back……sorry for the delay
我胡汉三又回来了……

zapcore/write_syncer.go
hnlq715: cancel would close the channel, and we do not need to wait for the goroutine to end apparently.

see the go comments below:

// The close built-in function closes a channel, which must be either
// bidirectional or send-only. It should be executed only by the sender,
// never the receiver, and has the effect of shutting down the channel after
// the last sent value is received. **After the last value has been received
// from a closed channel c**, any receive from c will succeed without
// blocking, returning the zero value for the channel element. The form
//	x, ok := <-c
// will also set ok to false for a closed channel.

zapcore/write_syncer.go
hnlq715: we do not need the make things so complicated, just simply get errors from Sync() func instead.

and the background goroutine just keep syncing until the cancel func is called.

zapcore/write_syncer.go
hnlq715: Maybe I should rename this func to CloseFunc? What do you think :-)

zapcore/write_syncer.go
```diff
@@ -75,6 +78,96 @@ func (s *lockedWriteSyncer) Sync() error {
...
+	}
+
+	// flush buffer every interval
+	// we do not need to exit this goroutine until closefunc called explicitly
+	go func() {
+		select {
+		case <-ticker.C:
+			// the background goroutine just keep syncing
+			// until the close func is called.
+			_ = ws.Sync()
```
amarine7882, line 134: Sorry if my comment is out of place here;

Is there a good way to collect these errors? If WriteSyncer makes a network call, like in a service I am currently writing, then these errors get lost which make it hard to debug

hnlq715: The timer loop here is just another way to sync data. We mostly call Write() to sync data directly, which would collect these errors.

amarine7882: I just noticed... this go routine will flush once after `flushInterval` then exit. There is no loop here.

hnlq715: thanks...my mistake...fixed

zapcore/write_syncer.go
prashantv: We don't need to wait for it to end, but I think it's a better API for callers that once `CloseFunc` returns, the "close" work is done.

E.g., if someone calls `defer CloseFunc` in their `main` function, we want to guarantee that we've done writing to the background

zapcore/write_syncer.go
prashantv: Calling `Sync` and returning that error makes sense, thanks.

zapcore/write_syncer.go
```diff
@@ -75,6 +78,98 @@ func (s *lockedWriteSyncer) Sync() error {
...
+
+	ws = &bufferWriterSyncer{
+		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
+		ticker:       ticker,
+	}
+
+	// flush buffer every interval
+	// we do not need to exit this goroutine until closefunc called explicitly
+	go func() {
+		for {
```
prashantv, line 130: do we have any test validating that this is called in a loop?

zapcore/write_syncer.go
```diff
@@ -75,6 +78,98 @@ func (s *lockedWriteSyncer) Sync() error {
...
+	defer s.Unlock()
+
+	// there are some logic internal for bufio.Writer here:
+	// 1. when the buffer is enough, data would not be flushed.
+	// 2. when the buffer is not enough, data would be flushed as soon as the buffer fills up.
+	// this would lead to log spliting, which is not acceptable for log collector
+	// so we need to flush bufferWriter before writing the data into bufferWriter
+	if len(bs) > s.bufferWriter.Available() && s.bufferWriter.Buffered() > 0 {
+		err := s.bufferWriter.Flush()
+		if err != nil {
```
prashantv, line 157: nit: can inline:
```suggestion
		if err := s.bufferWriter.Flush(); err != nil {
```

zapcore/write_syncer_bench_test.go
```diff
@@ -53,4 +56,51 @@ func BenchmarkMultiWriteSyncer(b *testing.B) {
 			}
 		})
 	})
+	b.Run("4 discarder with buffer", func(b *testing.B) {
```
prashantv, line 59: why is this test needed?

zapcore/write_syncer_bench_test.go
```diff
@@ -53,4 +56,51 @@ func BenchmarkMultiWriteSyncer(b *testing.B) {
...
+			for pb.Next() {
+				w.Write([]byte("foobarbazbabble"))
+			}
+		})
+	})
+}
+
+func BenchmarkWriteSyncer(b *testing.B) {
+	b.Run("write file with no buffer", func(b *testing.B) {
+		file, err := ioutil.TempFile(".", "*")
```
prashantv, line 78: nit: just pass `""` as the directory to use the default, and pass in `"log"` as the file argument (it'll automatically have values appended, no need for `*`)

zapcore/write_syncer_bench_test.go
```diff
@@ -53,4 +56,51 @@ func BenchmarkMultiWriteSyncer(b *testing.B) {
...
+		assert.Nil(b, err)
```
prashantv, line 79: here and throughout this change, use `assert.NoError` instead of `nil` when checking errors

zapcore/write_syncer_bench_test.go
hnlq715: To compare the performance between two situations below:
1. Writing to the discarder directly
2. Writing to the buffer directly and flushing to the discarder

zapcore/write_syncer_bench_test.go
hnlq715: alright, fixed

zapcore/write_syncer_bench_test.go
hnlq715: fixed

zapcore/write_syncer.go
hnlq715: Nice question, test case added.

hnlq715:
> Do we have tests to ensure that small writes are buffered and only written to the underying `Writer` periodically or after the size limit is hit?

Yes, we have test case for these situations above.

hnlq715:
any updates? 😃

michelaquino:
Any update about this PR?

zapcore/write_syncer.go
prashantv: > we may want to mention here that multiple `Write` calls will be batched

I don't think we documented this anywhere, and it's pretty important

zapcore/write_syncer.go
```diff
@@ -75,6 +78,99 @@ func (s *lockedWriteSyncer) Sync() error {
 	return err
 }
 
+type bufferWriterSyncer struct {
+	sync.Mutex
```
abhinav, line 82: Style: Please add an empty line between embedded fields and regular fields.

```suggestion
	sync.Mutex

```

https://github.com/uber-go/guide/blob/master/style.md#embedding-in-structs

zapcore/write_syncer.go
```diff
@@ -75,6 +78,99 @@ func (s *lockedWriteSyncer) Sync() error {
...
+				_ = ws.Sync()
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+
+	closefunc := func() error {
+		cancel()
+		return ws.Sync()
```
abhinav, line 141: We need to stop the ticker here first as well. Otherwise we'll leak the ticker goroutine.

```suggestion
		ticker.Stop()
		cancel()
		return ws.Sync()
```

I think it's also preferable for closefunc to block until the goroutine exits
so that we can be sure of a clean exit. We can do the final flush after the
goroutine exits.

```
stopped := make(chan struct{})
go func() {
  defer close(stopped)
  for { ... }
}

closefunc := func() error {
  ticker.Stop()
  cancel()
  <-stopped
  return ws.Sync()
}
```

zapcore/write_syncer_test.go
```diff
@@ -65,6 +70,80 @@ func TestAddSyncWriter(t *testing.T) {
...
 
+func TestBufferWriter(t *testing.T) {
+	// If we pass a plain io.Writer, make sure that we still get a WriteSyncer
+	// with a no-op Sync.
+	t.Run("sync", func(t *testing.T) {
+		buf := &bytes.Buffer{}
+		ws, close := Buffer(AddSync(buf), 0, 0)
+		defer close()
+		requireWriteWorks(t, ws)
+		assert.Equal(t, "", buf.String(), "Unexpected log calling a no-op Write method.")
```
abhinav, line 81: We can use `assert.Empty` here:

```suggestion
		assert.Empty(t, buf.String(), "Unexpected log calling a no-op Write method.")
```

zapcore/write_syncer_test.go
```diff
@@ -65,6 +70,80 @@ func TestAddSyncWriter(t *testing.T) {
...
+		assert.NoError(t, ws.Sync(), "Unexpected error calling a no-op Sync method.")
+		assert.Equal(t, "foo", buf.String(), "Unexpected log string")
+	})
+
+	t.Run("1 close", func(t *testing.T) {
+		buf := &bytes.Buffer{}
+		ws, close := Buffer(AddSync(buf), 0, 0)
+		requireWriteWorks(t, ws)
+		assert.Equal(t, "", buf.String(), "Unexpected log calling a no-op Write method.")
```
abhinav, line 90:
```suggestion
		assert.Empty(t, buf.String(), "Unexpected log calling a no-op Write method.")
```

zapcore/write_syncer_test.go
```diff
@@ -65,6 +70,80 @@ func TestAddSyncWriter(t *testing.T) {
...
+		buf := &bytes.Buffer{}
+		ws, close := Buffer(AddSync(buf), 0, 0)
+		defer close()
+		requireWriteWorks(t, ws)
+		assert.Equal(t, "", buf.String(), "Unexpected log calling a no-op Write method.")
+		assert.NoError(t, ws.Sync(), "Unexpected error calling a no-op Sync method.")
+		assert.Equal(t, "foo", buf.String(), "Unexpected log string")
+	})
+
+	t.Run("1 close", func(t *testing.T) {
```
abhinav, line 86: naming-wise, this is just a test of close. The number is less relevant because for the 2 case, it's about double wrapping.

```suggestion
	t.Run("close", func(t *testing.T) {
```

zapcore/write_syncer_test.go
```diff
@@ -65,6 +70,80 @@ func TestAddSyncWriter(t *testing.T) {
...
+		buf := &bytes.Buffer{}
+		ws, close := Buffer(AddSync(buf), 0, 0)
+		requireWriteWorks(t, ws)
+		assert.Equal(t, "", buf.String(), "Unexpected log calling a no-op Write method.")
+		close()
+		assert.Equal(t, "foo", buf.String(), "Unexpected log string")
+	})
+
+	t.Run("2 close", func(t *testing.T) {
```
abhinav, line 95:
```suggestion
	t.Run("wrap twice", func(t *testing.T) {
```

zapcore/write_syncer_test.go
```diff
@@ -65,6 +70,80 @@ func TestAddSyncWriter(t *testing.T) {
...
+		close1()
+		assert.Equal(t, "foo", buf.String(), "Unexpected log string")
+	})
+
+	t.Run("small buffer", func(t *testing.T) {
+		buf := &bytes.Buffer{}
+		bufsync, close1 := Buffer(AddSync(buf), 5, 0)
+		ws, close2 := Buffer(bufsync, 5, 0)
+		defer close1()
+		defer close2()
```
abhinav, line 111: Why do we have double wrapping in this test case? The close functionality for double wrapping is already tested above.
```suggestion
		ws, close := Buffer(AddSync(buf), 5, 0)
		defer close()
```

zapcore/write_syncer_test.go
```diff
@@ -65,6 +70,80 @@ func TestAddSyncWriter(t *testing.T) {
...
+		assert.Equal(t, "foo", buf.String(), "Unexpected log string")
+	})
+
+	t.Run("flush error", func(t *testing.T) {
+		ws, close := Buffer(AddSync(&errorWriter{}), 4, time.Nanosecond)
+		n, err := ws.Write([]byte("foo"))
+		require.NoError(t, err, "Unexpected error writing to WriteSyncer.")
+		require.Equal(t, 3, n, "Wrote an unexpected number of bytes.")
+		ws.Write([]byte("foo"))
+		assert.NotNil(t, close())
```
abhinav, line 124: We can use assert.Error here

```suggestion
		assert.Error(t, close(), "Expected close to fail.")
```

abhinav:
Also, seeing as this is the first piece of non-test code in Zap that spawns a goroutine, we should use [goleak](https://github.com/uber-go/goleak) to test for leaks.

mhughes:
Any Updates?     Appreciate all the work being done.

zapcore/write_syncer.go
hnlq715: Sure, we could do this through a channel.

zapcore/write_syncer.go
hnlq715: Sure, I'll add some comments here.

hnlq715:
Thanks for comments from @abhinav 
I would improve this close logic recently

UnderTreeTech:
Bro, cloud you provide some comparison charts of performance evaluation?

hnlq715:
> Fixes #663
> 
> This change adds `Buffer`, which wraps a `WriteSyncer` with buffering. This use `bufio.Writer` to buffer in-memory, and flush changes periodically (or when the buffer is full). The `Sync` method forces an immediate flush.
> 
> This can improve performance by amortizing any fixed overheads of the underlying `WriteSyncer`.
> 
> bechmark result
> 
> ```
> Running tool: /usr/local/opt/go/libexec/bin/go test -benchmem -run=^$ go.uber.org/zap/zapcore -bench ^(BenchmarkMultiWriteSyncer|BenchmarkWriteSyncer)$
> 
> goos: darwin
> goarch: amd64
> pkg: go.uber.org/zap/zapcore
> BenchmarkMultiWriteSyncer/2_discarder-4         	41213204	        66.5 ns/op	      16 B/op	       1 allocs/op
> BenchmarkMultiWriteSyncer/4_discarder-4         	19545279	        72.0 ns/op	      16 B/op	       1 allocs/op
> BenchmarkMultiWriteSyncer/4_discarder_with_buffer-4         	 8759560	       143 ns/op	      16 B/op	       1 allocs/op
> BenchmarkWriteSyncer/write_file_with_no_buffer-4            	  160742	      9367 ns/op	      16 B/op	       1 allocs/op
> BenchmarkWriteSyncer/write_file_with_buffer-4               	 7559112	       198 ns/op	      16 B/op	       1 allocs/op
> PASS
> ok  	go.uber.org/zap/zapcore	12.013s
> Success: Benchmarks passed.
> ```

@UnderTreeTech Here is the benchmark result :-)

zapcore/write_syncer_test.go
```diff
@@ -65,6 +71,80 @@ func TestAddSyncWriter(t *testing.T) {
 	assert.NoError(t, ws.Sync(), "Unexpected error calling a no-op Sync method.")
 }
 
+func TestBufferWriter(t *testing.T) {
+	goleak.VerifyNone(t)
```
prashantv, line 75:
```suggestion
	defer goleak.VerifyNone(t)
```

zapcore/write_syncer.go
prashantv: unresolved, since this doesn't look done

zapcore/write_syncer.go
```diff
@@ -75,6 +77,117 @@ func (s *lockedWriteSyncer) Sync() error {
 	return err
 }
 
+type bufferWriterSyncer struct {
```
prashantv, line 80: should we move this whole type and methods into a separate file?

zapcore/write_syncer.go
```diff
@@ -75,6 +77,117 @@ func (s *lockedWriteSyncer) Sync() error {
...
+		flushInterval = _defaultFlushInterval
+	}
+
+	ticker := time.NewTicker(flushInterval)
+
+	bws := &bufferWriterSyncer{
+		stop:         make(chan struct{}),
+		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
+		ticker:       ticker,
+	}
```
prashantv, line 133:
```suggestion
	bws := &bufferWriterSyncer{
		stop:         make(chan struct{}),
		bufferWriter: bufio.NewWriterSize(ws, bufferSize),
		ticker:       time.NewTicker(flushInterval),
	}
```

zapcore/write_syncer.go
```diff
@@ -75,6 +77,117 @@ func (s *lockedWriteSyncer) Sync() error {
...
+//     ws, closeWS := zapcore.Buffer(ws, 0, 0)
+//     defer closeWS()
+//     // ...
+//   }
+//
+// The buffer size defaults to 256 kB if set to zero.
+// The flush interval defaults to 30 seconds if set to zero.
+func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (_ WriteSyncer, close func() error) {
+	if _, ok := ws.(*bufferWriterSyncer); ok {
+		// no need to layer on another buffer
```
prashantv, line 115: the different `bufferWriteSyncer` may have different buffer sizes / flush intervals. we may end up skipping the layering, but could cause unexpected buffering/flushing

zapcore/write_syncer.go
```diff
@@ -75,6 +77,117 @@ func (s *lockedWriteSyncer) Sync() error {
...
+
+	go bws.flushLoop()
+
+	return bws, bws.close
+}
+
+// Write writes log data into buffer syncer directly, multiple Write calls will be batched,
+// and log data will be flushed to disk when the buffer is full or periodically.
+func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
+	// bufio is not goroutine safe, so add lock writer here
```
prashantv, line 143:
```suggestion
```

not sure this comment is useful

zapcore/write_syncer.go
```diff
@@ -75,6 +77,117 @@ func (s *lockedWriteSyncer) Sync() error {
...
+	s.Lock()
+	defer s.Unlock()
+
+	// there are some logic internal for bufio.Writer here:
+	// 1. when the buffer is enough, data would not be flushed.
+	// 2. when the buffer is not enough, data would be flushed as soon as the buffer fills up.
+	// this would lead to log spliting, which is not acceptable for log collector
+	// so we need to flush bufferWriter before writing the data into bufferWriter
```
prashantv, line 151: Let me try and simplify and clarify this comment,
```suggestion
	// To avoid partial writes from being flushed, we manually flush the existing buffer if:
	// * The current write doesn't fit into the buffer fully, and
	// * The buffer is not empty (since bufio will not split large writes when the buffer is empty)
```

zapcore/write_syncer.go
```diff
@@ -75,6 +77,117 @@ func (s *lockedWriteSyncer) Sync() error {
...
+			return 0, err
+		}
+	}
+
+	return s.bufferWriter.Write(bs)
+}
+
+// Sync flushes buffered log data into disk directly.
+func (s *bufferWriterSyncer) Sync() error {
+	// bufio is not goroutine safe, so add lock writer here
```
prashantv, line 163:
```suggestion
```

zapcore/write_syncer.go
```diff
@@ -75,6 +77,117 @@ func (s *lockedWriteSyncer) Sync() error {
...
+	return s.bufferWriter.Flush()
+}
+
+// flushLoop flushes the buffer at the configured interval until Close is
+// called.
+func (s *bufferWriterSyncer) flushLoop() {
+	for {
+		select {
+		case <-s.ticker.C:
+			_ = s.Sync()
```
prashantv, line 176: Should we add a note for why we ignore? Specifically that the underlying bufio writer stores any errors and returns them on future `Sync()`, and we return any error from `Sync()` as part of the `close`

zapcore/write_syncer.go
hnlq715: I think we should not support mulitple buffer layer to keep it simple.

zapcore/write_syncer.go
hnlq715: This file already contains Lock syncer and MultiWrite syncer, I think it is a proper place for Buffer syncer :-)

abhinav (approved):

mhughes:
Looking forward to this being merged.  Appreciate all the hard work @hnlq715 . @prashantv @abhinav  do you have an ETA for when this will make the mainline?  Appreciate it

UnderTreeTech:
Any updates?

msilberkasten-ring:
Looking forward to get this merged as well !

zapcore/write_syncer.go
```diff
@@ -75,6 +77,114 @@ func (s *lockedWriteSyncer) Sync() error {
...
+// Write writes log data into buffer syncer directly, multiple Write calls will be batched,
+// and log data will be flushed to disk when the buffer is full or periodically.
+func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
+	s.Lock()
+	defer s.Unlock()
+
+	// To avoid partial writes from being flushed, we manually flush the existing buffer if:
+	// * The current write doesn't fit into the buffer fully, and
+	// * The buffer is not empty (since bufio will not split large writes when the buffer is empty)
+	if len(bs) > s.bufferWriter.Available() && s.bufferWriter.Buffered() > 0 {
```
prashantv, line 147: this condition is not tested in the tests. even if I comment out this block, the tests don't notice any issues,
```
> git diff
diff --git a/zapcore/write_syncer.go b/zapcore/write_syncer.go
index dcb0247..844c109 100644
--- a/zapcore/write_syncer.go
+++ b/zapcore/write_syncer.go
@@ -144,11 +144,11 @@ func (s *bufferWriterSyncer) Write(bs []byte) (int, error) {
        // To avoid partial writes from being flushed, we manually flush the existing buffer if:
        // * The current write doesn't fit into the buffer fully, and
        // * The buffer is not empty (since bufio will not split large writes when the buffer is empty)
-       if len(bs) > s.bufferWriter.Available() && s.bufferWriter.Buffered() > 0 {
-               if err := s.bufferWriter.Flush(); err != nil {
-                       return 0, err
-               }
-       }
+       //if len(bs) > s.bufferWriter.Available() && s.bufferWriter.Buffered() > 0 && false {
+       //      if err := s.bufferWriter.Flush(); err != nil {
+       //              return 0, err
+       //      }
+       //}

        return s.bufferWriter.Write(bs)
 }
> go test
invalid value "unknown" for flag -level: unrecognized level: "unknown"
Usage of test:
  -level value
        minimum enabled logging level
PASS
ok      go.uber.org/zap 1.007s
```

zapcore/write_syncer.go
abhinav: Oh, I missed this before. I think silently ignoring the call here would be
surprising. We could extract and wrap the underlying buffer, but that might be
unnecessarily messy.

We should just double wrap here. WriteSyncers should usually not be
constructed in super complicated logic where risk of unintentional double
wrapping is too high.

zapcore/write_syncer.go
```diff
@@ -75,6 +77,114 @@ func (s *lockedWriteSyncer) Sync() error {
...
+//   }
+//
+// The buffer size defaults to 256 kB if set to zero.
+// The flush interval defaults to 30 seconds if set to zero.
+func Buffer(ws WriteSyncer, bufferSize int, flushInterval time.Duration) (_ WriteSyncer, close func() error) {
+	if _, ok := ws.(*bufferWriterSyncer); ok {
+		// no need to layer on another buffer
+		return ws, func() error { return nil }
+	}
+
```
abhinav, line 118:
```suggestion
```

zapcore/write_syncer_test.go
```diff
@@ -65,6 +71,80 @@ func TestAddSyncWriter(t *testing.T) {
 	assert.NoError(t, ws.Sync(), "Unexpected error calling a no-op Sync method.")
 }
 
+func TestBufferWriter(t *testing.T) {
+	defer goleak.VerifyNone(t)
+
+	// If we pass a plain io.Writer, make sure that we still get a WriteSyncer
+	// with a no-op Sync.
```
prashantv, line 78: The tests don't validate that the buffer writer doesn't acutally write to the underlying buffer straight away. We should ideally have a test that writes a log that should be significantly smaller than the buffer, and it should not be available in the underlying writer till after an explicit Sync or the background ticker flushes

zapcore/write_syncer_test.go
```diff
@@ -65,6 +71,80 @@ func TestAddSyncWriter(t *testing.T) {
...
+		bws.Lock()
+		assert.Equal(t, "foo", buf.String(), "Unexpected log string")
+		bws.Unlock()
+
+		// flush twice to validate loop logic
+		requireWriteWorks(t, ws)
+		ztest.Sleep(10 * time.Millisecond)
+		bws.Lock()
+		assert.Equal(t, "foofoo", buf.String(), "Unexpected log string")
+		bws.Unlock()
```
prashantv, line 144: why is the lock necessary here? there should be nothing to flush, so the `buf` should not be mutated while we read this (and if it does get mutated, that's likely a real race issue to investigate)?

zapcore/write_syncer_test.go
```diff
@@ -65,6 +71,80 @@ func TestAddSyncWriter(t *testing.T) {
...
+		ws, close := Buffer(AddSync(buf), 5, 0)
+		defer close()
+		requireWriteWorks(t, ws)
+		assert.Equal(t, "", buf.String(), "Unexpected log calling a no-op Write method.")
+		requireWriteWorks(t, ws)
+		assert.Equal(t, "foo", buf.String(), "Unexpected log string")
+	})
+
+	t.Run("flush error", func(t *testing.T) {
+		ws, close := Buffer(AddSync(&errorWriter{}), 4, time.Nanosecond)
```
prashantv, line 120: why such a short flush timer here? we're relying on the buffer size limit to flush rather than the background timer I think?

zapcore/write_syncer_test.go
hnlq715: Because we write the buf in goroutine 20, and read the buf in goroutine 19, maybe this race log says all of it.
And it should only reach race condition in this test case, because we need to check what the buf contains by `buf.String()`.
```
Read at 0x00c000117560 by goroutine 19:
  bytes.(*Buffer).String()
      /home/liqi/workspace/go/src/bytes/buffer.go:65 +0x411
  go.uber.org/zap/zapcore.TestBufferWriter.func6()
      /home/liqi/workspace/zap/zapcore/write_syncer_test.go:134 +0x1a4
  testing.tRunner()
      /home/liqi/workspace/go/src/testing/testing.go:1127 +0x202

Previous write at 0x00c000117560 by goroutine 20:
  bytes.(*Buffer).grow()
      /home/liqi/workspace/go/src/bytes/buffer.go:128 +0x484
  bytes.(*Buffer).Write()
      /home/liqi/workspace/go/src/bytes/buffer.go:172 +0x184
  go.uber.org/zap/zapcore.(*writerWrapper).Write()
      <autogenerated>:1 +0x87
  bufio.(*Writer).Flush()
      /home/liqi/workspace/go/src/bufio/bufio.go:607 +0x13c
  go.uber.org/zap/zapcore.(*bufferWriterSyncer).Sync()
      /home/liqi/workspace/zap/zapcore/write_syncer.go:161 +0xb8
  go.uber.org/zap/zapcore.(*bufferWriterSyncer).flushLoop()
      /home/liqi/workspace/zap/zapcore/write_syncer.go:173 +0x87

Goroutine 19 (running) created at:
  testing.(*T).Run()
      /home/liqi/workspace/go/src/testing/testing.go:1178 +0x796
  go.uber.org/zap/zapcore.TestBufferWriter()
      /home/liqi/workspace/zap/zapcore/write_syncer_test.go:128 +0x1cc
  testing.tRunner()
      /home/liqi/workspace/go/src/testing/testing.go:1127 +0x202

Goroutine 20 (running) created at:
  go.uber.org/zap/zapcore.Buffer()
      /home/liqi/workspace/zap/zapcore/write_syncer.go:133 +0x32b
  go.uber.org/zap/zapcore.TestBufferWriter.func6()
      /home/liqi/workspace/zap/zapcore/write_syncer_test.go:130 +0xde
  testing.tRunner()
      /home/liqi/workspace/go/src/testing/testing.go:1127 +0x202
```

zapcore/write_syncer_test.go
hnlq715: Absolutely, we can definitely set this timer to 0.

zapcore/write_syncer.go
hnlq715: Alright, we can just double wrap this buffer syncer directly.

abhinav (changes requested):
This PR has stalled because we need to follow up with some changes
before we merge, and have not had a chance to do that yet. Posting
this for visibility. The gist of the problem is:

There's a data race in the flush test for the buffered WriteSyncer.
Adding a lock on the underlying Buffer makes the test pass, but it
doesn't address the data race to a degree we are satisfied with.

We believe that the best way forward is to build upon the work done in
#897 and add the ability to build tickers to the Clock interface, and
with control of a fake ticker, exercise the flush path in the tests.
Note that because we're in zapcore, the Clock interface will have to be
moved to zapcore and re-exported from the top-level zap package.

Adding `NewTicker` to the Clock interface is not straightforward
because [`time.Ticker`][1] is a struct the internals of which we do not
have a lot of control over. So we have to decide whether the
`NewTicker` method on Clock would return a `*time.Ticker` or a custom
Ticker struct (similar to [benbjohnson/clock][2])

[1]: https://golang.org/pkg/time/#Ticker
[2]: https://pkg.go.dev/github.com/benbjohnson/clock#Ticker

Besides just the interface design, another thing to decide is the how
to instantiate the buffer. Previously, with the three arguments of the
Buffer function, two of which were optional, we were already pushing
it. With clock, which is also optional, we'd have 4 arguments to the
function, 3 of which can be omitted. At that point we'd want a
different way of instantiating the WriteSyncer. We considered a couple
ideas.

1. An exposed struct with unexported fields for internal state:

    ```
    type BufferedWriteSyncer struct {
      WriteSyncer
      
      Size int
      FlushInterval time.Duration
      Clock Clock
      
      // unexported fields for state
    }
    ```

    This would require checking if the internal state of the buffer had
    been initialized on every Write call. Given that we already have
    a lock on the WriteSyncer, it would be a simple boolean check so it
    would not be overly expensive.

    Usage to wrap a WriteSyncer with the default configuration would be:

    ```
    ws := &BufferedWriteSyncer{WriteSyncer: ws}
    ```

2. A configuration struct with a Wrap method:

```
type BufferConfig struct {
  Size int
  FlushInterval time.Duration
  Clock Clock
}

func (BufferConfig) Wrap(WriteSyncer) WriteSyncer
```

Usage to wrap a WriteSyncer with the default configuration would be:

```
ws := BufferConfig{}.Wrap(ws)
```

3. Functional options

```
type BufferOption
  func BufferSize(int) BufferOption
  func BufferFlushInterval(time.Duration) BufferOption
  func BufferClock(Clock) Clock

func Buffer(ws WriteSyncer, opts ...BufferOption) WriteSyncer
```

This would pollute the zapcore namespace too heavily so we're
disinclined to use it.

I'm personally leaning towards (1) but I'm not fully sold on it.

---

So in short, our problems are:

- Data race in the flush test
- Control over the source of time
- Clock interface design
- Buffer instantiation API

Unfortunately, as I mentioned before, we haven't had a chance to follow
up with the changes necessary for this, and I expect we won't get to it
for a couple more weeks at least.

hnlq715:
Thanks for the declaration, and I personally prefer the option 1 :-)

I would like to wait for your Clock interface implementation.

UnderTreeTech:
Any updates? @abhinav

abhinav:
> Any updates? @abhinav

Oops, I missed that. Yeah, we're working on it. #948 is the first piece of this, which will allow @moisesvega to implement the test without the data race.

moisesvega:
Here is the [PR](https://github.com/uber-go/zap/pull/952) with the new proposal Buffer instantiation API and the use of the new Clock Interface to remove the data race in the test cases.

hnlq715:
nice work :-)

abhinav:
We pulled these changes into #952 and developed on top of them, and merged the final version in #961.

Thanks for your contribution @hnlq715!

PR #786, opened 2020-02-13
Use Time.AppendFormat when possible

### LINKED ISSUE #783: Taking advantage of Time.AppendFormat
Hi,

I'm opening this issue to discuss a potential API change to the [buffer type](https://github.com/uber-go/zap/blob/master/buffer/buffer.go).

I have a custom encoder where times are encoded as strings formatted with the `Format` method from Go. Currently I have to do this:

```
enc.buf.AppendString(value.UTC().Format(time.RFC3339))
```

but it would be nice if we could just use the [AppendFormat](https://golang.org/pkg/time/#Time.AppendFormat) method to avoid an allocation.

I'm thinking of adding a method to the buffer type with this signature: `AppendTime(t time.Time, layout string)`.

Another solution would be to allow users to set the byte slice used by the buffer type so that I could write this:

```
buf := enc.buf.Bytes()
buf = value.AppendFormat(buf, time.RFC3339)
enc.buf.SetBytes(buf)
```

less elegant than having `AppendTime` but maybe a good solution if you don't want to special case times in the buffer type.

prashantv:
Fixes #783.

We can take advantage of Time.AppendFormat by adding a new exported
method to the JSON encoder, and upcasting when encoding time to use
the append-based method. This avoids an allocation to convert the time
to a string before appending to the buffer.

The benchmarks use the production config, which uses a nanos encoder
so to understand the performance difference, I tweaked
`BenchmarkZapJSON` to use RFC3339TimeEncoder and ran benchmarks:

```
> benchcmp -best old new
benchmark               old ns/op     new ns/op     delta
BenchmarkZapJSON-12     514           497           -3.31%

benchmark               old allocs     new allocs     delta
BenchmarkZapJSON-12     5              4              -20.00%

benchmark               old bytes     new bytes     delta
BenchmarkZapJSON-12     1297          1265          -2.47%
```

I also wrote a benchmark that only logs a simple message using the
RFC3339TimeEncoder,
```
func BenchmarkTimeEncoder(b *testing.B) {
        cfg := NewProductionConfig().EncoderConfig
        cfg.EncodeTime = zapcore.RFC3339TimeEncoder
        logger := New(
                zapcore.NewCore(
                        zapcore.NewJSONEncoder(cfg),
                        &ztest.Discarder{},
                        DebugLevel,
                ))
        b.ResetTimer()

        for i := 0; i < b.N; i++ {
                logger.Info("test")
        }
}
```

Results:
```
> benchcmp -best old new
benchmark                   old ns/op     new ns/op     delta
BenchmarkTimeEncoder-12     695           571           -17.84%

benchmark                   old allocs     new allocs     delta
BenchmarkTimeEncoder-12     1              0              -100.00%

benchmark                   old bytes     new bytes     delta
BenchmarkTimeEncoder-12     32            0             -100.00%
```

### DIFF
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -127,17 +127,26 @@ func encodeTimeLayout(t time.Time, layout string, enc PrimitiveArrayEncoder) {
 
 // ISO8601TimeEncoder serializes a time.Time to an ISO8601-formatted string
 // with millisecond precision.
+//
+// If enc supports AppendTimeLayout(t time.Time,layout string), it's used
+// instead of appending a per-formatted string value.
 func ISO8601TimeEncoder(t time.Time, enc PrimitiveArrayEncoder) {
 	encodeTimeLayout(t, "2006-01-02T15:04:05.000Z0700", enc)
 }
 
 // RFC3339TimeEncoder serializes a time.Time to an RFC3339-formatted string.
+//
+// If enc supports AppendTimeLayout(t time.Time,layout string), it's used
+// instead of appending a per-formatted string value.
 func RFC3339TimeEncoder(t time.Time, enc PrimitiveArrayEncoder) {
 	encodeTimeLayout(t, time.RFC3339, enc)
 }
 
 // RFC3339NanoTimeEncoder serializes a time.Time to an RFC3339-formatted string
 // with nanosecond precision.
+//
+// If enc supports AppendTimeLayout(t time.Time,layout string), it's used
+// instead of appending a per-formatted string value.
 func RFC3339NanoTimeEncoder(t time.Time, enc PrimitiveArrayEncoder) {
 	encodeTimeLayout(t, time.RFC3339Nano, enc)
 }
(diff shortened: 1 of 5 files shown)


zapcore/encoder.go
```diff
@@ -112,21 +112,34 @@ func EpochNanosTimeEncoder(t time.Time, enc PrimitiveArrayEncoder) {
 	enc.AppendInt64(t.UnixNano())
 }
 
+type appendTimeEncoder interface {
```
abhinav: This will become part of the contract for custom encoders. Can we export this?
Or we can document on the ISO8601, RFC3339, and RFC3339Nano time encoders that
if the encoder implements this interface, we will use that method.

zapcore/encoder.go
```diff
@@ -112,21 +112,43 @@ func EpochNanosTimeEncoder(t time.Time, enc PrimitiveArrayEncoder) {
...
+	AppendTimeLayout(time.Time, string)
+}
+
+func encodeTimeLayout(t time.Time, layout string, enc PrimitiveArrayEncoder) {
+	if enc, ok := enc.(appendTimeEncoder); ok {
```
jcorbin: fwiw: it's possible to just use an anonymous `duck, ok := val.(interface { quack() })` in cases like this where aliasing the interfaces serves no other purpose

prashantv: Yep, although the line looks kind of long and complicated.

However, I moved the interface into the method, since it's not needed outside of this method.

PR #791, opened 2020-02-21
EncoderConfig and AtomicLevel validation added

### LINKED ISSUE #777: Add Config validation to build stage to prevent runtime panics
When creating a custom zap.Config there seems to be only minimal validation of the configuration that results in runtime panics if certain properties are missing. This is dangerous as all required configuration properties should be checked at the build stage. 

There is probably a reason this has not been done, but opening an issue nonetheless.

## Example

```
	cfg := zap.Config{
		Encoding:    "json",
		OutputPaths: []string{"stderr"},
		EncoderConfig: zapcore.EncoderConfig{
			MessageKey: "msg",
		},
	}

	logger, err := cfg.Build()
	if err != nil {
		panic(err)
	}

	logger.Sugar().Errorw("Problem here", "foo", "bar") // Panic
```

This results in the following panic:

```

[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12696c6]

goroutine 1 [running]:
go.uber.org/atomic.(*Int32).Load(...)
	/Users/owner/go/pkg/mod/go.uber.org/atomic@v1.5.1/atomic.go:41
go.uber.org/zap.AtomicLevel.Level(...)
	/Users/owner/go/pkg/mod/go.uber.org/zap@v1.13.0/level.go:97
go.uber.org/zap.AtomicLevel.Enabled(0x0, 0x10b5c02, 0xc000098c30)
	/Users/owner/go/pkg/mod/go.uber.org/zap@v1.13.0/level.go:92 +0x6
go.uber.org/zap.(*SugaredLogger).log(0xc000055f00, 0xc000010002, 0x1312a3e, 0xc, 0x0, 0x0, 0x0, 0xc000055f30, 0x2, 0x2)
	/Users/owner/go/pkg/mod/go.uber.org/zap@v1.13.0/sugar.go:221 +0x1aa
go.uber.org/zap.(*SugaredLogger).Errorw(...)
	/Users/owner/go/pkg/mod/go.uber.org/zap@v1.13.0/sugar.go:191
main.main()
	/Users/owner/zaplog/zaplog.go:22 +0x27b
exit status 2
```

The issue here is we are missing a "Level"

```
Level:            zap.NewAtomicLevelAt(zapcore.DebugLevel),
```

However, there are other issues as well. Suppose you specify the "TimeKey", but not the "EncodeTime".

```
	cfg := zap.Config{
		Level:       zap.NewAtomicLevelAt(zapcore.DebugLevel),
		Encoding:    "json",
		OutputPaths: []string{"stderr"},
		EncoderConfig: zapcore.EncoderConfig{
			MessageKey: "msg",
			TimeKey:    "ts",
		},
	}
```

This configuration will build without error, but you again will be faced with `runtime error: invalid memory address or nil pointer dereference` on use.

YashishDua:
Resolves #777

### DIFF
--- a/config.go
+++ b/config.go
@@ -174,6 +174,13 @@ func (cfg Config) Build(opts ...Option) (*Logger, error) {
 		return nil, err
 	}
 
+	if cfg.Level == (AtomicLevel{}) {
+		cfg.Level = NewAtomicLevelAt(InfoLevel)
+		if cfg.Development {
+			cfg.Level = NewAtomicLevelAt(DebugLevel)
+		}
+	}
+
 	log := New(
 		zapcore.NewCore(enc, sink, cfg.Level),
 		cfg.buildOptions(errSink)...,
@@ -239,5 +246,12 @@ func (cfg Config) openSinks() (zapcore.WriteSyncer, zapcore.WriteSyncer, error) 
 }
 
 func (cfg Config) buildEncoder() (zapcore.Encoder, error) {
+	if cfg.EncoderConfig.TimeKey != "" && cfg.EncoderConfig.EncodeTime == nil {
+		cfg.EncoderConfig.EncodeTime = zapcore.EpochTimeEncoder
+		if cfg.Development {
+			cfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+		}
+	}
+
 	return newEncoder(cfg.Encoding, cfg.EncoderConfig)
 }
--- a/config.go
+++ b/config.go
@@ -244,9 +244,5 @@ func (cfg Config) openSinks() (zapcore.WriteSyncer, zapcore.WriteSyncer, error) 
 }
 
 func (cfg Config) buildEncoder() (zapcore.Encoder, error) {
-	if cfg.EncoderConfig.TimeKey != "" && cfg.EncoderConfig.EncodeTime == nil {
-		return nil, fmt.Errorf("missing EncodeTime in EncoderConfig")
-	}
-
 	return newEncoder(cfg.Encoding, cfg.EncoderConfig)
 }
--- a/config.go
+++ b/config.go
@@ -21,6 +21,7 @@
 package zap
 
 import (
+	"fmt"
 	"sort"
 	"time"
 
@@ -175,10 +176,7 @@ func (cfg Config) Build(opts ...Option) (*Logger, error) {
 	}
 
 	if cfg.Level == (AtomicLevel{}) {
-		cfg.Level = NewAtomicLevelAt(InfoLevel)
-		if cfg.Development {
-			cfg.Level = NewAtomicLevelAt(DebugLevel)
-		}
+		return nil, fmt.Errorf("missing Level")
 	}
 
 	log := New(
@@ -247,10 +245,7 @@ func (cfg Config) openSinks() (zapcore.WriteSyncer, zapcore.WriteSyncer, error) 
 
 func (cfg Config) buildEncoder() (zapcore.Encoder, error) {
 	if cfg.EncoderConfig.TimeKey != "" && cfg.EncoderConfig.EncodeTime == nil {
-		cfg.EncoderConfig.EncodeTime = zapcore.EpochTimeEncoder
-		if cfg.Development {
-			cfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
-		}
+		return nil, fmt.Errorf("missing EncodeTime in EncoderConfig")
 	}
 
 	return newEncoder(cfg.Encoding, cfg.EncoderConfig)
--- a/config.go
+++ b/config.go
@@ -175,7 +175,7 @@ func (cfg Config) Build(opts ...Option) (*Logger, error) {
 		return nil, err
 	}
 
-	if cfg.Level == AtomicLevel{} {
+	if cfg.Level == (AtomicLevel{}) {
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -174,6 +174,13 @@ func (cfg Config) Build(opts ...Option) (*Logger, error) {
 		return nil, err
 	}
 
+	if cfg.Level == (AtomicLevel{}) {
```
abhinav: The parentheses are unnecessary.

```suggestion
	if cfg.Level == AtomicLevel{} {
```

config.go
```diff
@@ -174,6 +174,13 @@ func (cfg Config) Build(opts ...Option) (*Logger, error) {
...
+		cfg.Level = NewAtomicLevelAt(InfoLevel)
+		if cfg.Development {
+			cfg.Level = NewAtomicLevelAt(DebugLevel)
```
abhinav: Should we return an `error` instead?

If we prefer to add a default behavior for Level when unspecified, we'd probably want to document it on Config.Level.

config.go
```diff
@@ -239,5 +246,12 @@ func (cfg Config) openSinks() (zapcore.WriteSyncer, zapcore.WriteSyncer, error)
 }
 
 func (cfg Config) buildEncoder() (zapcore.Encoder, error) {
+	if cfg.EncoderConfig.TimeKey != "" && cfg.EncoderConfig.EncodeTime == nil {
+		cfg.EncoderConfig.EncodeTime = zapcore.EpochTimeEncoder
```
abhinav: Same here. It's not obvious why we pick epoch for production and ISO8601 for development. Perhaps an error would be preferable?

prashantv:
Agree with @abhinav -- I think `Development` changing time behaviour might be surprising, since the current docs for `Development` is documented only to put the logger in development mode.

```
// Development puts the logger in development mode, which changes the
// behavior of DPanicLevel and takes stacktraces more liberally.
```

To avoid any surprises, returning an error seems best. We should also add unit tests to validate the new behaviour. Thanks!

config.go
YashishDua: I could see InfoLevel as default in ProductionConfig and DebugLevel in DevelopmentConfig, that's why thought of adding the same defaults here.

Let's return an error then!

config.go
YashishDua: NewDevelopmentEncoderConfig specifies ISO8601TimeEncoder by default. I have not made any explicit choices here. But, let's return an error.

config.go
YashishDua: Tried this earlier, but comes with `expected ';', found '{'`

config_test.go
```diff
@@ -106,3 +107,39 @@ func TestConfigWithInvalidPaths(t *testing.T) {
...
+			},
+			expectErr: "missing EncodeTime in EncoderConfig",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			cfg := tt.cfg
+			_, err := cfg.Build()
+			assert.Equal(t, err.Error(), tt.expectErr)
```
abhinav: Testify assertions are in the order: expected, then actual.

```suggestion
			assert.Equal(t, tt.expectErr, err.Error())
```

config_test.go
```diff
@@ -106,3 +107,39 @@ func TestConfigWithInvalidPaths(t *testing.T) {
...
+				},
+			},
+			expectErr: "missing EncodeTime in EncoderConfig",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			cfg := tt.cfg
+			_, err := cfg.Build()
```
abhinav: err.Error on the following line will panic if we don't first verify that the error is non-nil.
We should use require.Error here.

```suggestion
			_, err := cfg.Build()
			require.Error(t, err)
```

config_test.go
```diff
@@ -106,3 +107,39 @@ func TestConfigWithInvalidPaths(t *testing.T) {
...
 }
+
+func TestConfigWithMissingAttributes(t *testing.T) {
+	tests := []struct {
+		desc      string
+		cfg       Config
+		expectErr string
+	}{
+		{
+			desc: "missing_level",
```
abhinav: We don't need to add `_` to subtest case names here. testing.T will do it for us.

```suggestion
			desc: "missing level",
```

config_test.go
```diff
@@ -106,3 +107,39 @@ func TestConfigWithInvalidPaths(t *testing.T) {
...
+			cfg: Config{
+				Encoding: "json",
+			},
+			expectErr: "missing Level",
+		},
+		{
+			desc: "missing_encoder_time_in_encoder_config",
```
abhinav:
```suggestion
			desc: "missing encoder time in encoder config",
```

config.go
```diff
@@ -239,5 +244,9 @@ func (cfg Config) openSinks() (zapcore.WriteSyncer, zapcore.WriteSyncer, error)
 }
 
 func (cfg Config) buildEncoder() (zapcore.Encoder, error) {
+	if cfg.EncoderConfig.TimeKey != "" && cfg.EncoderConfig.EncodeTime == nil {
```
abhinav: It seems that newEncoder is the ultimate constructor for zapcore.Encoder given
an EncoderConfig. We should maybe move this check to the newEncoder function,
leaving buildEncoder unchanged.

Issue #803, opened 2020-03-21
Incorrect time value with zap.Time() when using a time far in the future


gkotian:
I'm trying to log a time value that is set to something very far in the future. It is supposedly the largest time value possible according to https://stackoverflow.com/a/25065327/793930.

I'm using the plain `Logger`, and creating the field for the log message using `zap.Time()`.
https://play.golang.org/p/Nc8jv_8NkCs

If you run the code in the playground, you'll see that `t1` is the same in both the zap log and in the `fmt.Println` print. However, `t2` is reported incorrectly in the zap log as:
```
1754-08-30T22:43:41.128Z
```

prashantv:
This is similar to #737

I think we should do the same fix as suggested there: if a value is out-of-range for `UnixNano`, then store a different representation internally.

abhinav:
Resolved in #804

PR #804, opened 2020-03-24
Fix handling of Time values out of UnixNano range

### LINKED ISSUE #737: inconsistent display of zap.Time field with zero time struct
When logging a zero initialized time.Time struct the output is not consistent:

```go
zap.L().With("time", time.Time{}).Debug("zero time")
```
Will output the time: `1754-08-30T22:43:41.128654848Z` because calling `UnixNano()` then `Unix(0, nsec)` is not symetric in this case.
The correct output would be `0001-01-01 00:00:00 +0000 UTC` (with the same layout as above).

see https://gist.github.com/patdowney/b64a4fee790ac66dd1a2 that exhibit the issue.

Looking at zap.Time I was wondering why you don't pass the Time struct in the Interface field instead of converting to an integer. Maybe this is to avoid making a new reference to traverse for the GC ?

https://github.com/uber-go/zap/blob/425214515ff452748375576b20c82524849177c6/field.go#L180-L184

https://github.com/uber-go/zap/blob/badef736563f8135e0add3232b06f58656da03c8/zapcore/field.go#L141-L147

prashantv:
Fixes #737, #803.

Time values are encoded by storing the UnixNano representation, but this
fails when the value of UnixNano is out of int64 range, see docs:

https://golang.org/pkg/time/#Time.UnixNano
> The result is undefined if the Unix time in nanoseconds cannot be
represented by an int64 (a date before the year 1678 or after 2262)

Fix this by storing values outside of UnixNano range as-is rather
than using UnixNano with a new Field type.

### DIFF
--- a/field.go
+++ b/field.go
@@ -33,7 +33,7 @@ import (
 type Field = zapcore.Field
 
 var (
-	_minTimeInt64 = time.Unix(0, 0)
+	_minTimeInt64 = time.Unix(0, math.MinInt64)
 	_maxTimeInt64 = time.Unix(0, math.MaxInt64)
 )
 
--- a/field.go
+++ b/field.go
@@ -32,6 +32,11 @@ import (
 // improves the navigability of this package's API documentation.
 type Field = zapcore.Field
 
+var (
+	_minTimeInt64 = time.Unix(0, 0)
+	_maxTimeInt64 = time.Unix(0, math.MaxInt64)
+)
+
 // Skip constructs a no-op field, which is often useful when handling invalid
 // inputs in other Field constructors.
 func Skip() Field {
@@ -339,6 +344,9 @@ func Stringer(key string, val fmt.Stringer) Field {
 // Time constructs a Field with the given key and value. The encoder
 // controls how the time is serialized.
 func Time(key string, val time.Time) Field {
+	if val.Before(_minTimeInt64) || val.After(_maxTimeInt64) {
+		return Field{Key: key, Type: zapcore.TimeFullType, Interface: val}
+	}
 	return Field{Key: key, Type: zapcore.TimeType, Integer: val.UnixNano(), Interface: val.Location()}
 }
 
--- a/zapcore/field.go
+++ b/zapcore/field.go
@@ -65,8 +65,11 @@ const (
 	Int8Type
 	// StringType indicates that the field carries a string.
 	StringType
-	// TimeType indicates that the field carries a time.Time.
+	// TimeType indicates that the field carries a time.Time that is
+	// representable by a UnixNano() stored as an int64.
 	TimeType
+	// TimeFullType indicates that the field carries a time.Time stored as-is.
+	TimeFullType
 	// Uint64Type indicates that the field carries a uint64.
 	Uint64Type
 	// Uint32Type indicates that the field carries a uint32.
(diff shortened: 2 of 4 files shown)


zapcore/field.go
```diff
@@ -65,8 +65,11 @@ const (
 	Int8Type
 	// StringType indicates that the field carries a string.
 	StringType
-	// TimeType indicates that the field carries a time.Time.
+	// TimeType indicates that the field carries a time.Time that is
+	// representable by a UnixNano() stored as an int64.
 	TimeType
+	// TimeFullType indicates that the field carries a time.Time stored as-is.
+	TimeFullType
```
abhinav: Are we sure introducing a new type is okay?
Would it make sense instead to store an invalid value like -1 in the integer field to indicate that the Interface field is set?

prashantv: Since this is defined like an enum, I think it's reasonable to create new values here, but it will break anyone who introspects the internals of a field and only handles the known values.

I don't think we should use the integer field as:
 - The full value range can be used and is valid for existing values. Negative values just mean a time before the epoch. We'd need to truncate the range, but that would cause the following issue.
 - the current `Interface` is storing the timezone, while the new implementation stores a `time.Time` so code that assumes that a timezone is stored for `TimeType` will fail badly (likely panic)

Both options could break callers, but this approach is cleaner, and will only break those who did not assume that the `Type` enum may have new values -- which callers should expect. See similar change in the Go standard library that extends an enum to include a new value:
https://github.com/golang/go/commit/e64241216dd141589144a07f7f68acd64dc108fe

abhinav: Makes sense.

field.go
```diff
@@ -32,6 +32,11 @@ import (
 // improves the navigability of this package's API documentation.
 type Field = zapcore.Field
 
+var (
+	_minTimeInt64 = time.Unix(0, 0)
```
abhinav:
```suggestion
	_minTimeInt64 = time.Unix(0, math.MinInt64)
```

zapcore/field_test.go
```diff
@@ -164,6 +165,10 @@ func TestFields(t *testing.T) {
 }
 
 func TestEquals(t *testing.T) {
+	timeOutOfRange := time.Unix(0, math.MaxInt64).Add(time.Nanosecond)
+	timeOutOfRangeNano := time.Unix(0, timeOutOfRange.UnixNano())
```
abhinav: We should try out of range in both directions.

prashantv: :+1: done

zapcore/field_test.go
```diff
@@ -164,6 +165,14 @@ func TestFields(t *testing.T) {
 }
 
 func TestEquals(t *testing.T) {
+	// Values outside the UnixNano range were encoded incorrectly (#737, #803).
+	timeOutOfRangeHigh := time.Unix(0, math.MaxInt64).Add(time.Nanosecond)
+	timeOutOfRangeLow := time.Unix(0, math.MinInt64).Add(-time.Nanosecond)
+	timeOutOfRangeHighNano := time.Unix(0, timeOutOfRangeHigh.UnixNano())
+	timeOutOfRangeLowNano := time.Unix(0, timeOutOfRangeLow.UnixNano())
+	require.NotEqual(t, timeOutOfRangeHigh, timeOutOfRangeHighNano, "should be different as value is >  UnixNano range")
+	require.NotEqual(t, timeOutOfRangeHigh, timeOutOfRangeHighNano, "should be different as value is <  UnixNano range")
```
abhinav: Should we use `Time.Equal` here? 
```suggestion
	require.False(t, timeOutOfRangeHigh.Equal(timeOutOfRangeHighNano), "should be different as value is >  UnixNano range")
	require.False(t, timeOutOfRangeHigh.Equal(timeOutOfRangeHighNano), "should be different as value is <  UnixNano range")
```

PR #806, opened 2020-03-26
Add WithCaller(), a generalized AddCaller()

### LINKED ISSUE #702: Would like negative version of AddCaller
If I have a logger and I need to derive a child logger which has the caller disabled, there doesn't seem to be a way to do it.

I had hoped to do something like:
```
child := parent.WithOptions(zap.DisableCaller())
```

danielbprice:
Fixes #702.

### DIFF
--- a/logger_test.go
+++ b/logger_test.go
@@ -339,7 +339,12 @@ func TestLoggerAddCaller(t *testing.T) {
 		options []Option
 		pat     string
 	}{
+		{opts(), `^undefined$`},
+		{opts(WithCaller(false)), `^undefined$`},
 		{opts(AddCaller()), `.+/logger_test.go:[\d]+$`},
+		{opts(AddCaller(), WithCaller(false)), `^undefined$`},
+		{opts(WithCaller(true)), `.+/logger_test.go:[\d]+$`},
+		{opts(WithCaller(true), WithCaller(false)), `^undefined$`},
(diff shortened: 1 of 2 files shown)


danielbprice:
FWiW I am not attached to the name SetCaller(); just give me feedback and I'll tweak it ASAP.  This issue has cropped up for me several times, so i am eager to fix it.

logger_test.go
```diff
@@ -339,7 +339,11 @@ func TestLoggerAddCaller(t *testing.T) {
 		options []Option
 		pat     string
 	}{
+		{opts(), `^undefined$`},
+		{opts(SetCaller(false)), `^undefined$`},
 		{opts(AddCaller()), `.+/logger_test.go:[\d]+$`},
+		{opts(SetCaller(true)), `.+/logger_test.go:[\d]+$`},
+		{opts(SetCaller(true), SetCaller(false)), `^undefined$`},
```
prashantv, line 346: can you do `AddCaller` followed by `SetCaller(false)` as well

abhinav:
How about `WithCaller(bool)`? That suggests inclusion of the caller with the other fields versus "SetCaller" which suggests you're setting the caller.

danielbprice:
Cool!  AFAICT there is nothing left for me to do.  But if I've missed something please let me know.

PR #808, opened 2020-03-27
Share callstack b/w Caller and Stack

abhinav:
A log entry that includes both, the caller and the stack trace requests
stack information twice: once to determine the caller and once to build
the stack trace.

This change optimizes this by requesting call stack information once and
sharing it for both cases.

---

I ran the benchmarks with `-count 3`.

**Before**

```
BenchmarkAddCallerHook-4         1262595               886 ns/op             248 B/op          3 allocs/op
BenchmarkAddCallerHook-4         1344021               889 ns/op             248 B/op          3 allocs/op
BenchmarkAddCallerHook-4         1000000              1085 ns/op             248 B/op          3 allocs/op
BenchmarkAddStack-4               483288              2160 ns/op             304 B/op          2 allocs/op
BenchmarkAddStack-4               550969              1993 ns/op             304 B/op          2 allocs/op
BenchmarkAddStack-4               614550              2000 ns/op             304 B/op          2 allocs/op
BenchmarkAddCallerAndStack-4      397208              2554 ns/op             552 B/op          5 allocs/op
BenchmarkAddCallerAndStack-4      462566              2578 ns/op             552 B/op          5 allocs/op
BenchmarkAddCallerAndStack-4      459589              2980 ns/op             552 B/op          5 allocs/op
```

**After**

```
BenchmarkAddCallerHook-4         1138980               960 ns/op             240 B/op          2 allocs/op
BenchmarkAddCallerHook-4         1096950               965 ns/op             240 B/op          2 allocs/op
BenchmarkAddCallerHook-4         1242375               978 ns/op             240 B/op          2 allocs/op
BenchmarkAddStack-4               868773              1388 ns/op             304 B/op          2 allocs/op
BenchmarkAddStack-4               857756              1438 ns/op             304 B/op          2 allocs/op
BenchmarkAddStack-4               865842              1348 ns/op             304 B/op          2 allocs/op
BenchmarkAddCallerAndStack-4      701884              1715 ns/op             544 B/op          4 allocs/op
BenchmarkAddCallerAndStack-4      698002              1709 ns/op             544 B/op          4 allocs/op
BenchmarkAddCallerAndStack-4      608336              1712 ns/op             544 B/op          4 allocs/op
```

On average,

1. Adding caller alone is ~14 nanoseconds (~1.5%) *slower*
2. Adding stack alone is ~659 nanoseconds (~32%) *faster*
3. Adding both, caller and stack is ~992 nanoseconds (~36%) *faster*

The second of these rightfully raises eyebrows because it was expected
to have no changes at best. A cursory investigation doesn't reveal
any obvious reason for the speed up unless there's something incredibly
wrong with the change.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -296,15 +296,33 @@ func (log *Logger) check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
 
 	// Thread the error output through to the CheckedEntry.
 	ce.ErrorOutput = log.errorOutput
+
+	addStack := log.addStack.Enabled(ce.Entry.Level)
+	if !log.addCaller && !addStack {
+		return ce
+	}
+
+	programCounters := newProgramCounters()
+	defer programCounters.Release()
+
+	pcs := programCounters.Callers(callerSkipOffset)
 	if log.addCaller {
-		ce.Entry.Caller = zapcore.NewEntryCaller(runtime.Caller(log.callerSkip + callerSkipOffset))
+		callerIdx := log.callerSkip
+		if len(pcs) <= callerIdx {
+			ce.Entry.Caller = zapcore.NewEntryCaller(0, "", 0, false)
+		} else {
+			frame, _ := runtime.CallersFrames(pcs[callerIdx:]).Next()
+			ce.Entry.Caller = zapcore.NewEntryCaller(frame.PC, frame.File, frame.Line, true)
+		}
+
 		if !ce.Entry.Caller.Defined {
 			fmt.Fprintf(log.errorOutput, "%v Logger.check error: failed to get caller\n", time.Now().UTC())
 			log.errorOutput.Sync()
 		}
 	}
-	if log.addStack.Enabled(ce.Entry.Level) {
-		ce.Entry.Stack = Stack("").String
+
+	if addStack {
+		ce.Entry.Stack = makeStacktrace(pcs)
 	}
 
 	return ce
--- a/logger.go
+++ b/logger.go
@@ -302,16 +302,27 @@ func (log *Logger) check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
 		return ce
 	}
 
+	// If a stack trace was requested, capture the entire stack.
+	stackLimit := 0
+
+	// For historical reasons, stack traces reported by Zap don't respect
+	// CallerSkip. See also #727.
+	stackSkip := callerSkipOffset
+
+	if !addStack {
+		stackLimit = 1
+		stackSkip += log.callerSkip
+	}
+
 	programCounters := newProgramCounters()
 	defer programCounters.Release()
 
-	pcs := programCounters.Callers(callerSkipOffset)
+	pcs := programCounters.Callers(stackSkip, stackLimit)
 	if log.addCaller {
-		callerIdx := log.callerSkip
-		if len(pcs) <= callerIdx {
+		if len(pcs) == 0 {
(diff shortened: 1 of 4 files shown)


stacktrace.go
```diff
@@ -113,8 +102,55 @@ type programCounters struct {
...
+			pc = make([]uintptr, limit)
+			pcs.pcs = pc
+		}
+	}
+
+	var numFrames int
+	for {
+		// Skip the call to runtime.Callers and programCounter.Callers
+		// so that the program counters start at our caller.
+		numFrames = runtime.Callers(skip+2, pc)
```
prashantv, line 140: can do `numFrames := ` since we don't use the value outside of the for loop and always replace it

stacktrace_test.go
```diff
@@ -73,3 +73,56 @@ func BenchmarkTakeStacktrace(b *testing.B) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			const withStackDepthName = "go.uber.org/zap.withStackDepth"
+
+			var trace string
+			withStackDepth(tt.stackDepth, func() {
+				pcs := newProgramCounters().Callers(1, tt.limit)
+				trace = makeStacktrace(pcs, false /* don't skip zap */)
```
prashantv, line 106: trying to avoid this boolean arg (as it's a test-only code path, and generally boolean args are not great)
presumably you need a fair bit of stack depth,, is it possible to use something outside of zap that takes in a callback, e.g.,
https://golang.org/pkg/bytes/#Map
hacky PoC: https://play.golang.org/p/bKrUH8kEreE

And keep calling that function recursively till we hit the limit we want, and that should add non-zap stack frames?

logger.go
```diff
@@ -302,16 +302,27 @@ func (log *Logger) check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
...
-	pcs := programCounters.Callers(callerSkipOffset)
+	pcs := programCounters.Callers(stackSkip, stackLimit)
 	if log.addCaller {
-		callerIdx := log.callerSkip
-		if len(pcs) <= callerIdx {
+		if len(pcs) == 0 {
 			ce.Entry.Caller = zapcore.NewEntryCaller(0, "", 0, false)
 		} else {
-			frame, _ := runtime.CallersFrames(pcs[callerIdx:]).Next()
+			frame, _ := runtime.CallersFrames(pcs).Next()
```
prashantv, line 325: if we are adding caller and stack, then the `stackSkip` passed in is 0 (e.g., no caller skip), so the caller value used here would be incorrect right, since we wouldn't be respecting the `log.callerSkip`

stacktrace.go
```diff
@@ -113,19 +113,40 @@ func (pcs *programCounters) Release() {
...
+// limit specifies how deep into the stack we go, zero meaning unlimited.
+func (pcs *programCounters) Callers(skip int, limit int) []uintptr {
+	pc := pcs.pcs
+	if limit > 0 {
+		// The shortened slice will never be placed back into the
+		// pool, but the longer one will.
+		if limit < len(pc) {
+			pc = pc[:limit]
+		} else {
+			pc = make([]uintptr, limit)
```
prashantv, line 131: this case never gets hit in prod right? anyway we can refactor the change to avoid adding a case that's not necessary in prod?

abhinav:
Closing until I have a chance to look at this again in the future.

PR #813, opened 2020-04-04
core/sampler: Support decision hook

shirchen:
This adds support for monitoring sampling decisions made by the Sampler
core with a `func(Entry, SamplingDecision)` where `SamplingDecision` is
a bit field.

To allow plumbing the hook to the sampler, this additionally deprecates
the `NewSampler` constructor in favor of `NewSamplerWithOptions`.

    type SamplerOption

    func SamplerHook(func(Entry, SamplingDecision))

    func NewSamplerWithOptions(/* ... */, opts ...SamplerOption)

This functionality is usable from the `zap` package via the new `Hook`
field of `zap.SamplingConfig`.

Refs T5056227

### DIFF
--- a/benchmarks/zap_test.go
+++ b/benchmarks/zap_test.go
@@ -116,7 +116,7 @@ func newZapLogger(lvl zapcore.Level) *zap.Logger {
 }
 
 func newSampledLogger(lvl zapcore.Level) *zap.Logger {
-	return zap.New(zapcore.NewSampler(
+	return zap.New(zapcore.NewSamplerWithOptions(
 		newZapLogger(zap.DebugLevel).Core(),
 		100*time.Millisecond,
 		10, // first
--- a/benchmarks/zap_test.go
+++ b/benchmarks/zap_test.go
@@ -116,7 +116,7 @@ func newZapLogger(lvl zapcore.Level) *zap.Logger {
 }
 
 func newSampledLogger(lvl zapcore.Level) *zap.Logger {
-	return zap.New(zapcore.NewSamplerWithOptions(
+	return zap.New(zapcore.NewSampler(
 		newZapLogger(zap.DebugLevel).Core(),
 		100*time.Millisecond,
 		10, // first
--- a/config.go
+++ b/config.go
@@ -32,15 +32,14 @@ import (
 // global CPU and I/O load that logging puts on your process while attempting
 // to preserve a representative subset of your logs.
 //
-// Hook is called whenever a Sampler makes a decision. Currently, whenever a
-// log is dropped.
+// If specified, the Sampler will invoke the Hook after each decision.
 //
 // Values configured here are per-second. See zapcore.NewSamplerWithOptions for
 // details.
 type SamplingConfig struct {
 	Initial    int `json:"initial" yaml:"initial"`
 	Thereafter int `json:"thereafter" yaml:"thereafter"`
-	Hook       func(zapcore.Entry, zapcore.SamplingDecision)
+	Hook       func(zapcore.Entry, zapcore.SamplingDecision) `json:"-" yaml:"-"`
 }
 
 // Config offers a declarative way to construct a logger. It doesn't do
--- a/config.go
+++ b/config.go
@@ -32,7 +32,8 @@ import (
 // global CPU and I/O load that logging puts on your process while attempting
 // to preserve a representative subset of your logs.
 //
-//
+// Hook is called whenever a Sampler makes a decision. Currently, whenever a
+// log is dropped.
 //
 // Values configured here are per-second. See zapcore.NewSamplerWithOptions for
 // details.
--- a/config.go
+++ b/config.go
@@ -37,8 +37,8 @@ import (
 // Values configured here are per-second. See zapcore.NewSamplerWithOptions for
 // details.
 type SamplingConfig struct {
-	Initial    int `json:"initial" yaml:"initial"`
-	Thereafter int `json:"thereafter" yaml:"thereafter"`
+	Initial    int                                           `json:"initial" yaml:"initial"`
+	Thereafter int                                           `json:"thereafter" yaml:"thereafter"`
 	Hook       func(zapcore.Entry, zapcore.SamplingDecision) `json:"-" yaml:"-"`
 }
 
--- a/config.go
+++ b/config.go
@@ -40,7 +40,7 @@ import (
 type SamplingConfig struct {
 	Initial    int `json:"initial" yaml:"initial"`
 	Thereafter int `json:"thereafter" yaml:"thereafter"`
-	Hook       func(zapcore.Entry, zapcore.SamplingDecision) error
+	Hook       func(zapcore.Entry, zapcore.SamplingDecision)
 }
 
 // Config offers a declarative way to construct a logger. It doesn't do
@@ -212,19 +212,19 @@ func (cfg Config) buildOptions(errSink zapcore.WriteSyncer) []Option {
 	if !cfg.DisableStacktrace {
 		opts = append(opts, AddStacktrace(stackLevel))
 	}
-	if cfg.Sampling != nil && cfg.Sampling.Hook == nil {
-		// Assign a default nop sampling hook.
-		cfg.Sampling.Hook = zapcore.NopSamplingHook
-	}
 
-	if cfg.Sampling != nil {
+	if scfg := cfg.Sampling; scfg != nil {
 		opts = append(opts, WrapCore(func(core zapcore.Core) zapcore.Core {
+			var samplerOpts []zapcore.SamplerOption
+			if scfg.Hook != nil {
+				samplerOpts = append(samplerOpts, zapcore.SamplerHook(scfg.Hook))
+			}
 			return zapcore.NewSamplerWithOptions(
 				core,
 				time.Second,
 				cfg.Sampling.Initial,
 				cfg.Sampling.Thereafter,
-				zapcore.SamplerHook(cfg.Sampling.Hook),
+				samplerOpts...,
 			)
 		}))
 	}
--- a/config.go
+++ b/config.go
@@ -36,6 +36,7 @@ import (
 type SamplingConfig struct {
 	Initial    int `json:"initial" yaml:"initial"`
 	Thereafter int `json:"thereafter" yaml:"thereafter"`
+	Hook       func(zapcore.Entry, zapcore.SamplingDecision)
 }
 
 // Config offers a declarative way to construct a logger. It doesn't do
@@ -209,9 +210,23 @@ func (cfg Config) buildOptions(errSink zapcore.WriteSyncer) []Option {
 	}
 
 	if cfg.Sampling != nil {
-		opts = append(opts, WrapCore(func(core zapcore.Core) zapcore.Core {
-			return zapcore.NewSampler(core, time.Second, int(cfg.Sampling.Initial), int(cfg.Sampling.Thereafter))
-		}))
+		if cfg.Sampling.Hook != nil {
+			opts = append(opts, WrapCore(func(core zapcore.Core) zapcore.Core {
+				return zapcore.NewSamplerWithOptions(
+					core,
+					time.Second,
+					cfg.Sampling.Initial,
+					cfg.Sampling.Thereafter,
+					zapcore.SamplerHook(cfg.Sampling.Hook),
+				)
+			}))
+		} else {
+			opts = append(opts, WrapCore(func(core zapcore.Core) zapcore.Core {
+				return zapcore.NewSampler(
+					core, time.Second, int(cfg.Sampling.Initial),
+					int(cfg.Sampling.Thereafter))
+			}))
+		}
 	}
 
 	if len(cfg.InitialFields) > 0 {
--- a/config.go
+++ b/config.go
@@ -32,11 +32,14 @@ import (
 // global CPU and I/O load that logging puts on your process while attempting
 // to preserve a representative subset of your logs.
 //
-// Values configured here are per-second. See zapcore.NewSampler for details.
+//
+//
+// Values configured here are per-second. See zapcore.NewSamplerWithOptions for
+// details.
 type SamplingConfig struct {
 	Initial    int `json:"initial" yaml:"initial"`
 	Thereafter int `json:"thereafter" yaml:"thereafter"`
-	Hook       func(zapcore.Entry, zapcore.SamplingDecision)
+	Hook       func(zapcore.Entry, zapcore.SamplingDecision) error
 }
 
 // Config offers a declarative way to construct a logger. It doesn't do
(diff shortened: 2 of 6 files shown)


benchmarks/zap_test.go
```diff
@@ -116,7 +116,7 @@ func newZapLogger(lvl zapcore.Level) *zap.Logger {
 }
 
 func newSampledLogger(lvl zapcore.Level) *zap.Logger {
-	return zap.New(zapcore.NewSampler(
```
shirchen, line 119: I'm gonna revert this to add back cc for NewSampler.

config_test.go
```diff
@@ -144,3 +145,72 @@ func TestConfigWithMissingAttributes(t *testing.T) {
 		})
 	}
 }
+
+func makeSamplerCountingHook() (func(zapcore.Entry, zapcore.SamplingDecision),
+	*atomic.Int64, *atomic.Int64) {
+	droppedCount := &atomic.Int64{}
+	sampledCount := &atomic.Int64{}
```
abhinav, line 152: nit: use `new` for pointers to zero value.
```suggestion
	droppedCount := new(atomic.Int64)
	sampledCount := new(atomic.Int64)
```

The other convention is `var` and `&` before returning:

```suggestion
	var dropped, sampled atomic.Int64
```

```
  return h, &dropped, &sampled
```

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
 	return 1
 }
 
+// SamplingDecision represents a decision made by sampler.
+type SamplingDecision uint8
```
abhinav, line 85: I'm guessing this is so that it takes one byte instead of 8?

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+	f(s)
+}
+
+// SamplerOption configures a Sampler option.
+type SamplerOption interface {
+	apply(*sampler)
+}
+
+// NopSamplingHook is the default hook used by sampler.
+func NopSamplingHook(_ Entry, _ SamplingDecision) {}
```
abhinav, line 107: can drop `_` when they're all `_`
```suggestion
func NopSamplingHook(Entry, SamplingDecision) {}
```

zapcore/sampler.go
```diff
@@ -100,13 +170,16 @@ type sampler struct {
...
+// Deprecated: use NewSamplerWithOptions.
 func NewSampler(core Core, tick time.Duration, first, thereafter int) Core {
 	return &sampler{
 		Core:       core,
 		tick:       tick,
 		counts:     newCounters(),
 		first:      uint64(first),
 		thereafter: uint64(thereafter),
+		hook:       NopSamplingHook,
 	}
```
abhinav, line 183: We should make it so that the old implementation calls the new one so that if we make changes to `sample` initialization, they don't have to go in two places.

```suggestion
	return NewSamplerWithOptions(core, tick, first, thereafter)
```

zapcore/sampler_bench_test.go
```diff
@@ -228,3 +229,50 @@ func BenchmarkSampler_Check(b *testing.B) {
...
+		if dec == LogDropped {
+			droppedCount.Inc()
+		} else if dec == LogSampled {
+			sampledCount.Inc()
+		}
+	}
+	return h, droppedCount, sampledCount
+}
+
+func BenchmarkSampler_CheckWithHook(b *testing.B) {
```
abhinav, line 246: optional: count dropped messages only to more accurately measure the use case we're trying to satisfy here.

config.go
```diff
@@ -32,10 +32,15 @@ import (
...
-// Values configured here are per-second. See zapcore.NewSampler for details.
+// Hook is called whenever a Sampler makes a decision. Currently, whenever a
+// log is dropped.
+//
+// Values configured here are per-second. See zapcore.NewSamplerWithOptions for
+// details.
 type SamplingConfig struct {
 	Initial    int `json:"initial" yaml:"initial"`
 	Thereafter int `json:"thereafter" yaml:"thereafter"`
+	Hook       func(zapcore.Entry, zapcore.SamplingDecision)
```
abhinav, line 43: Should add `json:"-" yaml:"-"` so the JSON/YAML moduels do not attempt to serialize this.
```suggestion
	Hook       func(zapcore.Entry, zapcore.SamplingDecision) `json:"-" yaml:"-"`
```

config.go
```diff
@@ -32,10 +32,15 @@ import (
 // global CPU and I/O load that logging puts on your process while attempting
 // to preserve a representative subset of your logs.
 //
-// Values configured here are per-second. See zapcore.NewSampler for details.
+// Hook is called whenever a Sampler makes a decision. Currently, whenever a
+// log is dropped.
```
abhinav, line 36: Second part no longer true. How about,

```suggestion
// If specified, the Sampler will invoke the Hook after each decision.
```

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
 	return 1
 }
 
+// SamplingDecision represents a decision made by sampler.
```
abhinav, line 84: nit: all types represent something. can do:

```suggestion
// SamplingDecision is a decision made by sampler.
```

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+type SamplingDecision uint8
+
+const (
+	// LogDropped means that a log was dropped.
```
abhinav, line 88:
```suggestion
	// LogDropped indicates that the Sampler dropped a log entry.
```

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+	LogDropped SamplingDecision = iota
+	// LogSampled means that a log was successfully sampled.
```
abhinav, line 90:
```suggestion
	// LogSampled indicates that the Sampler sampled a log entry.
```

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+)
+
+// optionFunc wraps a func so it satisfies the SamplerOption interface.
+type optionFunc func(*sampler)
+
+func (f optionFunc) apply(s *sampler) {
+	f(s)
+}
+
+// SamplerOption configures a Sampler option.
```
abhinav, line 101:
```suggestion
// SamplerOption configures a Sampler.
```

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+type SamplerOption interface {
+	apply(*sampler)
+}
+
+// NopSamplingHook is the default hook used by sampler.
```
abhinav, line 106: Do we actually need to export this? We'll use this if unset.

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+func NopSamplingHook(_ Entry, _ SamplingDecision) {}
+
+// SamplerHook registers a function  which will be called when Sampler makes a
+// decision. Currently a hook is called when a log is dropped and
+// zapcore.LogDropped decision is emitted.
```
abhinav, line 111: Second part no longer true.
```suggestion
// decision.
```

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+//
+// This hook is useful for side effects, for example emitting number of dropped
+// logs. Note, there is no access to Fields in this hook. In the future, this
+// hook can be expanded to emit whether this is first entry that was dropped,
+// first after a period, etc.
```
abhinav, line 116: I would avoid mentioning future plans in documentation. That's a comment-level thing, possibly referencing tickets.

```suggestion
// This hook may be used to get visibility into the performance of the sampler.
// For example, use it to track metrics of dropped versus sampled logs.
```

zapcore/sampler.go
```diff
@@ -81,17 +81,87 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+// NewSamplerWithOptions creates a Core that samples incoming entries, which
+// caps the CPU and I/O load of logging while attempting to preserve a
+// representative subset of your logs.
+//
+// Zap samples by logging the first N entries with a given level and message
+// each tick. If more Entries with the same level and message are seen during
+// the same interval, every Mth message is logged and the rest are dropped.
+//
+// Sampler also accepts an optional hook that can be used to count number of
+// dropped logs.
```
abhinav, line 132:
```suggestion
// Sampler can be configured to report sampling decisions with the SamplerHook
// option.
```

zapcore/sampler.go
shirchen: Maybe if someone wants to test this as NewSampler* constructors are public?

zapcore/sampler.go
shirchen: yup, as byte is an alias to uint8.

zapcore/sampler.go
shirchen: Good call.

zapcore/sampler.go
abhinav: NewSampler can be public but I mean the implementation of a no-op hook. Users don't really need access to that; they can write their own empty function, or just not provide the SamplingHook option.

zapcore/sampler.go
```diff
@@ -81,33 +81,96 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
 //
+// Sampler can be configured to report sampling decisions with the SamplerHook
+// option.
+//
 // Keep in mind that zap's sampling implementation is optimized for speed over
 // absolute precision; under load, each tick may be slightly over- or
 // under-sampled.
-func NewSampler(core Core, tick time.Duration, first, thereafter int) Core {
-	return &sampler{
+func NewSamplerWithOptions(core Core, tick time.Duration, first, thereafter int, opts ...SamplerOption) Core {
```
prashantv, line 134: If we want to reduce the added API surface (E.g., new function that looks similar + new types), we could add a `Build` method on the `SamplingConfig` struct, similar to the `Build` method on the `Configuration` struct:
https://pkg.go.dev/go.uber.org/zap?tab=doc#Config.Build

config_test.go
```diff
@@ -144,3 +145,72 @@ func TestConfigWithMissingAttributes(t *testing.T) {
 		})
 	}
 }
+
+func makeSamplerCountingHook() (func(zapcore.Entry, zapcore.SamplingDecision),
+	*atomic.Int64, *atomic.Int64) {
```
prashantv, line 150: can you name the returns? we also can do, `dropped, sampled *atomic.Int64` to avoid duplicating the type.

And a super nit: The way this declaration is broken up over multiple lines looks a little odd. We don't tend to force line length < 80, can we put the whole declaration on one line?

config_test.go
```diff
@@ -144,3 +145,72 @@ func TestConfigWithMissingAttributes(t *testing.T) {
...
+			Initial:    100,
+			Thereafter: 100,
+			Hook:       shook,
+		},
+		Encoding:         "json",
+		EncoderConfig:    NewProductionEncoderConfig(),
+		OutputPaths:      []string{"stderr"},
+		ErrorOutputPaths: []string{"stderr"},
+	}
+	expectN := 2 + 100 + 1 // 2 from initial logs, 100 initial sampled logs, 1 from off-by-one in sampler
```
prashantv, line 178: can we fix the off-by-one, seems a little odd, haven't looked into where it's coming from

zapcore/sampler.go
```diff
@@ -81,33 +81,96 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+type SamplingDecision uint8
 
-	counts            *counters
-	tick              time.Duration
-	first, thereafter uint64
+const (
+	// LogDropped indicates that the Sampler dropped a log entry.
+	LogDropped SamplingDecision = iota
+	// LogSampled indicates that the Sampler sampled a log entry.
+	LogSampled
```
prashantv, line 91: something to discuss
- it would be useful to know when a log was first dropped, and then first log after the sample again. e.g., if you want to add a special log that said "message X is now being dropped, no further logs"?
- doing this in the hook requires doing the same thing the sampler does, probably more inefficiently (a map by message)

zapcore/sampler.go
prashantv: Agree, not even sure if we need this in the production code, we could keep it in the test file?

abhinav: It's production. The sampler defaults the hook field to NopSamplingHook.

zapcore/sampler.go
abhinav: Wouldn't we still want a means of building the sampler directly?
Config.Build is porcelain for zap.New. Anything you can do with Config.Build,
you can do with zap.New.
IMO SamplingConfig would need to match that: the underlying constructor would
need to provide the same functionality—unless we decide to deprecate direct
constructors and switch to Config structs.

zapcore/sampler.go
abhinav: Just an idea: If we make SamplingDecision a bitfield, we could do that separately and re-use sampler logic

```
LogDropped SamplingDecision = 1 << iota
LogSampled
LogDropInitial
LogDropThereafter
```

config_test.go
shirchen: Yea, I didn't verify the comment and copied from another test. Fixed explanation.

zapcore/sampler.go
prashantv: I see, in that case, can we unexport and keep it here?

zapcore/sampler.go
prashantv: true, I'm not sure how valuable the direct structs have been vs configs, but don't have to make that change here. let's keep this as-is

config_test.go
```diff
@@ -144,3 +145,72 @@ func TestConfigWithMissingAttributes(t *testing.T) {
...
+		Encoding:         "json",
+		EncoderConfig:    NewProductionEncoderConfig(),
+		OutputPaths:      []string{"stderr"},
+		ErrorOutputPaths: []string{"stderr"},
+	}
+	expectN := 2 + 101 // out of 203 logs: 3 types (debug, info, warn), 2 (debug + warn) + 101 (100 initial + 200th thereafter)
+	expectRe := `{"level":"info","caller":"zap/config_test.go:\d+","msg":"info","k":"v","z":"zz"}` + "\n" +
+		`{"level":"warn","caller":"zap/config_test.go:\d+","msg":"warn","k":"v","z":"zz"}` + "\n"
+	expectDropped := 99  // 200 - 100 initial - 1 thereafter
+	expectSampled := 103 // 2 from initial + 100 + 1 thereafter
```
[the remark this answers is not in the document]
shirchen: `expecttN` is not a samplerHook, but a zap hook which only measures what is sent. i regret carrying over a hook from another test now as it seems to cause confusion. :)

zapcore/sampler.go
```diff
@@ -81,33 +81,97 @@ func (c *counter) IncCheckReset(t time.Time, tick time.Duration) uint64 {
...
+
+// nopSamplingHook is the default hook used by sampler.
+func nopSamplingHook(Entry, SamplingDecision) {}
+
+// SamplerHook registers a function  which will be called when Sampler makes a
+// decision.
+//
+// This hook may be used to get visibility into the performance of the sampler.
+// For example, use it to track metrics of dropped versus sampled logs.
+func SamplerHook(hook func(entry Entry, dec SamplingDecision)) SamplerOption {
```
abhinav, line 115: Worth adding a sample of how to use the sampling decision?

```suggestion
// For example, use it to track metrics of dropped versus sampled logs.
//
//  var dropped atomic.Int64
//  zapcore.SamplerHook(func(ent zapcore.Entry, dec zapcore.SamplingDecision) {
//    if dec&zapcore.LogDropped > 0 {
//      dropped.Inc()
//    }
//  })
func SamplerHook(hook func(entry Entry, dec SamplingDecision)) SamplerOption {
```

config_test.go
```diff
@@ -144,3 +145,72 @@ func TestConfigWithMissingAttributes(t *testing.T) {
...
+			Initial:    100,
+			Thereafter: 100,
+			Hook:       shook,
+		},
+		Encoding:         "json",
+		EncoderConfig:    NewProductionEncoderConfig(),
+		OutputPaths:      []string{"stderr"},
+		ErrorOutputPaths: []string{"stderr"},
+	}
+	expectN := 2 + 101 // out of 203 logs: 3 types (debug, info, warn), 2 (debug + warn) + 101 (100 initial + 200th thereafter)
```
abhinav, line 178: I'm confused by these numbers. What exactly is expectN?

shirchen: see https://github.com/uber-go/zap/pull/813#discussion_r409186070

zapcore/sampler_test.go
```diff
@@ -37,6 +37,7 @@ import (
 
 func fakeSampler(lvl LevelEnabler, tick time.Duration, first, thereafter int) (Core, *observer.ObservedLogs) {
 	core, logs := observer.New(lvl)
+	// Keep using deprecated constructor for cc.
```
prashantv, line 40: what is `cc` and why can't this be updated?

shirchen: code coverage

PR #839, opened 2020-06-14
Optimize the process of getting/putting a jsonEncoder from/to _jsonPool

tylitianrui:
Any item stored in the Pool may be removed automatically at any time without notification.  it will  be better ，if  the process of getting/putting a jsonEncoder is modified.

### DIFF
--- a/zapcore/json_encoder.go
+++ b/zapcore/json_encoder.go
@@ -40,10 +40,7 @@ var _jsonPool = sync.Pool{New: func() interface{} {
 }}
 
 func getJSONEncoder() *jsonEncoder {
-	return _jsonPool.Get().(*jsonEncoder)
-}
-
-func putJSONEncoder(enc *jsonEncoder) {
+	enc := _jsonPool.Get().(*jsonEncoder)
 	if enc.reflectBuf != nil {
 		enc.reflectBuf.Free()
 	}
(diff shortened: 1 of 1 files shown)


prashantv:
Thanks for the contribution @tylitianrui 

Would you be able to explain the reasoning a little more, I wasn't able to follow the advantage of setting fields in get vs put.

tylitianrui:
a jsonEncoder  stored in   _jsonPool  may  not  be  used becase of  GC.  if  a  jsonEncoder were  put   into  _jsonPool just   before  GC. IT would be removed  instead of reused.  Resetting  the   jsonEncoder  , when  Putting  into  _jsonPool,   is  unnecessary.   Resetting   jsonEncoder , when getting from _jsonPool,   it is  better.

prashantv:
> a jsonEncoder stored in _jsonPool may not be used becase of GC. if a jsonEncoder were put into _jsonPool just before GC. IT would be removed instead of reused. Resetting the jsonEncoder , when Putting into _jsonPool, is unnecessary. Resetting jsonEncoder , when getting from _jsonPool, it is better.

Thanks for the details. There's a couple of additional considerations as well:
 * For performance, we only need to reset for "dirty" objects, but now it resets even for new encoders.
 * Resetting the object as part of put makes it more likely that misuses are detected.

The misuse detection is especially useful for correctness, so before we change this, I'd like to see what sort of performance improvement this approach leads to.

tylitianrui:
thanks. I agree with you.

PR #843, opened 2020-06-23
Honor CallerSkip when taking stack traces & add the StackSkip field

### LINKED ISSUE #512: CallerSkip does not apply to stacktraces captured by zap

### LINKED ISSUE #727: Add an option for setting stack skip
Sometimes you log from internal functions and want the stack trace to point to their caller, for example: a panic recovery middleware/wrapper. This is especially important when directing errors logged to zap to an error collection service such as Sentry/Rollbar/Bugsnag/etc as they rely on the stack trace to identify and merge similar errors.

I'm proposing a new field factory `func StackSkip(key string, skip int) Field` that also allows controlling the stack skip, that is the number of frames to skip from the top of the stack. And a new option `func SetStackSkip(skip int) Option` that can be used to set the stack skip for the automatic stack traces.

segevfiner:
* Honor `CallerSkip` when taking a stack trace. Both the caller and stack trace will now point to the same frame.
* Add `StackSkip` which is similar to `Stack` but allows skipping frames from the top of the stack.

This removes the internal behavior of skipping zap specific stack frames when taking a stack trace, and instead relies on the same behavior used to calculate the number of frames to skip for getting the caller to also skip frames in the stack trace.

Fixes #512, fixes #727

### DIFF
--- a/field.go
+++ b/field.go
@@ -371,11 +371,8 @@ func Stack(key string) Field {
 	return String(key, takeStacktrace(0))
 }
 
-// StackSkip constructs a field that stores a stacktrace of the current
-// goroutine under provided key, skipping the given number of frames from the
-// top of the stacktrace. Keep in mind that taking a stacktrace is eager and
-// expensive (relatively speaking); this function both makes an allocation and
-// takes about two microseconds.
+// StackSkip constructs a field similarly to Stack, but also skips the given
+// number of frames from the top of the stacktrace.
 func StackSkip(key string, skip int) Field {
 	// Returning the stacktrace as a string costs an allocation, but saves us
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
--- a/field.go
+++ b/field.go
@@ -364,26 +364,17 @@ func Timep(key string, val *time.Time) Field {
 // expensive (relatively speaking); this function both makes an allocation and
 // takes about two microseconds.
 func Stack(key string) Field {
-	// Returning the stacktrace as a string costs an allocation, but saves us
-	// from expanding the zapcore.Field union struct to include a byte slice. Since
-	// taking a stacktrace is already so expensive (~10us), the extra allocation
-	// is okay.
-	return String(key, takeStacktrace(0))
+	return StackSkip(key, 1) // skip Stack
 }
 
 // StackSkip constructs a field similarly to Stack, but also skips the given
 // number of frames from the top of the stacktrace.
 func StackSkip(key string, skip int) Field {
-	// Skip the call to StackSkip
-	if skip != 0 {
-		skip++
-	}
-
 	// Returning the stacktrace as a string costs an allocation, but saves us
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	return String(key, takeStacktrace(skip))
+	return String(key, takeStacktrace(skip+1)) // skip StackSkip
 }
 
 // Duration constructs a field with the given key and value. The encoder
--- a/field.go
+++ b/field.go
@@ -368,7 +368,20 @@ func Stack(key string) Field {
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	return String(key, takeStacktrace())
+	return String(key, takeStacktrace(nil))
+}
+
+// StackSkip constructs a field that stores a stacktrace of the current
+// goroutine under provided key, skipping the given number of frames from the
+// top of the stacktrace. Keep in mind that taking a stacktrace is eager and
+// expensive (relatively speaking); this function both makes an allocation and
+// takes about two microseconds.
+func StackSkip(key string, skip int) Field {
+	// Returning the stacktrace as a string costs an allocation, but saves us
+	// from expanding the zapcore.Field union struct to include a byte slice. Since
+	// taking a stacktrace is already so expensive (~10us), the extra allocation
+	// is okay.
+	return String(key, takeStacktrace(&skip))
 }
 
 // Duration constructs a field with the given key and value. The encoder
--- a/field.go
+++ b/field.go
@@ -374,6 +374,11 @@ func Stack(key string) Field {
 // StackSkip constructs a field similarly to Stack, but also skips the given
 // number of frames from the top of the stacktrace.
 func StackSkip(key string, skip int) Field {
+	// Skip the call to StackSkip
+	if skip != 0 {
+		skip++
+	}
+
 	// Returning the stacktrace as a string costs an allocation, but saves us
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
--- a/field.go
+++ b/field.go
@@ -368,7 +368,7 @@ func Stack(key string) Field {
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	return String(key, takeStacktrace(nil))
+	return String(key, takeStacktrace(0))
 }
 
 // StackSkip constructs a field that stores a stacktrace of the current
@@ -381,7 +381,7 @@ func StackSkip(key string, skip int) Field {
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	return String(key, takeStacktrace(&skip))
+	return String(key, takeStacktrace(skip))
 }
 
 // Duration constructs a field with the given key and value. The encoder
--- a/field_test.go
+++ b/field_test.go
@@ -259,6 +259,6 @@ func TestStackField(t *testing.T) {
 	f := Stack("stacktrace")
 	assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
 	assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
-	assert.Equal(t, takeStacktrace(nil), f.String, "Unexpected stack trace")
+	assert.Equal(t, takeStacktrace(0), f.String, "Unexpected stack trace")
 	assertCanBeReused(t, f)
 }
--- a/field_test.go
+++ b/field_test.go
@@ -259,6 +259,6 @@ func TestStackField(t *testing.T) {
 	f := Stack("stacktrace")
 	assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
 	assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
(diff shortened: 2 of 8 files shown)


stacktrace.go
```diff
@@ -43,7 +43,7 @@ var (
 	_zapStacktraceVendorContains = addPrefix("/vendor/", _zapStacktracePrefixes...)
 )
 
-func takeStacktrace() string {
+func takeStacktrace(skip *int) string {
```
prashantv, line 46: since we default to `0` anyway, can we make this an `int` and pass in `0` if we don't want any skip?

field.go
```diff
@@ -368,7 +368,20 @@ func Stack(key string) Field {
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	return String(key, takeStacktrace())
+	return String(key, takeStacktrace(nil))
+}
+
+// StackSkip constructs a field that stores a stacktrace of the current
```
prashantv, line 374: Rather than duplicate the comments, I'd suggest just referring to `Stack`, `StackSkip is similar to Stack, but also skips the given number of frames from the top of the stacktrace" or something similar

segevfiner:
>     * Do we need `StackSkip`: It's is not used or needed for the logger to skip stack frames based on caller skip, so doesn't seem too related to the change.

We don't have too, it's just for completeness. Zap takes a stack trace, and the user can take a stack using the `Stack` field and place it where he wants as well. So as Zap itself can take a stack trace while skipping frames, it makes sense to me that the user can also take a stack with a skip using a field and place it where he wants.

>     * I think leaving the stacktrace pruning as-is is best for backwards compatibility
>
> I think using an option is the right thing for backwards compatibility, although in this case, I think it may be more of a bug that the stacktrace attached skips zap stackframes, but doesn't take callerSkip into apply.

That's why I kept it for the time being and why it's being a flag. Though I wonder if that flag is on should we also do zap stackframe pruning or rely on `callerSkip` being set correctly...

prashantv:
I discussed whether we should default `UseCallerSkipForStacktrace` with @abhinav last week, and we agreed that:
 * This is better behaviour by default
 * There may be some use-cases that are depending stack traces to include the frames, but it doesn't seem common

Given that, we agreed on:
 * Always use caller skip for stack trace as part of this change (this is effectively a bug fix)
 * If we get reports from users who rely on stack traces to *not* inherit skip, we'll add a future option (the implementation should allow for that flexibility, but not add the option yet)

Would you be able to update your change to match the above behaviour @segevfiner? Thanks!

segevfiner:
@prashantv I can. I will just need to find some time for it. Effectively I'll be removing the option and moving it to always use the new behavior.

The only problem part is the zap stack frame trimming which I kept to preserve the old behavior when the option is false, which I will then need to remove along with its tests, so if you would want to restore the option, you might have to restore that code later on. It will probably also fixing some tests and passing the skip correctly in the `Stack` field rather then relying on the old behavior.

field.go
```diff
@@ -368,7 +368,22 @@ func Stack(key string) Field {
...
-	return String(key, takeStacktrace())
+	return String(key, takeStacktrace(0))
+}
+
+// StackSkip constructs a field similarly to Stack, but also skips the given
+// number of frames from the top of the stacktrace.
+func StackSkip(key string, skip int) Field {
+	// Skip the call to StackSkip
+	if skip != 0 {
+		skip++
```
abhinav, line 379: minor stylistic thing: Comments on their own lines should be full sentences ending with periods, or they can be inline if they're not full sentences.

```suggestion
	// Skip the call to StackSkip.
	if skip != 0 {
		skip++
```

Or,

```suggestion
	if skip != 0 {
		skip++ // skip StackSkip
```

field.go
```diff
@@ -368,7 +368,22 @@ func Stack(key string) Field {
 	// from expanding the zapcore.Field union struct to include a byte slice. Since
 	// taking a stacktrace is already so expensive (~10us), the extra allocation
 	// is okay.
-	return String(key, takeStacktrace())
+	return String(key, takeStacktrace(0))
```
abhinav, line 371: Rather than having Stack and StackSkip be nearly the same, I think it may be better if Stack called StackSkip.

```
	return StackSkip(1) // skip Stack
```

segevfiner:
Made the requested change.

stacktrace.go
```diff
@@ -22,36 +22,28 @@ package zap
...
 	defer buffer.Free()
 	programCounters := _stacktracePool.Get().(*programCounters)
 	defer _stacktracePool.Put(programCounters)
 
 	var numFrames int
 	for {
-		// Skip the call to runtime.Counters and takeStacktrace so that the
+		// Skip the call to runtime.Callers and takeStacktrace so that the
 		// program counters start at the caller of takeStacktrace.
 		numFrames = runtime.Callers(2, programCounters.pcs)
```
prashantv, line 48: why do we implement `skip` manually below, can't we just add it here?

stacktrace.go
```diff
@@ -63,17 +55,15 @@ func takeStacktrace() string {
 	}
 
 	i := 0
-	skipZapFrames := true // skip all consecutive zap frames at the beginning.
```
prashantv, line 66: I'm not 100% sure what the impact of removing this is -- are all callers of `takeStacktrace` correctly passing skip so we no longer need this? Is there any scenario where it might not be correct?

segevfiner: That's one of the things I tried changing the tests for. I already didn't use it when enabling this feature since the behavior with both skip and it was weird, I removed it now since it is not used with this being always enabled.

I explicitly changed the few places calling it so they pass a correct skip value. Also, Zap already took care to compensate for differences in call stack depth between `Logger` and `SugaredLogger` for the caller, so I reused that here.

stacktrace.go
segevfiner: Leftover from still having skip zap frames.

stacktrace_ext_test.go
```diff
@@ -120,6 +120,31 @@ func TestStacktraceFiltersVendorZap(t *testing.T) {
...
 }
 
+func TestStacktraceWithoutCallerSkip(t *testing.T) {
+	withLogger(t, func(logger *zap.Logger, out *bytes.Buffer) {
+		func() {
+			logger.Error("test log")
+		}()
+
+		require.Contains(t, out.String(), "TestStacktraceWithoutCallerSkip.", "Should not skip too much")
+		require.Contains(t, out.String(), "TestStacktraceWithoutCallerSkip", "Should not skip too much")
```
abhinav, line 130: If we pass the first check, we'll always also pass the second check. We can probably just drop the second.
```suggestion
		require.Contains(t, out.String(), "TestStacktraceWithoutCallerSkip.", "Should not skip too much")
```

stacktrace_test.go
```diff
@@ -29,47 +29,46 @@ import (
...
-	nonZapFrames := []string{
-		"github.com/uber/tchannel-go.NewChannel",
-		"go.uber.org/not-zap.New",
-		"go.uber.org/zapext.ctx",
-		"go.uber.org/zap_ext/ctx.New",
-	}
+func TestTakeStacktraceWithSkip(t *testing.T) {
+	trace := takeStacktrace(1)
+	lines := strings.Split(trace, "\n")
+	require.True(t, len(lines) > 0, "Expected stacktrace to have at least one frame.")
```
abhinav, line 46: minor: Can use NotEmpty.
```suggestion
	require.NotEmpty(t, lines, "Expected stacktrace to have at least one frame.")
```

stacktrace_test.go
```diff
@@ -29,47 +29,46 @@ import (
...
+	assert.Contains(
+		t,
+		lines[0],
+		"testing.",
+		"Expected stacktrace to start with the test runner (skipping our own frame) %s.", lines[0],
```
abhinav, line 51: minor: Testify will include the `lines[0]` in the output so we don't need to.
```suggestion
		"Expected stacktrace to start with the test runner (skipping our own frame).",
```

stacktrace_test.go
```diff
@@ -29,47 +29,46 @@ import (
...
-			require.False(t, isZapFrame(f), f)
-		}
-	})
+func TestTakeStacktraceWithSkipInnerFunc(t *testing.T) {
+	var trace string
+	func() {
+		trace = takeStacktrace(2)
+	}()
+	lines := strings.Split(trace, "\n")
+	require.True(t, len(lines) > 0, "Expected stacktrace to have at least one frame.")
```
abhinav, line 61:
```suggestion
	require.NotEmpty(t, lines, "Expected stacktrace to have at least one frame.")
```

stacktrace_test.go
```diff
@@ -29,47 +29,46 @@ import (
...
+	assert.Contains(
+		t,
+		lines[0],
+		"testing.",
+		"Expected stacktrace to start with the test function (skipping the test function) %s.", lines[0],
```
abhinav, line 66:
```suggestion
		"Expected stacktrace to start with the test function (skipping the test function.",
```

field_test.go
```diff
@@ -259,6 +260,23 @@ func TestStackField(t *testing.T) {
...
+
+func TestStackSkipField(t *testing.T) {
+	f := StackSkip("stacktrace", 1)
+	assert.Equal(t, "stacktrace", f.Key, "Unexpected field key.")
+	assert.Equal(t, zapcore.StringType, f.Type, "Unexpected field type.")
+	assert.Equal(t, takeStacktrace(1), f.String, "Unexpected stack trace")
+	assertCanBeReused(t, f)
+}
+
+func TestStackSkipFieldWithSkip(t *testing.T) {
```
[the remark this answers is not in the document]
segevfiner: There is. On tests `Stack` which does indeed call `StackSkip` but we want to see that it calls it correctly and produces a stack trace pointing to the right location. The other calls the new `StackSkip` which we also want to test calling directly.

logger.go
```diff
@@ -304,7 +304,7 @@ func (log *Logger) check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
 		}
 	}
 	if log.addStack.Enabled(ce.Entry.Level) {
-		ce.Entry.Stack = Stack("").String
+		ce.Entry.Stack = StackSkip("", log.callerSkip+callerSkipOffset).String
```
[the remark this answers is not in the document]
segevfiner: Note that that's the assumption that taking the caller also makes, so it is simply repeated here.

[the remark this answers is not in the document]
prashantv: Yep, since we're using the same assumption when setting `entry.Caller`, I think that makes sense. Thanks for verifying.

prashantv:
The change looks good! Can we update the title / summary to reflect the changes made during code review (we now always respect caller skip when taking stack traces as part of an entry)

segevfiner:
@prashantv Updated.

PR #845, opened 2020-06-27
stop time.Now().UTC() escape to heap

amarjeetanandsingh:
### Problem
* `time.Now().UTC()` use to escape to heap because it was passed as argument for an interface param
* Old `go build -gcflags='-m -m'` shows :- 
```
./logger.go:302:96: time.Now().UTC() escapes to heap
```
### Solution
Passing error message as string argument for string param of `fmt.Fprintf`  stopped `escape to heap`
**Result**:
```
./logger.go:302:36: (*Logger).check time.Time(~R0).String() + " Logger.check error: failed to get caller\n" does not escape
```

Am not sure if this change worth merging because it doesn't make any significant difference. However, I think it improves something. So thought of raising this one-liner PR.


amarjeetanandsingh:
closing it because it fails the `make lint` check 

```
logger.go:303:33: printf-style function with dynamic format string and no further arguments should use print-style function instead (SA1006)
```

Issue #851, opened 2020-07-27
Unexpected memory usage


wyxloading:
Here cloned a `jsonEncoder` (which actually get from `_jsonPool`), but never put back to `_jsonPool` after use.  

https://github.com/uber-go/zap/blob/53a387079b46cce51bbb6d360d9c57f86977addb/zapcore/console_encoder.go#L131-L145

Perhaps we need something like below to decrease memory allocations.

```go
context := c.jsonEncoder.Clone().(*jsonEncoder) 
defer func() {
    context.buf.Free()
    putJSONEncoder(context)
}
```

prashantv:
From a quick glance, this seems reasonable. Can you send a PR, and ensure the race detector doesn't find any issues once the above change is made. Thanks!

wyxloading:
I would like to send PR for this (FYI, PR link #852)

abhinav:
#852 was merged so we can resolve this.

PR #852, opened 2020-07-28
consoleEncoder: put cloned jsonEncoder back to pool

wyxloading:
This PR put back the cloned `jsonEncoder` in `consoleEncoder` to `_jsonPool` as mentioned in #851 

Tested and got no errors ( with race detector )
* `make test` 
* `go test -race -run="^$$" -bench="BenchmarkZapConsole" ./zapcore`

### DIFF
--- a/zapcore/console_encoder.go
+++ b/zapcore/console_encoder.go
@@ -131,6 +131,8 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer, 
 func (c consoleEncoder) writeContext(line *buffer.Buffer, extra []Field) {
 	context := c.jsonEncoder.Clone().(*jsonEncoder)
 	defer func() {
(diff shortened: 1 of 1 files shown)


zapcore/console_encoder.go
```diff
@@ -130,7 +130,10 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer,
 
 func (c consoleEncoder) writeContext(line *buffer.Buffer, extra []Field) {
 	context := c.jsonEncoder.Clone().(*jsonEncoder)
-	defer context.buf.Free()
+	defer func() {
+		context.buf.Free()
```
prashantv, line 134:
```suggestion
		// putJSONEncoder assumes the buffer is still used, but we write out the buffer so
		// we can free it.
		context.buf.Free()
```

PR #854, opened 2020-08-04
Show "<nil>" for nil Stringer.

andy-retailnext:
In encodeStringer, instead of returning an error when a panic
occurs when calling String() on a nil pointer, use the string value
"<nil>" like the fmt package does.

It is not always possible to handle this case by fixing the
implementation of String to not panic. This requires implementing
String with a pointer receiver, which doesn't work if you need to
be able to call String on non-addressable values.

### DIFF
--- a/zapcore/field.go
+++ b/zapcore/field.go
@@ -208,7 +208,12 @@ func addFields(enc ObjectEncoder, fields []Field) {
 func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) (err error) {
 	defer func() {
 		if v := recover(); v != nil {
-			err = fmt.Errorf("PANIC=%v", v)
+			val := reflect.ValueOf(stringer)
+			if val.Kind() == reflect.Ptr && val.IsNil() {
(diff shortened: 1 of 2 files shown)


zapcore/field.go
```diff
@@ -208,7 +208,12 @@ func addFields(enc ObjectEncoder, fields []Field) {
 func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) (err error) {
 	defer func() {
 		if v := recover(); v != nil {
-			err = fmt.Errorf("PANIC=%v", v)
+			val := reflect.ValueOf(stringer)
```
prashantv, line 211: This block looks very similar to the logic in the fmt package uses to determine if `nil` should be used,
https://golang.org/src/fmt/print.go#L540

Should we reference this?

abhinav, line 211:
```suggestion
			// See also https://golang.org/src/fmt/print.go#L540
			val := reflect.ValueOf(stringer)
```

PR #861, opened 2020-09-01
Add options to customize Fatal behaviour for better testability

### LINKED ISSUE #846: Make it easy to test Fatal logs without requiring indirection at the call site
Related to #494

There's no clean way to test code that uses `Fatal`, 

Rather than exposing the internal/exit package, we could expose an option that overrides the exit function used by zap.

prashantv:
Fixes #846.

There's currently no easy way to test Fatal from outside of zap, as it
triggers an os.Exit(1). Add options to customize the behaviour (allowing
for panic, or Goexit), which can be used with `recover()` in tests.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -49,6 +49,7 @@ type Logger struct {
 	addStack  zapcore.LevelEnabler
 
 	callerSkip int
+	onFatal    zapcore.CheckWriteAction // default is WriteThenFatal.
 }
 
 // New constructs a new Logger from the provided zapcore.Core and Options. If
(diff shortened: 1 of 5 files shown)


zapcore/entry.go
```diff
@@ -163,6 +164,8 @@ const (
 	WriteThenPanic
 	// WriteThenFatal causes a fatal os.Exit after Write.
 	WriteThenFatal
+	// WriteThenGoexit runs runtime.Goexit after Write.
+	WriteThenGoexit
```
abhinav, line 168: Comment on CheckWriteAction states:

> Actions are ordered in increasing severity.

So this probably needs to go before panic.

logger.go
```diff
@@ -280,7 +281,11 @@ func (log *Logger) check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
 	case zapcore.PanicLevel:
 		ce = ce.Should(ent, zapcore.WriteThenPanic)
 	case zapcore.FatalLevel:
-		ce = ce.Should(ent, zapcore.WriteThenFatal)
+		onFatal := log.onFatal
+		if onFatal == 0 {
```
abhinav, line 285:
```suggestion
		if onFatal == zapcore.WriteThenNoop {
```

logger.go
```diff
@@ -49,6 +49,7 @@ type Logger struct {
 	addStack  zapcore.LevelEnabler
 
 	callerSkip int
+	onFatal    zapcore.CheckWriteAction // default is WriteThenFatal.
```
abhinav, line 52:
```suggestion
	onFatal    zapcore.CheckWriteAction // default is WriteThenFatal
```

This isn't a full sentence so we don't need a trailing `.`

PR #865, opened 2020-09-18
Byte alignment, optimization from 96 to 80 byte

wziww:
Optimization `Logger` struct

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -39,11 +39,11 @@ import (
 // safety over brevity. For most applications, the SugaredLogger strikes a
 // better balance between performance and ergonomics.
 type Logger struct {
-	onFatal zapcore.CheckWriteAction // default is WriteThenFatal
+	core zapcore.Core
 
 	development bool
 	addCaller   bool
-	core        zapcore.Core
+	onFatal     zapcore.CheckWriteAction // default is WriteThenFatal
 
 	name        string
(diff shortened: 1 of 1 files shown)


prashantv (commented):
Thank you for the contribution @wziww 

The change is correct, but it has some unnecessary reordering. For example, there's no advantage in moving `core` to not be the first field, and having it there is a little cleaner.

The main improvement is from grouping the bool types + the `CheckWriteAction`, no need to reorder the `zapcore.Core`.

wziww:
> Thank you for the contribution @wziww
> 
> The change is correct, but it has some unnecessary reordering. For example, there's no advantage in moving `core` to not be the first field, and having it there is a little cleaner.
> 
> The main improvement is from grouping the bool types + the `CheckWriteAction`, no need to reorder the `zapcore.Core`.

Yeah you right.

prashantv (approved):

rabbbit:
Is there an easy/automated way you were aligning this, or is it just manual checking? :)

wziww:
> Is there an easy/automated way you were aligning this, or is it just manual checking? :)

I do think manual checking is the best, try to form a habit while coding.
But we can also use `reflect` to build some small tools to check it.

abhinav (approved):

PR #866, opened 2020-09-20
feat: Add Exit Logging func 

ysicing:
Useful for running exit log a message.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -50,7 +50,11 @@ type Logger struct {
 
 	callerSkip int
 	onFatal    zapcore.CheckWriteAction // default is WriteThenFatal
+
+	exitFunc exitFunc // default is os.Exit
 }
+
+type exitFunc func(int)
 
 // New constructs a new Logger from the provided zapcore.Core and Options. If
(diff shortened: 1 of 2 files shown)


prashantv:
Thanks for the contribution @ysicing. Can you please provide more context on your use-case, and what advantages there are by having zap handle this, as opposed to calling `os.Exit` explicitly,
```go
logger.Error(...)
os.Exit(code)
```

Integrating this functionality into zap has some drawbacks:
 * Semantics on level are unclear (the other methods have their own level, but `Exit` doesn't), so it's inconsistent.
   * Hardcoding a level doesn't seem ideal, users may want to use other levels. E.g., `info` may be appropriate for `os.Exit(0)`.
 * Testability concerns: without additional APIs, exit calls could not be easily tested.

Since there are many exit scenarios, and it's possible for the user to call `os.Exit` themselves, I'm not sure what the benefits are of handling this within zap.

ysicing:
This may not be an elegant way. In some cases, you may need to exit after printing the log. Maybe we can hook it up

PR #867, opened 2020-09-24
field/error: Handle panic in Error()

FMLS:
We shouldn't panic if the `Error()` method of an `error` panics.

```
type T struct{ msg string }

func (t *T) Error() string { return t.msg }

// The following panics.
var n *T = nil
log.Info("panic", zap.Error(n))
```

### DIFF
--- a/error.go
+++ b/error.go
@@ -21,8 +21,9 @@
 package zap
 
 import (
-	"go.uber.org/zap/zapcore"
 	"sync"
+
+	"go.uber.org/zap/zapcore"
 )
 
 var _errArrayElemPool = sync.Pool{New: func() interface{} {
--- a/error.go
+++ b/error.go
@@ -21,9 +21,8 @@
 package zap
 
 import (
-	"sync"
-
 	"go.uber.org/zap/zapcore"
+	"sync"
 )
 
 var _errArrayElemPool = sync.Pool{New: func() interface{} {
--- a/zapcore/error.go
+++ b/zapcore/error.go
@@ -49,7 +49,7 @@ func encodeError(key string, err error, enc ObjectEncoder) (retErr error) {
 	defer func() {
 		if rerr := recover(); rerr != nil {
 			// If it's a nil pointer, just say "<nil>". The likeliest causes are a
-			// Stringer that fails to guard against nil or a nil pointer for a
+			// error that fails to guard against nil or a nil pointer for a
 			// value receiver, and in either case, "<nil>" is a nice result.
 			if v := reflect.ValueOf(err); v.Kind() == reflect.Ptr && v.IsNil() {
 				enc.AddString(key, "<nil>")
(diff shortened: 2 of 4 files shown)


prashantv (changes requested):
Thank you for the contribution @FMLS 

Using `reflect` check for `nil` isn't always right, as it's possible to have a `nil` value that implements the `error` interface,
https://play.golang.org/p/5T8S35zdGsH

To handle nil in this case, please follow the same logic used by `Stringer` to detect a nil panic (which uses `recover`, similar to the standard library).

zapcore/field.go
```diff
@@ -209,7 +209,7 @@ func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) (retErr
 	// Try to capture panics (from nil references or otherwise) when calling
 	// the String() method, similar to https://golang.org/src/fmt/print.go#L540
 	defer func() {
-		if err := recover(); err != nil {
+		if rerr := recover(); rerr != nil {
```
abhinav, line 212: can we undo this rename?

```suggestion
		if err := recover(); err != nil {
```

zapcore/field.go
```diff
@@ -218,7 +218,7 @@ func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) (retErr
 				return
 			}
 
-			retErr = fmt.Errorf("PANIC=%v", err)
+			retErr = fmt.Errorf("PANIC=%v", rerr)
```
abhinav, line 221:
```suggestion
			retErr = fmt.Errorf("PANIC=%v", err)
```

zapcore/error.go
```diff
@@ -42,13 +43,29 @@ import (
...
+		}
+	}()
+
 	basic := err.Error()
 	enc.AddString(key, basic)
 
 	switch e := err.(type) {
 	case errorGroup:
-		return enc.AddArray(key+"Causes", errArray(e.Errors()))
+		retErr = enc.AddArray(key+"Causes", errArray(e.Errors()))
```
abhinav, line 68: we don't need to do this. the defer and the named return will still capture the value.

```suggestion
		return enc.AddArray(key+"Causes", errArray(e.Errors()))
```

zapcore/error.go
```diff
@@ -57,7 +74,7 @@ func encodeError(key string, err error, enc ObjectEncoder) error {
 			enc.AddString(key+"Verbose", verbose)
 		}
 	}
-	return nil
+	return retErr
```
abhinav, line 77:
```suggestion
	return nil
```

zapcore/error.go
```diff
@@ -42,13 +43,29 @@ import (
...
 //    ],
 //  }
-func encodeError(key string, err error, enc ObjectEncoder) error {
+func encodeError(key string, err error, enc ObjectEncoder) (retErr error) {
+	// Try to capture panics (from nil references or otherwise) when calling
+	// the Error() method
+	defer func() {
+		if rerr := recover(); rerr != nil {
+			// If it's a nil pointer, just say "<nil>". The likeliest causes are a
+			// Stringer that fails to guard against nil or a nil pointer for a
```
abhinav, line 52:
```suggestion
			// error that fails to guard against nil or a nil pointer for a
```

error.go
```diff
@@ -21,9 +21,8 @@
 package zap
 
 import (
-	"sync"
-
 	"go.uber.org/zap/zapcore"
+	"sync"
 )
```
abhinav, line 26: this doesn't need to change

```suggestion
	"sync"

	"go.uber.org/zap/zapcore"
)
```

abhinav (approved):

prashantv (approved):

Issue #853, opened 2020-08-03
Feature Request: toml tags for configs


Prepodavan:
can i hope to see this feauture anytime later?
some fields of EncoderConfig have json+yaml tags that differs from field name (but have no toml tag). that forces me to use configs with different keys in json and toml.
it impossible to use toml for zap.Config (with same key as in json) since without tags toml can't recognize that some fields of EncoderConfig is encoding.TextUnmarshler so they have nil and panics when logger tries to use it
code for reproducing
```golang
package main

import (
	"fmt"
	"github.com/BurntSushi/toml"
	"go.uber.org/zap"
)

func main() {
	var cfg1 zap.Config
	_, err1 := toml.Decode(`
level = "debug"
encoding = "json"
outputPaths = ["stdout"]
errorOutputPaths = ["stderr"]

    [encoderConfig]
        messageKey = "message"
        levelKey = "level"
        encodeLevel = "lowercase"	# real field name
`, &cfg1)

	var cfg2 zap.Config
	_, err2 := toml.Decode(`
level = "debug"
encoding = "json"
outputPaths = ["stdout"]
errorOutputPaths = ["stderr"]

    [encoderConfig]
        messageKey = "message"
        levelKey = "level"
        levelEncoder = "lowercase" 		# json+yaml tag (provided by example of json config from doc)
`, &cfg2)

	fmt.Println(err1, err2) // nil nil
	l1, err1 := cfg1.Build()
	l2, err2 := cfg2.Build()
	fmt.Println(err1, err2) // nil nil
	l1.Debug("hello world") // its fine
	l2.Debug("hello world") // panic as long as cfg2.EncoderConfig.EncodeLevel is nil as long as
	// not recognized as encoding.TextUnmarshaler
}
```

LeviMatus:
If this is something the maintainers are okay with considering, then I can put up a PR adding toml tags to the configs.

prashantv:
We are fine with adding toml tags to zap, and taking on a toml test-only dependency. Please avoid adding toml dependencies to the non-test code.

LeviMatus:
In my continued work on #868, looking at the most commonly used TOML parsing/encoding libraries ([BurntSushi/toml](https://github.com/BurntSushi/toml), [pelletier/go-toml](https://github.com/pelletier/go-toml), and [naoina/toml](https://github.com/naoina/toml)), integration into the existing code isn't as simple as initially thought. The main issue is 

```toml
[timeEncoder]
[...]

PR #868, opened 2020-10-01
feat: add support for toml tags

LeviMatus:
#### Added
+ TOML tags to `Config`, `SamplingConfig`, and `zapcore.EncoderConfig`.
+ Example test in `example_test.go` which introduces a test-dependency on `github.com/BurntSushi/toml` for decoding TOML info a `Config` struct.

See also: #853

### DIFF
--- a/config.go
+++ b/config.go
@@ -37,9 +37,9 @@ import (
 // Values configured here are per-second. See zapcore.NewSamplerWithOptions for
 // details.
 type SamplingConfig struct {
-	Initial    int                                           `json:"initial" yaml:"initial"`
-	Thereafter int                                           `json:"thereafter" yaml:"thereafter"`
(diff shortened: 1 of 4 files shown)


prashantv (commented):
Thanks for the contribution @LeviMatus 

I found that there was a field in zapcore/encoder.go that has yaml tags but not toml tags,
```
Layout string `json:"layout" yaml:"layout"`
```

We might need to add some logic for TOML as well, I'm not sure if TOML can take advantage of the existing `UnmarshalYAML` or `UnmarshalJSON` methods.

LeviMatus:
Good catch @prashantv. I just spent some time trying an `UnmarshalTOML` implementation but ran into some issues with different TOML packages when handling the configs with the `timeEncoder = { layout = "06/01/02 03:04pm" }` scenario. I'll need to look this over some more.

PR #871, opened 2020-10-23
Add "UnmarshalFlag" to x/zapcore/Level

witriew:
https://pkg.go.dev/github.com/jessevdk/go-flags#Unmarshaler requires a
`UnmarshalFlag` to satisfy the interface for parsing purposes.  Provide this
method so that a x/zapcore/Level could be used as a type when trying to parse
flags.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module go.uber.org/zap
 go 1.13
 
 require (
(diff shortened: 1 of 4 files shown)


witriew:
Adding in methods to satisfy non-stdlib interfaces is probably not preferred.

Abandoning this PR.

PR #881, opened 2020-11-22
Add support for grpclog.LoggerV2

### LINKED ISSUE #534: zapgrpc upgrade for grpclog.LoggerV2

ash2k:
Fixes #534

Closes https://github.com/uber-go/zap/pull/538


ash2k:
I don't see how coverage can be improved further. The only two lines that are not executed are the ones that call the real `Fatal`.

prashantv (commented):
I don't think we want this significantly larger API surface added to the zap logger.

I think it's best for a wrapper to live outside of zap to support the gRPC v2 logger interface.

ash2k:
Hm, fine. I assumed you wanted it because #534 was open for 3 years and nobody closed it, and the PR too.

prashantv:
> Hm, fine. I assumed you wanted it because #534 was open for 3 years and nobody closed it, and the PR too.

Apologies for leaving the task open, I didn't realize how much of an  API change it would result in.

prashantv:
By API change, I mean the additional methods that are added to the Sugar logger. The number of methods for each level increases from the 3 per-level (which all behave differently) to 4 per-level (where the Ln suffixed methods behave the same).

Once an API is added, it must be maintained for backwards compatibility, and I don't think think this added API surface makes sense to add directly to the zap Sugar logger.

ash2k:
> By API change, I mean the additional methods that are added to the Sugar logger. 

But this PR does not add any methods to the Sugar logger?

prashantv:
For some reason, I thought we were modifying the Sugar logger, which was a misread on my part.

Given this doesn't add methods to the sugar logger, but to the gRPC-specific logger, I think this is reasonable, reopening for review.

Sorry for the confusion @ash2k

prashantv (commented):
The `WithDebug` option still makes sense for the V1 logger Print* methods, but not for the V2 logger. We can either:
 * clarify in the documentation for `WithDebug`
 * Add a separate constructor + type for the V2 logger (which takes a separate option type)

The latter increases the API surface, but may be clearer (especially if there's other options in future that will only be applicable to V2).

I don't have strong opinions on either approach though, thoughts @ash2k?

ash2k:
@prashantv 

> clarify in the documentation for WithDebug

I'd go for this option as it's less work and does not add API surface - this can always be done later if needed but cannot be undone, so I think it's better to keep API as is.

Could you suggest what should be documented there? That the option is only applicable to v1 API?

prashantv:
> @prashantv
> 
> > clarify in the documentation for WithDebug
> 
> I'd go for this option as it's less work and does not add API surface - this can always be done later if needed but cannot be undone, so I think it's better to keep API as is.
> 
> Could you suggest what should be documented there? That the option is only applicable to v1 API?

Yep, specifically that `WithDebug` will only affect the Print* methods, which are only used in the gRPC v1 Logger APIs. Thanks

ash2k:
@prashantv done

zapgrpc/zapgrpc.go
```diff
@@ -36,65 +39,129 @@ func (f optionFunc) apply(log *Logger) {
...
+	if l.fatalToWarn {
+		l.delegate.Warnf(format, args...)
+	} else {
+		l.delegate.Fatalf(format, args...)
+	}
+}
+
+// V implements grpclog.LoggerV2.
+func (l *Logger) V(level int) bool {
+	return l.delegate.Desugar().Core().Enabled(zapcore.Level(level))
```
prashantv, line 166: I'm not sure the gRPC levels are directly convertible to zap levels,

Based on the code here,
https://github.com/grpc/grpc-go/blob/v1.35.0/grpclog/loggerv2.go#L79

It looks like gRPC maps:
 - 0=info
 - 1=warn
 - 2=error
 - 3=fatal

zap is similar, but slightly different
https://pkg.go.dev/go.uber.org/zap@v1.16.0/zapcore#Level
 - 0=info
 - 1=warn
 - 2=error
 - 3=dpanic

We should convert gRPC's fatal (3) to a 5 to map to zap.

Can we add tests to validate the gRPC / zap levels map correctly as well?

zapgrpc/zapgrpc.go
```diff
@@ -36,65 +39,129 @@ func (f optionFunc) apply(log *Logger) {
...
+	l.Print(args...)
+}
+
+// Info implements grpclog.LoggerV2.
+func (l *Logger) Info(args ...interface{}) {
+	l.delegate.Info(args...)
+}
+
+// Infoln implements grpclog.LoggerV2.
+func (l *Logger) Infoln(args ...interface{}) {
```
prashantv, line 102: The *Ln methods are slightly different from the non-Ln methods, the Ln methods always add spaces between all arguments, while the non-Ln methods only add spaces between objects if the objects are not both strings,
https://golang.org/pkg/fmt/#Println

See related issue on zap which doesn't have *Ln methods,
https://github.com/uber-go/zap/issues/889

I think the existing `Println` method is similarly incorrect, just didn't catch this till now

prashantv (commented):
Sorry for the delay in reviewing, I wanted to make sure I gave this a thorough look, and noticed a couple of small issues.

zapgrpc/zapgrpc.go
ash2k: @prashantv Done, PTAL.

zapgrpc/zapgrpc.go
ash2k: Hm, I was not aware of such nuance. Done.

zapgrpc/zapgrpc.go
```diff
@@ -21,7 +21,30 @@
...
 
-import "go.uber.org/zap"
+import (
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	// See https://github.com/grpc/grpc-go/blob/v1.35.0/grpclog/loggerv2.go#L77-L86
+
```
prashantv, line 31:
```suggestion
// See https://github.com/grpc/grpc-go/blob/v1.35.0/grpclog/loggerv2.go#L77-L86
const (
```

zapgrpc/zapgrpc.go
```diff
@@ -36,65 +59,148 @@ func (f optionFunc) apply(log *Logger) {
...
+// V implements grpclog.LoggerV2.
+func (l *Logger) V(level int) bool {
+	return l.delegate.Desugar().Core().Enabled(grpc2zapLvl[level])
 }
 
-// Println implements grpclog.Logger.
-func (l *Logger) Println(args ...interface{}) {
-	l.print(l.log, args...)
+// addSpaces always adds spaces between arguments like https://golang.org/pkg/fmt/#Println
+func addSpaces(args []interface{}) []interface{} {
```
prashantv, line 190: instead of this, we could use `Sprintln(args)` and trim the trailing newline and use the string instead

zapgrpc/zapgrpc.go
```diff
@@ -21,7 +21,30 @@
...
+	grpcLvlInfo  = 0
+	grpcLvlWarn  = 1
+	grpcLvlError = 2
+	grpcLvlFatal = 3
+)
+
+var (
+	// grpc2zapLvl maps gRPC log levels to zap log levels.
+	// See https://pkg.go.dev/go.uber.org/zap@v1.16.0/zapcore#Level
+	grpc2zapLvl = map[int]zapcore.Level{
```
prashantv, line 41: nit: can we use camel case + full word here `_grpcToZapLevel` (since it's not used in a lot of places, better to keep it clear). we also prefixes our globals with `_` in zap

prashantv (approved):
Couple of minor comments, but overall looks good

zapgrpc/zapgrpc.go
ash2k: We could, but I think this would lead to more memory allocations. This code only allocates once, only when needed and the perfect slice size. `Sprintln()` would turn all arguments into strings, which is arbitrarily expensive in terms of both CPU and memory. I think it'd better to keep the current code.

zapgrpc/zapgrpc.go
ash2k: Done

zapgrpc/zapgrpc.go
prashantv: We have to call `Sprint*` at some point to convert the arguments into a string for the logger API. With the current implementation, doing `Sprintln` here would be an additional allocation, but we can actually avoid a second `Sprint` in the case we receive a single string argument,
https://github.com/uber-go/zap/pull/913

With the above change merged in, `Sprintln` will actually be more efficient, as we convert the arguments to a string once (1 alloc), vs creating a slice here with the strings, and then converting them using `Sprint` later (2 allocs).

zapgrpc/zapgrpc_test.go
```diff
@@ -33,44 +34,186 @@ import (
...
+		"hello world",
+		"",
+		"foo",
+		"foo bar",
 	}, func(logger *Logger) {
+		logger.Info("hello")
+		logger.Infof("%s world", "hello")
+		logger.Infoln()
+		logger.Infoln("foo")
+		logger.Infoln("foo", "bar")
```
prashantv, line 51: for additional validation, we should add a case with a mix of strings + other types too, e.g.,
```go
log.Infoln("s1", "s2", 1, 2, 3, "s3", 4, "s5", 6)
```

zapgrpc/zapgrpc.go
ash2k: The above is true, but consider a case where arguments are expensive to turn into strings and the logger is disabled. This is a common scenario - verbose debug logging that outputs bulky objects with all the details. This case would perform much worse if we use `Sprint` here.

As a user, I would expect the logger to evaluate (i.e. turn into strings) arguments if and only if it will then actually log. As a user, I would not expect to need to guard my log statements with `if is_log_level_enabled` checks but just rely on the above.

prashantv: Agree that we should avoid expensive operations if the logger is disabled (without pushing any checks to the caller). However, this implementation still allocates for the slice even if the log is disabled. An alternate option is to add level checks into the gRPC logger explicitly, so it skips doing any work if the level is disabled.  If the level is enabled, then we can use `Sprintln` to serialize the arguments. Would that work?

ash2k:
```go
// Fatalln implements grpclog.LoggerV2.
func (l *Logger) Fatalln(args ...interface{}) {
	if l.delegate.Desugar().Core().Enabled(zapcore.FatalLevel) {
		l.Fatal(fmt.Sprintln(args...))
	}
}
```
This would allocate too in the `Desugar()` call 🤷 I've made the change, PTAL.

ash2k: I've extracted the `LevelEnabler` into a field to avoid this allocation.

zapgrpc/zapgrpc.go
```diff
@@ -36,65 +60,158 @@ func (f optionFunc) apply(log *Logger) {
...
+func (l *Logger) V(level int) bool {
+	return l.levelEnabler.Enabled(_grpcToZapLevel[level])
 }
 
-// Println implements grpclog.Logger.
-func (l *Logger) Println(args ...interface{}) {
-	l.print(l.log, args...)
+func sprintln(args []interface{}) string {
+	s := fmt.Sprintln(args...)
+	return s[:len(s)-1]
```
prashantv, line 216:
```suggestion
	// Drop the new line character added by Sprintln 
	return s[:len(s)-1]
```

zapgrpc/zapgrpc_test.go
```diff
@@ -33,44 +34,210 @@ import (
...
+			t.Run(fmt.Sprintf("%s %d", zapLvl, grpcLvl), func(t *testing.T) {
+				checkLevel(t, zapLvl, true, func(logger *Logger) bool {
+					return logger.V(grpcLvl)
+				})
+			})
+		}
+	}
+}
+
+func TestLoggerVFalseExpected(t *testing.T) {
```
prashantv, line 193: suggest merging this test with the one above using a test table like:
```
tests := struct{
  zapLevel zapcore.Level
  grpcEnabled []int 
  grpcDisabled []int
}
```

prashantv (approved):
Looks good with a couple of minor test suggestions.

zapgrpc/zapgrpc_test.go
ash2k: Done

abhinav (approved):
LGTM.

I think it would be valuable to add a test that tests directly against grpc-go
so that we know we're interface compliant.

But we would also like to not add a dependency on grpc-go in Zap. This
probably requires a submodule. I can add that in a follow up PR.

abhinav:
RE: coverage: The failing cases are the ones for Fatal. I think there's a way to organize this that addresses that issue. I'll add a commit, just a moment.

prashantv (approved):

PR #885, opened 2020-11-24
Add NumbersAsStrings to EncoderConfig

### LINKED ISSUE #884: Proposal: Add an "always encode numbers as strings" option to zapcore.EncoderConfig
The reason of this proposal is that in json the only number format defined is float64. so int64 numbers in json log might lose precision from the log ingestion. Take this log line as an example:

```json
{"level":"DEBUG","timestamp":"2020-11-24T20:50:50.393889219Z","caller":"testing/testing.go:1123","message":"test int64","int64":955216105955032459,"int64-field":955216105955032459}
```

The log ingestion vendor we use doesn't do any special handling on json numbers, so the int64 numbers becomes `955216105955032400` in their UI.

The proposal is to add an option to encode them automatically into:

```json
{"level":"DEBUG","timestamp":"2020-11-24T20:50:50.393889219Z","caller":"testing/testing.go:1123","message":"test int64","int64":"955216105955032459","int64-field":"955216105955032459"}
```

fishy:
This partially fixes https://github.com/uber-go/zap/issues/884.

This implementation is limited to top level fields, like:

    Int64("int64", 123)
    Float64("float64", 456)

Or sugared:

    With("int64", 123, "float64", float64(456), "slice", []int{1, 2, 3})

It does NOT work with reflected fields like:

    Reflect("slice", []int{1, 2, 3})
    With("struct", struct{ Int64 int64 }{Int64: 123})

As for reflected fields we use go's stdlib json encoder directly, which
doesn't provide such feature.

### DIFF
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -333,6 +333,13 @@ type EncoderConfig struct {
 	// Configures the field separator used by the console encoder. Defaults
 	// to tab.
 	ConsoleSeparator string `json:"consoleSeparator" yaml:"consoleSeparator"`
+	// If set to true, all numbers will be wrapped by double quotation marks.
+	// This is usally useful in JSON logging to make sure that you don't lose
+	// precision on number fields during log ingester's json decoding process.
+	// NOTE: currently this only works for top level fields, e.g.
+	// Int64("key", 123) or sugared With("key", 456). It doesn't work with
+	// reflected fields like Reflect("key", []int{1, 2, 3}).
+	NumbersAsStrings bool `json:"numbersAsStrings" yaml:"numbersAsStrings"`
 }
 
 // ObjectEncoder is a strongly-typed, encoding-agnostic interface for adding a
--- a/zapcore/json_encoder.go
+++ b/zapcore/json_encoder.go
@@ -248,7 +248,13 @@ func (enc *jsonEncoder) AppendDuration(val time.Duration) {
 
 func (enc *jsonEncoder) AppendInt64(val int64) {
 	enc.addElementSeparator()
+	if enc.EncoderConfig.NumbersAsStrings {
+		enc.buf.AppendByte('"')
+	}
 	enc.buf.AppendInt(val)
+	if enc.EncoderConfig.NumbersAsStrings {
+		enc.buf.AppendByte('"')
+	}
 }
 
 func (enc *jsonEncoder) AppendReflected(val interface{}) error {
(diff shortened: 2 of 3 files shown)


abhinav:
Thanks for the PR, @fishy. Although this is probably solvable externally with a
custom wrapper around the JSON encoder, I agree that this is something Zap
should ship a solution for.

However, I'm not sure that this is the API surface we'd want to expose for
this. Maybe we should provide a hook to control how numbers are encoded? Or if
we don't foresee the need to encode numbers differently besides these two
representations, we probably want it to take the form of an enum.

What do you think, @prashantv?

prashantv:
> Thanks for the PR, @fishy. Although this is probably solvable externally with a
> custom wrapper around the JSON encoder, I agree that this is something Zap
> should ship a solution for.
> 
> However, I'm not sure that this is the API surface we'd want to expose for
> this. Maybe we should provide a hook to control how numbers are encoded? Or if
> we don't foresee the need to encode numbers differently besides these two
> representations, we probably want it to take the form of an enum.
> 
> What do you think, @prashantv?

Agree, I'm not sure about the API surface. I also don't like that it only appleis in some scenarios such top-level fields, but not for reflected fields (or other nested fields).

If this is a general problem for JSON encoding, I'd expect to see an option in the Go standard library in the JSON encoder, and then consistently enable it for all encodings (top-level, nested, or reflected).

fishy:
So in JavaScript spec, there's only one type of number defined, which is 64-bit floats: https://www.w3schools.com/js/js_numbers.asp. JSON inherits that from JavaScript.

In reality, decoder/unmarshaler implementations can (and often choose to) extend that to higher precision. For example, in Go [`encoding/json.Decoder.UseNumbers`](https://golang.org/pkg/encoding/json/#Decoder.UseNumber) can be used to keep int64 precision when possible ([here is an example](https://play.golang.org/p/Yq9yUviDTMJ)).

But that extension is not part of the json specification so it's not always guaranteed (for example, the log ingestion vendor we use doesn't do that, and we can't blame them for that because it's not part of the spec).

The reason Go's json library provided this extension as an option is probably why they don't provide an option on the encoder side to encode numbers as strings.

I certainly agree that it would be great to be able to do it for reflected fields, but that's something a lot more complex (e.g. we probably need to either push go to provide that option, which is unlikely to happen, or switch to a different json encoding library for reflected fields that provides this feature), and this partial fix will cover the majority of the cases, so I don't think this partial fix should be blocked by the complete fix.

prashantv:
Sorry for the delay in getting back to you @fishy.

I'm still not sure zap should provide this specific option of encoding numbers as strings, but perhaps a more general extension mechanism could be used to implement this feature.

This use-case could be solved with an encoder-decorator, that can wrap the `AddInt*` and similar methods to encode them using any format (e.g., using `AddString`). This is technically possible today if you create an encoder, wrap it, then create your own core/logger, but it's a little bit painful.

We could provide options for wrapping encoders as part of construction that could simplify the above, what do you think of this approach?

fishy:
@prashantv thanks. we are currently working around it by wrapping the core: https://github.com/reddit/baseplate.go/blob/master/log/core_wrapper.go

if there's a new API to only wrap the encoder I think that would be better than wrapping the whole core.

fishy:
@prashantv is their an issue/PR I can follow for the progress of the new encoder-decorator api?

Issue #889, opened 2020-12-15
The string type does not automatically add spaces.


FH0:
```go
a := "a"
zap.S().Debug(a, a)
fmt.Println(a, a)
```

zap will print "aa"
fmt will print "a a"

johnaoss:
@FH0 The [SugaredLogger.Debug documentation states](https://pkg.go.dev/go.uber.org/zap#SugaredLogger.Debug) that `Debug` uses `fmt.Sprint` internally for creating strings. 

The behaviour of [fmt.Sprint](https://golang.org/pkg/fmt/#Sprint) is:
> Sprint formats using the default formats for its operands and returns the resulting string. **Spaces are added between operands when neither is a string**.

Whereas you show [fmt.Println](https://golang.org/pkg/fmt/#Println) which says:
> Println formats using the default formats for its operands and writes to standard output. **Spaces are always added between operands** and a newline is appended. It returns the number of bytes written and any write error encountered.

FH0:
There may be people like me who would choose another log library for this issue.

PR #894, opened 2020-12-18
change fmt.Sprint to fmt.Sprintln

FH0:
https://github.com/uber-go/zap/issues/889

### DIFF
--- a/sugar.go
+++ b/sugar.go
@@ -225,7 +225,8 @@ func (s *SugaredLogger) log(lvl zapcore.Level, template string, fmtArgs []interf
 	// Format with Sprint, Sprintf, or neither.
 	msg := template
 	if msg == "" && len(fmtArgs) > 0 {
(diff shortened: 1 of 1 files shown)


abhinav (changes requested):
This constitutes a breaking change. We strongly prefer to avoid a breaking change.

Perhaps we can add `*ln` variants of methods to the SugaredLogger: Infoln, Warnln, and so on.

alxn:
Also logging typically doesn't need to be terminated with a new line.

FH0:
> This constitutes a breaking change. We strongly prefer to avoid a breaking change.
> 
> Perhaps we can add `*ln` variants of methods to the SugaredLogger: Infoln, Warnln, and so on.

ok

FH0:
> Also logging typically doesn't need to be terminated with a new line.

The goal is to automatically add spaces. Check out the issue I mentioned.

PR #897, opened 2020-12-24
Add WithClock Option

### LINKED ISSUE #694: feature request : interface around time.Now
we use some clock testing packages like 
https://github.com/jonboulle/clockwork
that relies on the tested code not calling directly time.Now()
can we have some interface in the encoder configuration ?

for example TimeSource interface that implemented the time.Now(), and in the zap encoder configuration an option to put your own interface to supply time ?

ernado:
Allow passing current time source.

Fix #694

### DIFF
--- a/clock.go
+++ b/clock.go
@@ -31,6 +31,8 @@ type Clock interface {
 // systemClock implements default Clock that uses system time.
 type systemClock struct{}
 
+var _systemClock Clock = systemClock{}
+
 func (systemClock) Now() time.Time {
 	return time.Now()
 }
--- a/clock.go
+++ b/clock.go
@@ -22,9 +22,9 @@ package zap
 
 import "time"
 
-// Clock represents time source.
+// Clock is a source of time for logged entries.
 type Clock interface {
-	// Now returns current time.
+	// Now returns the current local time.
 	Now() time.Time
 }
 
--- a/clock.go
+++ b/clock.go
@@ -1,0 +1,36 @@
+// Copyright (c) 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zap
+
+import "time"
+
+// Clock represents time source.
+type Clock interface {
+	// Now returns current time.
+	Now() time.Time
+}
+
+// systemClock implements default Clock that uses system time.
+type systemClock struct{}
+
+func (systemClock) Now() time.Time {
+	return time.Now()
+}
--- a/clock_test.go
+++ b/clock_test.go
@@ -1,0 +1,46 @@
+// Copyright (c) 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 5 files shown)


time_source_test.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
abhinav, line 1:
```suggestion
// Copyright (c) 2020 Uber Technologies, Inc.
```

abhinav (commented):
Hello, and thanks for the contribution! This is a good
change, however, in the interest of future expansion,
perhaps the source of time should be an interface? And we
could call it Clock.

```
type Clock interface { Now() time.Time }

func WithClock(Clock) Option
```

The rationale behind that is that interfaces can be upcast,
so it lets us expand to more clock-based functionality in
the future.

For example, if we end up needing `time.Sleep`, we can add a
`Sleeper` interface with a sleep method and use that if the
`Clock` supports it.

    // hypothetical sleep code
    if sleeper, ok := logger.clock.(Sleeper); ok {
      sleeper.Sleep(t)
    }

ernado:
I like this approach even more :)
I've updated pull request to use `Clock` interface, thank you! 
Not sure about naming in comments, hope they are fine too.

clock.go
```diff
@@ -0,0 +1,36 @@
...
+	// Now returns current time.
+	Now() time.Time
+}
+
+// systemClock implements default Clock that uses system time.
+type systemClock struct{}
+
+func (systemClock) Now() time.Time {
+	return time.Now()
+}
```
abhinav, line 36: Converting a `struct` to an `interface` costs an allocation.
We can front-load this allocation once by declaring an unexported systemClock
instance with the Clock type and using that instead.

```suggestion
// systemClock implements default Clock that uses system time.
type systemClock struct{}

var _systemClock Clock = systemClock{}

func (systemClock) Now() time.Time {
	return time.Now()
}
```

logger.go
```diff
@@ -71,6 +72,7 @@ func New(core zapcore.Core, options ...Option) *Logger {
 		core:        core,
 		errorOutput: zapcore.Lock(os.Stderr),
 		addStack:    zapcore.FatalLevel + 1,
+		clock:       systemClock{},
```
abhinav, line 75:
```suggestion
		clock:       _systemClock,
```

clock.go
```diff
@@ -0,0 +1,36 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zap
+
+import "time"
+
+// Clock represents time source.
+type Clock interface {
+	// Now returns current time.
```
abhinav, line 27:
```suggestion
// Clock is a source of time for logged entries.
type Clock interface {
	// Now returns the current local time.
```

options.go
```diff
@@ -138,3 +138,10 @@ func OnFatal(action zapcore.CheckWriteAction) Option {
 		log.onFatal = action
 	})
 }
+
+// WithClock configures the Logger to use Clock while creating entries.
```
abhinav, line 142:
```suggestion
// WithClock specifies the clock used by the logger to determine the current
// time for logged entries. Defaults to the system clock with time.Now.
```

abhinav (approved):
LGTM besides minor suggestions and doc changes.

We still call `time.Now()` in CheckedEntry.Write when writing Zap-internal errors to ErrorOutput, but I think that's fine because this is a source of time for logged entries only.

CC @prashantv

clock_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+type constantClock struct {
+	now time.Time
+}
+
+func (c constantClock) Now() time.Time { return c.now }
+
+func TestWithClock(t *testing.T) {
+	date := time.Date(2077, 1, 23, 10, 15, 13, 441, time.UTC)
+	clock := constantClock{now: date}
```
abhinav, line 40: We can simplify this a bit.

```suggestion
type constantClock time.Time

func (c constantClock) Now() time.Time { return time.Time(c) }

func TestWithClock(t *testing.T) {
	date := time.Date(2077, 1, 23, 10, 15, 13, 441, time.UTC)
	clock := constantClock(date)
```

logger.go
```diff
@@ -307,7 +309,7 @@ func (log *Logger) check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
 	if log.addCaller {
 		frame, defined := getCallerFrame(log.callerSkip + callerSkipOffset)
 		if !defined {
-			fmt.Fprintf(log.errorOutput, "%v Logger.check error: failed to get caller\n", time.Now().UTC())
+			fmt.Fprintf(log.errorOutput, "%v Logger.check error: failed to get caller\n", log.clock.Now().UTC())
```
abhinav, line 312: I think if we're using the system clock for CheckedEntry.Write's error cases, we might want to do the same here.

```suggestion
			fmt.Fprintf(log.errorOutput, "%v Logger.check error: failed to get caller\n", time.Now().UTC())
```

logger.go
```diff
@@ -26,7 +26,6 @@ import (
 	"os"
 	"runtime"
 	"strings"
```
abhinav, line 28:
```suggestion
	"strings"
	"time"
```

logger.go
```diff
@@ -72,7 +72,7 @@ func New(core zapcore.Core, options ...Option) *Logger {
 		core:        core,
 		errorOutput: zapcore.Lock(os.Stderr),
 		addStack:    zapcore.FatalLevel + 1,
-		clock:       systemClock{},
+		clock:       _systemClock,
```
ernado, line 76: AFAIK there will be no allocation anyway

abhinav: Converting the `systemClock{}` struct to the `Clock` interface will cause an allocation.

We can verify this with escape analysis of the old version:

```
$ go build -gcflags=-m 2>&1 | grep logger.go
[...]
./logger.go:75:3: systemClock literal escapes to heap
[...]
```

The new version doesn't have that.

emretanriverdi (approved):

prashantv (commented):
Change looks good, although there's still calls to `time.Now()` that have not been updated,
```
> git grep 'time.Now' | grep '\.go:' | grep -v '_test\.go'
logger.go:              Time:       time.Now(),
logger.go:                      fmt.Fprintf(log.errorOutput, "%v Logger.check error: failed to get caller\n", time.Now().UTC())
zapcore/entry.go:                       fmt.Fprintf(ce.ErrorOutput, "%v Unsafe CheckedEntry re-use near Entry %+v.\n", time.Now(), ce.Entry)
zapcore/entry.go:                       fmt.Fprintf(ce.ErrorOutput, "%v write error: %v\n", time.Now(), err)
```

Can you check to see if there's any other cases of time being used that need to be updated?

We'll likely want to add a ticker for #782 as well (which we can do as a follow-up).

abhinav:
Discussed with @prashantv. For the error cases, we can use the timestamps from the entries. Incoming.

prashantv (approved):

PR #900, opened 2020-12-31
1. add SetCore(), Clone(), CloneWithNewCore() function

romberli:
1. add SetCore(), Clone(), CloneWithNewCore() function

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -250,6 +250,21 @@ func (log *Logger) Core() zapcore.Core {
 	return log.core
 }
 
+func (log *Logger) SetCore(core zapcore.Core) {
+	log.core = core
+}
+
+func (log *Logger) Clone() *Logger {
+	return log.clone()
+}
+
+func (log *Logger) CloneWithNewCore(core zapcore.Core) *Logger {
+	c := log.clone()
+	c.SetCore(core)
+
+	return c
+}
+
 func (log *Logger) clone() *Logger {
 	copy := *log
 	return &copy


logger.go
```diff
@@ -250,6 +250,21 @@ func (log *Logger) Core() zapcore.Core {
 	return log.core
 }
 
+func (log *Logger) SetCore(core zapcore.Core) {
+	log.core = core
```
prashantv, line 254: This is racy, since there may be concurrent calls to the logger that are accessing the core.

in general, I don't think mutability of the underlying core is something we should do

logger.go
```diff
@@ -250,6 +250,21 @@ func (log *Logger) Core() zapcore.Core {
...
+}
+
+func (log *Logger) Clone() *Logger {
```
prashantv, line 257: this can be achieved via `logger.With()` but generally isn't needed, as loggers are not mutable.

logger.go
```diff
@@ -250,6 +250,21 @@ func (log *Logger) Core() zapcore.Core {
...
+	return log.clone()
+}
+
+func (log *Logger) CloneWithNewCore(core zapcore.Core) *Logger {
```
prashantv, line 261: You can use `logger.WithOptions(zap.WrapCore(...))` to create a new logger with the core wrapped or replaced completely

prashantv (changes requested):
Can you please provide some additional context on what you are trying to do? Some of the new APIs are already possible (though the API may not be obvious), but generally we don't want loggers to be mutable after creation, and any changes can be made by creating a new logger.

romberli:
@prashantv, thanks for your guide, close this pr.

PR #903, opened 2021-01-08
http: support additional content type

### LINKED ISSUE #902: http: more human friendly request body
First of all, thank you for the great library.

We are using the [https://godoc.org/go.uber.org/zap#AtomicLevel.ServeHTTP](AtomicLevel.ServeHTTP) endpoint for setting the log levels dynamically.

For operators, this is a bit cumbersome, because they have to specify a valid JSON object.

By default, curl uses  `application/x-www-form-urlencoded` as the content type in a PUT request.
I think it would be nice to accept that as content type in `ServeHTTP` because it is a lot easier to type.
Following would be the command to set the logging level to debug:

```
curl -X PUT localhost:3000/log/level -d level=debug
```

If the content type is not `application/x-www-form-urlencoded` just keep the previous behavior, and expect a JSON payload.

If this is something that would be accepted, I'm be happy to work on this.
I'm just not sure if this qualifies as a breaking change or not.

Oncilla:
Support `application/x-www-form-urlencoded` as an additional content
type for `AtomicLevel.ServeHTTP`.

This is the default content type for `curl -X POST`.

With this change, interacting with the HTTP endpoint is a bit more
user friendly:

```
curl -X PUT localhost:8080/log/level -d level=debug
```

Additionally, the unit tests for the HTTP handler are transformed to
a table driven approach.

fixes #902

---
This change is [Reviewable](https://reviewable.io/reviews/uber-go/zap/903)

### DIFF
--- a/http_handler.go
+++ b/http_handler.go
@@ -44,13 +44,18 @@ import (
 //
 //    Content-Type: application/x-www-form-urlencoded
 //
-// With this content type, the request body is expected to be URL encoded like:
+// With this content type, the level can be provided through the request body or
+// a query parameter. The log level is URL encoded like:
 //
 //    level=debug
 //
-// This is the default content type for a curl PUT request. An example curl
-// request could look like this:
+// The request body takes precedence over the query parameter, if both are
+// specified.
 //
+// This content type is the default for a curl PUT request. Following are two
+// example curl requests that both set the logging level to debug.
+//
+//    curl -X PUT localhost:8080/log/level?level=debug
 //    curl -X PUT localhost:8080/log/level -d level=debug
 //
 // For any other content type, the payload is expected to be JSON encoded and
--- a/http_handler.go
+++ b/http_handler.go
@@ -69,60 +69,68 @@ func (lvl AtomicLevel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Error string `json:"error"`
 	}
 	type payload struct {
-		Level *zapcore.Level `json:"level"`
+		Level zapcore.Level `json:"level"`
 	}
 
 	enc := json.NewEncoder(w)
 
 	switch r.Method {
-
 	case http.MethodGet:
-		current := lvl.Level()
-		enc.Encode(payload{Level: &current})
-
+		enc.Encode(payload{Level: lvl.Level()})
 	case http.MethodPut:
-		requestedLvl, err := func(body io.Reader) (*zapcore.Level, error) {
-			switch r.Header.Get("Content-Type") {
-			case "application/x-www-form-urlencoded":
-				pld, err := ioutil.ReadAll(body)
-				if err != nil {
-					return nil, err
-				}
-				values, err := url.ParseQuery(string(pld))
-				if err != nil {
-					return nil, err
-				}
-				lvl := values.Get("level")
-				if lvl == "" {
-					return nil, fmt.Errorf("must specify logging level")
-				}
-				var l zapcore.Level
-				if err := l.UnmarshalText([]byte(lvl)); err != nil {
-					return nil, err
-				}
-				return &l, nil
-			default:
-				var pld payload
-				if err := json.NewDecoder(r.Body).Decode(&pld); err != nil {
-					return nil, fmt.Errorf("malformed request body: %v", err)
-				}
-				if pld.Level == nil {
-					return nil, fmt.Errorf("must specify logging level")
-				}
-				return pld.Level, nil
-			}
-		}(r.Body)
+		requestedLvl, err := lvl.decodePutRequest(r.Header.Get("Content-Type"), r.Body)
 		if err != nil {
 			w.WriteHeader(http.StatusBadRequest)
 			enc.Encode(errorResponse{Error: err.Error()})
 			return
 		}
-		lvl.SetLevel(*requestedLvl)
-		enc.Encode(payload{Level: requestedLvl})
+		lvl.SetLevel(requestedLvl)
+		enc.Encode(payload{Level: lvl.Level()})
 	default:
 		w.WriteHeader(http.StatusMethodNotAllowed)
 		enc.Encode(errorResponse{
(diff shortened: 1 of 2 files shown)


http_handler.go
```diff
@@ -53,25 +81,44 @@ func (lvl AtomicLevel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+				}
+				lvl := values.Get("level")
+				if lvl == "" {
+					return nil, fmt.Errorf("must specify logging level")
+				}
+				var l zapcore.Level
+				if err := l.UnmarshalText([]byte(lvl)); err != nil {
+					return nil, err
+				}
+				return &l, nil
```
abhinav, line 103: Optionally, this can be shortened.
```suggestion
				err := l.UnmarshalText([]byte(lvl))
				return &l, err
```

http_handler_test.go
```diff
@@ -22,110 +22,124 @@ package zap_test
...
-	assertCodeMethodNotAllowed(t, code)
-	assertJSONError(t, body)
+func TestAtomicLevelServeHTTP(t *testing.T) {
+	tests := map[string]struct {
+		Method        string
+		ContentType   string
+		Body          string
+		ExpectedCode  int
+		ExpectedLevel zapcore.Level
+	}{
```
abhinav, line 44: Thanks for making this a table test. If you wouldn't mind, we have a preference for slice-based table tests rather than maps:

```
tests :=  []struct {
  desc string
  method string
  contentType string
  ...
}{
  ...
}

for _, tt := range tests {
  ..
```

(See also https://github.com/uber-go/guide/blob/master/style.md#test-tables)

http_handler_test.go
```diff
@@ -22,110 +22,124 @@ package zap_test
...
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			lvl := zap.NewAtomicLevel()
+			lvl.SetLevel(zapcore.InfoLevel)
+
+			ts := httptest.NewServer(lvl)
+			defer ts.Close()
+
+			req, err := http.NewRequest(test.Method, ts.URL, strings.NewReader(test.Body))
```
abhinav, line 117: can we name the test server srv or server? 

```suggestion
			server := httptest.NewServer(lvl)
			defer server.Close()

			req, err := http.NewRequest(test.Method, server.URL, strings.NewReader(test.Body))
```

http_handler_test.go
```diff
@@ -22,110 +22,124 @@ package zap_test
...
+		},
+		"POST URL": {
+			Method:       http.MethodPost,
+			ExpectedCode: http.StatusMethodNotAllowed,
+			ContentType:  "application/x-www-form-urlencoded",
+			Body:         "level=warn",
+		},
+	}
+
+	for name, test := range tests {
```
abhinav, line 109: For table tests, consider renaming the test cases to `tt`.

  for _, tt := range tests {

http_handler.go
```diff
@@ -53,25 +81,44 @@ func (lvl AtomicLevel) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-			enc.Encode(errorResponse{Error: errmess})
+			enc.Encode(errorResponse{Error: err.Error()})
 			return
 		}
-
-		lvl.SetLevel(*req.Level)
-		enc.Encode(req)
-
+		lvl.SetLevel(*requestedLvl)
+		enc.Encode(payload{Level: requestedLvl})
```
abhinav, line 121: Do we want a JSON output even for non-JSON input?

Maybe this is okay and we can add a plain text output later if we need it and the client has an `Accepts: text/plain`?

CC @prashantv

abhinav (commented):
Thanks for the change! The core logic looks mostly good. I have some suggestions around style.

Oncilla:
Thanks for the swift feedback. I hope I get around to address the comments this week.

http_handler.go
Oncilla: I prefer being explicit here. Especially after switching to value type.

It is not immediately obvious whether the returned value is pre- or post-UnmarshalText without knowing that part of the spec by hart. (TBH I would need to look that up)

http_handler.go
Oncilla: If we want to return non-JSON in the response, I would definitively go for content type negotiation through the `Accepts` header.

In any case, I would vote for this to be part of a follow-up PR if it is desired.

http_handler_test.go
Oncilla: Done

http_handler.go
```diff
@@ -23,59 +23,114 @@ package zap
...
 			Error: "Only GET and PUT are supported.",
 		})
 	}
 }
+
+func (lvl AtomicLevel) decodePutRequest(contentType string, body io.Reader) (zapcore.Level, error) {
+	if contentType == "application/x-www-form-urlencoded" {
+		return lvl.decodePutURL(body)
+	}
+	return lvl.decodePutJSON(body)
```
abhinav, line 101: Minor: It looks like these methods could be top-level functions since they don't do anything with the AtomicLevel. I'm gonna change that.

abhinav (approved):

abhinav:
Sorry for the delay in reviewing. This looks great!
I'm going to get a second review on this before merging.
Thanks for the change!

prashantv (commented):
Change looks good, although one change we can make to simplify, but also provide an easier way to set the level: use `ParseForm()` and `FormValue(..)` to get the level field if it exists. It would simplify the code, while also allowing URL parameters to be used. Thoughts?

Oncilla:
@prashantv nice. Did not know that exists.

Should I also add tests for the URL params encoding?

prashantv:
> Should I also add tests for the URL params encoding?

Yes, that would be great if you could add a test for passing the parameter in a URL (and mention it in the docs). Thanks!

Oncilla:
Done.

prashantv (approved):
Thanks for the contribution!

PR #912, opened 2021-02-05
Support multi-field encoding using zap.InlineObject

### LINKED ISSUE #876: Add strict fields into log
Hey there,
Let's say if I have a struct

```
type logFormat struct {
	Level        string    `json:"level,omitempty"`
	ClientIP     string    `json:"client_ip,omitempty"`
	Method       string    `json:"method,omitempty"`
	Path         string    `json:"path,omitempty"`
}
```

if there a way add it to zap logger in flat without sub field?

I was able easy to do this:
```
{"level":"info","ts":1605346064.461162,"caller":"logger/logger.go:109","msg":"","log":{"level":"WARNING","client_ip":"::1","method":"GET","path":"/api"}}
```

with ```sugar.Infow("", zap.Reflect("log", log))```

but would be nice to construct something like this instead without using individual zap fields
```
{"level":"info","ts":1605346064.461162,"caller":"logger/logger.go:109","msg:"","client_ip":"::1","method":"GET","path":"/api"}
```

Thanks

prashantv:
Fixes #876

Currently, a `zap.Field` can only represent a single key-value. Add
`zap.InlineObject` so to allow adding multiple fields to the current
namespace from a type implementing `zap.ObjectMarshaler`.

This also solves a more general problem: a single `zap.Field` can now
be used to add multiple key/value pairs.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -165,6 +165,29 @@ func ExampleNamespace() {
 	// {"level":"info","msg":"tracked some metrics","metrics":{"counter":1}}
 }
 
+type request struct {
+	URL string
+	IP  string
+}
+
+func (r request) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	enc.AddString("url", r.URL)
+	enc.AddString("ip", r.IP)
+	return nil
+}
+
+func ExampleObject() {
+	logger := zap.NewExample()
+	defer logger.Sync()
+
+	req := &request{"/test", "127.0.0.1"}
+	logger.Info("new request, in nested object", zap.Object("req", req))
+	logger.Info("new request, inline", zap.InlineObject(req))
+	// Output:
+	// {"level":"info","msg":"new request, in nested object","req":{"url":"/test","ip":"127.0.0.1"}}
+	// {"level":"info","msg":"new request, inline","url":"/test","ip":"127.0.0.1"}
+}
+
 func ExampleNewStdLog() {
 	logger := zap.NewExample()
 	defer logger.Sync()
--- a/field.go
+++ b/field.go
@@ -400,6 +400,15 @@ func Object(key string, val zapcore.ObjectMarshaler) Field {
 	return Field{Key: key, Type: zapcore.ObjectMarshalerType, Interface: val}
 }
 
+// InlineObject is similar to Object, but does not nest the object under a field
+// name, but adds the fields to the current namespace inline.
+func InlineObject(val zapcore.ObjectMarshaler) Field {
+	return zapcore.Field{
+		Type:      zapcore.InlineObjectMarshalerType,
+		Interface: val,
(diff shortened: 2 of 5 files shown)


example_test.go
```diff
@@ -165,6 +165,29 @@ func ExampleNamespace() {
...
+func ExampleObject() {
+	logger := zap.NewExample()
+	defer logger.Sync()
+
+	req := &request{"/test", "127.0.0.1"}
+	logger.Info("new request, in nested object", zap.Object("req", req))
+	logger.Info("new request, inline", zap.InlineObject(req))
+	// Output:
+	// {"level":"info","msg":"new request, in nested object","req":{"url":"/test","ip":"127.0.0.1"}}
+	// {"level":"info","msg":"new request, inline","url":"/test","ip":"127.0.0.1"}
```
abhinav, line 188: I think we need a test somewhere to also verify how this behaves when used inside an object. Like,

```
func (foo) MarshalLogObject(enc zapcore.ObjectEncoder) error {
  InlineObject(...).AddTo(enc)
}
```

field.go
```diff
@@ -400,6 +400,15 @@ func Object(key string, val zapcore.ObjectMarshaler) Field {
 	return Field{Key: key, Type: zapcore.ObjectMarshalerType, Interface: val}
 }
 
+// InlineObject is similar to Object, but does not nest the object under a field
+// name, but adds the fields to the current namespace inline.
+func InlineObject(val zapcore.ObjectMarshaler) Field {
```
abhinav, line 405: Is there a different name we can use here? How do you feel about `zap.Inline`?

example_test.go
prashantv: Updated the example test, the `MarshalLogObject` now adds both an inline object as well as a separate object

field.go
prashantv: `Inline` is fine with me, I originally had `Object` in the name to match the interface we accept, but renamed everything to just `Inline`

field.go
```diff
@@ -400,6 +400,15 @@ func Object(key string, val zapcore.ObjectMarshaler) Field {
 	return Field{Key: key, Type: zapcore.ObjectMarshalerType, Interface: val}
 }
 
+// Inline is similar to Object, but does not nest the object under a field
+// name, but adds the fields to the current namespace inline.
```
abhinav, line 404: This sentence reads weird with two "but"s. How about,

    // Inline constructs a Field that is similar to Object, but it
    // will add the elements of the provided ObjectMarshaler to the
    // current namespace.

abhinav (approved):

field.go
prashantv: Didn't notice the two "but"'s, updated.

abhinav (approved):

PR #913, opened 2021-02-05
Optimize Sugar logger for calls with a single string arg

prashantv:
Currently, the Sugar logger uses fmt.Sprint in all cases when the
template is empty. However, this call is unnecessary if there's
a single string type argument, as we can use it directly.

With this optimization, we reduce the cost and avoid an unnecessary
alloc:
```
> benchcmp pre post
benchmark                         old ns/op     new ns/op     delta
BenchmarkSugarSingleStrArg-10     636           570           -10.38%

benchmark                         old allocs     new allocs     delta
BenchmarkSugarSingleStrArg-10     1              0              -100.00%
```

### DIFF
--- a/sugar.go
+++ b/sugar.go
@@ -222,17 +222,28 @@ func (s *SugaredLogger) log(lvl zapcore.Level, template string, fmtArgs []interf
 		return
 	}
 
-	// Format with Sprint, Sprintf, or neither.
-	msg := template
-	if msg == "" && len(fmtArgs) > 0 {
-		msg = fmt.Sprint(fmtArgs...)
-	} else if msg != "" && len(fmtArgs) > 0 {
-		msg = fmt.Sprintf(template, fmtArgs...)
-	}
-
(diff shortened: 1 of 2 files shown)


rabbbit (approved):

rabbbit:
>  However, this call is unnecessary if there's a single string type argument,

Why would anyone pass in no template, but a single string argument? zap.Info("", "hello")..?

abhinav (approved):

abhinav:
> Why would anyone pass in no template, but a single string argument? zap.Info("", "hello")..?

IIRC, internally Zap calls that function with an empty template for the `Sprint`-equivalent cases.

prashantv:
> > Why would anyone pass in no template, but a single string argument? zap.Info("", "hello")..?
> 
> IIRC, internally Zap calls that function with an empty template for the `Sprint`-equivalent cases.

Yep, exactly. `zap.Info("hello")` will actually go through this path (the first argument is the template if used with `Sprintf`).

PR #915, opened 2021-02-09
Remove outdated information, make dependencies is no longer necessary.

dherbst:
This commit https://github.com/uber-go/zap/commit/672336fffe89dc92675790ac2d8284be067b9328#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L21-L38 removed the `dependencies` target and this step is no longer necessary.


prashantv (approved):
LGTM, thanks for the cleanup

Issue #654, opened 2018-11-28
Support for logging contextual metadata


yurishkuro:
# Requirement

When my application is propagating `context.Context` object and that context include metadata about a distributed transaction, such as distributed tracing information (trace_id, span_id, etc.) from Jaeger or Zipkin, I want to be able to tag all logs with those metadata fields, similar to how MDC works in [Java loggers][https://logback.qos.ch/manual/mdc.html].

# What's preventing me from doing that in Zap

It is currently possible to do in Zap via `With([]Fields)` scoping of the Logger, however:
  * if it's done at the call sites, then it is easy to forget to do so
  * if the scoped Logger is created at the top of the request handler, then this logger needs to be passed through the application logic, which does not play well with dependency injection since many components tend to accept the logger only at initialization.

# Proposal

1. Create a new type of logger, e.g. `ContextLogger`, similar to `SugarLogger`, which has methods similar to the regular `Logger`, but always taking `context.Context` as the first argument.
   * Could also create `SugarContextLogger`
2. Add mechanism for optional `context.Context` encoding in the Entry/Encoders.

yurishkuro:
@akshayjshah @prashantv please let me know if this is an acceptable direction, and I can follow up with full PRs.

prashantv:
Rather than adding new types which are essentially duplicates of the current interface, I think I'd prefer a new smaller type -- it accepts a context and returns the existing logger types:
```go
type ContextLogger struct{...}

func (*ContextLogger) Ctx(context.Context) zap.Logger
```

We could have a duplicate that returns a `SugaredLogger`, or additional methods. 

It also allows the user to select when to pay the cost of serializing the context, rather than being forced to pay it at every log call.

```go
// if you only log a couple of times, you may want to take the cost of context serialization each time
ctxLogger.Ctx(ctx).Info(...)
ctxLogger.Ctx(ctx).Info(...)

// if you're logging many times, you can cache the serialized context in the logger
logger := ctxLogger.Ctx(ctx)
logger.Info(..)
logger.Info(..)
logger.Info(..)
logger.Warn(..)
logger.Error(..)
```

We can still add lint rules to ensure that if there's a `ctx` in scope, that the user uses a `ContextLogger`.

The `ContextLogger` type can also live outside of zap, nothing about it ties it to zap (it just needs to return a `zap.Logger` with fields with the context's contents).

yurishkuro:
[...]

PR #919, opened 2021-02-24
Add LoggerWithCtx

vmihailenco:
Updates https://github.com/uber-go/zap/issues/654

This PR adds `Entry.Ctx context.Context` field that allows associating a log entry with a `context.Context`. Which opens a possibility to send zap logs as OpenTelemetry events. See [example](https://github.com/uptrace/uptrace-go/tree/master/example/zap)

This is supposed to be used as this:

```go
logger, _ := zap.NewDevelopment()

func Zap(ctx context.Context) zap.LoggerWithCtx {
    return logger.Ctx(ctx)
}

func main() {
    ctx := context.Background()
    Zap(ctx).Error("hello", zap.String("foo", "bar"))
}
```

I've added a `LoggerWithCtx` wrapper so `logger.Ctx(ctx)` does not allocate. Alternatively `context.Context` can be added to the existing `*Logger`, but that would require cloning `*Logger` whenever context is set (and thus an allocation).


vmihailenco:
Gentle ping. Any thoughts on this?

Issue #918, opened 2021-02-18
AddFloat will remove the 0 at the end of the float


liyechen:
For example, here is a struct:
``` go
type Log struct {
	Afloat64 float64 `json:"a"`
}

func (e Log) MarshalLogObject(enc zapcore.ObjectEncoder) error {
	enc.AddFloat64("a", e.Afloat64)
	return nil
}

// some instance
logA := Log{
	Afloat64: 2.0
}
```
When using `zap.Logger` to record some log here,  the json will look like `{ "a": 2 }` instead of `{ "a": 2.0 }` because in function [AppendFloat](https://github.com/uber-go/zap/blob/master/buffer/buffer.go#L74) using `strconv.AppendFloat(b.bs, f, 'f', -1, bitSize)`, the parameter `-1` here will remove all 0s at the end of the float.

**Will you consider to add some functions to support like logger can log fix length of a float?**
For example:
```go
type Log struct {
	Afloat64 float64 `json:"a"`
}

func (e Log) MarshalLogObject(enc zapcore.ObjectEncoder) error {
	enc.AddFloat64WithFixLength("a", e.Afloat64, 5)
	return nil
}
```
The json will look like `{ "a": 2.00000 }`.

prashantv:
We can't add new methods to the `ObjectEncoder` interface in zap v1, as adding new methods to the interface would be a backwards-incompatible change (existing implementations would no longer implement the interface).

I think this logic could live in the encoder however. The encoder could have options on how much precision to maintain, and all floats encoded by zap would then inherit the same precision.

Would that approach work for your use-case?

liyechen:
> We can't add new methods to the `ObjectEncoder` interface in zap v1, as adding new methods to the interface would be a backwards-incompatible change (existing implementations would no longer implement the interface).
> 
> I think this logic could live in the encoder however. The encoder could have options on how much precision to maintain, and all floats encoded by zap would then inherit the same precision.
> 
> Would that approach work for your use-case?

Absolutely yes. Actually, options should be better. Appreciate that if this can be supported.

prashantv:
Great, we're open to patches implementing this functionality. I think the changes needed are:
 * Add a field on `EncoderConfig` to control the precision. We need to default to `-1` to match the existing behaviour
 * Add a new method to [buffer](https://pkg.go.dev/go.uber.org/zap/buffer#Buffer.AppendFloat) to append float with a specified precision
[...]

PR #923, opened 2021-02-25
add precision to EncoderConfig with default value as -1

liyechen:
This is my implementation of this [issue](https://github.com/uber-go/zap/issues/918). 
The problem is I can not control float in reflections.
I think if this needs to take effect in **reflections**, much more jobs need to be done.
Check this out and see if this can work.
Thanks a lot !
@prashantv

### DIFF
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -74,6 +74,12 @@ func (b *Buffer) AppendFloat(f float64, bitSize int) {
 	b.bs = strconv.AppendFloat(b.bs, f, 'f', -1, bitSize)
 }
 
+// AppendFloatWithPrecision appends a float with precision to the underlying buffer. It doesn't quote NaN
+// or +/- Inf.
+func (b *Buffer) AppendFloatWithPrecision(f float64, bitSize int, prec int) {
+	b.bs = strconv.AppendFloat(b.bs, f, 'f', prec, bitSize)
+}
+
 // Len returns the length of the underlying byte slice.
 func (b *Buffer) Len() int {
 	return len(b.bs)
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -31,6 +31,7 @@ import (
 // Alternate line endings specified in EncoderConfig can override this
 // behavior.
 const DefaultLineEnding = "\n"
+const DefaultPrecision = -1
 
 // OmitKey defines the key to use when callers want to remove a key from log output.
 const OmitKey = ""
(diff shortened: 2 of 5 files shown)


buffer/buffer.go
```diff
@@ -74,6 +74,12 @@ func (b *Buffer) AppendFloat(f float64, bitSize int) {
 	b.bs = strconv.AppendFloat(b.bs, f, 'f', -1, bitSize)
 }
 
+// AppendFloatWithPrecision appends a float with precision to the underlying buffer. It doesn't quote NaN
+// or +/- Inf.
+func (b *Buffer) AppendFloatWithPrecision(f float64, bitSize int, prec int) {
```
prashantv, line 79: `AppendFloat` takes `prec, bitSize int` as the last 2 arguments, let's do the same here for consistency

zapcore/encoder.go
```diff
@@ -333,6 +334,9 @@ type EncoderConfig struct {
 	// Configures the field separator used by the console encoder. Defaults
 	// to tab.
 	ConsoleSeparator string `json:"consoleSeparator" yaml:"consoleSeparator"`
+
+	// float precision
+	Precision int `json:"precision" yaml:"precision"`
```
prashantv, line 339: Should we make this more explicitly `FloatPrecision`?

zapcore/encoder.go
```diff
@@ -333,6 +334,9 @@ type EncoderConfig struct {
 	// Configures the field separator used by the console encoder. Defaults
 	// to tab.
 	ConsoleSeparator string `json:"consoleSeparator" yaml:"consoleSeparator"`
+
+	// float precision
```
prashantv, line 338: can we add more details for API docs

zapcore/json_encoder.go
```diff
@@ -82,6 +82,9 @@ func NewJSONEncoder(cfg EncoderConfig) Encoder {
 }
 
 func newJSONEncoder(cfg EncoderConfig, spaced bool) *jsonEncoder {
+	if cfg.Precision == 0 {
```
prashantv, line 85: What if users want precision = 0?

prashantv (commented):
> I think if this needs to take effect in reflections, much more jobs need to be done.

Good point, this is also a similar issue with https://github.com/uber-go/zap/pull/885

I'm not sure if the solution proposed there will work here, since there's no field value that will force any precision.

I think encoder config with a note that this only applies to non-reflect encoded floats is the only reasonable option.

cc @abhinav any thoughts?

zapcore/json_encoder.go
liyechen: I've changed other codes according to comments up there except this one.
Actually, if the user did not set `FloatPrecision`, the value of `FloatPrecision` here is 0, I can not judge if the user wants 0 or the default value.
Do you have some good suggestions?

PR #926, opened 2021-03-05
optimize sructs with memory alignment / padding

chemidy:
I think optimizing sructs with memory alignment / padding for a logger is interesting  : 

- buffer/buffer.go : struct with 32 pointer bytes to 16 bytes
- config.go : struct with 24 pointer bytes to 8 bytes
- internal/exit/exit.go : struct with 16 pointer bytes to 8 bytes
- internal/ztest/writer.go : struct with 56 pointer bytes to 32 bytes
- logger.go : struct with 96 pointer bytes to 72 bytes
- zapcore/entry.go : 
   - struct with 48 pointer bytes to 24 bytes
   - struct with 128 pointer bytes to 96 bytes
   - struct with 168 pointer bytes to 136 bytes
- zapcore/field.go : struct with 64 pointer bytes to 40 byes
- zapcore/write_syncer.go : struct with 24 pointer bytes to 16 bytes
- zapgrpc/zapgrpc.go : struct with 40 pointer bytes to 32 bytes

### DIFF
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -33,8 +33,8 @@ const _size = 1024 // by default, create 1 KiB buffers
 // Buffer is a thin wrapper around a byte slice. It's intended to be pooled, so
 // the only way to construct one is via a Pool.
 type Buffer struct {
-	bs   []byte
 	pool Pool
+	bs   []byte
 }
 
 // AppendByte writes a single byte to the Buffer.
--- a/config.go
+++ b/config.go
@@ -37,9 +37,9 @@ import (
 // Values configured here are per-second. See zapcore.NewSamplerWithOptions for
 // details.
 type SamplingConfig struct {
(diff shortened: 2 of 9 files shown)


prashantv:
How were these generated? I checked the first one listed,
> buffer/buffer.go : struct with 32 pointer bytes to 16 bytes

I was surprised to see a reduction, as it contains:
 * Pool struct (which itself contains a single pointer)
 * `[]byte` which is a pointer + 2 ints (see https://golang.org/pkg/reflect/#SliceHeader)

On a 64-bit architecture, I'd expect the pointers + ints to all be 8 bytes each, so the final size would be 32 bytes.  Reordering these fields won't help, and a quick check shows that the size of both is the same:
https://play.golang.org/p/jCWnhsVSXHi

Are you confident that these changes result in size reductions?

prashantv:
Closing as it's not clear that this change reduces size, how it was generated, and the questions above are unanswered.

PR #943, opened 2021-05-15
Expose additional filters for observed logs

jkanywhere:
There are various ways  I examine logs when testing how my library uses `zap`.
Filtering by log level seems generically useful.
Other filters may not be re-usable enough to include directly in `zap`, so expose the ability to
filter by an arbitrary function.

### DIFF
--- a/zaptest/observer/observer.go
+++ b/zaptest/observer/observer.go
@@ -80,21 +80,21 @@ func (o *ObservedLogs) AllUntimed() []LoggedEntry {
 
 // FilterMessage filters entries to those that have the specified message.
 func (o *ObservedLogs) FilterMessage(msg string) *ObservedLogs {
-	return o.filter(func(e LoggedEntry) bool {
+	return o.Filter(func(e LoggedEntry) bool {
 		return e.Message == msg
 	})
 }
 
 // FilterMessageSnippet filters entries to those that have a message containing the specified snippet.
 func (o *ObservedLogs) FilterMessageSnippet(snippet string) *ObservedLogs {
-	return o.filter(func(e LoggedEntry) bool {
+	return o.Filter(func(e LoggedEntry) bool {
 		return strings.Contains(e.Message, snippet)
 	})
 }
 
 // FilterField filters entries to those that have the specified field.
 func (o *ObservedLogs) FilterField(field zapcore.Field) *ObservedLogs {
-	return o.filter(func(e LoggedEntry) bool {
+	return o.Filter(func(e LoggedEntry) bool {
 		for _, ctxField := range e.Context {
 			if ctxField.Equals(field) {
 				return true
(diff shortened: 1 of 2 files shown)


zaptest/observer/observer.go
```diff
@@ -116,7 +123,9 @@ func (o *ObservedLogs) FilterFieldKey(key string) *ObservedLogs {
 	})
 }
 
-func (o *ObservedLogs) filter(match func(LoggedEntry) bool) *ObservedLogs {
+// Filter filters entries to those for which the provided function returns
+// true.
+func (o *ObservedLogs) Filter(match func(LoggedEntry) bool) *ObservedLogs {
```
abhinav, line 128: minor: consider renaming `match` to `keep` to make it clear that entries that this function returns true for will be kept in the returned ObservedLogs.

zaptest/observer/observer.go
```diff
@@ -116,7 +123,9 @@ func (o *ObservedLogs) FilterFieldKey(key string) *ObservedLogs {
 	})
 }
 
-func (o *ObservedLogs) filter(match func(LoggedEntry) bool) *ObservedLogs {
+// Filter filters entries to those for which the provided function returns
+// true.
```
abhinav, line 127: Probably worth clarifying that this returns a new ObservedLogs.

How about,

```suggestion
// Filter returns a copy of this ObservedLogs containing only those entries
// for which the provided function returns true.
```

abhinav (commented):
FilterLevel is an obvious yes.

Arbitrary Filter, I'm not opposed to exposing. @prashantv?

zaptest/observer/observer.go
```diff
@@ -78,23 +78,30 @@ func (o *ObservedLogs) AllUntimed() []LoggedEntry {
 	return ret
 }
 
+// FilterLevel filters entries to those logged at exactly the given level.
```
prashantv, line 81: optional: In other places in zap, level is the minimum level but it's an exact match.

To avoid confusion, should this be include "Exact" or "Same" in the name, `FilterLevelExact`?

prashantv (commented):
> FilterLevel is an obvious yes.
> 
> Arbitrary Filter, I'm not opposed to exposing. @prashantv?

I think it's useful, +1 for allowing arbitrary filter.

zaptest/observer/observer.go
jkanywhere: 👍 
I can imagine a future `FilterLevelAbove` or `FilterLevelAtLeast`.

zaptest/observer/observer_test.go
```diff
@@ -219,6 +227,24 @@ func TestFilters(t *testing.T) {
...
 			want:     logs[7:9],
 		},
+		{
+			msg: "filter by arbitrary function",
+			filtered: sink.Filter(func(e LoggedEntry) bool {
+				return len(e.Context) > 1
+			}),
+			want: func() []LoggedEntry {
+				// Do not modify logs slice.
+				w := []LoggedEntry{}
```
abhinav, line 237: nit: can use nil slice or a pre-allocated slice here:
```suggestion
				w := make([]LoggedEntry, 0, len(logs))
```

Or,

```suggestion
				var w []LoggedEntry
```

abhinav (approved):
LGTM w/ nit

prashantv (approved):

PR #948, opened 2021-05-21
Add NewTicker to the Clock Interface

moisesvega:
Changes in this diff: 

- Move Clock interface to zapcore and re-exported from the top-level zap package
- Add NewTicker method as part of Clock Interface

With this changes Clock interface will be available inside of zapcore and the NewTicker method will allow us to control time.Ticker when it is necessary. 

This is a stacked PR to addressed the comments on https://github.com/uber-go/zap/pull/782 . Next PR will be doing a refactor of WriteSyncer.


clock.go
```diff
@@ -20,13 +20,14 @@
...
+
+	"go.uber.org/zap/zapcore"
+)
 
 // Clock is a source of time for logged entries.
-type Clock interface {
-	// Now returns the current local time.
-	Now() time.Time
-}
+type Clock zapcore.Clock
```
abhinav, line 30: Do we need to keep the zap-level Clock type? We haven't shipped this yet so we can move this type to zapcore without breaking anything.

Also, if we do want to keep it around, we probably want to use a type alias.

```suggestion
type Clock = zapcore.Clock
```

The current form declares a new type Clock that is distinct from zapcore.Clock, while the type alias resolves to zapcore.Clock at compile time.

zapcore/clock.go
```diff
@@ -0,0 +1,34 @@
...
+	"time"
+)
+
+// Clock is a source of time for logged entries.
+type Clock interface {
+	// Now returns the current local time.
+	Now() time.Time
+	// NewTicker returns *time.Ticker that holds a channel
+	// that delivers ``ticks'' of a clock.
+	NewTicker(time.Duration) *time.Ticker
```
abhinav, line 33: Is this enough for a mock/test implementation? I think it would help to include a test implementation in this PR to understand how this gives us control over the Ticker.

clock_test.go
```diff
@@ -32,6 +32,9 @@ import (
 type constantClock time.Time
 
 func (c constantClock) Now() time.Time { return time.Time(c) }
+func (c constantClock) NewTicker(d time.Duration) *time.Ticker {
+	return time.NewTicker(d)
```
abhinav, line 36: Behaviorally, I would assume that a constantClock will never advance the ticker since time never moves.

zapcore/clock.go
moisesvega: I added a testcase that showcase this.

zapcore/clock_test.go
```diff
@@ -0,0 +1,61 @@
...
+type controlledClock struct {
+	mockClock *clock.Mock
+}
+
+func (c *controlledClock) Now() time.Time {
+	return c.mockClock.Now()
+}
+
+func (c *controlledClock) NewTicker(d time.Duration) *time.Ticker {
+	return &time.Ticker{C: c.mockClock.Ticker(d).C}
```
abhinav, line 41: This works but we can shorten this by embedding clock.Mock and only overriding NewTicker:

```
type controlledClock struct{
  *clock.Mock
}

func newControlledClock() *controlledClock

func (*controlledClock) NewTicker(...) *time.Ticker
```

zapcore/clock_test.go
```diff
@@ -0,0 +1,61 @@
...
+func TestMockClock(t *testing.T) {
+	var n int32
+	ctrlMock := &controlledClock{mockClock: clock.NewMock()}
+
+	// Create a channel to increment every microsecond.
+	go func() {
+		ticker := ctrlMock.NewTicker(1 * time.Microsecond)
+		for {
+			<-ticker.C
+			atomic.AddInt32(&n, 1)
```
abhinav, line 53: This is a goroutine leak. If we use go.uber.org/goleak here (which we do in #782).
We need to create a second `quit` channel that this `select`s on.

```
for {
  select {
  case <-quit:
    ticker.Stop()
    return
    
  case <-ticker.C:
    // ...
  }
}
```

This might also have to use a second channel to indicate that the goroutine has finished to wait for it to exit.

```
done := make(chan struct{})
go func() {
    defer close(done)
    // ...
}()
defer func() { <-done }() // wait for end
```

zapcore/clock_test.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}()
+	time.Sleep(1 * time.Microsecond)
```
abhinav, line 56: This shouldn't be necessary. My guess is that it's to wait for the ticker to start up.
We can probably get rid of this by starting the ticker right above the `go func()`, or as an argument to `go func`.

```
go func(ticker *time.Ticker) {
  // ...
}(ctrlMock.NewTicker(..))
```

zapcore/clock_test.go
```diff
@@ -0,0 +1,61 @@
...
+	return &time.Ticker{C: c.mockClock.Ticker(d).C}
+}
+
+func TestMockClock(t *testing.T) {
+	var n int32
+	ctrlMock := &controlledClock{mockClock: clock.NewMock()}
+
+	// Create a channel to increment every microsecond.
+	go func() {
+		ticker := ctrlMock.NewTicker(1 * time.Microsecond)
```
abhinav, line 50: code style: can be just `time.Microsecond` instead of `1 * time.Microsecond`.

zapcore/clock_test.go
```diff
@@ -0,0 +1,61 @@
...
+func (c *controlledClock) Now() time.Time {
+	return c.mockClock.Now()
+}
+
+func (c *controlledClock) NewTicker(d time.Duration) *time.Ticker {
+	return &time.Ticker{C: c.mockClock.Ticker(d).C}
+}
+
+func TestMockClock(t *testing.T) {
+	var n int32
```
abhinav, line 45: Instead of using `sync/atomic`, we should use [go.uber.org/atomic](https://pkg.go.dev/go.uber.org/atomic#Int32) here.


```
var n atomic.Int32

// ...

<-ticker.C
n.Inc()

// ...

assert.Equal(t, 42, n.Load())
```

See also https://github.com/uber-go/guide/blob/master/style.md#use-gouberorgatomic

zapcore/clock_test.go
```diff
@@ -0,0 +1,61 @@
...
+		ticker := ctrlMock.NewTicker(1 * time.Microsecond)
+		for {
+			<-ticker.C
+			atomic.AddInt32(&n, 1)
+		}
+	}()
+	time.Sleep(1 * time.Microsecond)
+
+	// Move clock forward.
+	ctrlMock.mockClock.Add(10 * time.Microsecond)
```
abhinav, line 59: If we embed clock.MockClock inside ctrlMock as suggested above, this `.mockClock` can be dropped.

zapcore/clock_test.go
```diff
@@ -0,0 +1,61 @@
...
+	assert.Equal(t, atomic.LoadInt32(&n), int32(10))
```
abhinav, line 60: For `assert.Equal`, the arguments are `(expected, got)`, not `(got, expected)`.

zapcore/clock_test.go
```diff
@@ -0,0 +1,65 @@
...
+			case <-ticker.C:
+				n.Inc()
+			}
+		}
+	}(ctrlMock.NewTicker(time.Microsecond))
+
+	// Move clock forward.
+	ctrlMock.Add(2 * time.Microsecond)
+	assert.Equal(t, int32(2), n.Load())
+	quit <- true
```
abhinav, line 64: For channels that'll only ever fire once, `close` the channel instead of posting to it. (See also [this](https://abhinavg.net/posts/fire-once-antipattern/).)

abhinav: Separately, please also see [this](https://github.com/uber-go/zap/pull/948#discussion_r637262837) comment.

We need to wait for the goroutine to actually exit once we tell it to stop before the test finishes.

Consider incorporating [goleak](https://github.com/uber-go/goleak) with a TestMain in zapcore to detect issues like this.

abhinav:
RE: lint failures: you need to run `go mod tidy` if this is a new depenedency.

abhinav (approved):

zapcore/clock.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"time"
+)
+
+// Clock is a source of time for logged entries.
+type Clock interface {
+	// Now returns the current local time.
+	Now() time.Time
+	// NewTicker returns *time.Ticker that holds a channel
+	// that delivers ``ticks'' of a clock.
```
mh-park, line 32: super nit: 
```suggestion
	// that delivers "ticks" of a clock.
```
Not sure what kind of quotes you wanted here, but the current ticks don't look correct. Ignore me if I'm wrong.

mh-park (approved):
LGTM!

clock.go
```diff
@@ -36,3 +37,7 @@ var _systemClock Clock = systemClock{}
 func (systemClock) Now() time.Time {
 	return time.Now()
 }
+
```
sywhang, line 40: Do these exported names not need comments?

moisesvega: We don't need comments here as the `systemClock` is not being exported, it is just a private variable satisfying an interface. 

Most likely I will move this systemClock to zapcore and exported it, if that happens I will add the proper comments for the exported type.

sywhang: Ah, I see. Thanks for the clarification!

sywhang (approved):

clock_test.go
```diff
@@ -42,3 +46,31 @@ func TestWithClock(t *testing.T) {
...
+	done := make(chan struct{})
+	defer func() {
+		<-done
+		assert.Equal(t, target, n.Load())
+	}() // wait for end
+
+	quit := make(chan struct{})
+	// Create a channel to increment every millisecond.
+	go func(ticker *time.Ticker) {
+		defer close(done)
```
abhinav, line 62: Do we need this to be a separate goroutine?
For this one, since we're blocking until three ticks, we can just do this all inline. It would simplify book keeping and it doesn't need to be an atomic because there would be no concurrency:

```
target := 3
var n int

ticker := _systemClock.NewTicker(..)
for range ticker.C {
  n++
  if n == target {
    ticker.Stop()
    return
  }
}
```

(We don't need the `assert.Equal` either because if the logic is faulty, the function will just loop forever.)

PR #949, opened 2021-05-22
Don't panic in `sweetenFields`

### LINKED ISSUE #896: Panic on wrong number of key value pairs
I forgot to add a key to a value on 
`func (s *SugaredLogger) Warn(msg string, keysAndValues ...interface{}) {`
And it caused me a panic crash of the app. 

Shouldn´t be wise to protect this error from happening?
The code compiles normally, you deploy on production... it stays there for months.. then the log is printed and the system crashes because a log was wrong. (not my case just hypothetical.)

This is a thing that can be protected. We always have the human factor to account for. 

https://github.com/uber-go/zap/blob/a68efdbdd15b7816de33cdbe7e6def2a559bdf64/sugar.go#L264

atrn0:
Hello team 👋.
This PR fixes #896.

### DIFF
--- a/sugar.go
+++ b/sugar.go
@@ -266,7 +266,7 @@ func (s *SugaredLogger) sweetenFields(args []interface{}) []Field {
 
 		// Make sure this element isn't a dangling key.
 		if i == len(args)-1 {
-			s.base.DPanic(_oddNumberErrMsg, Any("ignored", args[i]))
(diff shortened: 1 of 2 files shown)


sywhang (approved):
👍 This LGTM but I'd like to wait for @abhinav to go over it since he has more context on this issue.

abhinav (approved):
LGTM, thanks!

PR #950, opened 2021-05-25
clock: Move to zapcore

abhinav:
We don't need to export the Clock interface from the top-level Zap
package. This is a low-level API so we can keep it in zapcore only.


moisesvega (approved):
👍

PR #952, opened 2021-05-25
Add buffered write syncer

moisesvega:
This diff addresses comments in https://github.com/uber-go/zap/pull/782 

- Add BufferedWriteSyncer
- Use new Clock Interface for testing 

API for BufferedWriteSyncer:

```
type BufferedWriteSyncer struct {
	WriteSyncer

	Size          int
	FlushInterval time.Duration
	Clock         Clock

	// unexported fields for state
}

buf := &bytes.Buffer{}
ws := &BufferedWriteSyncer{WriteSyncer: AddSync(buf)} //  BufferedWriteSyncer with the default values
ws.Write([]byte("foo"))

 // Close closes the buffer, cleans up background goroutines, and flushes remaining, unwritten data.
ws.Close()
```


zapcore/buffered_write_syncer_bench_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func BenchmarkBufferedWriteSyncer(b *testing.B) {
+	b.Run("write file with buffer", func(b *testing.B) {
+		file, err := ioutil.TempFile("", "log")
+		assert.NoError(b, err)
+		defer file.Close()
+		defer os.Remove(file.Name())
```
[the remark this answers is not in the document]
moisesvega: Looks like there are some issues with this `b.TempDir()` on go 1.15 there is a know issue https://github.com/golang/go/issues/41062 about this, this issue was solved on 1.16 but exist in 1.15.

For now I will use ioutil as a workaround.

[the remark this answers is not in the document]
abhinav: Oh fun. Nice catch

zapcore/write_syncer_bench_test.go
```diff
@@ -39,7 +42,7 @@ func BenchmarkMultiWriteSyncer(b *testing.B) {
 			}
 		})
 	})
-	b.Run("4", func(b *testing.B) {
+	b.Run("4 discarder", func(b *testing.B) {
```
shirchen, line 45: This looks like a 2 discarder? Any difference between here and on line 33?

zapcore/write_syncer_bench_test.go
```diff
@@ -53,4 +56,38 @@ func BenchmarkMultiWriteSyncer(b *testing.B) {
...
+				w.Write([]byte("foobarbazbabble"))
+			}
+		})
+	})
+}
+
+func BenchmarkWriteSyncer(b *testing.B) {
+	b.Run("write file with no buffer", func(b *testing.B) {
+		file, err := ioutil.TempFile("", "log")
+		assert.NoError(b, err)
```
shirchen, line 81: any reason for catching errors before defering the file removal? Do you even need to manually Close or does os.Remove do that for you?

zapcore/buffered_write_syncer_bench_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func BenchmarkBufferedWriteSyncer(b *testing.B) {
+	b.Run("write file with buffer", func(b *testing.B) {
+		file, err := ioutil.TempFile("", "log")
+		require.NoError(b, err)
```
shirchen, line 34: same as below, maybe you need to defer before checking for error? dunno if file can get created if TempFile errors. can it?

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if s.Clock == nil {
+		s.Clock = DefaultClock
+	}
+	s.ticker = s.Clock.NewTicker(flushInterval)
+
+	s.writer = bufio.NewWriterSize(s.WriteSyncer, size)
+	s.stop = make(chan struct{})
+	s.initialized = true
+	go s.flushLoop()
```
shirchen, line 75: dunno, i recall having discussions with @abhinav about this before, but i think i prefer if go routines start not in loadConfig but below, in Write function. However, I may be wrong here.

zapcore/write_syncer_bench_test.go
moisesvega: This benchmark already exist in master branch. The idea to add this text to provide more context in the benchmark. 

Here the `4 discarder` refers that we are benchmarking `NewMultiwriterSyncer` with 4 `ztest.Discarder{}` 

With the new `BufferedWriteSyncer` there is a need to make a distinction to make sure the benchmark shows the proper information between benchmarks. 
```
BenchmarkMultiWriteSyncer/2_discarder-4         	41213204	        66.5 ns/op	      16 B/op	       1 allocs/op
BenchmarkMultiWriteSyncer/4_discarder-4         	19545279	        72.0 ns/op	      16 B/op	       1 allocs/op
BenchmarkMultiWriteSyncer/4_discarder_with_buffer-4         	 8759560	       143 ns/op	      16 B/op	       1 allocs/op
```

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+// A BufferedWriteSyncer is a WriteSyncer that can also flush any buffered data
+// with the ability to change the buffer size, flush interval and Clock.
+// The default values are; Size 256kb, FlushInterval 30s.
+type BufferedWriteSyncer struct {
+	WriteSyncer
+
+	Size          int
+	FlushInterval time.Duration
```
abhinav, line 36: minor nit: we should specify the defaults above the fields as we document the fields:

```
// Size specifies the maximum amount of data the writer will buffer before
// flushing. Defaults to 256 kB.

// FlushInterval specifies how often the writer should flush data if there
// have been no writes. Defaults to 30 seconds.
```

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,130 @@
...
+	Clock         Clock
```
abhinav, line 37: let's document this too

```
// Clock, if specified, provides control of the source of time for the
// writer. Uses the system clock by default.
```

zapcore/buffered_write_syncer.go
abhinav: In this case, loadConfig is like our constructor. Compare the following two:

```
func NewFoo(...) *Foo { /* initialize internals */ }

func (*Foo) DoThing() { ... }

// versus

type Foo struct{}

func (*Foo) loadConfig() { /* initialize internals */ }

func (f *Foo) DoThing() {
  if !f.loaded { f.loadConfig() }
  ...
}
```

So loadConfig is effectively doing the job of the constructor, in which case it makes sense to spawn the goroutine there because all initialization should happen in the same place. (Spawning the goroutine is part of initialization here because there will only ever be one goroutine per instance of this object, tied to that object. If we were spawning goroutines on a per-request basis, then that would be in Write.)

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,130 @@
...
+const (
+	// _defaultBufferSize specifies the default size used by Buffer.
+	_defaultBufferSize = 256 * 1024 // 256 kB
+
+	// _defaultFlushInterval specifies the default flush interval for
+	// Buffer.
+	_defaultFlushInterval = 30 * time.Second
+)
+
+func (s *BufferedWriteSyncer) loadConfig() {
```
abhinav, line 56: minor naming: can we name this init or initialize because the bool is called initialize. Or we can name the bool "loaded" and call this "load", but I'm more in favor of init or initialize.

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,130 @@
...
+	if flushInterval == 0 {
+		flushInterval = _defaultFlushInterval
+	}
+
+	if s.Clock == nil {
+		s.Clock = DefaultClock
+	}
+	s.ticker = s.Clock.NewTicker(flushInterval)
+
+	s.writer = bufio.NewWriterSize(s.WriteSyncer, size)
```
abhinav, line 72: minor optimization on the WriteSyncer:
since BufferedWriteSyncer is already thread-safe, we can optimize this a bit if s.WriteSyncer was wrapped with [`zapcore.Lock`](https://github.com/uber-go/zap/blob/v1.17.0/zapcore/write_syncer.go#L56).

```
writer := s.WriteSyncer
if w, ok := writer.(*lockedWriteSyncer); ok {
  writer = w.ws // don't double lock
}
s.writer = bufio.NewWriterSize(writer, size)
```

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,130 @@
...
+package zapcore
+
+import (
+	"bufio"
+	"sync"
+	"time"
+)
+
+// A BufferedWriteSyncer is a WriteSyncer that can also flush any buffered data
+// with the ability to change the buffer size, flush interval and Clock.
```
abhinav, line 30: We should probably mention that the WriteSyncer is safe for concurrent use.

```
// BufferedWriteSyncer is safe for concurrent use. You don't need to use
// zapcore.Lock for WriteSyncers with BufferedWriteSyncer.
```

zapcore/buffered_write_syncer_bench_test.go
abhinav: Na, you shouldn't defer until after the check. Consider,

```
foo, err := acquireSomeResource()
defer foo.Close()
if err != nil { return err }
```

The contract for cases like this is that `foo` is unusable and/or invalid if `err != nil`. So if you `defer` first, then it'll try to run `foo.Close()` when you return, but since `err != nil`, `foo` was invalid so `foo.Close` will probably fail -- might even panic.

abhinav (commented):
Sorry for the delay. I'll look at the next iteration quicker!

zapcore/buffered_write_syncer_bench_test.go
shirchen: sweet, im totally off here, thanks for TIL.

zapcore/write_syncer_bench_test.go
shirchen: Sorry, what i meant is that i dont see a difference between benchmarks on 33:44 and 45:58. They look the same to me.

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,131 @@
...
+		if err := s.writer.Flush(); err != nil {
+			return 0, err
+		}
+	}
+
+	return s.writer.Write(bs)
+}
+
+// Sync flushes buffered log data into disk directly.
+func (s *BufferedWriteSyncer) Sync() error {
```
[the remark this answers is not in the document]
moisesvega: Now that we have a reference to the wrapped `WriteSyncer`, Should we enforce `WriteSyncer` to not be nil at initialization?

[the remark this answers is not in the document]
abhinav: If we validate non-nil, will we error on Write, or panic?

FWIW, comparable APIs like [LimitedReader](https://golang.org/src/io/io.go?s=15635:15692#L456) just assume you gave them a non-nil value, and let it panic if not set.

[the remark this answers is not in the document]
moisesvega: Sounds reasonable as the outcome won't change even if we check for non-nil values. It will panic if not set.

zapcore/write_syncer_bench_test.go
moisesvega: These test are practically the same just with more `ztest.Discarder{}`. For context this test exist since 2017

shirchen: derp, the last 2 are hidden in the review. sorry, it took me a minute to see that they get cut off in the code review. carry on.

zapcore/buffered_write_syncer_test.go
```diff
@@ -0,0 +1,117 @@
...
+		requireWriteWorks(t, ws)
+		assert.Equal(t, "", buf.String(), "Unexpected log calling a no-op Write method.")
+		requireWriteWorks(t, ws)
+		assert.Equal(t, "foo", buf.String(), "Unexpected log string")
+		assert.NoError(t, ws.Close())
+	})
+
+	t.Run("with lockedWriteSyncer", func(t *testing.T) {
+		buf := &bytes.Buffer{}
+		ws := &BufferedWriteSyncer{WriteSyncer: &lockedWriteSyncer{ws: AddSync(buf)}, Size: 5}
```
abhinav, line 80: For the sake of hiding internal details, we should probably use `zapcore.Lock` here to build the lockedWriteSyncer.
```suggestion
		ws := &BufferedWriteSyncer{WriteSyncer: Lock(AddSync(buf)), Size: 5}
```

abhinav (commented):
Mostly LGTM. For the wrap-twice case, can we add a test with an explicit sync to verify that the inner BufWriter was also flushed?

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// Write writes log data into buffer syncer directly, multiple Write calls will be batched,
+// and log data will be flushed to disk when the buffer is full or periodically.
+func (s *BufferedWriteSyncer) Write(bs []byte) (int, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if !s.initialized {
+		s.initialize()
```
shirchen, line 102: This is a bool, so having this check in the hot path should be cheap, so this is fine. In any case, it's too late to redesign the API.

zapcore/buffered_write_syncer_bench_test.go
```diff
@@ -0,0 +1,49 @@
...
+		defer file.Close()
```
shirchen, line 35: I think this can return an error which we need to catch. Also in example in https://golang.org/pkg/io/ioutil/#TempFile i dont think they manually close the file.

shirchen (commented):
lgtm, just nits about closing files in tests and not needing to manually closing files.

zapcore/buffered_write_syncer_bench_test.go
abhinav: Which example are you talking about? Both examples linked there close the files at the end; they just don't use defer.
ioutil.TempFile returns an open file so we always need to close it.

RE: error: yeah, we can/should handle that. We can probably combine the two defers into a single one with two NoErrors:

```
defer func() {
  assert.NoError(t, file.Close())
  assert.NoError(t, os.Remove(file.Name())
}()
```

shirchen: Question is whether a separate explicit `file.Close` is needed or `os.Remove(file.Name())` is necessary. When @moisesvega experimented, it looked like Removing will close the file for you so a separate Close statement may be redundant and unnecessary.

abhinav: I would think that from a hygiene perspective, we'd always want to close the file descriptor once we were done with it regardless of what we were doing with the file afterwards. We shouldn't leave the file descriptor open when we get around to deleting the file; that just seems icky.

zapcore/buffered_write_syncer_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	t.Run("wrap twice", func(t *testing.T) {
+		buf := &bytes.Buffer{}
+		bufsync := &BufferedWriteSyncer{WriteSyncer: AddSync(buf)}
+		ws := &BufferedWriteSyncer{WriteSyncer: bufsync}
+		requireWriteWorks(t, ws)
+		assert.Equal(t, "", buf.String(), "Unexpected log calling a no-op Write method.")
+		assert.NoError(t, ws.Close())
+		assert.NoError(t, bufsync.Close())
+		assert.Equal(t, "foo", buf.String(), "Unexpected log string")
```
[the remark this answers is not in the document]
moisesvega: 👍

zapcore/buffered_write_syncer_bench_test.go
moisesvega: The file descriptor does not close automatically after removing the file. If we don't close the file descriptor we would rely on the Garbage Collector to close it. This could lead to a file descriptor leak. Best practice here is run `file.Close` then `os.Remove`.

I will add the asserts inside of a defer function just to make sure that `close` and `remove` run without errors.

shirchen: thanks!

shirchen (approved):
lgtm but wait for @abhinav 's stamp too.

abhinav (approved):

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+}
+
+// Close closes the buffer, cleans up background goroutines, and flushes
+// remaining, unwritten data. This will not close the underlying WriteSyncer.
+func (s *BufferedWriteSyncer) Close() error {
+	s.ticker.Stop()
+	close(s.stop)
+	return s.Sync()
```
prashantv, line 146: Can we wait for the background goroutine to stop before returning here? It leads to:
 * no goroutine running after `Close` guaranteed
 * avoids odd cases where a `Sync` can happen *after* the `Close` returns

abhinav:
No, we don't want a merge commit for this. Reverting.

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,147 @@
...
+			return
+		}
+	}
+}
+
+// Close closes the buffer, cleans up background goroutines, and flushes
+// remaining, unwritten data. This will not close the underlying WriteSyncer.
+func (s *BufferedWriteSyncer) Close() error {
+	s.ticker.Stop()
+	close(s.stop)
```
prashantv, line 145: Since this will panic if we do it twice, do we want to explicitly call out that `Close` must only be called once, or put some protection here in case `Close` is called twice?

zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,147 @@
...
+			// and we return any error from Sync() as part of the close
+			_ = s.Sync()
+		case <-s.stop:
+			return
+		}
+	}
+}
+
+// Close closes the buffer, cleans up background goroutines, and flushes
+// remaining, unwritten data. This will not close the underlying WriteSyncer.
```
prashantv, line 142: The warning that this will not close the underlying `WriteSyncer` makes me wonder if `Close` is a little surprising. Would this be better named something else like `Stop` to make it very clear that it doesn't close the underlying writer?

abhinav:
Reopened in #961  @prashantv

zapcore/buffered_write_syncer.go
abhinav: Reasonable. Renaming.

moisesvega: Sounds good. `Stop` makes more sense in the context of this specific writer.

PR #959, opened 2021-06-04
Add support for closing namespaces

fxposter:


prashantv:
Since this change modifies an interface, it is not backwards compatible, and so we cannot accept it for zap 1.0.

It may help to explain your use-case so we can suggest alternatives, or add this to the 2.0 wishlist.

prashantv:
Closing as this is not something we can accept for zap 1.0.

PR #961, opened 2021-06-08
zapcore: Add Buffered Writer

abhinav:
Add a BufferedWriteSyncer that buffers writes in memory before flushing
them to the underlying WriteSyncer at some interval or when a
configured amount of data has been buffered -- whichever comes first.

---

This is a copy of #952 (which itself is a fork of #782) because it was
accidentally merged instead of squashed, and wasn't yet fully reviewed.


zapcore/buffered_write_syncer.go
```diff
@@ -0,0 +1,187 @@
...
+	s.ticker = s.Clock.NewTicker(flushInterval)
+
+	// Unpack to the underlying WriteSyncer if we have a lockedWriteSyncer
+	// to avoid double-locking. Note that there's a risk here if the user
+	// tries to use the Lock-ed WriteSyncer directly in addition to using
+	// it with BufferedWriteSyncer, so we declare that the wrapped
+	// WriteSyncer is only ours to use once wrapped.
+	if w, ok := s.WS.(*lockedWriteSyncer); ok {
+		s.WS = w.ws
+	}
```
abhinav, line 104: I'm unsure about this bit. Maybe we shouldn't unpack the lock? @prashantv @moisesvega WDYT?

moisesvega: I am in favor of not unpacking `lockedWriteSyncer` and let the users decide if they need to unpack `lockedWriteSyncer` or any typer of `WriteSyncer` for their use case.

akshayjshah:
This is awesome! Glad to see that buffering finally made it in :)

PR #971, opened 2021-06-24
Add zapio.Writer

### LINKED ISSUE #929: Possible to set exec.Command Stdout and Stderr to zap?
Is it possible to use a zap.Logger for anything that provides its own Stdout / Stderr?

Given something like this:
```
cmd := exec.Command("echo", "hello world")
cmd.Stdout = zap.LoggerSomething
cmd.Stderr = zap.LoggerSomething
cmd.Run()
```

It doesn't look like it's possible to fetch the io.Writer from zap, but wondering if its possible another way?

abhinav:
This adds a new zapio package that provides a `Writer`. The writer
implements `io.WriteCloser` and `zapcore.WriteSyncer`.

It works by splitting the input on newlines, flushing to the logger as
new lines are encountered, and buffering input otherwise.

So for example, if write "foobar\n" is split across multiple Write calls
"foo" and "bar\n", instead of emitting two separate logs for "foo"
and "bar", the Writer will buffer the input until the newline is
encountered and write a single log for "foobar".

Performance:

```
name             time/op
Writer/single-4  384ns ± 3%
Writer/splits-4  488ns ±26%

name             alloc/op
Writer/single-4  16.0B ± 0%
Writer/splits-4  16.0B ± 0%

name             allocs/op
Writer/single-4   2.00 ± 0%
Writer/splits-4   2.00 ± 0%
```

Resolves #929

### DIFF
--- a/zapio/writer.go
+++ b/zapio/writer.go
@@ -1,0 +1,108 @@
+// Copyright (c) 2021 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package zapio
+
+import (
+	"bytes"
+	"io"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// Writer is an io.Writer that writes to the provided Zap logger, splitting log
+// messages on line boundaries.
+//
+// Writer must be closed when finished to flush buffered data to the logger.
+type Writer struct {
+	Log   *zap.Logger   // log to write to
+	Level zapcore.Level // log level to write at
+
+	buff bytes.Buffer
+}
+
+var (
+	_ zapcore.WriteSyncer = (*Writer)(nil)
+	_ io.Closer           = (*Writer)(nil)
+)
+
+// Write writes the provided bytes to the underlying logger at the configured
+// log level and returns the length of the bytes.
+func (w *Writer) Write(bs []byte) (n int, err error) {
+	// Skip all checks if the level isn't enabled.
+	if !w.Log.Core().Enabled(w.Level) {
+		return len(bs), nil
+	}
+
+	n = len(bs)
+	for len(bs) > 0 {
+		bs = w.writeLine(bs)
+	}
+
+	return n, nil
+}
+
+// writeLine writes a single line from the input, returning the remaining,
+// unconsumed bytes.
+func (w *Writer) writeLine(line []byte) (remaining []byte) {
+	idx := bytes.IndexByte(line, '\n')
+	if idx < 0 {
+		// If there are no newlines, buffer the entire string.
+		w.buff.Write(line)
+		return nil
+	}
+
+	// Split on the newline, buffer and flush the left.
+	line, remaining = line[:idx], line[idx+1:]
+	w.buff.Write(line)
+
+	// Log empty messages in the middle of the stream so that we don't lose
+	// information when the user writes "foo\n\nbar".
+	w.flush(true /* allowEmpty */)
+
+	return remaining
+}
+
+// Close closes the writer, flushing any buffered data in the process.
+func (w *Writer) Close() error {
+	return w.Sync()
+}
+
+// Sync flushes the buffered data from the writer, even if it doesn't end with
+// a newline.
+func (w *Writer) Sync() error {
+	// Don't allow empty messages on explicit Sync calls or on Close
+	// because we don't want an extraneous empty message at the end of the
+	// stream -- it's common for files to end with a newline.
+	w.flush(false /* allowEmpty */)
+	return nil
+}
+
+// flush flushes the buffered data to the logger, allowing empty messages only
+// if the bool is set.
+func (w *Writer) flush(allowEmpty bool) {
+	if allowEmpty || w.buff.Len() > 0 {
+		if ce := w.Log.Check(w.Level, w.buff.String()); ce != nil {
+			ce.Write()
+		}
+	}
+	w.buff.Reset()
+}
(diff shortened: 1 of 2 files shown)


zapio/writer.go
```diff
@@ -0,0 +1,108 @@
...
+	_ io.Closer           = (*Writer)(nil)
+)
+
+// Write writes the provided bytes to the underlying logger at the configured
+// log level and returns the length of the bytes.
+func (w *Writer) Write(bs []byte) (n int, err error) {
+	// Skip all checks if the level isn't enabled.
+	if !w.Log.Core().Enabled(w.Level) {
+		return len(bs), nil
+	}
```
prashantv, line 53: (no action needed, just an observation) If we used a constructor, we could do this check once.

I don't think this is going to affect performance in a meaningful way so not worth changing to constructors for this reason though.

zapio/writer.go
```diff
@@ -0,0 +1,108 @@
...
+	idx := bytes.IndexByte(line, '\n')
+	if idx < 0 {
+		// If there are no newlines, buffer the entire string.
+		w.buff.Write(line)
+		return nil
+	}
+
+	// Split on the newline, buffer and flush the left.
+	line, remaining = line[:idx], line[idx+1:]
+	w.buff.Write(line)
```
prashantv, line 75: optional performance optimization:
if the buffer is empty, we could write the contents directly to the logger without copying it to `w.buff`

zapio/writer.go
```diff
@@ -0,0 +1,108 @@
...
+	// stream -- it's common for files to end with a newline.
+	w.flush(false /* allowEmpty */)
+	return nil
+}
+
+// flush flushes the buffered data to the logger, allowing empty messages only
+// if the bool is set.
+func (w *Writer) flush(allowEmpty bool) {
+	if allowEmpty || w.buff.Len() > 0 {
+		if ce := w.Log.Check(w.Level, w.buff.String()); ce != nil {
```
prashantv, line 103: can we save some allocations using a `strings.Builder` instead of a `bytes.Buffer`?

zapio/writer.go
```diff
@@ -0,0 +1,108 @@
...
+	_ zapcore.WriteSyncer = (*Writer)(nil)
+	_ io.Closer           = (*Writer)(nil)
+)
+
+// Write writes the provided bytes to the underlying logger at the configured
+// log level and returns the length of the bytes.
+func (w *Writer) Write(bs []byte) (n int, err error) {
+	// Skip all checks if the level isn't enabled.
+	if !w.Log.Core().Enabled(w.Level) {
+		return len(bs), nil
```
sywhang, line 52: Doesn't Write usually return len of bytes written? If so, shouldn't this be 0?

```suggestion
		return 0, nil
```

zapio/writer.go
```diff
@@ -0,0 +1,108 @@
...
+func (w *Writer) writeLine(line []byte) (remaining []byte) {
+	idx := bytes.IndexByte(line, '\n')
+	if idx < 0 {
+		// If there are no newlines, buffer the entire string.
+		w.buff.Write(line)
+		return nil
+	}
+
+	// Split on the newline, buffer and flush the left.
+	line, remaining = line[:idx], line[idx+1:]
```
sywhang, line 74: if `idx` == `len(bytes)-1`, wouldn't `line[idx+1:]` panic?

zapio/writer.go
abhinav: I considered a constructor first but I realized that meant that we need at least two parameters: level and logger. To leave room for parameters in the future, we would have had to use a parameter struct or functional options, and that seems like a lot of surface area for a simple writer.

I figured we can start with the minimal struct-only approach. We can always add a constructor later; it should be easy to continue to maintain the behavior of respecting the Logger and Level fields since any constructor-based variant with any unexported fields will still always need a Logger and a Level.

zapio/writer.go
```diff
@@ -0,0 +1,108 @@
...
+		bs = w.writeLine(bs)
+	}
+
+	return n, nil
+}
+
+// writeLine writes a single line from the input, returning the remaining,
+// unconsumed bytes.
+func (w *Writer) writeLine(line []byte) (remaining []byte) {
+	idx := bytes.IndexByte(line, '\n')
```
moisesvega, line 66: Is there a possibility to get more than one `\n` as part of the input ?

zapio/writer.go
abhinav: I wrote a benchmark and tested that. strings.Builder is actually slower here because it throws away the buffer on Reset. (Because it has no guarantee you're not using the `String()` it returned elsewhere.)

Given that Zap will not use the string given to Check once we call ce.Write, it might be okay for us to cast the `[]byte` to string unsafely.

zapio/writer.go
abhinav: Oh, that's a good idea. That'll be the case a majority of the time anywa.

zapio/writer.go
abhinav: That's a valid point, in that we're not actually writing those bytes, but per https://golang.org/pkg/io/#Writer

> Write must return a non-nil error if it returns n < len(p).

So this is less to detect "how much you wrote in reality" but more "how much were you able to write before it failed". If the logger is configured to not log at this level, it's a no-op, but we're still consuming the text, so I figured we'd want to tell the caller we consumed everything.

The alternative here is to return an error stating that the logger is not configured to log at this level, but I suspect that's undesirable.

zapio/writer.go
abhinav: Na, subslicing like `foo[x:]` works fine when `x == len(foo)`, but not if `x > len(foo)`.

https://play.golang.org/p/5f_L9-TqAUs

zapio/writer.go
abhinav: Yeah, if there are multiple consecutive newlines in the input, we'll retain them in the output. One of the test cases checks for this:

```
		{
			desc: "blank lines in the middle",
			writes: []string{
				"foo\n\nbar\nbaz",
			},
			want: []zapcore.Entry{
				{Level: zap.InfoLevel, Message: "foo"},
				{Level: zap.InfoLevel, Message: ""},
				{Level: zap.InfoLevel, Message: "bar"},
				{Level: zap.InfoLevel, Message: "baz"},
			},
		},
```

zapio/writer.go
abhinav: So I tried the unsafe cast. Turns out we can't do that after all.
We have no guarantee that a custom Core isn't holding onto the string.
In fact, I discovered the issue because zaptest/observer holds onto the string.

zapio/writer.go
sywhang: Ok, that makes sense. I agree that telling the caller that we consumed everything would be better than returning an error. Thanks!

zapio/writer.go
prashantv: Agree, makes sense to start with the struct

zapio/writer_test.go
```diff
@@ -0,0 +1,215 @@
...
+			// Turn []observer.LoggedEntry => []zapcore.Entry
+			got := make([]zapcore.Entry, observed.Len())
+			for i, ent := range observed.AllUntimed() {
+				got[i] = ent.Entry
+			}
+			assert.Equal(t, tt.want, got, "Logged entries do not match.")
+		})
+	}
+}
+
```
prashantv, line 159: can you add a test to ensure that `Sync()` doesn't flush out empty lines

prashantv (approved):
Can you paste the benchmark results as well?

One other minor question: does it make sense to use this writer in place of `loggerWriter`?
https://github.com/uber-go/zap/blob/v1.17.0/global.go#L160

abhinav:
Yeah, they were in the individual commits. I'll add them to the PR description.

abhinav:
> Can you paste the benchmark results as well?

Yeah, they were in the individual commits. I'll add them to the PR description.

zapio/writer_test.go
abhinav: Ack

abhinav:
> One other minor question: does it make sense to use this writer in place of `loggerWriter`?
> https://github.com/uber-go/zap/blob/v1.17.0/global.go#L160

That's a good point, except we'll end up with a cycle between zap and
zapio.  I don't have an immediate good solution for this, except maybe
moving zapio.Writer to the zap package. I originally elected to put it
in zapio because zap is already a pretty densely populated package.

Another way we can break the zapio->zap dependency by making the
`*zap.Logger` references in zapio use an interface but that seems like
it'll be limiting, especially if it means that we can never introduce a
dependency from zapio to zap.

So short of moving it to the zap package, we might have to live with
the duplication. Honestly, it might be justifiable with the argument
that "that's for Zap's own internal usage, and this is for everyone
else."

PR #975, opened 2021-06-29
Discard counters for out-of-bounds levels

### LINKED ISSUE #972: Zap + logr == not very expressive, open to fixes?
One of my side-projects is https://github.com/go-logr/logr (an API for libraries to use to log) and alongside it lives the bridge to zap (https://github.com/go-logr/zapr).

Logr uses integer log levels - V(3) is "more debuggy" than V(2), which is more than V(1).  Zap doesn't natively support that, but it seems to work, with a small number of caveats.  So far I have found 1, actually.

`zapcore/sampler.go` does an indexed lookup by log level.  When I use Logr verbose levels > 1 (which maps to Debug in zap) I get an out-of-bounds panic.

A quick hack of adding some number of extra debug levels (just add 100 indices to that array and accesses by +100) makes a trivial test pass with V(2), V(3), etc.

Before I really dig in, I wanted to see if this is something zap maintainers are willing to entertain ?  I imagine it would ultimately manifest as a set of small changes here and there to handle the extra levels, plus tests.  E.g. the level name already renders as "Level(-2)", so someone thought about this a little already.

thockin:
Previously this just crashed.

Fixes #972

### DIFF
--- a/zapcore/sampler.go
+++ b/zapcore/sampler.go
@@ -197,12 +197,14 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
 		return ce
 	}
 
-	counter := s.counts.get(ent.Level, ent.Message)
-	n := counter.IncCheckReset(ent.Time, s.tick)
-	if n > s.first && (n-s.first)%s.thereafter != 0 {
-		s.hook(ent, LogDropped)
-		return ce
+	if ent.Level >= _minLevel && ent.Level <= _maxLevel {
+		counter := s.counts.get(ent.Level, ent.Message)
+		n := counter.IncCheckReset(ent.Time, s.tick)
+		if n > s.first && (n-s.first)%s.thereafter != 0 {
+			s.hook(ent, LogDropped)
+			return ce
+		}
+		s.hook(ent, LogSampled)
 	}
-	s.hook(ent, LogSampled)
 	return s.Core.Check(ent, ce)
 }
--- a/zapcore/sampler_test.go
+++ b/zapcore/sampler_test.go
@@ -225,8 +225,22 @@ func TestSamplerRaces(t *testing.T) {
 	wg.Wait()
 }
 
-func TestSamplerBounds(t *testing.T) {
-	// Prove that out-of-bounds levels don't crash.
-	sampler, _ := fakeSampler(DebugLevel-1, time.Minute, 2, 3)
-	writeSequence(sampler, 1, DebugLevel-1)
+func TestSamplerUnknownLevels(t *testing.T) {
+	// Prove that out-of-bounds levels don't panic.
+	unknownLevels := []Level{
+		DebugLevel - 1,
+		FatalLevel + 1,
+	}
+
+	for _, lvl := range unknownLevels {
+		t.Run(lvl.String(), func(t *testing.T) {
+			sampler, logs := fakeSampler(lvl, time.Minute, 2, 3)
+			for i := 1; i < 10; i++ {
+				writeSequence(sampler, i, lvl)
+			}
+
+			// Expect no sampling for unknown levels.
+			assertSequence(t, logs.TakeAll(), lvl, 1, 2, 3, 4, 5, 6, 7, 8, 9)
+		})
+	}
 }
--- a/zapcore/sampler_test.go
+++ b/zapcore/sampler_test.go
@@ -224,3 +224,9 @@ func TestSamplerRaces(t *testing.T) {
 	close(start)
 	wg.Wait()
 }
+
+func TestSamplerBounds(t *testing.T) {
+	// Prove that out-of-bounds levels don't crash.
+	sampler, _ := fakeSampler(DebugLevel-1, time.Minute, 2, 3)
+	writeSequence(sampler, 1, DebugLevel-1)
+}


zapcore/sampler.go
```diff
@@ -42,8 +42,13 @@ func newCounters() *counters {
 	return &counters{}
 }
 
+var nullCounter counter
+
 func (cs *counters) get(lvl Level, key string) *counter {
 	i := lvl - _minLevel
+	if i < 0 {
```
prashantv, line 49: Though it's unlikely, checking the upper bound would be helpful here too
```suggestion
	if i < 0 || i >= len(cs) {
```

zapcore/sampler.go
```diff
@@ -42,8 +42,13 @@ func newCounters() *counters {
...
+		return &nullCounter
```
prashantv, line 50: The behaviour here is that all unknown level / any message will be sampled as if they were a single level + message, which I don't think is the intended behaviour. It will likely lead to over-sampling

Looking at the implementation, a simple alternative could be to min/max `i` so it's within the bounds:
```
if i < 0 {
  i = 0
}
if i >= len(cs) {
  i = len(cs) - 1
}
```

Alternatively, we could have a separate set of counters for "unknown level", but I think the above is a little bit simpler.

zapcore/sampler.go
thockin: ACK

zapcore/sampler.go
thockin: I'll admit I don't know enough about zap to say what the "right" answer is or even exactly what's being sampled.  I assumed that the counts for out-of-bounds level just weren't interesting (since they don't have names).

If the desire is just to saturate at the lowest/highest named levels then your solution works.  But that seems like it might confuse users?

If you want to actually count per-level, then maybe we want a `map[int][_countersPerLevel]counter`, but then we need some atomic accessor.  To avoid that in most cases, you could keep the array and fall back to the map only if the level is out-of-bounds?  E.g.

```
    // ...
    if i >= 0 && i < len(cs.named) {
        return cs.named[i][j]
    }
    return cs.getUnnamed(i, j)
}

func (cs *counters) getUnnamed(i, j int) *counter {
    cs.mu.Lock()
    defer cs.mu.Unlock()
    if cs.unnamed[i] == nil {
        cs.unnamed[i] = [_countersPerLevel]counter{}
    }
    return &cs.unnamed[i][j]
}
```

Maybe simpler is my original plan - just add some fixed number of counters at one or both ends and document it.  Your `counters` type gets bigger either way, but I am not confident to say if that matters or not.

I'll approach it however you think is best - my need is small, but I don't want to have disproportionate impact.

prashantv: My alternative proposal was mostly to localize this change to where we use the level to convert to an index, but I agree that affecting other levels may be surprising.

I think the preferred behaviour is for the sampler to ignore unknown levels. Instead of making any change here, we can instead do the level check at the caller,
https://github.com/uber-go/zap/blob/v1.18.1/zapcore/sampler.go#L200

If the level is unsupported, then we "fail open" by calling the underlying core's `Check`.

thockin:
Please take a look - is that what you had in mind?

prashantv (approved):
Yep, this is what I had in mind, thanks!

Change looks good to me, I've pushed a couple of small test changes (more levels + validate no sampling).

Tagging @abhinav to take a look as well.

abhinav (approved):

thockin:
Hello.  Any chance we can get a new tag, so I can pin to this version?

zapcore/sampler.go
```diff
@@ -197,12 +197,14 @@ func (s *sampler) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
 		return ce
 	}
 
-	counter := s.counts.get(ent.Level, ent.Message)
-	n := counter.IncCheckReset(ent.Time, s.tick)
-	if n > s.first && (n-s.first)%s.thereafter != 0 {
-		s.hook(ent, LogDropped)
-		return ce
+	if ent.Level >= _minLevel && ent.Level <= _maxLevel {
```
nmiculinic, line 200: Maybe I'm dropping here like a martian, but wouldn't it make more sense to clip the `ent.Level` to [_minLevel, _maxLevel] range? 

Will this behavior cause log entries outside valid range to bypass sampling, thus always be emitted?

prashantv: That was one of the options considered, see https://github.com/uber-go/zap/pull/975#discussion_r660809497 but there was concern that it could confuse users.

Yes, the current approach causes log entries with unknown levels to bypass sampling.

thockin:
My feeling is that super-verbose levels are special and not "normal" and,
as such, sampling is probably not appropriate.

That's how I justify it to myself anyway :)

BTW: How about a new tag ?  :)

zapcore/sampler.go
nmiculinic: In typical usecase for this feature would be integration with klog/logr. Those out-of-bounds entries are going to be the most common ones actually, and ones which should be sampled...probably....maybe when user actually uses those levels they are interested in all logs and will bypass sampling anyway?

thockin:
That was my feeling.  logr V(1) == zap.Debug (so sampled).  logr V(2) ==
zap.Level(-2).  Now I will admit that some k8s components run at V=2 by
default, but that itself seems like a problem :)

abhinav:
> BTW: How about a new tag ? :)

Oops, missed this. We'll tag a release next week. (We prefer not to tag new releases of core components on Fridays.) Thanks, @thockin!

thockin:
Thanks!

PR #978, opened 2021-07-09
Added support for console encoder to configure whether it should json escape strings

jlstevo:
Allowing console encoders to avoid json escaping strings allows an easy way for the default formatting to provide more legible error output.


fighterlyt:
no one deal with this ?

abhinav (changes requested):
Hello, sorry but we're not convinced that Zap should generate invalid JSON strings like this. Even `fmt.Sprintf` with `%q` will escape the objects. For this level of output customization, it seems that what we want is not a JSON-based encoder but something else entirely.

PR #983, opened 2021-07-29
Simplify check for writing errors to ErrorOutput

lancoLiu:


zapcore/entry.go
```diff
@@ -219,11 +219,9 @@ func (ce *CheckedEntry) Write(fields ...Field) {
 	for i := range ce.cores {
 		err = multierr.Append(err, ce.cores[i].Write(ce.Entry, fields))
 	}
-	if ce.ErrorOutput != nil {
-		if err != nil {
-			fmt.Fprintf(ce.ErrorOutput, "%v write error: %v\n", ce.Time, err)
-			ce.ErrorOutput.Sync()
-		}
+	if ce.ErrorOutput != nil && err != nil {
```
prashantv, line 222:
```suggestion
	if err != nil && ce.ErrorOutput != nil {
```

prashantv (approved):
Thanks for the simplification!

zapcore/entry.go
prashantv: This may be a little clearer logically -- if there's an error, *then* we care about whether we have an `ErrorOutput`.

lancoLiu: thanks. I agree with you.

prashantv:
Thanks for the contribution @lancoLiu

PR #984, opened 2021-07-30
Optimize size of BufferedWriteSyncer by moving stopped bool field

lancoLiu:


prashantv (approved):
Keeping the 2 `bool` fields next to each other avoids wasting an additional word. Verified that the size of BufferedWriteSyncer drops from 104 bytes to 96 bytes with this change on amd64.

PR #989, opened 2021-08-12
Added SkipLineEnding to zapcore.EncoderConfig

lruggieri:
Setting`config.EncoderConfig.SkipLineEnding = true` avoids adding any further symbol to the end of the line.

My use-case is the one of a CGO library internally using zap and redirecting the logs to a custom C/C++ function. The calling function expects a string as output and does not expect a newline, or any other symbol, added.

### DIFF
--- a/zapcore/console_encoder.go
+++ b/zapcore/console_encoder.go
@@ -125,13 +125,7 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer, 
 		line.AppendString(ent.Stack)
 	}
 
-	if !c.SkipLineEnding {
-		if c.LineEnding != "" {
-			line.AppendString(c.LineEnding)
-		} else {
-			line.AppendString(DefaultLineEnding)
-		}
-	}
+	line.AppendString(c.LineEnding)
 	return line, nil
 }
 
--- a/zapcore/console_encoder.go
+++ b/zapcore/console_encoder.go
@@ -125,10 +125,12 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer, 
 		line.AppendString(ent.Stack)
 	}
 
-	if c.LineEnding != "" {
-		line.AppendString(c.LineEnding)
-	} else {
-		line.AppendString(DefaultLineEnding)
+	if !c.SkipLineEnding {
+		if c.LineEnding != "" {
+			line.AppendString(c.LineEnding)
+		} else {
+			line.AppendString(DefaultLineEnding)
+		}
 	}
 	return line, nil
 }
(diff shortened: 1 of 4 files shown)


manjari25 (approved):

zapcore/console_encoder.go
```diff
@@ -125,10 +125,12 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer,
 		line.AppendString(ent.Stack)
 	}
 
-	if c.LineEnding != "" {
-		line.AppendString(c.LineEnding)
-	} else {
-		line.AppendString(DefaultLineEnding)
+	if !c.SkipLineEnding {
```
sywhang, line 128: It'd be preferable to change `DefaultLineEnding` to an empty string instead of doing this because this is essentially adding an additional boolean check for configs that didn't specify the SkipLineEnding every time we encode.

sywhang (commented):
@lruggieri thanks for submitting this PR! 

There is actually a room for improvement in the existing code as well - if we compute the "real" line ending in `NewJsonEncoder` or `NewConsoleEncoder` by checking this if condition just once, we can simply concatenate the "real" line ending here without having to do any if check for each subsequent encoding attempts. 

Could you please modify this PR in such a way? Otherwise I can take it from here and make these changes.

zapcore/console_encoder.go
lruggieri: Wouldn't changing the default value mean breaking backward compatibility? I suppose there are people expecting to see their logs going to a new line every time, and pulling a change like that would break that behaviour, I think.

sywhang: sorry, I should've been more clear w/ my comment - I meant that you'd just need to change DefaultLineEnding to an empty string *if* `SkipLineEnding` is set to true.

lruggieri: I see. So merging with your other comment below if would be changing `SkipLineEnding ` to a var and computing the "real" `LineEnding` at `NewXEncoder` time, right?

sywhang: `SkipLineEnding` can stay in `EncoderConfig`; What I'd imagine would change is somewhere along the lines of:

```
 func newJSONEncoder(cfg EncoderConfig, spaced bool) *jsonEncoder {
+       if c.SkipLineEnding {
+               cfg.LineEnding = ""
+       } else if cfg.LineEnding == "" {
+               cfg.LineEnding = DefaultLineEnding
+       }
```

And in these lines, we'd simply need to do:
`line.AppendString(cfg.LineEnding)` instead of doing the if checks over and over.

`NewConsoleEncoder` uses `newJSONEncoder` so you wouldn't need to change anything else.

lruggieri: should be handled within [the new commit](https://github.com/uber-go/zap/pull/989/commits/6fa2cff0b691b932c5b8a17f6f1f97f4a1697fdb)

sywhang (approved):
This looks good to me! Thanks again for the PR and working with us to improve the existing code as well :)

PR #999, opened 2021-09-07
CI: Test with Go 1.16, 1.17 on Windows and Linux

abhinav:
Update the GitHub workflow configuration to test on both, Linux and
Windows with the latest releases of Go.


shvc:
any update?

Issue #1002, opened 2021-09-08
Bug: Incorrect encoding of float32 numbers


hemantjadon:
`zap.Float32` field types incorrectly formats the float32 numbers. The precision (bitSize) information is lost while converting float32.
Eg. `2.71` is formatted as `2.7100000381469727`

```go
package main

import (
	"go.uber.org/zap"
)

func main() {
    l, _ := zap.NewProduction()

    l.Info("test", zap.Float32("key", 2.71))
}
```

### Got output
```
{"level":"info","ts":1257894000,"caller":"sandbox3599029287/prog.go:10","msg":"test","key":2.7100000381469727}
```

### Expected output
```
{"level":"info","ts":1257894000,"caller":"sandbox3599029287/prog.go:10","msg":"test","key":2.71}
```

### Reference
https://play.golang.org/p/3IkJONIZ_xN

hemantjadon:
This is due to the fact, that correct bitSize information is not reaching the `enc.appendFloat` method, as on `zapcore/json_encoder.go#L300`, `float32` value is being converted to a `float64`, and then is passed to `enc.AppendFloat64`, instead of being passed to `enc.AppendFloat32`

```go
package main

import (
	"fmt"
	"strconv"
)

func main() {
	var f64 float64 = 2.71
	var f32 float32 = 2.71
	fmt.Println(strconv.FormatFloat(f64, 'f', -1, 64))          // 2.71
	fmt.Println(strconv.FormatFloat(float64(f32), 'f', -1, 32)) // 2.71
	fmt.Println(strconv.FormatFloat(float64(f32), 'f', -1, 64)) // 2.7100000381469727
}

```

### Reference
https://play.golang.org/p/5_f8UMKIFMT

abhinav:
Thanks for reporting! We'll try to have this fixed shortly.

(internal ref: GO-860)

manjari25:
Released on v1.19.1

PR #1003, opened 2021-09-08
json: Fix float32 encoding and add test

manjari25:
This commit fixes incorrect float32 encoding. A test 
case is also added.

Refs: 
* https://github.com/uber-go/zap/issues/1002
* Internal ticket: GO-860


sywhang (approved):
LGTM

abhinav (approved):

PR #1004, opened 2021-09-08
Update changelog

manjari25:
This commit adds an unreleased section to the changelog 
to track things that need to be released. It also lists the
unreleased changes since the last release to this section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,7 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
-and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 ### Fixed
-* [#1001][]: Fix complex number encoding with negative imaginary part. Thanks to @hemantjadon.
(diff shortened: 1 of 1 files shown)


abhinav (approved):

CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
 # Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
```
abhinav, line 5: So this changelog doesn't quite adhere to keepachangelog.

The format here is a bit bespoke at this point. 

Unless we want to go through the trouble of migrating the format, I would only mention that we follow semver.

manjari25: Updated.

CHANGELOG.md
```diff
@@ -1,4 +1,15 @@
 # Changelog
+All notable changes to this project will be documented in this file.
+
+This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+### Fixed
+* [#1001][]: Fix complex number encoding with negative imaginary part. Thanks to @hemantjadon.
+* [#1003][]: Fix float32 encoding.
```
abhinav, line 9: Something more specific here? It's not that float32 was entirely broken; only that the preciseness was lost.

```suggestion
* [#1003][]: JSON: Fix inaccurate precision when encoding float32.
```

CHANGELOG.md
```diff
@@ -1,4 +1,15 @@
 # Changelog
+All notable changes to this project will be documented in this file.
+
+This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+### Fixed
+* [#1001][]: Fix complex number encoding with negative imaginary part. Thanks to @hemantjadon.
```
abhinav, line 8: Let's prefix this with component:

```suggestion
* [#1001][]: JSON: Fix complex number encoding with negative imaginary part. Thanks to @hemantjadon.
```

abhinav (approved):

PR #1009, opened 2021-09-10
CI: Test with Go 1.17

abhinav:
Add Go 1.17 to the list of versions we test with. `gofmt` all files to
add the new `//go:build` directives meant to replace `// +build`.

This supersedes the Go version upgrade in #999. That PR will be narrowed
down just to Windows support.

### DIFF
--- a/global_go112.go
+++ b/global_go112.go
@@ -19,6 +19,7 @@
 // THE SOFTWARE.
 
 // See #682 for more information.
+//go:build go1.12
 // +build go1.12
 
 package zap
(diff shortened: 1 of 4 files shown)


global_go112.go
```diff
@@ -19,6 +19,7 @@
 // THE SOFTWARE.
 
 // See #682 for more information.
+//go:build go1.12
```
rabbbit, line 22: can we just kill these now?

moisesvega (approved):

rabbbit (approved):

PR #1011, opened 2021-09-10
JSON: Fix complex64 encoding precision

### LINKED ISSUE #1010: Bug: Incorrect encoding of complex64 numbers
`zap.Complex64` field types incorrectly formats the `complex64` numbers. The precision (bitSize) information is lost while converting `complex64`.
Eg. `2.71+3.14i` is formatted as `2.7100000381469727+3.140000104904175i`

```go
package main

import (
	"go.uber.org/zap"
)

func main() {
    l, _ := zap.NewProduction()

    l.Info("test", zap.Complex64("key", 2.71+3.14i))
}
```

### Got output
{"level":"info","ts":1257894000,"caller":"sandbox265920417/prog.go:10","msg":"test","key":"2.7100000381469727+3.140000104904175i"}

### Expected output
{"level":"info","ts":1257894000,"caller":"sandbox265920417/prog.go:10","msg":"test","key":"2.71+3.14i"}

### Reference

https://play.golang.org/p/YH_NmnaRgBs

abhinav:
Converting `complex64` to `complex128` meant that
the real and imaginary components of the `complex64` value
were encoded as `float64` values which,
similar to #1002, encoded them with incorrect precision.

Fix this by adding a precision parameter during encoding to specify
the precision with which the components should be encoded.

Guard against other such cases where there's loss of precision
by adding tests for several such corner cases for
complex, float, int, and uint.

This has no significant performance impact on my laptop.
(The drop is likely a false positive.)

```
name                          old time/op  new time/op  delta
ZapJSONFloat32AndComplex64-4   491ns ±22%   442ns ± 6%  -10.00%  (p=0.000 n=9+9)
```

Fixes #1010


zapcore/json_encoder.go
```diff
@@ -301,28 +309,28 @@ func (enc *jsonEncoder) AppendUint64(val uint64) {
...
+func (enc *jsonEncoder) AddInt8(k string, v int8)       { enc.AddInt64(k, int64(v)) }
+func (enc *jsonEncoder) AddUint(k string, v uint)       { enc.AddUint64(k, uint64(v)) }
+func (enc *jsonEncoder) AddUint32(k string, v uint32)   { enc.AddUint64(k, uint64(v)) }
+func (enc *jsonEncoder) AddUint16(k string, v uint16)   { enc.AddUint64(k, uint64(v)) }
+func (enc *jsonEncoder) AddUint8(k string, v uint8)     { enc.AddUint64(k, uint64(v)) }
+func (enc *jsonEncoder) AddUintptr(k string, v uintptr) { enc.AddUint64(k, uint64(v)) }
+func (enc *jsonEncoder) AppendComplex64(v complex64)    { enc.appendComplex(complex128(v), 32) }
+func (enc *jsonEncoder) AppendComplex128(v complex128)  { enc.appendComplex(complex128(v), 64) }
+func (enc *jsonEncoder) AppendFloat64(v float64)        { enc.appendFloat(v, 64) }
+func (enc *jsonEncoder) AppendFloat32(v float32)        { enc.appendFloat(float64(v), 32) }
```
sywhang, line 324: Sorry, I should've probably commented on this when `float32` encoding was being changed... But I think there may be a chance we're taking a performance hit here because now the payloads are not all 64-bytes anymore, whereas before everything was 64 bytes, so we may be taking a hit on alignment (and therefore cache locality+fragmentation for buffers in the pool).

Have we done any benchmarks with these changes? Just to be clear, I think incorrect encoding is still a bigger issue than a perf hit, but at least we should understand what impact it had.

abhinav: Fair. I'll add a benchmark and results.

sywhang (approved):
LGTM

PR #1013, opened 2021-09-16
feat: add config  option handler

tylitianrui:
during my work. I  use  `Development` mode  ( zap.NewDevelopmentConfig )  with  `Encoding = 'console'` and `EncoderConfig = XXXXEncoderConfig`,so i  have to change code like it
```
cnf := NewDevelopmentConfig()
cnf.Encoding = "console"
cnf.EncoderConfig = zapcore.EncoderConfig{ 
}
```
or  i  new a config  with  many  attributes to be assigned like the below:
```
c = &Config{
			Level:             AtomicLevel{},
			Development:       false,
			DisableCaller:     false,
			DisableStacktrace: false,
			Sampling:          &SamplingConfig{},
			Encoding:          encoding,
			EncoderConfig:     zapcore.EncoderConfig{
				MessageKey:    "",
				LevelKey:      "",
				TimeKey:       "",
				NameKey:       "",
				CallerKey:     "",
				FunctionKey:   "",
				StacktraceKey: "",
				LineEnding:    "",
				EncodeLevel: func( zapcore.Level,  zapcore.PrimitiveArrayEncoder) {
				},
				EncodeTime: func( time.Time,  zapcore.PrimitiveArrayEncoder) {
				},
				EncodeDuration: func( time.Duration,  zapcore.PrimitiveArrayEncoder) {
				},
				EncodeCaller: func( zapcore.EntryCaller,  zapcore.PrimitiveArrayEncoder) {
				},
				EncodeName: func( string,  zapcore.PrimitiveArrayEncoder) {
				},
				ConsoleSeparator: "",
			},
			OutputPaths:       []string{},
			ErrorOutputPaths:  []string{},
			InitialFields:     map[string]interface{}{},
		}

```
i dont think it is so graceful. 
why not  support   `option`  to change  attributes which users want .

### DIFF
--- a/config.go
+++ b/config.go
@@ -117,8 +117,8 @@ func NewProductionEncoderConfig() zapcore.EncoderConfig {
 //
 // It uses a JSON encoder, writes to standard error, and enables sampling.
 // Stacktraces are automatically included on logs of ErrorLevel and above.
-func NewProductionConfig() Config {
-	return Config{
+func NewProductionConfig(opts ...ConfOptionHandler) Config {
+	c := Config{
 		Level:       NewAtomicLevelAt(InfoLevel),
 		Development: false,
 		Sampling: &SamplingConfig{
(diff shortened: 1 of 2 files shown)


abhinav (changes requested):
I may be misunderstanding the problem,
but can we not use the following to do what WithConfEncoding would give us?

```
cfg := NewDevelopmentConfig()
cfg.Encoding = "console"
```

PR #1033, opened 2021-12-08
sampler: Support thereafter of zero

### LINKED ISSUE #1032: Sampling core doesn't seem to have any effect
I'm trying to limit the excessive output of some very active logs, the same two messages repeating (same logger) thousands of times per second.

I figured that wrapping my custom core in a `NewSamplerWithOptions()` would do the trick:

```go
// ...

core := zapcore.NewTee(
	zapcore.NewCore(consoleEncoder, consoleOut, zap.DebugLevel),
	zapcore.NewCore(jsonEncoder, webOut, zap.InfoLevel),
)

core = zapcore.NewSamplerWithOptions(core, 5*time.Second, 1, 1)

logger := zap.New(core)

// ...
```

I thought I could expect about 2 log messages per 5 seconds with this configuration, but I'm still seeing a firehose of thousands of the same logs every second, completely obliterating the performance of the app (writing the logs to the outputs is slowing it down).

Am I using it wrong? :sweat_smile:

abhinav:
The Zap sampling logger accepts three configuration parameters:
interval, first, and thereafter.

After we see `first` log with the same message in `interval`,
the sampler kicks in.
Following that, we let through every `thereafter`-th log.

So for example,

    NewSamplerWithOptions(core, time.Second, 100, 50)

This will allow 100 logs with the same message in a second,
and following that, every 50th message with that message.

In #1032, the user wanted the sampler to reject *all* duplicate messages
once the limit was reached, but our sampler panics if `thereafter = 0`.

This change adds support for setting `thereafter` to 0,
dropping all messages in that interval once the limit is reached.
It also adds further explanation to the documentation to address the
misuse in #1032.

Resolves #1032

### DIFF
--- a/zapcore/sampler.go
+++ b/zapcore/sampler.go
@@ -133,10 +133,21 @@ func SamplerHook(hook func(entry Entry, dec SamplingDecision)) SamplerOption {
 // each tick. If more Entries with the same level and message are seen during
 // the same interval, every Mth message is logged and the rest are dropped.
 //
+// For example,
+//
+//   core = NewSamplerWithOptions(core, time.Second, 10, 5)
+//
+// This will log the first 10 log entries with the same level and message
+// in a one second interval as-is. Following that, it will allow through
+// every 5th log entry with the same level and message in that interval.
+//
+// If thereafter is zero, the Core will drop all log entries after the first N
+// in that interval.
+//
 // Sampler can be configured to report sampling decisions with the SamplerHook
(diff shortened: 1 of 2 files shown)


moisesvega (approved):

sywhang (approved):
:shipit:

mholt:
Thank you! The updated docs and support for M=0 are much clearer to me.

PR #1039, opened 2021-12-14
feat: add support for custom ReflectType encoder

### LINKED ISSUE #1034: feat: Custom `interface{}` object Encoder
## Description

We log a lot of structs using zap. And zap uses `json.Encoder` for encoding these `interface{}` (or `ReflectType`) fields.
It would be great if we can use something else other than `encoding/json` lib.

For instance, [json-iterator](https://github.com/json-iterator/go) is faster than std lib and provides other customizations on top of it.

This would provide 2 benefits:
1. Better performance
2. Logging `interface{}` objects in a custom way.

For example: [json-iterator](https://github.com/json-iterator/go) provides the functionality to have custom tags instead of `json`.

```go
type UserData struct {
  Name string `json:"name" log:"name"`
  Age int64 `json:"age" log:"age"`
  Email string `json:"email" log:"-"`
  Address string `json:"address" log:"-"`
}
```

With this feature, we can control what we want to log. 
In the above example, we want to only log `Name` and `Age` but don't want to log `Email` and `Address`. 
Since this is using its own instance, it won't affect other serializations/deserializations taking place somewhere else.

## On Implementation

I figured one way we can do this is by replacing the type of `reflectEnc` variable to a newly defined `ReflectTypeEncoder` interface in `json_encoder.go`.

This interface will expose mainly 2 methods:

1. `WithBuffer(*buffer.Buffer)`: which will provide the encoder the Buffer where it needs to write the data
2. `Encode(interface{})`: which encodes the object and writes it to the provided buffer above

The configuration can be provided while initializing `Logger`:

```go
jsonIterConfig := jsoniter.Config{
  EscapeHTML:             false,
  TagKey: "log",
}

zapConfig := zap.NewProductionConfig()
zapConfig.EncoderConfig.EncodeReflectType = NewJsonIterReflectTypeEncoder(jsonIterConfig)
```

If not provided, we can continue using the default `encoding/json` one so won't be a breaking change.

LMK if there are any concerns with this approach.
Or there are other ways I can do this.
If you're okay with it, I can raise a PR. 

Note: This may also solve #993

psrajat:
Hi Team!
This fixes #1034

### DIFF
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -334,7 +334,7 @@ type EncoderConfig struct {
 	EncodeName NameEncoder `json:"nameEncoder" yaml:"nameEncoder"`
 	// Configure the encoder for interface{} type objects
 	// If not provided, objects are encoded using json.Encoder
-	NewReflectedEncoder func(io.Writer) ReflectedEncoder `json:"newReflectedEncoder" yaml:"newReflectedEncoder"`
+	NewReflectedEncoder func(io.Writer) ReflectedEncoder `json:"-" yaml:"-"`
 	// Configures the field separator used by the console encoder. Defaults
 	// to tab.
 	ConsoleSeparator string `json:"consoleSeparator" yaml:"consoleSeparator"`
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -332,7 +332,7 @@ type EncoderConfig struct {
 	// Unlike the other primitive type encoders, EncodeName is optional. The
 	// zero value falls back to FullNameEncoder.
 	EncodeName NameEncoder `json:"nameEncoder" yaml:"nameEncoder"`
-	// Configure the encoder for interface{} type objects
+	// Configure the encoder for interface{} type objects.
 	// If not provided, objects are encoded using json.Encoder
 	NewReflectedEncoder func(io.Writer) ReflectedEncoder `json:"-" yaml:"-"`
 	// Configures the field separator used by the console encoder. Defaults
--- a/zapcore/encoder.go
+++ b/zapcore/encoder.go
@@ -22,6 +22,7 @@ package zapcore
 
 import (
 	"encoding/json"
+	"io"
 	"time"
 
 	"go.uber.org/zap/buffer"
@@ -331,6 +332,9 @@ type EncoderConfig struct {
 	// Unlike the other primitive type encoders, EncodeName is optional. The
 	// zero value falls back to FullNameEncoder.
 	EncodeName NameEncoder `json:"nameEncoder" yaml:"nameEncoder"`
+	// Configure the encoder for interface{} type objects
+	// If not provided, objects are encoded using json.Encoder
+	NewReflectedEncoder func(io.Writer) ReflectedEncoder `json:"newReflectedEncoder" yaml:"newReflectedEncoder"`
 	// Configures the field separator used by the console encoder. Defaults
 	// to tab.
 	ConsoleSeparator string `json:"consoleSeparator" yaml:"consoleSeparator"`
--- a/zapcore/json_encoder.go
+++ b/zapcore/json_encoder.go
@@ -22,7 +22,7 @@ package zapcore
 
 import (
 	"encoding/base64"
-	"encoding/json"
+	"io"
 	"math"
 	"sync"
 	"time"
@@ -64,7 +64,7 @@ type jsonEncoder struct {
 
 	// for encoding generic values by reflection
 	reflectBuf *buffer.Buffer
-	reflectEnc *json.Encoder
+	reflectEnc ReflectedEncoder
 }
 
 // NewJSONEncoder creates a fast, low-allocation JSON encoder. The encoder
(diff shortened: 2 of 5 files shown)


zapcore/encoder.go
```diff
@@ -331,6 +332,9 @@ type EncoderConfig struct {
 	// Unlike the other primitive type encoders, EncodeName is optional. The
 	// zero value falls back to FullNameEncoder.
 	EncodeName NameEncoder `json:"nameEncoder" yaml:"nameEncoder"`
+	// Configure the encoder for interface{} type objects
+	// If not provided, objects are encoded using json.Encoder
+	NewReflectedEncoder func(io.Writer) ReflectedEncoder `json:"newReflectedEncoder" yaml:"newReflectedEncoder"`
```
abhinav, line 337: Since this a function that is not JSON/YAML encodeable, we'll have to use `-` here.
```suggestion
	NewReflectedEncoder func(io.Writer) ReflectedEncoder `json:"-" yaml:"-"`
```

zapcore/json_encoder.go
```diff
@@ -152,10 +152,14 @@ func (enc *jsonEncoder) AddInt64(key string, val int64) {
...
-
-		// For consistency with our custom JSON encoder.
-		enc.reflectEnc.SetEscapeHTML(false)
+		// If no EncoderConfig.NewReflectedEncoder is provided by the user, then use default
+		var newReflectedEncoder func(io.Writer) ReflectedEncoder
+		if enc.NewReflectedEncoder == nil {
+			newReflectedEncoder = GetDefaultReflectedEncoder()
+		} else {
+			newReflectedEncoder = enc.NewReflectedEncoder
+		}
```
abhinav, line 161: We should do this in `NewJSONEncoder`? We can set `NewReflectedEncoder` in there once if it's nil, and then we don't have to check it on the hot path.

zapcore/json_encoder.go
```diff
@@ -152,10 +152,14 @@ func (enc *jsonEncoder) AddInt64(key string, val int64) {
...
-		enc.reflectEnc = json.NewEncoder(enc.reflectBuf)
-
-		// For consistency with our custom JSON encoder.
-		enc.reflectEnc.SetEscapeHTML(false)
+		// If no EncoderConfig.NewReflectedEncoder is provided by the user, then use default
+		var newReflectedEncoder func(io.Writer) ReflectedEncoder
+		if enc.NewReflectedEncoder == nil {
+			newReflectedEncoder = GetDefaultReflectedEncoder()
+		} else {
+			newReflectedEncoder = enc.NewReflectedEncoder
```
abhinav, line 160: Would you please add a test for the custom reflection encoder case?
We don't want us to introduce a new dependency for that,
so a fake implementation of the function will suffice.

zapcore/reflected_encoder.go
```diff
@@ -0,0 +1,52 @@
...
+
+package zapcore
+
+import (
+	"encoding/json"
+	"io"
+)
+
+// A ReflectedEncoder handles the serialization of Field which have FieldType = ReflectType and writes them to the underlying data stream
+// The underlying data stream is provided by zap during initialization. See EncoderConfig.NewReflectedEncoder
```
abhinav, line 29:
```suggestion
// The underlying data stream is provided by Zap during initialization. See EncoderConfig.NewReflectedEncoder.
```

zapcore/reflected_encoder.go
```diff
@@ -0,0 +1,52 @@
...
+type ReflectedEncoder interface {
+	// Encode encodes and writes to the underlying data stream
```
abhinav, line 31:
```suggestion
	// Encode encodes and writes to the underlying data stream.
```

zapcore/reflected_encoder.go
```diff
@@ -0,0 +1,52 @@
...
+	Encode(interface{}) error
+}
+
+func GetDefaultReflectedEncoder() func(writer io.Writer) ReflectedEncoder {
```
abhinav, line 35: We would prefer not to export this function. We also don't need the wrapper type because json.Encoder satisfies the ReflectedEncoder interface.

So something like the following should suffice:

```
func defaultReflectedEncoder(w io.Writer) ReflectedEncoder {
  enc := json.NewEncoder(w)
  enc.SetEscapeHTML(false)
  return enc
}
```

zapcore/json_encoder_test.go
```diff
@@ -171,3 +172,65 @@ func TestJSONEmptyConfig(t *testing.T) {
...
+type emptyReflectedEncoder struct {
+	writer io.Writer
+}
+
+func (enc *emptyReflectedEncoder) Encode(obj interface{}) error {
+	_, err := enc.writer.Write([]byte("{}"))
+	return err
+}
+
+func TestJSONCustomReflectionEncoder(t *testing.T) {
```
sywhang, line 186: minor nit: 
```suggestion
func TestJSONCustomReflectedEncoder(t *testing.T) {
```

zapcore/json_encoder_test.go
```diff
@@ -171,3 +172,65 @@ func TestJSONEmptyConfig(t *testing.T) {
...
+		NewReflectedEncoder: func(writer io.Writer) zapcore.ReflectedEncoder {
+			return &emptyReflectedEncoder{
+				writer: writer,
+			}
+		},
+	})
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			buf, err := enc.EncodeEntry(zapcore.Entry{
```
sywhang, line 224:
```suggestion
	for _, tt := range tests {
	        tt := tt
		t.Run(tt.name, func(t *testing.T) {
		        t.Parallel()
			buf, err := enc.EncodeEntry(zapcore.Entry{
```

sywhang (approved):

zapcore/json_encoder_test.go
abhinav: Oh, we can do that, although for that the encoder has to be instantiated per-test.
(Encoder isn't safe for concurrent use.)
I'll do it.

abhinav (approved):

abhinav:
Thanks!

PR #1044, opened 2022-01-04
added missing modules download

jklaw90:
I missed two more modules in [PR-1043](https://github.com/uber-go/zap/pull/1043)
Now the action builds skipping the all downloads from those modules.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -38,7 +38,9 @@ jobs:
     - name: Download Dependencies
       run: |
         go mod download
-        cd tools && go mod download
+        cd tools && go mod download && cd -
+        cd benchmarks && go mod download && cd -
+        cd zapgrpc/internal/test && go mod download && cd -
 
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -38,7 +38,9 @@ jobs:
     - name: Download Dependencies
       run: |
         go mod download
-        cd tools && go mod download
+        cd tools && go mod download && cd -
+        cd benchmarks && go mod download && cd -
+        cd zapgrpc/internal/test && go mod download && cd -
```
abhinav, line 43: We can use subshells to avoid cding back.

```suggestion
        (cd tools && go mod download)
        (cd benchmarks && go mod download)
        (cd zapgrpc/internal/test && go mod download)
```

jklaw90: good call!

abhinav (approved):

PR #1047, opened 2022-01-13
zapcore: Add ParseLevel

Techassi:
This adds the public ParseLevel function which enables the user to
construct a log level based on ASCII text input.

This re-uses the UnmarshalText method of Level for the heavy-lifting,
but it avoids the user having to declare a local `Level` variable first.

```go
var level zapcore.Level
err := level.UnmarshalText([]byte("debug"))

// versus
level, err := zapcore.ParseLevel("debug")
```

#### Usage

```go
level, err := zapcore.LevelFromString("info") // zapcore.InfoLevel, nil

level, err := zapcore.LevelFromString("DEBUG") // zapcore.DebugLevel, nil

level, err := zapcore.LevelFromString("FOO") // -2, "invalid level"
```

One thing I'm not quite sure is the returned Level value of `-2` when the user provided a invalid log level as text. Usually I would go for `-1` but this is already used by `zapcore.DebugLevel`. Let me know if you know any better solution.

### DIFF
--- a/zapcore/level.go
+++ b/zapcore/level.go
@@ -26,10 +26,7 @@ import (
 	"fmt"
 )
 
-var (
-	errUnmarshalNilLevel = errors.New("can't unmarshal a nil *Level")
-	errInvalidLevel      = errors.New("invalid level")
-)
+var errUnmarshalNilLevel = errors.New("can't unmarshal a nil *Level")
 
 // A Level is a logging priority. Higher levels are more important.
 type Level int8
@@ -58,30 +55,19 @@ const (
 	_maxLevel = FatalLevel
 )
 
-// LevelFromString returns a level based on the lower-case or all-caps ASCII
-// representation of the log level.
+// UnmarshalLevel unmarshals a level based on the lower-case or all-caps ASCII
+// representation of the log level. If the provided ASCII representation is
+// invalid an error is returned.
 //
 // This is particularly useful when dealing with text input to configure log
 // levels.
-func LevelFromString(text string) (Level, error) {
-	switch text {
-	case "debug", "DEBUG":
-		return DebugLevel, nil
-	case "info", "INFO", "": // make the zero value useful
-		return InfoLevel, nil
-	case "warn", "WARN":
-		return WarnLevel, nil
-	case "error", "ERROR":
-		return ErrorLevel, nil
-	case "dpanic", "DPANIC":
-		return DPanicLevel, nil
-	case "panic", "PANIC":
-		return PanicLevel, nil
-	case "fatal", "FATAL":
-		return FatalLevel, nil
-	default:
-		return -2, errInvalidLevel
+func UnmarshalLevel(text string) (Level, error) {
+	level := DebugLevel
+	err := level.UnmarshalText([]byte(text))
+	if err != nil {
+		return level, err
 	}
+	return level, nil
 }
 
 // String returns a lower-case ASCII representation of the log level.
--- a/zapcore/level.go
+++ b/zapcore/level.go
@@ -26,7 +26,10 @@ import (
 	"fmt"
 )
 
-var errUnmarshalNilLevel = errors.New("can't unmarshal a nil *Level")
+var (
+	errUnmarshalNilLevel = errors.New("can't unmarshal a nil *Level")
+	errInvalidLevel      = errors.New("invalid level")
+)
 
 // A Level is a logging priority. Higher levels are more important.
 type Level int8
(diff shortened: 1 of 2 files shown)


abhinav (commented):
Hey there! Thanks for the contribution.

I'm not opposed to a convenience function, but I think:

1. It should be named UnmarshalLevel to clearly state that we're using the unmarshaling logic, which brings me to
2. It should not re-implement the unmarshaling logic. It should use the UnmarshalText method

Techassi:
I adjusted the implementation.

#### Adjusted Usage

```go
level, err := zapcore.UnmarshalLevel("info") // zapcore.InfoLevel, nil

level, err := zapcore.UnmarshalLevel("DEBUG") // zapcore.DebugLevel, nil

level, err := zapcore.UnmarshalLevel("FOO") // zapcore.DebugLevel, "unrecognized level: FOO"
```

I was also thinking about adding a `zap.NewAtomicLevelFrom(string) AtomicLevel` convenience function to mirror the newly added `UnmarshalLevel` function.

zapcore/level.go
```diff
@@ -55,6 +55,21 @@ const (
...
 )
 
+// UnmarshalLevel unmarshals a level based on the lower-case or all-caps ASCII
+// representation of the log level. If the provided ASCII representation is
+// invalid an error is returned.
+//
+// This is particularly useful when dealing with text input to configure log
+// levels.
+func UnmarshalLevel(text string) (Level, error) {
+	level := DebugLevel
```
abhinav, line 65: We should use the zero value as the default here

```suggestion
	var level Level
```

zapcore/level.go
```diff
@@ -55,6 +55,21 @@ const (
...
+	err := level.UnmarshalText([]byte(text))
+	if err != nil {
+		return level, err
+	}
+	return level, nil
```
abhinav, line 70: minor change: we can simplify this:
```suggestion
	return level, err
```

zapcore/level.go
```diff
@@ -55,6 +55,21 @@ const (
 	_maxLevel = FatalLevel
 )
 
+// UnmarshalLevel unmarshals a level based on the lower-case or all-caps ASCII
+// representation of the log level. If the provided ASCII representation is
+// invalid an error is returned.
+//
+// This is particularly useful when dealing with text input to configure log
+// levels.
+func UnmarshalLevel(text string) (Level, error) {
```
abhinav, line 64: I know I suggested Unmarshal but I realize now that Parse might be more appropriate.
Similar to time.Parse, for example.

```suggestion
// ParseLevel parses a level based on the lower-case or all-caps ASCII
// representation of the log level. If the provided ASCII representation is
// invalid an error is returned.
//
// This is particularly useful when dealing with text input to configure log
// levels.
func ParseLevel(text string) (Level, error) {
```

abhinav (commented):
Could you please add a couple test cases for the new function in level_test.go? The cases you demonstrated in the comments should suffice.

RE: AtomicLevelFromString: Seems like it might be unnecessary since you can just call that on the result of UnmarshalLevel?

abhinav:
Oh, re: this:

> One thing I'm not quite sure is the returned Level value of -2 when the user provided a invalid log level as text. Usually I would go for -1 but this is already used by zapcore.DebugLevel. Let me know if you know any better solution.

The contract for `ParseLevel`, like most functions that return `error` as their last argument, is that if `err != nil`, the other value is invalid. So the user should not assume anything about the returned level if `err != nil`.

Techassi:
I added the tests.

RE: AtomicLevelFromString: This seems like a similar convenience function as `ParseLevel`. The current usage is:

```go
level := zap.NewAtomicLevel()
err := level.UnmarshalText(byte("debug"))
```

We could condense this into:

```go
// Or zap.NewAtomicLevelFrom()
level := zap.NewAtomicLevelFromString("debug") // level = DebugLevel
```

abhinav:
> I added the tests.
> 
> RE: AtomicLevelFromString: This seems like a similar convenience function as `ParseLevel`. The current usage is:
> 
> ```go
> level := zap.NewAtomicLevel()
> err := level.UnmarshalText(byte("debug"))
> ```
> 
> We could condense this into:
> 
> ```go
> // Or zap.NewAtomicLevelFrom()
> level := zap.NewAtomicLevelFromString("debug") // level = DebugLevel
> ```

Fair enough. If you have need for it, a zap.ParseAtomicLevel would be fine.

Techassi:
Great! I will add this asap.

PS: Thanks for the great input on the tests as this is the first time I work with tests in Go.

abhinav:
Thanks!

PR #1048, opened 2022-01-13
zap: Add ParseAtomicLevel func

Techassi:
As discussed in #1047 this PR adds a `zap.ParseAtomicLevel()` func which enables the user to construct a log level based on ASCII text input.

#### Usage

```go
level, err := zap.ParseAtomicLevel("info") // InfoLevel, nil

level, err := zap.ParseAtomicLevel("DEBUG") // DebugLevel, nil

level, err := zap.ParseAtomicLevel("FOO") // InfoLevel, "unrecognized level: "FOO""
```

### DIFF
--- a/level.go
+++ b/level.go
@@ -86,7 +86,7 @@ func NewAtomicLevelAt(l zapcore.Level) AtomicLevel {
 	return a
 }
 
-// ParseAtomicLevel parses a level based on the lower-case or all-caps ASCII
+// ParseAtomicLevel parses an AtomicLevel based on a lowercase or all-caps ASCII
 // representation of the log level. If the provided ASCII representation is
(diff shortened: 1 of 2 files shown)


level.go
```diff
@@ -86,6 +86,23 @@ func NewAtomicLevelAt(l zapcore.Level) AtomicLevel {
 	return a
 }
 
+// ParseAtomicLevel parses a level based on the lower-case or all-caps ASCII
```
sywhang, line 89:
```suggestion
// ParseAtomicLevel parses an AtomicLevel based on a lowercase or all-caps ASCII
```

sywhang (approved):
Thank you for the contributions! 👍

PR #1053, opened 2022-01-25
Avoid CallersFrames from escaping

prashantv:
The `runtime.CallersFrames` method doesn't do much other than returning a struct pointer wrapping the passed in `pcs`. Calling this method twice ends up simplifying some of the logic (not having to format the first stack separately from the remaining), and ends up brining the performance of `TakeStacktrace` similar to `master`.

```
# benchstat comparing 3 runs of master vs this branch, each run using benchtime=10s
name            old time/op    new time/op    delta
TakeStacktrace    1.47µs ± 2%    1.48µs ± 2%   ~     (p=1.000 n=3+3)

name            old alloc/op   new alloc/op   delta
TakeStacktrace      496B ± 0%      496B ± 0%   ~     (all equal)

name            old allocs/op  new allocs/op  delta
TakeStacktrace      2.00 ± 0%      2.00 ± 0%   ~     (all equal)
```

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -332,7 +332,7 @@ func (log *Logger) check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
 		return ce
 	}
 
-	frame, more := stack.Next()
+	frame := stack.First()
 
 	if log.addCaller {
 		ce.Caller = zapcore.EntryCaller{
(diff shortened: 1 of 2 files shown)


abhinav (approved):
For the case #1052 optimizes, this is slightly slower but still better overall. Worth it versus regressing on the other cases.

```
name \ time/op            master.txt  1052.txt    1053.txt
AddCallerAndStacktrace-8  869ns ± 4%  670ns ± 2%  761ns ± 2%

name \ alloc/op           master.txt  1052.txt    1053.txt
AddCallerAndStacktrace-8   665B ± 0%   449B ± 0%   657B ± 0%

name \ allocs/op          master.txt  1052.txt    1053.txt
AddCallerAndStacktrace-8   5.00 ± 0%   3.00 ± 0%   4.00 ± 0%
```

PR #1058, opened 2022-02-04
Check if EncodeLevel is set to avoid nil pointer dereference

aerosol:
Hello,

Apologies in advance, this is my first go PR.

This PR checks if `EncodeLevel` is set before calling it on `final` encoder clone. This prevents panic on nil pointer dereference in case it hasn't been supplied by the user. For an untrained eye, the stacktrace wasn't making it super clear that `EncodeLevel` _must_ be provided along with `LevelKey`, so hopefully an explicit check hints at it for someone reading the source code. 

I wasn't sure if maybe providing a sensible default here would've been better, but looking at the [console logger](https://github.com/uber-go/zap/blob/a9218159e1d1842b4df993bfa840e3a0c18a0c34/zapcore/console_encoder.go#L83), a similar check is made.

Thanks.

### DIFF
--- a/zapcore/json_encoder_test.go
+++ b/zapcore/json_encoder_test.go
@@ -137,7 +137,6 @@ func TestJSONEncodeEntry(t *testing.T) {
 }
 
 func TestNoEncodeLevelSupplied(t *testing.T) {
-
 	enc := zapcore.NewJSONEncoder(zapcore.EncoderConfig{
 		MessageKey:     "M",
 		LevelKey:       "L",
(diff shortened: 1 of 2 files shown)


abhinav (approved):
Thanks for the fix!

zapcore/json_encoder_test.go
```diff
@@ -136,6 +136,36 @@ func TestJSONEncodeEntry(t *testing.T) {
 	}
 }
 
+func TestNoEncodeLevelSupplied(t *testing.T) {
+
```
abhinav, line 140:
```suggestion
```

PR #1071, opened 2022-03-19
Add Objects and ObjectValues field constructors

abhinav:
This pull request adds two new field constructors for slices, parameterized
over the types of the values in the slices:

```go
func Objects[T zapcore.ObjectMarshaler](key string, value []T) Field
```

This turns a slice of any Zap marshalable object into a Zap field.

```go
func ObjectValues[T any, P objectMarshalerPtr[T]](key string, values []T) Field
```

This is a variant of Objects for the case where `*T` implements
`zapcore.ObjectMarshaler` but we have a `[]T`.

Together, these two field constructors obviate the need for writing
`ArrayMarshaler` implementations for slices of objects:

```go
// Before //////////////

type userArray []*User

func (uu userArray) MarshalLogArray(enc zapcore.ArrayEncoder) error {
    for _, u := range uu {
        if err := enc.AppendObject(u); err != nil {
            return err
        }
    }
    return nil
}

var users []*User = ..
zap.Array("users", userArray(users))

// Now /////////////////

var users []*User = ..
zap.Objects("users", users)
```

### Backwards compatibility

Both new field constructors are hidden behind a build tag, but use of type
parameters requires us to bump the `go` directive in the `go.mod` file to 1.18.

Note that this *does not* break Zap on Go 1.17.
Per the [documentation for the `go` directive](https://go.dev/ref/mod#go-mod-file-go),

> If an older Go version builds one of the module’s packages and encounters a
> compile error, the error notes that the module was written for a newer Go
> version.

It only breaks our ability to run `go mod tidy` inside Zap from Go 1.17,
which is okay because with the `go` directive, we're stating that we're
developing Zap while on Go 1.18.

### Linting

staticcheck support for type parameters is expected in a few weeks.
Meanwhile, this disables staticcheck for the `make lint` target.
(I considered switching linting to Go 1.17 until then, but that isn't
enough because the formatting linter expects valid Go code--which type
parameters aren't in 1.17.)

### DIFF
--- a/array_go118.go
+++ b/array_go118.go
@@ -1,0 +1,54 @@
+// Copyright (c) 2022 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


array_go118.go
```diff
@@ -0,0 +1,124 @@
...
+func ObjectValues[T any, P objectMarshalerPtr[T]](key string, values []T) Field {
+	return Array(key, objectValues[T, P](values))
+}
+
+type objectValues[T any, P objectMarshalerPtr[T]] []T
+
+func (os objectValues[T, P]) MarshalLogArray(arr zapcore.ArrayEncoder) error {
+	for i := range os {
+		// It is necessary for us to explicitly reference the "P" type.
+		// We cannot simply pass "&o" to AppendObject because its type
```
sywhang, line 113:
```suggestion
		// We cannot simply pass "&os[i]" to AppendObject because its type
```

sywhang (approved):
neat :) left a minor nit on the comment but LGTM otherwise.

PR #1075, opened 2022-03-30
zapcore/encoder_test: Use yaml.v3

abhinav:
encoder_test decodes YAML to verify the behavior of EncoderConfig.
This behavior is the same between yaml.v2 and yaml.v3.
Since dependencies that use YAML have already moved on to yaml.v3,
we should as well.

Note that this drops yaml.v2 from our go.mod entirely.


prashantv (approved):
🥳

sywhang (approved):

PR #1079, opened 2022-04-08
Add WithOptions to SugaredLogger

### LINKED ISSUE #1072: Add WithOptions() for SugaredLogger
Currently there is WithOptions() only for unsugared loggers, so Desugar() and then Sugar() needs to be called in order to add options to a SugaredLogger.

craigpastro:
Closes #1072.

I added a simple test for this feature. One thought to add more comprehensive tests would be to replace the following line in the [`withSugar`](https://github.com/uber-go/zap/blob/212dcb966b47f20c741c821bf5f89f127a8ca7b9/common_test.go#L43) function in `common_test.go`:
```
withLogger(t, e, opts, func(logger *Logger, logs *observer.ObservedLogs) { f(logger.Sugar(), logs) })
```
with
```
withLogger(t, e, nil, func(logger *Logger, logs *observer.ObservedLogs) { f(logger.Sugar().WithOptions(opts...), logs) })
```
but maybe that is too much. WDYT?

Thank you!

### DIFF
--- a/sugar.go
+++ b/sugar.go
@@ -61,8 +61,8 @@ func (s *SugaredLogger) Named(name string) *SugaredLogger {
 	return &SugaredLogger{base: s.base.Named(name)}
 }
 
-// WithOptions clones the base Logger, applies the supplied Options, wraps
-// the resulting Logger, and returns it. It's safe to use concurrently.
+// WithOptions clones the current SugaredLogger, applies the supplied Options,
+// and returns the result. It's safe to use concurrently.
 func (s *SugaredLogger) WithOptions(opts ...Option) *SugaredLogger {
 	base := s.base.clone()
 	for _, opt := range opts {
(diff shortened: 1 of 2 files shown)


sugar.go
```diff
@@ -61,6 +61,16 @@ func (s *SugaredLogger) Named(name string) *SugaredLogger {
 	return &SugaredLogger{base: s.base.Named(name)}
 }
 
+// WithOptions clones the base Logger, applies the supplied Options, wraps
+// the resulting Logger, and returns it. It's safe to use concurrently.
```
abhinav, line 65: I think this is reporting too much of the implementation.
We can simplify this:

```suggestion
// WithOptions clones the current SugaredLogger, applies the supplied Options,
// and returns the result. It's safe to use concurrently.
```

abhinav (approved):
Thanks!

PR #1088, opened 2022-04-22
Add WithFatalHook/CheckedEntry.After to replace OnFatal/CheckedEntry.Should

### LINKED ISSUE #1086: Fatal should deterministically calculate the POSIX return code
The _Fatal_ method is calling `os.Exit(1)` regardless of a message. This isn't the best _UX_ choice. The _POSIX_ return code could be used for quick assessment about a process failure. When the retcode is always equal `1` for any error, this can't be utilized.

Tools like Docker or Kubernetes report back the return code of the process. By using a deterministic algorithm to calculate the _POSIX_ return code, a quick debugging is possible. Users could indicate if the process is failing because of the same reason, without looking into logs.

The best would be to rely on _Fatal_'s message or on an _error_ object if it's attached as a _Field_.

cardil:
Fixes #1086 

## Changes

- Deterministic calculate the POSIX return code for Fatal method

### DIFF
--- a/logger_test.go
+++ b/logger_test.go
@@ -213,15 +213,8 @@ func TestLoggerAlwaysFatals(t *testing.T) {
 	// Users can disable writing out fatal-level logs, but calls to logger.Fatal()
 	// should still terminate the process.
 	withLogger(t, FatalLevel+1, nil, func(logger *Logger, logs *observer.ObservedLogs) {
-		stub := exit.WithStub(func() { logger.Fatal("foo") })
+		stub := exit.WithStub(func() { logger.Fatal("") })
 		assert.True(t, stub.Exited, "Expected calls to logger.Fatal to terminate process.")
-		assert.Equal(t, 1, stub.Code, "Expected calls to logger.Fatal to terminate process with predictable retcode.")
-
-		logger = logger.WithOptions(OnFatal(zapcore.WriteThenPosixExit))
-		err := errors.New("bar")
-		stub = exit.WithStub(func() { logger.Fatal("foo", Error(err)) })
-		assert.True(t, stub.Exited, "Expected calls to logger.Fatal to terminate process.")
-		assert.Equal(t, 129, stub.Code, "Expected calls to logger.Fatal to terminate process with predictable retcode.")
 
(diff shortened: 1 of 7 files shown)


sywhang:
Thanks for the contribution @cardil! It mostly looked good; i just made a few tweaks to remove a race condition and some breaking changes in the change. 

@abhinav PTAL when you get a chance.

logger_test.go
```diff
@@ -579,6 +579,29 @@ func TestLoggerCustomOnFatal(t *testing.T) {
...
+	h := &customWriteHook{}
+	withLogger(t, InfoLevel, opts(WithFatalHook(h)), func(logger *Logger, logs *observer.ObservedLogs) {
+		recovered := make(chan interface{})
+		go func() {
+			defer func() {
+				recovered <- recover()
+			}()
+			logger.Fatal("")
+		}()
+		<-recovered
```
abhinav: The goroutine or channel isn't necessary.
The logger.Fatal won't panic because we're overriding it.

zapcore/entry.go
```diff
@@ -164,10 +170,28 @@ const (
...
+		panic(ce.Message)
+	case WriteThenFatal:
+		exit.Exit()
+	}
+}
+
+var _ CheckWriteHook = WriteThenNoop
+var _ CheckWriteHook = WriteThenGoexit
+var _ CheckWriteHook = WriteThenPanic
+var _ CheckWriteHook = WriteThenFatal
```
abhinav: The check is type-level, not value-level so we can just verify that CheckWriteAction implements the interface.

abhinav (approved):

zapcore/entry.go
```diff
@@ -152,6 +151,13 @@ type Entry struct {
 	Stack      string
 }
 
+// CheckWriteHook allows to customize the action to take after a Fatal log entry
+// is processed.
+type CheckWriteHook interface {
+	// OnWrite gets invoked when an entry is written
+	OnWrite(*CheckedEntry, []Field)
```
prashantv, line 158: It may be useful to document that the fields here are not all fields on the logger, but just the fields specific to the message.

PR #1118, opened 2022-06-22
Logger: add Log method

### LINKED ISSUE #1084: Specify log level by parameter
Hello!

Would it be possible to have something like the following added to the logger ?

```
func (log *Logger) Log(level zapcore.Level, msg string, fields ...Field) {
	if ce := log.check(level, msg); ce != nil {
		ce.Write(fields...)
	}
}
```

this would allow anyone to also make dynamic calls i.e. `log.Log(ErrorLevel, "message", fields...)`. 

Thanks!

craigpastro:
Unfortunately I don't see how this could be used internally without breaking caller skip.

Resolve #1084.

### DIFF
--- a/sugar.go
+++ b/sugar.go
@@ -279,7 +279,9 @@ func (s *SugaredLogger) log(lvl zapcore.Level, template string, fmtArgs []interf
 	}
 
 	msg := getMessage(template, fmtArgs)
-	s.base.Log(lvl, msg, s.sweetenFields(context)...)
+	if ce := s.base.Check(lvl, msg); ce != nil {
+		ce.Write(s.sweetenFields(context)...)
+	}
 }
 
 // logln message with Sprintln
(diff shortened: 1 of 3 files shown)


sugar.go
```diff
@@ -279,9 +279,7 @@ func (s *SugaredLogger) log(lvl zapcore.Level, template string, fmtArgs []interf
 	}
 
 	msg := getMessage(template, fmtArgs)
-	if ce := s.base.Check(lvl, msg); ce != nil {
-		ce.Write(s.sweetenFields(context)...)
-	}
+	s.base.Log(lvl, msg, s.sweetenFields(context)...)
```
abhinav, line 282: I would revert this and the other change in this file.

The Check version has the advantage that we don't spend CPU on `sweetenFields` unless we're actually going to log the message. So if the log level is too low or the message gets sampled, we won't pay to serialize the fields.

craigpastro: Ah, that is a good point. Thank you! Reverted.

abhinav (approved):
> Unfortunately I don't see how this could be used internally without breaking caller skip.

Yeah, fair. I don't see an obvious way to do that without some refactoring.
This should be fine as-is.

CC @sywhang @mway

sywhang (approved):

abhinav:
Thanks!

PR #1128, opened 2022-07-12
Refactor to reuse Log method

sywhang:
This refactors the implementation of the Log method to reuse it
in the other level variants of the method (i.e. `Debug`, `Info`, etc.).

To do this, a skip argument was added to `check` to allow its callers to
specify how many frames to skip in the log, and a private `log` method
was added which calls this `check` method with the appropriate frame
skip count.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -187,47 +187,46 @@ func (log *Logger) With(fields ...Field) *Logger {
 // is enabled. It's a completely optional optimization; in high-performance
 // applications, Check can help avoid allocating a slice to hold fields.
 func (log *Logger) Check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
-	return log.check(lvl, msg)
+	return log.check(lvl, msg, 0)
+}
+
+func (log *Logger) log(lvl zapcore.Level, msg string, skip int, fields ...Field) {
+	if ce := log.check(lvl, msg, skip); ce != nil {
+		ce.Write(fields...)
+	}
 }
 
 // Log logs a message at the specified level. The message includes any fields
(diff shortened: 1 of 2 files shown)


logger.go
```diff
@@ -187,47 +187,46 @@ func (log *Logger) With(fields ...Field) *Logger {
...
 // applications, Check can help avoid allocating a slice to hold fields.
 func (log *Logger) Check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
-	return log.check(lvl, msg)
+	return log.check(lvl, msg, 0)
+}
+
+func (log *Logger) log(lvl zapcore.Level, msg string, skip int, fields ...Field) {
+	if ce := log.check(lvl, msg, skip); ce != nil {
+		ce.Write(fields...)
+	}
```
mway, line 196: nit: Let's move `Logger.log` below the exported definitions (maybe right above or below `Logger.check`).

logger.go
```diff
@@ -187,47 +187,46 @@ func (log *Logger) With(fields ...Field) *Logger {
 // is enabled. It's a completely optional optimization; in high-performance
 // applications, Check can help avoid allocating a slice to hold fields.
 func (log *Logger) Check(lvl zapcore.Level, msg string) *zapcore.CheckedEntry {
-	return log.check(lvl, msg)
+	return log.check(lvl, msg, 0)
```
mway, line 190: I might be misreading, but should this be `1` instead? And for log methods below, should those be `3`? (There was previously a `const` offset of `2`, but we've since added another intermediate call.)

Can we add a test to validate the skips?

PR #1134, opened 2022-07-21
chore: use errors.New instead of format function without args for error

sashamelentyev:


sywhang (approved):
LGTM, thanks for the contribution.

Issue #1078, opened 2022-04-05
Leveled fields: Fields present only at certain log levels


abhinav:
An issue that users sometimes have is this:

They have a log statement like the following,

```
logger.Info("processing request", zap.String("fromUser", fromUser))
```

They would like to include additional information in that log statement, but only if the debug level is enabled.

Currently, the options are:

1. Copy paste the log statement

```
if debugLevel {
  logger.Info("processing request", zap.String("fromUser", fromUser), zap.Object("request", request))
} else {
  logger.Info("processing request", zap.String("fromUser", fromUser))
}
```

2. Add another log statement

```
logger.Info("processing request", zap.String("fromUser", fromUser))
logger.Debug("processing request", zap.Object("request", request))
```

3. custom helpers like the following:

```
func levelField(log *zap.Logger, lvl zap.Level, field zap.Field) zap.Field {
  if !log.Core().Enabled(lvl) { return zap.Skip() }
  return field
}
```

```
logger.Info("processing request",
  zap.String("fromUser", fromUser),
  levelField(logger, zap.DebugLevel, zap.Object("request", request)))
```

I think it might be valuable for Zap to natively provide some concept of "leveled fields"—fields present only at certain log levels—based on something like the levelField helper above.

I don't have an exact design in mind, but Zap-native support for something like this would likely be able to drop that extra `log` argument.

abhinav:
Summarizing discussion with @mway here:

Matt proposed the following options:

1. Add options to existing field constructors.

    ```go
    func String(key, value string, opts ...FieldOption)
    
    type FieldOption
      func FieldLevel(Level) FieldOption
    ```

2. Parallel constructors.

    ```go
    func String(key, value string)
    func StringLevel(key, value string, lvl Level)
    ```

3. Method on zapcore.Field.

    ```go
    package zapcore
    
    type Field struct{ ... }
        func (Field) IfLevel(Level) Field
    ```

4. Name-based filtering at the logger level:

    ```go
    func WithFieldLevelFilters(map[string]Level) Option
    ```

---

My thoughts on the options:

1. Breaking change. I would like to avoid that for such a small feature.
2. Too much noise. It'll double the number of field constructors.
3. I like this. Maybe we can come up with a better name.
4. Stringly-typed API. Would prefer to avoid. Requires field name to be stable/unique across the application.

So *if* we were going to solve this problem, based on current options, I lean (3).
I'm also okay with saying that this isn't something Zap should support natively at this time,
[...]

PR #1135, opened 2022-07-21
feat: add debug fields

sashamelentyev:
#1078


sashamelentyev:
If this ok, i add all level field functions

PR #1139, opened 2022-08-01
zapcore/BufferedWriteSyncer: More docs

abhinav:
Better document how to use BufferedWriteSyncer in a real application.

Highlight that a `Stop()` method call should be deferred,
and how to change the default parameters of a BufferedWriteSyncer.

### DIFF
--- a/zapcore/buffered_write_syncer.go
+++ b/zapcore/buffered_write_syncer.go
@@ -43,6 +43,37 @@ const (
 //
 // BufferedWriteSyncer is safe for concurrent use. You don't need to use
 // zapcore.Lock for WriteSyncers with BufferedWriteSyncer.
+//
+// To set up a BufferedWriteSyncer, construct a WriteSyncer for your log
+// destination (*os.File is a valid WriteSyncer), wrap it with
+// BufferedWriteSyncer, and defer a Stop() call for when you no longer need the
+// object.
+//
+//   func main() {
+//     ws := ... // your log destination
+//     bws := &zapcore.BufferedWriteSyncer{WS: ws}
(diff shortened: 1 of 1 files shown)


sywhang (approved):
👍

joni-jones:
Thanks @abhinav for the documentation.

I'd like to make a few suggestions.

As documentation for `BufferedWriteSyncer` wasn't there, I was researching its usage on my own.
This is how I implemented it.

```go
var cfg zap.Config
var enc zapcore.Encoder
cfg = zap.NewProductionConfig()
enc = zapcore.NewJSONEncoder(cfg.EncoderConfig)
sink, closeOut, err := zap.Open(cfg.OutputPaths...)
if err != nil {
    _, err = fmt.Fprintf(os.Stderr, "Could not get log Output paths." + err.Error())
}
errSink, _, err := zap.Open(cfg.ErrorOutputPaths...)
if err != nil {
    _, err = fmt.Fprintf(os.Stderr, "Could not get log Error Output paths." + err.Error())
    closeOut()
}

ws := &zapcore.BufferedWriteSyncer{WS: sink}
Logger = zap.New(
    zapcore.NewCore(enc, ws, cfg.Level),
    BuildOptions(cfg, errSink)...,
)
```

I still want to use standard `lockedWriteSyncer` but wrapped by `BufferedWriteSyncer`, so the logs will be written only when the buffer is full or after a specified period of time.

As, `Config.openSinks` and `Config.buildOptions` are not available for the client code, I had to mimic `Config.openSinks` behavior and completely copy `Config.buildOptions`. Is there a better way to do it?

PR #1143, opened 2022-08-04
Expand LevelEnabler interface by method Level() Level

ghost:


abhinav (changes requested):
Thanks for the contribution, @krupyansky, but adding a new method to an existing interface is a breaking change. We're not currently entertaining breaking changes to Zap.

If you share what problem you're trying to solve with this, we can work with you to try to find a solution to your problem.

krupyansky:
> Thanks for the contribution, @krupyansky, but adding a new method to an existing interface is a breaking change. We're not currently entertaining breaking changes to Zap.
> 
> If you share what problem you're trying to solve with this, we can work with you to try to find a solution to your problem.

I need an external package to be able to get the logging level from zap logger.

For logrus, this problem is solved in the following way:
https://pkg.go.dev/github.com/sirupsen/logrus#Logger.GetLevel
https://pkg.go.dev/github.com/sirupsen/logrus#Entry

I need a similar feature from the logger zap.

abhinav:
Superseded by #1147.
Thanks for the initial contribution and opening the discussion around it, @krupyansky!

krupyansky:
Thank you for your response and feedback, @abhinav!

PR #1146, opened 2022-08-15
Replace os.TempDir usage with t.TempDir

prashantv:
Current tests can leave files around if tests fail, and even though they
use rand, it's not seeded, different runs can share the same output file
which causes other failures (e.g., file exists when it's not expected
to).  Avoid this by using  t.TempDir which automatically removed after
tests have run.

### DIFF
--- a/zapcore/buffered_write_syncer_bench_test.go
+++ b/zapcore/buffered_write_syncer_bench_test.go
@@ -30,12 +30,11 @@ import (
 
 func BenchmarkBufferedWriteSyncer(b *testing.B) {
 	b.Run("write file with buffer", func(b *testing.B) {
-		file, err := os.CreateTemp("", "test.log")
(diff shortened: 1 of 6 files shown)


zapcore/buffered_write_syncer_bench_test.go
```diff
@@ -30,7 +30,7 @@ import (
 
 func BenchmarkBufferedWriteSyncer(b *testing.B) {
 	b.Run("write file with buffer", func(b *testing.B) {
-		file, err := os.CreateTemp("", "log")
+		file, err := os.CreateTemp("", "test.log")
```
abhinav, line 33: Why doesn't this use b.TempDir()?

```suggestion
		file, err := os.CreateTemp(b.TempDir(), "test.log")
```

abhinav (approved):
lgtm minus one missed case. will fix

PR #1147, opened 2022-08-15
zapcore: Add LevelOf(LevelEnabler), UnknownLevel

### LINKED ISSUE #1144: Getting level from zap logger
How can I get log level from zap logger?

For logrus, this problem is solved in the following way:
https://pkg.go.dev/github.com/sirupsen/logrus#Logger.GetLevel
https://pkg.go.dev/github.com/sirupsen/logrus#Entry

I need a similar feature from zap logger.

abhinav:
Add a new function LevelOf that reports the minimum enabled log level
for a LevelEnabler.

This works by looping through all known Zap levels in-order,
returning the newly introduced UnknownLevel if none of the known levels
are enabled.

A LevelEnabler or Core implementation may implement the `Level() Level`
method to override and optimize this behavior.
AtomicLevel already implemented this method.
This change adds the method to all Core implementations shipped with
Zap.

Note:
UnknownLevel is set at FatalLevel+1 to account for the possibility that
users of Zap are using DebugLevel-1 as their own custom log level--even
though this isn't supported, it's preferable not to break these users.
Users are less likely to use FatalLevel+1 since Fatal means "exit the
application."

Resolves #1144
Supersedes #1143, which was not backwards compatible

### DIFF
--- a/internal/level_enab.go
+++ b/internal/level_enab.go
@@ -1,0 +1,35 @@
+// Copyright (c) 2022 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package internal
+
+import "go.uber.org/zap/zapcore"
+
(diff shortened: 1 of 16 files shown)


internal/level_enab.go
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2022 Uber Technologies, Inc.
```
mway, line 1: nit: `level_enabler.go`

zapcore/hook_test.go
```diff
@@ -42,6 +43,11 @@ func TestHooks(t *testing.T) {
 
 	for _, tt := range tests {
 		fac, logs := observer.New(tt.coreLevel)
+
+		// sanity check
+		require.Equal(t, tt.coreLevel, LevelOf(fac),
+			"Original logger has the wrong level")
```
mway, line 49: style nit: All on one line, or one per line (here and below)

zapcore/level.go
```diff
@@ -53,6 +53,11 @@ const (
 
 	_minLevel = DebugLevel
 	_maxLevel = FatalLevel
+
+	// UnknownLevel is an invalid value for Level.
+	//
+	// Core implementations may panic if they see messages of this level.
+	UnknownLevel = _maxLevel + 1
```
mway, line 60: Can we reorganize these so that we have an invalid zero value? e.g.

```go
const (
  InvalidLevel = iota
  DebugLevel
  // ...
  FatalLevel

  _minLevel = DebugLevel
  _maxLevel = FatalLevel
)
```

mway, line 60: nit: `s/Unknown/Invalid/`

sywhang: Wouldn't this be a breaking change because the integer value of DebugLevel (and everything else) would change?

zapcore/tee_test.go
```diff
@@ -49,6 +49,41 @@ func TestTeeUnusualInput(t *testing.T) {
...
+		},
+		{
+			desc: "debug and warn",
+			give: []Core{warnLogger, debugLogger},
+			want: DebugLevel,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
```
sywhang, line 80: nit: t.Parallel()

sywhang (approved):

zapcore/level.go
abhinav: I lean towards not changing values of published constants, even if the line around whether it's a breaking change is a bit blurry. Such a change _will_ break someone, and we don't get much in exchange.

sywhang (approved):

mway (approved):

PR #1149, opened 2022-08-16
Wrap Open error and update TestOpen to check for fs.ErrNotExist

prashantv:
This is a prefactor for Windows support (#621).

`TestOpen` currently relies on a hardcoded error "no such file or
directory" if a file is not found. However, this error message is
OS-specific.  We can now rely on `errors.Is(err, fs.ErrNotExist)`
if we wrap errors using `%w` instead of `%v`.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -70,7 +70,7 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
 	for _, path := range paths {
 		sink, err := newSink(path)
 		if err != nil {
-			openErr = multierr.Append(openErr, fmt.Errorf("couldn't open sink %q: %v", path, err))
+			openErr = multierr.Append(openErr, fmt.Errorf("couldn't open sink %q: %w", path, err))
 			continue
 		}
 		writers = append(writers, sink)
--- a/writer.go
+++ b/writer.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Uber Technologies, Inc.
+// Copyright (c) 2016-2022 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/writer.go
+++ b/writer.go
@@ -70,7 +70,7 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
 	for _, path := range paths {
 		sink, err := newSink(path)
 		if err != nil {
-			openErr = multierr.Append(openErr, fmt.Errorf("couldn't open sink %q: %w", path, err))
+			openErr = multierr.Append(openErr, fmt.Errorf("open sink %q: %w", path, err))
 			continue
 		}
 		writers = append(writers, sink)
--- a/writer_test.go
+++ b/writer_test.go
@@ -23,14 +23,15 @@ package zap
 import (
 	"errors"
 	"io"
+	"io/fs"
 	"net/url"
 	"os"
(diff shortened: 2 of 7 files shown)


writer.go
```diff
@@ -70,7 +70,7 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
 	for _, path := range paths {
 		sink, err := newSink(path)
 		if err != nil {
-			openErr = multierr.Append(openErr, fmt.Errorf("couldn't open sink %q: %v", path, err))
+			openErr = multierr.Append(openErr, fmt.Errorf("couldn't open sink %q: %w", path, err))
```
abhinav, line 73: while we're playing here, consider also changing the error message to conform to [Code Review Comments > Error Strings](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

```suggestion
			openErr = multierr.Append(openErr, fmt.Errorf("open sink %q: %w", path, err))
```

writer_test.go
```diff
@@ -50,49 +51,106 @@ func TestOpenNoPaths(t *testing.T) {
...
+				defer cleanup()
+			}
 
-		if len(tt.errs) == 0 {
-			assert.NoError(t, err, "Unexpected error opening paths %v.", tt.paths)
-		} else {
-			msg := err.Error()
-			for _, expect := range tt.errs {
-				assert.Contains(t, msg, expect, "Unexpected error opening paths %v.", tt.paths)
+			if len(tt.wantNotFoundPaths) == 0 && tt.wantOtherErr == "" {
```
abhinav, line 134: optional: These are three nearly exclusive code paths.
We can use a switch here.

```
switch {
case len(tt.wantNotFoundPaths) > 0:
  // ..

case len(tt.wantOtherErr) > 0:
  // ...

default:
  assert.NoError(t, err, ...)
}
```

Alternatively, given that the shared code is just a couple lines, we can also separate these into two or three separate test functions: success, not-found-paths, and other-errors.

writer_test.go
```diff
@@ -50,49 +51,106 @@ func TestOpenNoPaths(t *testing.T) {
...
 			}
-		}
+
+			require.Error(t, err)
+			if len(tt.wantNotFoundPaths) > 0 {
+				errs := multierr.Errors(err)
+				require.Len(t, errs, len(tt.wantNotFoundPaths))
+				for i, err := range errs {
+					assert.ErrorIs(t, err, fs.ErrNotExist)
+					assert.Contains(t, err.Error(), tt.wantNotFoundPaths[i], "missing path in error")
```
abhinav, line 145: minor: can use assert.ErrorContains here.
```suggestion
					assert.ErrorContains(t, err, tt.wantNotFoundPaths[i], "missing path in error")
```

(If the method is missing, we might need to update testify.)

writer_test.go
```diff
@@ -50,49 +51,106 @@ func TestOpenNoPaths(t *testing.T) {
...
+				}
+			}
+
+			if tt.wantOtherErr != "" {
+				msg := err.Error()
+				assert.Contains(t, msg, tt.wantOtherErr, "Unexpected error opening paths %v.", tt.paths)
```
abhinav, line 151: same, can use assert.ErrorContains.
```suggestion
				assert.ErrorContains(t, err, tt.wantOtherErr, "Unexpected error opening paths %v.", tt.paths)
```

abhinav (approved):
Thanks! Left a few minor comments.

abhinav (approved):

PR #1155, opened 2022-08-22
Add Stringers field constructor for array of objects implementing fmt.Stringer method

saurabh95:
Add a new Zap field constructor that encodes an array of zero or more
objects which implement fmt.Stringer interface into a Zap array.

Usage:

//	type Request struct{ ... }
//	func (a Request) String() string
//
//	var requests []Request = ...
//	logger.Info("sending requests", zap.Stringers("requests", requests))

Credits: @zmanji, @abhinav for the suggestions

### DIFF
--- a/array_go118.go
+++ b/array_go118.go
@@ -23,7 +23,11 @@
 
 package zap
 
-import "go.uber.org/zap/zapcore"
+import (
+	"fmt"
+
+	"go.uber.org/zap/zapcore"
+)
 
 // Objects constructs a field with the given key, holding a list of the
 // provided objects that can be marshaled by Zap.
@@ -119,6 +123,34 @@ func (os objectValues[T, P]) MarshalLogArray(arr zapcore.ArrayEncoder) error {
 		if err := arr.AppendObject(p); err != nil {
 			return err
 		}
+	}
+	return nil
+}
+
+// Stringers constructs a field with the given key, holding a list of the
+// output provided by the value's String method
+//
+// Given an object that implements String on the value receiver, you
+// can log a slice of those objects with Objects like so:
+//
+//	type Request struct{ ... }
+//	func (a Request) String() string
+//
+//	var requests []Request = ...
+//	logger.Info("sending requests", zap.Stringers("requests", requests))
+//
+// Note that these objects must implement fmt.Stringer directly.
+// That is, if you're trying to marshal a []Request, the String method
+// must be declared on the Request type, not its pointer (*Request).
+func Stringers[T fmt.Stringer](key string, values []T) Field {
+	return Array(key, stringers[T](values))
+}
+
+type stringers[T fmt.Stringer] []T
+
+func (os stringers[T]) MarshalLogArray(arr zapcore.ArrayEncoder) error {
+	for _, o := range os {
+		arr.AppendString(o.String())
 	}
 	return nil
 }
--- a/array_go118_test.go
+++ b/array_go118_test.go
@@ -25,6 +25,7 @@ package zap
 
 import (
 	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


zmanji:
This looks good to me, I wonder if we should add some sort of benchmark to show the benefit of this, since I believe the only way before was to use `zap.Any`.

array_go118_test.go
```diff
@@ -179,3 +179,48 @@ func TestObjectsAndObjectValues_marshalError(t *testing.T) {
...
+
+func TestStringers(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		desc string
+		give Field
+		want []any
+	}{
+		{
```
abhinav, line 199: Can we also add a test for a `[]fmt.Stringer`? That should also work.

array_go118.go
```diff
@@ -122,3 +126,27 @@ func (os objectValues[T, P]) MarshalLogArray(arr zapcore.ArrayEncoder) error {
 	}
 	return nil
 }
+
+// Stringers constructs a field with the given key, holding a list of the
+// output provided by the value's String method
+//
+// Given an object that implements String on the value receiver, you
```
abhinav, line 133: Same caveat as Objects applies here. We should add a comment about it:

```
// Note that these objects must implement fmt.Stringer directly.
// That is, if you're trying to marshal a []Request, the String method
// must be declared on the Request type, not its pointer (*Request).
```

In a future change, we can add an ObjectValues equivalent if necessary that will let the String method be on `*Request` for a `[]Request`.

abhinav (commented):
nice. pretty straightforward.

CC @sywhang @mway

array_go118_test.go
```diff
@@ -179,3 +179,48 @@ func TestObjectsAndObjectValues_marshalError(t *testing.T) {
 		})
 	}
 }
+
+type stringerObject struct{
```
sywhang, line 183: nit: lint 
```suggestion
type stringerObject struct {
```

sywhang (approved):
👍 Thanks for the contribution. Changes LGTM modulo @abhinav's comments and a lint failure

abhinav (approved):
Thanks!

Pedro-Pessoa:
Hi! I'm wondering if there was a specific reason as to why generics were used in this. Wouldn't this accomplish the same result?

```go
func Stringers(key string, values []fmt.Stringer) zap.Field {
	return zap.Array(key, stringers(values))
}

type stringers []fmt.Stringer

func (os stringers) MarshalLogArray(arr zapcore.ArrayEncoder) error {
	for _, o := range os {
		arr.AppendString(o.String())
	}
	return nil
}
```

Thanks!

mway:
> Hi! I'm wondering if there was a specific reason as to why generics were used in this. Wouldn't this accomplish the same result?
> 
> ```go
> func Stringers(key string, values []fmt.Stringer) zap.Field {
> 	return zap.Array(key, stringers(values))
> }
> 
> type stringers []fmt.Stringer
> 
> func (os stringers) MarshalLogArray(arr zapcore.ArrayEncoder) error {
> 	for _, o := range os {
> 		arr.AppendString(o.String())
> 	}
> 	return nil
> }
> ```
> 
> Thanks!

You can't implicitly convert between `[]T` and `[]InterfaceThatTImplements` by default (you have to allocate storage and copy the contents). So if you have a type `Foo` that implements `fmt.Stringer`:

```go
foos := []Foo{ /* ... */ }
```

We then have to handle it as follows (without generics):

```go
func Stringers(key string, values []fmt.Stringer) zap.Field {
  // ...
}

// Stringers(foos) <- does not compile

converted := make([]fmt.Stringer, len(foos))
for i, f := range foos {
  converted[i] = f
}

Stringers(converted) // compiles
```

Whereas with generics, we can handle both `[]fmt.Stringer` and `[]T` (where `T` implements `fmt.Stringer`):

```go
func Stringers[T fmt.Stringer](key string, values []T) zap.Field {
  // ...
}

var (
  a = []fmt.Stringer{ /* ... */ }
  b = []Foo{ /* ... */ }
)

Stringers(a) // compiles
Stringers(b) // compiles
```

Thus the generic version gives callers more flexibility by (1) letting them use wider types than `fmt.Stringer` in `[]T` and (2) not requiring conversions.

Pedro-Pessoa:
That makes perfect sense. Thanks for the explanation!

Issue #1000, opened 2021-09-07
Windows support


abhinav:
Per #999, it appears that Zap, or at least zap.Config does not yet support Windows.
I suspect that besides filepath munging, we should work on WIndows just fine.
Opening this to track fixing this.

Ref #994

Issue #994, opened 2021-08-30
Question: WIndows - How to specify full path to log file name


stuartstein777:
I have this:

```go
func getLogger() (*zap.Logger, error) {
	dir, err := os.Executable()
	if err != nil {
		panic(err)
	}
	fullpath := filepath.Join(filepath.Dir(dir), "proxy.log")
	cfg := zap.NewProductionConfig()
	cfg.OutputPaths = []string{
		fullpath,
	}
	return cfg.Build()
}
```
fullpath here is `C:\Users\stuarts\Source\sentilan-agent-proxy-api\proxy.log`

When I try to log I get a nil dereference error. The logger that is returned is nil.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x7ff65d95a787]
``

However, this works.

```
func getLogger() (*zap.Logger, error) {
	cfg := zap.NewProductionConfig()
	cfg.OutputPaths = []string{
		"proxy.log",
	}
	return cfg.Build()
}
```

This logs to proxy.log in the same folder as the .exe, But we are running this s a service and its causing the file to be logged to windows system folder (as this is the working folder when its run a service).

What am I doing wrong here? I'd like to specify a full path to log file.

**Comment by abhinav** [comment]
Hey there! It looks like Zap (or at least zap.Config) doesn't quite work on Windows yet (experimented in #999).

I'll create a separate task for Windows support, but meanwhile, you should be able to work around this issue by constructing a logger directly with `zap.New` instead of `zap.Config`. Given the above, the rough equivalent is:

```go
enc := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())

ws, err := os.OpenFile(fullpath, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
if err != nil { ... }

core := zapcore.NewCore(enc, ws, zapcore.InfoLevel)

logger := zap.New(core)
```

PR #1159, opened 2022-08-25
Open absolute paths as files, limited Windows support

prashantv:
Ref #994, #1000

Alternate approach to #999, this approach brings absolute path support to Windows.
It does so by checking for absolute paths and handling them using the file factory directly.

To test different paths without trying to actually open them (UNC paths hit the network), this change also
prefactors the sink registry into a separate type that allows stubbing of the `os.OpenFile` call.

Note: This change does not bring full Windows support -- many tests still fail, and Windows file URIs don't work.


sink.go
```diff
@@ -70,16 +55,22 @@ func (e *errSinkNotFound) Error() string {
...
-// have a factory registered. Zap automatically registers a factory for the
-// "file" scheme.
-func RegisterSink(scheme string, factory func(*url.URL) (Sink, error)) error {
-	_sinkMutex.Lock()
-	defer _sinkMutex.Unlock()
+type nopCloserSink struct{ zapcore.WriteSyncer }
+
+func (nopCloserSink) Close() error { return nil }
+
+func newSinkRegistry() *sinkRegistry {
```
abhinav, line 64: nit: ordering: should the type definition for sinkRegistry be closer to the construtor?
we could move it right above this.

sink_test.go
```diff
@@ -87,11 +98,10 @@ func TestRegisterSinkErrors(t *testing.T) {
 		{"http*", "not a valid scheme"},
 	}
 
+	r := newSinkRegistry()
```
abhinav, line 101: There does not appear to be any shared state between individual test cases.
Should this be defined inside the subtest?

sink.go
```diff
@@ -45,6 +45,7 @@ type Sink interface {
 type sinkRegistry struct {
 	mu        sync.Mutex
 	factories map[string]func(*url.URL) (Sink, error) // keyed by scheme
+	openFile  func(string, int, os.FileMode) (*os.File, error)
```
abhinav, line 49: consider a comment adding that this is the same signature as os.OpenFile.

```suggestion
	openFile  func(string, int, os.FileMode) (*os.File, error) // == os.OpenFile
```

sink_windows_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWindowsPaths(t *testing.T) {
+	sr := newSinkRegistry()
+
+	var openFilename string
+	sr.openFile = func(filename string, _ int, _ os.FileMode) (*os.File, error) {
+		openFilename = filename
+		return nil, assert.AnError
```
abhinav, line 38: Not a blocker, but a function shared between subtests that modifies its lexical scope smells a bit.

How do you feel about pushing this setup inside the test logic as well:

```
t.Run(..., func(t *testing.T) {
  sr := newSinkRegistry()

  var openFilename string
  sr.openFile = func(...) (*os.File, error) {
    openFilename = filename
    return nil, assert.AnError
  }

  // ...

  assert.Equal(t, tt.path, openFilename, ...)
})
```

abhinav (approved):
Thanks! No blockers, left a few suggestions.

prashantv:
Thanks for the suggestions, updated to incorporate them.

abhinav (approved):

PR #1185, opened 2022-10-06
SugaredLogger: Turn error into zap.Error 

### LINKED ISSUE #1184: SugaredLogger: Turn error into zap.Error

With Zap's sugared logger, if you provide a plain error object to Infow/Errorw/etc.,
it does not desugar the error into a zap.Field.

```go
log.Warnw("something happened", err) // bad

log.Warnw("something happened", "err", err) // good
// desugars to
log.Desugar().Warn("something happened", zap.NamedError("err", err))
```

This is inconvenient for the majority case where there will be only one error.

We should turn the first case into,

```go
log.Desugar().Warn("something happened", zap.Error(err))
```

We could even be smart about it in desugarFields
and make sure there's only one instance of "err".
Others have to have a field name.

**Describe alternatives you've considered**

Add "err" next to every err argument.

No

craigpastro:
Resolves #1184.

Please let me know what you think! Thanks.

### DIFF
--- a/sugar.go
+++ b/sugar.go
@@ -349,7 +349,14 @@ func (s *SugaredLogger) sweetenFields(args []interface{}) []Field {
 			continue
 		}
 
-		// Make sure this element isn't a dangling key.
+		// If it is an error.
+		if err, ok := args[i].(error); ok {
+			fields = append(fields, Error(err))
+			i++
+			continue
+		}
+
+		// If it is not a singular error, make sure this element isn't a dangling key.
 		if i == len(args)-1 {
 			s.base.Error(_oddNumberErrMsg, Any("ignored", args[i]))
 			break
--- a/sugar.go
+++ b/sugar.go
@@ -337,11 +337,13 @@ func (s *SugaredLogger) sweetenFields(args []interface{}) []Field {
 		return nil
 	}
 
-	// Allocate enough space for the worst case; if users pass only structured
-	// fields, we shouldn't penalize them with extra allocations.
-	fields := make([]Field, 0, len(args))
-	var invalid invalidPairs
-	var seenError bool
+	var (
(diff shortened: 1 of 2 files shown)


abhinav (approved):
This looks great! Thanks!

@mway @sywhang what do you think?

sywhang (approved):
This looks good to me. Thanks for doing this work!

sugar.go
```diff
@@ -340,6 +341,7 @@ func (s *SugaredLogger) sweetenFields(args []interface{}) []Field {
 	// fields, we shouldn't penalize them with extra allocations.
 	fields := make([]Field, 0, len(args))
 	var invalid invalidPairs
+	var seenError bool
```
mway, line 344: nit: let's combine these into a `var` block

sugar_test.go
```diff
@@ -198,9 +214,10 @@ func TestSugarStructuredLogging(t *testing.T) {
 	}
 
 	// Common to all test cases.
+	err := errors.New("qux")
 	context := []interface{}{"foo", "bar"}
-	extra := []interface{}{"baz", false}
-	expectedFields := []Field{String("foo", "bar"), Bool("baz", false)}
+	extra := []interface{}{err, "baz", false}
+	expectedFields := []Field{String("foo", "bar"), Error(err), Bool("baz", false)}
```
mway, line 220: nit: let's combine these into a `var` block here as well

mway (approved):
This is great, thanks! Couple of style nits, but otherwise lgtm.

sugar.go
craigpastro: Sure! Thank you. Done.

sugar_test.go
craigpastro: Combined into a var block here too. Thank you!

PR #1189, opened 2022-10-19
Use strings.ToUpper in Level CapitalString method

sashamelentyev:


abhinav (changes requested):
Thanks for the proposal @sashamelentyev, but we would advise against this change.
While the manual mapping is slightly more boilerplate, it's allocation-free for the supported levels.

If someone uses the [CapitalLevelEncoder](https://github.com/uber-go/zap/blob/ee349f9e2995499f66628dc573382939b6941b09/zapcore/encoder.go#L60), this method is on the log encoding path, so we want to avoid the extra allocation of upper-casing the `String()` for every log entry.

PR #1216, opened 2023-01-02
chore: add test to ensure that build Development e Production will never fail

robsonpeixoto:


sywhang (commented):
Hi @robsonpeixoto , 

Thanks for the contribution, but we already test this in `TestConfig` in `config_test.go`, so I don't see any reason this test needs to be added. I'll close this PR as such.

PR #1235, opened 2023-02-17
ci: Test with Go 1.19 and 1.20

abhinav:
With the release of Go 1.20, only Go 1.19 and 1.20 are supported.
Change CI to only test against Go 1.19 and 1.20.

In the process, update the outdated GitHub Actions:

- setup-go v3 has built-in opt-in caching
  so we can drop the manual cache action
- checkout v3 has updated dependencies
- codecov v1 is deprecated and v3 should be used


r-hang (approved):

abhinav:
Derp. Can't run setup-go before checkout because it needs go.sum files present.

sywhang (approved):

r-hang (approved):

PR #1239, opened 2023-02-17
Add govulncheck as a CI step

sywhang:
This adds [govulncheck](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) as a CI step to Zap to scan for known vulnerabilities.

Internal Ref: GO-1877.


Makefile
```diff
@@ -72,3 +76,7 @@ updatereadme:
 .PHONY: tidy
 tidy:
 	@$(foreach dir,$(MODULE_DIRS),(cd $(dir) && go mod tidy) &&) true
+
+.PHONY: vulncheck
+vulncheck:
```
abhinav, line 81: Needs a dependency on GOVULNCHECK
```suggestion
vulncheck: $(GOVULNCHECK)
```

r-hang (approved):

PR #1241, opened 2023-03-05
remove strings.ToLower(s) ,avoid  redundant loops

tylitianrui:
there are loops  of   string `s` in function `strings.ToLower(s)` . And  function `normalizeScheme()` traverses the same string ,again. I think it  is  unnecessary.  removing strings.ToLower(s)  can  avoid  redundant loops.

### DIFF
--- a/sink.go
+++ b/sink.go
@@ -159,21 +159,29 @@ func (sr *sinkRegistry) newFileSinkFromPath(path string) (Sink, error) {
 
 func normalizeScheme(s string) (string, error) {
 	// https://tools.ietf.org/html/rfc3986#section-3.1
-	s = strings.ToLower(s)
-	if first := s[0]; 'a' > first || 'z' < first {
+	var b strings.Builder
+	b.Grow(len(s))
+	if first := s[0]; 'A' > first || 'z' < first || ('Z' < first && 'a' < first) {
 		return "", errors.New("must start with a letter")
 	}
-	for i := 1; i < len(s); i++ { // iterate over bytes, not runes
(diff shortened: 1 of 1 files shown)


abhinav:
Thanks for the contribution, @tylitianrui.
Unfortunately, this change increases the number of allocations, rather than decreate.
I wrote a quick benchmark and ran it before and after this change:

```go
func BenchmarkNormalizeScheme(b *testing.B) {
	tests := []string{
		"file",
		"FILE",
		"FiLe",
	}

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for _, tt := range tests {
			normalizeScheme(tt)
		}
	}
}
```

Here's the before/after:

```
name               old time/op    new time/op    delta
NormalizeScheme-2     241ns ± 7%     272ns ± 7%   ~     (p=0.100 n=3+3)

name               old alloc/op   new alloc/op   delta
NormalizeScheme-2     8.00B ± 0%    28.00B ± 0%   ~     (p=0.100 n=3+3)

name               old allocs/op  new allocs/op  delta
NormalizeScheme-2      2.00 ± 0%      4.00 ± 0%   ~     (p=0.100 n=3+3)
```

Allocations go up by two per operation.
```

prashantv (commented):
Even if we could make this code path faster, since it's not on the critical path (logger initialization is typically a one-off), I'd prefer to keep the code simpler.

tylitianrui:
@abhinav @prashantv   thanks  for  view

Issue #1229, opened 2023-02-11
Add `slog` to performance benchmarks


Dentrax:
> [slog](https://pkg.go.dev/golang.org/x/exp/slog) is a new standard library package which could provide structured logging with levels. [^1]

* Add slog package into [benchmarks](https://github.com/uber-go/zap/tree/master/benchmarks)
* * Add slog package into [Performance](https://github.com/uber-go/zap#performance) section at README. (I think it's auto-generated)

I can submit a PR for this, let's discuss before.

[^1]: https://www.youtube.com/watch?v=gd_Vyb5vEw0

sywhang:
Thanks for the suggestion. We'll have someone take a look at this shortly.

JacobOaks:
Internal Ref: GO-1885

ChannyClaus:
i took a stab at it: https://github.com/uber-go/zap/pull/1245 let me know if it looks good

wasn't sure how to generate the snippets in the `README.md`, though.

Dentrax:
https://github.com/golang/go/issues/56345#issuecomment-1470506816

sywhang:
Closing this as it was completed with #1245

robsonpeixoto:
Are there plans to add zapslog?

PR #1245, opened 2023-03-10
benchmark: add slog

ChannyClaus:
https://github.com/uber-go/zap/issues/1229

not sure how to generate the output in the `README.md`, though - maybe someone can help me out with that?


abhinav:
Thanks, @ChannyClaus!
The README is generated by running `make updatereadme`.
It'll run all benchmarks and update numbers in the README.

abhinav:
Thanks!

> seems like the alignment changed but this is what the make target did (would you like me to manually align these?)

No, this is fine. It's valid Markdown.
Someone must have manually aligned the table, but it's not required because it's auto generated.

It's curious that 'standard library' reports 8 nanoseconds for the last benchmark.
That doesn't seem right -- unless that operation is being discarded completely.
... I just checked, yeah, that's what happening.

https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/log/log.go;l=201-203

log.Printf will no-op early if built with an io.Discard. That messes up this benchmark.

A workaround for this is to use Zap's ztest.Discarder, which log.Printf will not detect as being an io.Discard.

I've pushed a commit doing this to your branch; would you please re-run? Thanks!

ChannyClaus:
> 

nice catch - made the fix! standard library benchmark looks slow as expected now 😆

abhinav (approved):
CC @prashantv @sywhang

README.md
```diff
@@ -64,42 +64,43 @@ id="anchor-versions">[1](#footnote-versions)</sup>
...
-| :zap: zap           |    44 ns/op |      +0%      |    0 allocs/op    |
-| :zap: zap (sugared) |    74 ns/op |     +68%      |    1 allocs/op    |
-| go-kit              |  5072 ns/op |   +11427%     |   56 allocs/op    |
-| log15               | 22634 ns/op |   +51341%     |   70 allocs/op    |
-| apex/log            | 28775 ns/op |   +65298%     |   53 allocs/op    |
-| logrus              | 29047 ns/op |   +65916%     |   68 allocs/op    |
+| Package | Time | Time % to zap | Objects Allocated |
+| :------ | :--: | :-----------: | :---------------: |
+| :zap: zap (sugared) | 265 ns/op | -31% | 1 allocs/op
+| :zap: zap | 386 ns/op | +0% | 0 allocs/op
```
rabbbit, line 83: This one looks wrong?

rabbbit: As in, sugared is somehow faster than non-sugared.

Also, looks like previously someone (manually?) sorted zerolog to be on top?

abhinav: Huh, yeah that's weird. I haven't had a chance to dig into it yet.

Yeah, I think when this was last updated, the author aligned the table and sorted them by time rather than "zap on top and then everything else sorted" which is what the script does.

ChannyClaus: hmm i re-generated the readme again a bunch of times and now i'm consistently getting the expected result (i.e. sugared logger is slower by ~20-30%)

i suspect that i likely unintentionally strained my cpu doing something else on my laptop while the sugared benchmark was running (sorry! noob in benchmarking)

sywhang (approved):
Thank you for doing this work!

PR #1246, opened 2023-03-11
[Feature] Add  experiment `zapslog` package to integrate with slog

knight42:
Ref https://github.com/uber-go/zap/discussions/1179

This PR provides a implementation of `slog.Handler` in a separate experiment `exp/zapslog` module as suggested by @abhinav.


exp/zapslog/example_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+func Example_slog() {
+	logger := zap.NewExample(zap.IncreaseLevel(zap.InfoLevel))
+	defer logger.Sync()
+
+	sl := zapslog.New(logger)
+
+	sl.Info("user", "name", "Al", "secret", Password("secret"))
+	sl.Error("oops", net.ErrClosed, "status", 500)
+	sl.LogAttrs(context.TODO(), slog.LevelError, "oops",
```
abhinav, line 50: We don't need to use context.TODO.
We can use context.Background instead -- it represents the process' lifetime, which is appropriate here.

```suggestion
	ctx := context.Backgrond()
	sl.LogAttrs(ctx, slog.LevelError, "oops",
```

exp/zapslog/example_test.go
```diff
@@ -0,0 +1,69 @@
...
+		slog.Any("err", net.ErrClosed), slog.Int("status", 500))
+	sl.Info("message",
+		slog.Group("group",
+			slog.Float64("pi", 3.14),
+			slog.Duration("1min", time.Minute),
+		),
+	)
+	sl.WithGroup("s").LogAttrs(context.TODO(), slog.LevelWarn, "warn", slog.Uint64("u", 1), slog.Any("m", map[string]any{
```
abhinav, line 58:
```suggestion
	sl.WithGroup("s").LogAttrs(ctx, slog.LevelWarn, "warn", slog.Uint64("u", 1), slog.Any("m", map[string]any{
```

exp/zapslog/example_test.go
```diff
@@ -0,0 +1,69 @@
...
+		"foo": "bar",
+	}))
+	sl.LogAttrs(context.TODO(), slog.LevelDebug, "not show up")
```
abhinav, line 61:
```suggestion
	sl.LogAttrs(ctx, slog.LevelDebug, "not show up")
```

exp/zapslog/go.mod
```diff
@@ -0,0 +1,13 @@
+module go.uber.org/zap/exp/zapslog
```
abhinav, line 1: I wasn't planning on the subpackage being a submodule, but I'm not opposed to it.
@mway @prashantv @sywhang what do you think?

- Pro: Independently versioned. Can tag `exp/zapslog/v0.1.0` (or `exp/v0.1.0` if we move the submodule definition one level up) independently of Zap's 1.0 instead of saying "go.uber.org/zap/exp is part of the main Zap module, but it doesn't comply with semver."
- Con: Adds slightly more complexity to the repository. Have to use replace directives or go.work files for cross-cutting changes (because exp will otherwise only have access to the released version), and have to be sure to release Zap before exp if a change in exp depends on a new change in Zap.

exp/zapslog/slog.go
```diff
@@ -0,0 +1,132 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+//go:build zapslog
+
```
abhinav, line 22: We don't need this build tag.
```suggestion
```

exp/zapslog/slog.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"context"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"golang.org/x/exp/slog"
+)
+
+type handler struct {
+	logger *zap.Logger
```
abhinav, line 34: @knight42 I think there was a misunderstanding.
We want to implement this interface around zapcore.Core.

slog.Handler and zapcore.Core are at the same level
similar to how slog.Logger and zap.Logger are at the same level.

exp/zapslog/slog.go
```diff
@@ -0,0 +1,132 @@
...
+
+var _ slog.Handler = (*handler)(nil)
+
+// group holds all the Attrs saved in a slog.GroupValue.
+type group struct {
+	attrs []slog.Attr
+}
+
+func (g *group) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	for _, attr := range g.attrs {
```
abhinav, line 45: Two things:

- We don't need a struct here. Just a `type group []slog.Attr` should be enough.
- 'group' is too general a name for package-level scope. We should qualify it to something more specific that is less likely to collide with a variable or package name elsewhere. `groupObject` may be better.

```suggestion
// groupObject holds all the Attrs saved in a slog.GroupValue.
type groupObject []slog.Attr

func (gs groupObject) MarshalLogObject(enc zapcore.ObjectEncoder) error {
	for _, attr := range gs {
```

exp/zapslog/slog.go
```diff
@@ -0,0 +1,132 @@
...
+		return zap.Time(attr.Key, attr.Value.Time())
+	case slog.KindUint64:
+		return zap.Uint64(attr.Key, attr.Value.Uint64())
+	case slog.KindGroup:
+		return zap.Object(attr.Key, &group{attrs: attr.Value.Group()})
+	case slog.KindLogValuer:
+		return convertAttrToField(slog.Attr{
+			Key:   attr.Key,
+			Value: attr.Value.Resolve(),
+		})
```
abhinav, line 73: This will resolve the LogValuer eagerly, which will hurt anyone using LogValuer to defer expensive operations.

Ideally, we want to use a wrapper struct that calls LogValue() at serialization time,
but that may not be 100% straightforward because a LogValuer doesn't know if it's a group or a primitive type until calling LogValue(),
whereas Zap wants to know that in advance.
This may require a lazy field-equivalent in Zap.
We can defer designing that to a follow-up. 
CC @mway @prashantv @sywhang

exp/zapslog/slog.go
```diff
@@ -0,0 +1,132 @@
...
+func (h *handler) WithGroup(name string) slog.Handler {
+	return &handler{
+		logger: h.logger.With(zap.Namespace(name)),
+	}
+}
+
+// New returns a *slog.Logger which writes to the supplied zap Logger.
+func New(logger *zap.Logger) *slog.Logger {
+	const slogCallerDepth = 1
+	const loggerWriterDepth = 2
```
abhinav, line 128: These are helpful, thanks.

abhinav (commented):
Thanks for this, @knight42!
I've taken an initial pass over it, including some thinking-out-loud.

There are a couple open questions here that may merit some discussion, so I've tagged other maintainers.

exp/zapslog/go.mod
prashantv: More modules does seem more complex, though separating exp (where we probably will keep it at 0.x and make breaking changes) seems like a good idea. So submodule at exp is my vote.

exp/zapslog/slog.go
prashantv: +1 on a adding a lazy-field equivalent in zap separately.

exp/zapslog/slog.go
knight42: @abhinav Oh thanks for pointing it out, so what we actually provide in this package is sth like [slog.JSONHandler](https://pkg.go.dev/golang.org/x/exp/slog#JSONHandler), right? Should we provide a similar [HandlerOptions](https://pkg.go.dev/golang.org/x/exp/slog#HandlerOptions) struct? 

What I currently have in mind is defining a public function in the package like the following, does it need further improvement?
```go
func NewHandlerFromCore(core zapcore.Core) slog.Handler {
  ...
}
```

exp/zapslog/go.mod
knight42: Per the discussion, I have made the `exp` a submodule.

exp/zapslog/slog.go
abhinav: Yeah, that sounds right, @knight42.
Suggested changes:

- We don't need "FromCore" there since it'll be defined in zapslog, so that we're dealing with Zap abstractions will already be in the name's context.
- We could maybe return a `*Handler` struct instead of an interface should we need flexibility there in the future.

So perhaps:

```go
package zapslog

type Handler struct{ ... }

var _ slog.Handler = (*Handler)(nil)

func NewHandler(core zapcore.Core) *Handler
```

(In case you're not aware, that `var _` line is a [compile-time check to verify interface compliance](https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance).)

exp/zapslog/go.mod
abhinav: Excellent, thanks!

exp/zapslog/slog.go
```diff
@@ -0,0 +1,145 @@
...
+func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
+	ent := zapcore.Entry{
+		Level:   convertSlogLevel(record.Level),
+		Time:    record.Time,
+		Message: record.Message,
+		// FIXME: do we need to set the following fields?
+		// LoggerName:
+		// Caller:
+		// Stack:
+	}
```
knight42, line 106: cc @abhinav . Do we need to complete the `Entry`?

abhinav: Those three fields are optional. We can start without them.
I can see Caller and LoggerName being filled in the future.
(We could probably fill the Caller from [Record.PC](https://pkg.go.dev/golang.org/x/exp/slog#Record.PC),
but doing that cheaply adds some complexity so we can do without for now.)

However, this does bring up the need for customization.
A plain `NewHandler` may not be sufficient for that.

I've pushed a change to your branch with the following new API:

```go
type HandlerOptions struct{ LoggerName string }

func (HandlerOptions) New(zapcore.Core) *Handler
```

The `func NewHandler(..)` just calls this now to get default options.
This mirrors the JSONHandler, NewJSONHandler, HandlerOptions setup.

This leaves us room for future customization llke:
opting into adding a caller name, custom level mapping, etc.

knight42: Thank you! The `HandlerOptions` is commonly used for customization, though I thought we might want to consistently follow the functional options pattern in zap.

> I can see Caller and LoggerName being filled in the future.
(We could probably fill the Caller from [Record.PC](https://pkg.go.dev/golang.org/x/exp/slog#Record.PC),

After some trial & error, I found I could add the Caller in this PR as well, plz see the commit https://github.com/uber-go/zap/pull/1246/commits/1ae59453aac0c7f3e7fec679ca167393f7b47218

exp/zapslog/doc.go
```diff
@@ -0,0 +1,27 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package zapslog provides an implementation of slog.Handler which writes to
+// the supplied zapcore.Core.
+//
+// Since the slog proposal has not been officially accepted by the creation of this package,
+// we do not want Zap's standard packages to take a dependency on `golang.org/x/exp/slog`.
+// Instead, we provide this separate module as a way for users to integrate Zap with slog.
+package zapslog // import "go.uber.org/zap/exp/slog"
```
mway, line 27: nit: Why doesn't this import match the package name? If we want this import path, we should rename the package (and directory), else we should use the package name.

exp/zapslog/example_test.go
```diff
@@ -0,0 +1,68 @@
...
+package zapslog_test
+
+import (
+	"context"
+	"net"
+	"time"
+
+	"go.uber.org/zap"
+	"golang.org/x/exp/slog"
+
```
mway, line 30: nit: remove newline

exp/zapslog/example_test.go
```diff
@@ -0,0 +1,68 @@
...
+		slog.Any("err", net.ErrClosed), slog.Int("status", 500))
+	sl.Info("message",
+		slog.Group("group",
+			slog.Float64("pi", 3.14),
+			slog.Duration("1min", time.Minute),
+		),
+	)
+	sl.WithGroup("s").LogAttrs(ctx, slog.LevelWarn, "warn", slog.Uint64("u", 1), slog.Any("m", map[string]any{
+		"foo": "bar",
+	}))
```
mway, line 59: style nit: all params on one line, or each on their own line
```suggestion
	sl.WithGroup("s").LogAttrs(
		ctx,
		slog.LevelWarn,
		"warn", // message
		slog.Uint64("u", 1),
		slog.Any("m", map[string]any{
			"foo": "bar",
		}),
	)
```

mway (commented):
Thank you for implementing this!

mway:
Thinking about it more, as long as we address the import name comment, we can worry about the other nits later.

exp/zapslog/doc.go
knight42: Oh nice catch! I will fix this typo.

knight42:
@mway Hi I have addressed your comments, PTAL

PR #1251, opened 2023-03-18
Leveled fields: Fields present only at certain log levels

ChannyClaus:
https://github.com/uber-go/zap/issues/1078

Added just `DebugField` for now per the suggestions in the issue. Extending this should be very straight-forward if need be.

### DIFF
--- a/zapcore/core.go
+++ b/zapcore/core.go
@@ -92,17 +92,7 @@ func (c *ioCore) Check(ent Entry, ce *CheckedEntry) *CheckedEntry {
 }
 
 func (c *ioCore) Write(ent Entry, fields []Field) error {
-	// https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
-	filteredFields := fields[:0]
-	for _, field := range fields {
-		// include the field if the level is unset or
-		// the field level is higher than the entry's level.
-		if field.Level == nil || *field.Level >= ent.Level {
-			filteredFields = append(filteredFields, field)
-		}
-	}
-
-	buf, err := c.enc.EncodeEntry(ent, filteredFields)
+	buf, err := c.enc.EncodeEntry(ent, fields)
 	if err != nil {
 		return err
 	}
--- a/zapcore/entry.go
+++ b/zapcore/entry.go
@@ -249,8 +249,19 @@ func (ce *CheckedEntry) Write(fields ...Field) {
 	ce.dirty = true
 
 	var err error
+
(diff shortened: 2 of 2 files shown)


zapcore/field.go
```diff
@@ -106,6 +106,7 @@ type Field struct {
 	Type      FieldType
 	Integer   int64
 	String    string
+	Level     *Level
```
ChannyClaus, line 109: Making this attribute a pointer seemed the cleanest way to implement this - doing so would make all existing fields' `Level` attribute be `nil`.

field.go
```diff
@@ -35,8 +35,17 @@ type Field = zapcore.Field
 var (
 	_minTimeInt64 = time.Unix(0, math.MinInt64)
 	_maxTimeInt64 = time.Unix(0, math.MaxInt64)
+
+	DebugLevelField = zapcore.DebugLevel
```
ChannyClaus, line 39: needed to duplicate this in order to be able to pass the address of the `DebugLevel` for assignment (https://github.com/uber-go/zap/pull/1251/files#diff-b494c03820a49f9efc491691ed31a2acc16d4cebf39d679ffd74bf97bbee968bR45)

`DebugLevel` doesn't work because https://github.com/uber-go/zap/blob/master/level.go#L32 already uses that name in the same package.

prashantv (commented):
Doing this in the ioCore means that this won't work with other cores. For example, this won't work with `zaptest/observer`. 

Edit: I think the implementation of this could live in `CheckedEntry` to work with all cores.

ChannyClaus:
> Doing this in the ioCore means that this won't work with other cores. For example, this won't work with `zaptest/observer`.

ah darn - thanks for taking a look! will look into it when i get a chance ;.)

ChannyClaus:
@prashantv moved the filtering over to where you pointed out in the comment - sorry about the incorrect change earlier.

not super sure if https://github.com/uber-go/zap/pull/1251/files#diff-4a689f7bcb3f92ee8d87b3d9070215a7fcb37680843003657a60357c6a821d90R273 is also necessary since all invocations of `OnWrite` seem to _not_ use the field argument:
```
$ git grep 'func.*OnWrite'
logger_test.go:func (h *customWriteHook) OnWrite(_ *zapcore.CheckedEntry, _ []Field) {
zapcore/entry.go:func (a CheckWriteAction) OnWrite(ce *CheckedEntry, _ []Field) {
zapcore/entry_test.go:func (c *customHook) OnWrite(_ *CheckedEntry, _ []Field) {
```

PR #1253, opened 2023-03-19
Use sync/atomic, drop external atomics

abhinav:
Switches all uses of go.uber.org/atomic to types defined in sync/atomic.
These types have been available since Go 1.19,
which is the minimum supported version of Go for Zap.

This allows us to drop one dependency from Zap.
The version of multierr was bumped to pick up uber-go/multierr#72
which made a similar change to that package.

This was nearly a drop-in, save the following:

- There's no `Int*.Inc`; we have to use `Int*.Add(1)`
- There's no `Bool.Toggle`; we have to implement the CAS loop manually

This change has no user-facing consequences
because nearly all atomic usages are internal to Zap.
The only atomic exposed (AtomicLevel) is a wrapper type
for which the behavior does not change.


mway (approved):
lgtm!

PR #1262, opened 2023-03-20
internal: Add and use a generic pool instead of using sync.Pool directly

mway:
This replaces discrete typed `sync.Pool` wrappers with a generic wrapper so that pools have a strongly-typed API rather than doing callsite type assertions.


internal/pool/pool.go
```diff
@@ -0,0 +1,64 @@
...
+
+// Package pool provides internal pool utilities.
+package pool
+
+import (
+	"sync"
+)
+
+// A Constructor is a function that creates a T when a [Pool] is empty.
+type Constructor[T any] func() T
```
abhinav, line 29: Consider deleting this newtype. 
It'll almost never be used as `Constructor[T]`, only as an alias to `func() T` which is (a) shorter, and (b) doesn't have a runtime artifact. (Newtypes result in a new type at runtime versus aliases.)

internal/pool/pool.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// New returns a new [Pool] for T, and will use ctor to construct new Ts when
+// the pool is empty.
+func New[T any](ctor Constructor[T]) *Pool[T] {
+	return &Pool[T]{
+		pool: sync.Pool{
+			New: func() any {
+				return ctor()
+			},
```
abhinav, line 45: Not a blocker by any means, but FYI:
staticcheck likely will not be able to detect [SA6002](https://staticcheck.io/docs/checks/#SA6002) for a custom generic pool, so we'll just have to be careful about it

internal/pool/pool.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Get gets a new T from the pool, or creates a new one if the pool is empty.
+func (p *Pool[T]) Get() T {
+	if x, ok := p.pool.Get().(T); ok {
+		return x
+	}
+
+	// n.b. This branch is effectively unreachable.
+	return p.ctor()
```
abhinav, line 58: Maybe we should drop the unreachable branch.
A failure here means a catastrophic bug.
Perhaps a simple panicking cast is sufficient:
```suggestion
	return p.pool.Get().(T)
```

internal/pool/pool.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Put puts x into the pool.
```
abhinav, line 61: nit:

```suggestion
// Put returns a value to the pool.
```

(I mean, there's nothing stopping someone from putting a non-pool-generated value into the pool, but I figure primary usage is to put values made by Get() back.)

internal/pool/pool_internals_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+			// Override the internal pool to provide unexpected types.
+			pool.pool.New = func() any {
+				return tt.input
+			}
+
+			pool.pool.Put(tt.input)
+			pool.Get()
+			require.Equal(t, tt.expectCall, called)
+		})
```
abhinav, line 79: Going to this length to hit the unreachable case seems not worth. 
We should not handle the impossible case, and let it panic if the impossible happens.

internal/pool/pool_norace_test.go
```diff
@@ -0,0 +1,56 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+//go:build !race
```
abhinav, line 21: ?

internal/pool/pool_norace_test.go
```diff
@@ -0,0 +1,56 @@
...
+		return bytes.NewBuffer([]byte("new"))
+	})
+	p.Put(bytes.NewBuffer([]byte(t.Name())))
+
+	// Ensure that we always get the expected value.
+	for i := 0; i < 1_000; i++ {
+		func() {
+			buf := p.Get()
+			defer p.Put(buf)
+			require.Equal(t, t.Name(), buf.String())
```
abhinav, line 48: Ah, I see what this is doing. It took a second.
I would appreciate a comment above pool.New explaining what this test is trying to verify and how.

abhinav (approved):
Mostly nits and maintainability suggestions. LGTM otherwise.

internal/pool/pool.go
mway: Yeah, that'd be my preference too - I just figured I'd start a little more conservatively. Will do.

internal/pool/pool.go
mway: Yeah, very true. I'll add a comment.

internal/pool/pool.go
mway: 😆 I had the same internal debate and decided to be a pedant (hence "puts into"), but I don't have a strong preference.

internal/pool/pool_norace_test.go
mway: It's easier to test pool correctness deterministically without `-race`, because when race is enabled, `sync.Pool.Put` will randomly drop the returning object on the floor (it does a `fastrand(4)==0` check). I don't think it's useful to test the behavior of the underlying `sync.Pool`, but we do want it to have predictable behavior for tests.

We brute force it (given that it's improbable to drop everything on the floor over a sufficient number of attempts), but it felt like more work than necessary.

I'll noodle on this for a little bit and see if I can come up with a non-race-dependent version.

internal/pool/pool_race_test.go
```diff
@@ -0,0 +1,69 @@
...
+type pooledValue struct {
+	n int64
+}
+
+func TestNew(t *testing.T) {
+	// n.b. [sync.Pool] will randomly drop re-pooled objects when race is
+	//      enabled, so rather than testing nondeterminsitic behavior, we use
+	//      this test solely to prove that there are no races. See pool_test.go
+	//      for correctness testing.
+
```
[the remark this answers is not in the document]
mway: Ultimately I decided that brute force is probably fine here since we can't really affect the behavior of `sync.Pool.Put` when `-race` is used.

We can go the race/non-race route if necessary, it's just more work for probably minimal gain, since we trust `sync.Pool` to do the right thing and only really care about quasi-deterministic behavior for the test.

abhinav (approved):

PR #1263, opened 2023-03-20
slog: Support inline group fields per spec

mway:
* Adds support for inline groups per the slog spec: if a group's attr key is empty, its fields should be merged inline rather than nested within an object namespace.
* Adds some coarse optimization by pooling temporary `[]zap.Field` storage.

Before:

```
# go test ./zapslog/ -count 5 -run xxx -bench .
goos: linux
goarch: amd64
pkg: go.uber.org/zap/exp/zapslog
cpu: AMD EPYC 7B13
BenchmarkSlog-96          526584              2468 ns/op             243 B/op          3 allocs/op
BenchmarkSlog-96          446736              2751 ns/op             243 B/op          3 allocs/op
BenchmarkSlog-96          538926              2636 ns/op             243 B/op          3 allocs/op
BenchmarkSlog-96          409718              2610 ns/op             243 B/op          3 allocs/op
BenchmarkSlog-96          566422              2762 ns/op             243 B/op          3 allocs/op
PASS
ok      go.uber.org/zap/exp/zapslog     8.525s
```

After:

```
# go test ./zapslog/ -count 5 -run xxx -bench .
goos: linux
goarch: amd64
pkg: go.uber.org/zap/exp/zapslog
cpu: AMD EPYC 7B13
BenchmarkSlog-96          512924              2443 ns/op              24 B/op          1 allocs/op
BenchmarkSlog-96          397900              2796 ns/op              24 B/op          1 allocs/op
BenchmarkSlog-96          414337              2967 ns/op              24 B/op          1 allocs/op
BenchmarkSlog-96          517803              2335 ns/op              24 B/op          1 allocs/op
BenchmarkSlog-96          422788              2814 ns/op              24 B/op          1 allocs/op
PASS
ok      go.uber.org/zap/exp/zapslog     8.033s
```


exp/zapslog/slog.go
```diff
@@ -74,43 +110,62 @@ type groupObject []slog.Attr
...
+		converted := convertAttrToFields(attr)
+		converted.AddTo(enc)
+		converted.Return()
 	}
 	return nil
 }
 
-func convertAttrToField(attr slog.Attr) zapcore.Field {
+func convertAttrToFields(attr slog.Attr) *fieldSlice {
+	fields := _fieldSlices.Get()
```
abhinav, line 121: I think this will be less prone to use-after-free bugs if "get from pool" and "defer return to pool" are next to each other.
Consider turning `convertAttrToFields` into `addAttrAsFields(enc, attr)`, moving the AddTo inside the function, which will allow a `defer fields.Return()` right after the Get.

If that's unavoidable, we can Get from pool up above in MarshalLogObject outside the for loop,
pass into `convertAttrToFields(attr, &converted)`, and then reset and re-use between each iteration:

```go
converted := _fieldSlices.Get()
defer converted.Return()
for _, attr := range gs {
  converted.Reset()
  convertAttrToFields(attr, &converted)
  converted.AddTo(enc)
}
```

exp/zapslog/slog.go
```diff
@@ -25,10 +25,46 @@ import (
...
+type fieldSlice []zapcore.Field
+
+func (s *fieldSlice) Append(f zapcore.Field) {
+	*s = append(*s, f)
+}
+
+func (s *fieldSlice) AddTo(enc zapcore.ObjectEncoder) {
+	for i := 0; i < s.Len(); i++ {
+		(*s)[i].AddTo(enc)
+	}
```
abhinav, line 42: Any reason not to use `range` here?
```suggestion
	for _, f := range (*s) {
		f.AddTo(enc)
	}
```

exp/zapslog/slog.go
```diff
@@ -157,33 +213,52 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
...
 
-	fields := make([]zapcore.Field, 0, record.NumAttrs())
+	fields := _fieldSlices.Get()
+	defer fields.Return()
+
 	record.Attrs(func(attr slog.Attr) {
-		fields = append(fields, convertAttrToField(attr))
+		converted := convertAttrToFields(attr)
+		fields.Extend(converted)
+		converted.Return()
```
abhinav, line 222: I think if you take my suggestion to make `convert` take a `*fieldSlice`,
this won't need to borrow a new slice at all.
```suggestion
		convertAttrToFields(attr, &fields)
```

exp/zapslog/slog.go
```diff
@@ -157,33 +213,52 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
...
-	fields := make([]zapcore.Field, len(attrs))
-	for i, attr := range attrs {
-		fields[i] = convertAttrToField(attr)
+	fields := _fieldSlices.Get()
+	defer fields.Return()
+
+	for _, attr := range attrs {
+		converted := convertAttrToFields(attr)
+		fields.Extend(converted)
+		converted.Return()
```
abhinav, line 238: Same
```suggestion
		convertAttrToFields(attr, &fields)
```

exp/zapslog/slog.go
```diff
@@ -157,33 +213,52 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
...
 // the receiver's existing groups.
 func (h *Handler) WithGroup(group string) slog.Handler {
-	return h.withFields(zap.Namespace(group))
+	// Stack-allocate here: the number of options is known at compile time.
+	tmp := [...]zapcore.Field{
+		zap.Namespace(group),
+	}
+
+	cloned := *h
+	cloned.core = h.core.With(tmp[:])
```
abhinav, line 253: I'm 75% certain that the manual allocation won't have any effect.
It will behave the same as `h.core.With(zap.Namespace(group))`:
if the slice escapes, it'll be heap allocated anyway, and if it doesn't escape, it was going to be stack allocated anyway.

exp/zapslog/slog.go
```diff
@@ -157,33 +213,52 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
...
+	return &cloned
 }
 
 // withFields returns a cloned Handler with the given fields.
-func (h *Handler) withFields(fields ...zapcore.Field) *Handler {
+func (h *Handler) withFields(fields *fieldSlice) *Handler {
+	defer fields.Return()
```
abhinav, line 259: This is super sus.
As a reader, it's super not clear why/how this is being freed here.

exp/zapslog/slog.go
```diff
@@ -25,10 +25,46 @@ import (
...
+func (s *fieldSlice) Reset() {
+	*s = (*s)[:0]
+}
+
+func (s *fieldSlice) Return() {
+	s.Reset()
+	_fieldSlices.Put(s)
+}
+
+var _fieldSlices = pool.New(func() *fieldSlice {
```
abhinav, line 62: Consider adding a clarifying comment that the pooled value MUST be a pointer type to avoid allocations.

mway:
Thanks for the pass, just a rough draft atm. Agree with your comments, will rework some of this soon.

abhinav:
Superseded by #1408

Issue #1267, opened 2023-04-05
zapgrpc implements Experimental grpclog.DepthLoggerV2


forsaken628:
**Is your feature request related to a problem? Please describe.**

grpc log caller incorrect

**Describe the solution you'd like**

zapgrpc implements Experimental grpclog.DepthLoggerV2

**Describe alternatives you've considered**

**Is this a breaking change?**
We do not accept breaking changes to the existing API. Please consider if your proposed solution is backwards compatible. If not, we can help you make it backwards compatible, but this must be considered when we consider new features.

no

**Additional context**

yurishkuro:
possible workaround:
```
	grpclog.SetLoggerV2(zapgrpc.NewLogger(
		logger.WithOptions(
			zap.AddCallerSkip(5), // ensure the actual caller:lineNo is shown
		)))
```

PR #1268, opened 2023-04-05
[Feature] zapgrpc.Logger implements Experimental DepthLoggerV2

forsaken628:
https://github.com/uber-go/zap/issues/1267

and fix verbosity 
https://github.com/grpc-ecosystem/go-grpc-middleware/pull/452


sywhang:
@forsaken628 thanks for the submission, do you have any additional info on why this is needed on Zap's side? The issue you created doesn't have most of its fields filled out properly.

forsaken628:
The current zapgrpc package is poorly maintained and outdated. This PR just fills in the parts that this package should have. Or zapgrpc should be marked as deprecated.

frbvianna:
Any updates on this? It definitely looks like zapgrpc package is outdated compared to `grpc-go-middleware` past version, the grpclog verbosity level variable `GRPC_GO_LOG_VERBOSITY_LEVEL` currently does not seem to be working at all.

Issue #1200, opened 2022-11-12
Return logger name


djarbz:
**Is your feature request related to a problem? Please describe.**
I am writing an application that overall should have a user specified log level, ie INFO.
But they can overwrite the log level for certain "actions" that would be handled on a case by case basis.
Since we are unable to change the log level of a cloned logger, I am instantiating a new logger set to the DEBUG level.
Currently, this new logger does not have the inherited name for the existing logger.
I would like a function such as `zapLog.GetName()` that would return a string of the current log name that I can then pass on to the new logger.

**Is this a breaking change?**
I do not believe this would be a breaking change as it would add a function to the current interface.

prashantv:
It seems reasonable to add `Name() string` to `*Logger` that exposes the name.

rexywork:
Hi, I try to create PR for this issues here:
https://github.com/uber-go/zap/pull/1273

Please check the comment so I can double check what I need to do

rnemeth90:
Is this being actively worked on? If not, please assign to me and I'll submit a PR.

prashantv:
This was released in https://github.com/uber-go/zap/releases/tag/v1.25.0

PR #1273, opened 2023-04-17
Add Name() function to Logger struct

rexywork:
Related to https://github.com/uber-go/zap/issues/1200

This PR add `Name()` function to Logger struct which will return the Logger's Name.
By default, Logger's name is empty if not set so for Unnamed Logger it will return empty string.

Please let me know which file test that I should include for testing this function and I will include the unit test.
Thank you.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -281,6 +281,12 @@ func (log *Logger) Core() zapcore.Core {
 	return log.core
 }
 
+// Name returns the Logger's underlying name.
+// Return empty string for unnamed Logger.
+func (log *Logger) Name() string {
+	return log.name
+}
+
 func (log *Logger) clone() *Logger {
 	copy := *log
 	return &copy
--- a/logger.go
+++ b/logger.go
@@ -281,8 +281,8 @@ func (log *Logger) Core() zapcore.Core {
 	return log.core
 }
 
-// Name returns the Logger's underlying name.
-// Return empty string for unnamed Logger.
+// Name returns the Logger's underlying name,
+// or an empty string if the logger is unnamed.
 func (log *Logger) Name() string {
 	return log.name
 }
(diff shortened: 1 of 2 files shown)


logger.go
```diff
@@ -281,6 +281,12 @@ func (log *Logger) Core() zapcore.Core {
 	return log.core
 }
 
+// Name returns the Logger's underlying name.
+// Return empty string for unnamed Logger.
```
abhinav, line 285: optional change:

```suggestion
// Name returns the Logger's underlying name,
// or an empty string if the logger is unnamed.
```

abhinav (approved):
LGTM.
CC @mway @prashantv @sywhang

mway (approved):
lgtm

sywhang (approved):
Thanks for the PR!

rexywork:
@abhinav @sywhang @mway Seems I need to add test for the `Name()` function to pass the coverage.
After looking the test, I found `func TestLoggerNames(t *testing.T)` here: https://github.com/uber-go/zap/blob/master/logger_test.go#L319
Should we just adjust this test? For example, in [here](https://github.com/uber-go/zap/blob/master/logger_test.go#L319) and [here](https://github.com/uber-go/zap/blob/master/logger_test.go#L351), we can change:
`logs.AllUntimed()[0].LoggerName` to `log.Name()` (Instead of getting the LoggerName from the Entry, we can get directly from Logger variables)
What do you think? If you agree, I can push new changes. Thank you

mway:
Instead of replacing (because we still want to validate that entries have the correct logger name), just adding a second check against the expected logger name should be fine.

Thank you!

rexywork:
@mway Ah I see, I will add a line check in the same test function since it kind of weird to separate as new test function.
Thank you for the quick response!

rexywork:
@mway @abhinav @sywhang I already adjust the test and the coverage check already passed.
Please help me review and kindly help to merge if there are no additional request. Thank you

mway:
Thanks @rexywork!

PR #1274, opened 2023-04-17
Make DPanic fatal in zaptest

WGH-:
The intention of `DPanic` is to catch errors that should never happen. Unless you added `zaptest.WrapOptions(zap.Development())`, they were previously silently ignored.

This is a breaking change. But you can opt out of this using `zaptest.Development(false)` option.

## Summary by CodeRabbit

- **New Features**
  - Added an option to disable development mode for the logger.
- **Documentation**
  - Updated logger documentation to reflect the new default behavior and the option to opt out of development mode.
- **Tests**
  - Added tests to verify logger behavior when development mode is enabled or disabled.

### DIFF
--- a/zaptest/logger.go
+++ b/zaptest/logger.go
@@ -33,8 +33,9 @@ type LoggerOption interface {
 }
 
 type loggerOptions struct {
-	Level      zapcore.LevelEnabler
-	zapOptions []zap.Option
+	Level       zapcore.LevelEnabler
+	zapOptions  []zap.Option
+	development bool
 }
 
 type loggerOptionFunc func(*loggerOptions)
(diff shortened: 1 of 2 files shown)


abhinav (commented):
Thanks! I agree that the zaptest logger should panic/run in development mode. This was an oversight. For opting out, we could maybe add a zaptest.LoggerOption that allows setting it to false?

```go
zaptest.NewLogger(t) // will panic
zaptest.NewLogger(t, zaptest.Development(false)) // won't panic
```

There's a question of whether this is a breaking change, though.
Based on the intent of this API, I would consider this more a bugfix than a breaking change, but I'd like to hear thoughts on this from other maintainers.

CC @sywhang @mway @prashantv

sywhang:
IMO this is a breaking change in behavior; but since this will only affect tests I think it might be worth considering. 

Adding an option for opting out of the panic behavior would save many people from suddenly seeing a bunch of their tests break so I'm +1 on that suggestion.

WGH-:
So I've just added `zaptest.NoDevelopment()` option to opt out of this.

(yeah, sorry, I kinda forgot for two years)

prashantv:
Agree that this is worth changing as it only impacts tests, and it should have been this way from the start.

Issue #1275, opened 2023-04-24
zap.String with ~string generic


bcspragu:
**Is your feature request related to a problem? Please describe.**

Frequently, we'll want to log something defined as:

```go
// Imagine this for basically any primary key in an ecosystem, e.g. AccountID, ContactID, etc
type UserID string
```

And to log it presently, we'll have to do something like:

```go
logger.Debug("user hit admin endpoint",
  zap.String("path", r.URL.Path),
  zap.String("user_id", string(u.ID)),
  zap.String("account_id", string(acct.ID)))
```

Which includes a lot of `string(...)` conversions, that are mostly unnecessary now that generics exist.

**Describe the solution you'd like**

The solution I'm proposing is to change the type of `zap.String` from:

```go
func String(k string, v string) zap.Field
```

to

```go
func String[T ~string](k string, v T) zap.Field
```

This is purely a convenience thing, but it removes a lot of casting that clutters up our logging code. AFAICT, this should be backwards compatible, and generics are already in use in the latest release. The same strategy could be applied to all the other primitive-type `zap.Field` helpers to simplify end-user code a bit. 

**Describe alternatives you've considered**

The `zap.Stringer` helper is closely related, but doesn't help the `type X string` use case, which is a pretty common idiom in Go.

**Is this a breaking change?**

Nope! Should totally be backwards compatible. Existing `string(x)` casts will work fine, but future ones can be reduced to just `x`.

**Additional context**

This change would take about 37 seconds to implement, happy to contribute that if folks are interested. Also, apologies if this has come up before, I tried searching issues for it, but came up empty handed. Searching for `"~string"` didn't give me anything relevant AFAICT.

abhinav:
Thanks for the issue, @bcspragu.
I agree that this would be useful to have, although I don't think we can change the signature of the existing String function.
While it won't break any call sites, it will break function references. e.g.

```go
s := zap.String
log.Info("foo", s("k", "v"))
```

The above will fail to compile with the suggested change.

However, I still think this functionality is valuable.
We can maybe add it as a new function?
As an opening suggestion: `Str` for stringy things?

```go
func Str[T ~string](k string, v T) Field {
  return String(k, string(v))
}
```

If desirable, we can even expand that type constraint in the future to `~string | fmt.Stringer` and dispatch to `zap.String` or `zap.Stringer` dynamically. But we don't have to do that just yet. Expanding the type constraints of a function that expects a type constraint is backwards-compatible.

[...]

PR #1281, opened 2023-05-01
Introduce Str + Strs generic string helpers

bcspragu:
Implements #1275 by adding new `zapfield.Str` and `zapfield.Strs` types, which operate on generic string-like types and use the underlying `zap.String` and `zap.Strings` codepaths.

In #1275, the decision was to put these in a separate `zap/exp/zapfield` package and I didn't see an existing one, so I made a new `zapfield` package.

Since this is a separate package, I had to copy a few test + non-test helpers here. If that's undesirable, I can export them in a shared helper package somewhere and have both places use that

### DIFF
--- a/exp/zapfield/zapfield.go
+++ b/exp/zapfield/zapfield.go
@@ -1,3 +1,23 @@
+// Copyright (c) 2016-2023 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


sywhang (approved):
Thanks! This looks good to me.

abhinav (approved):
LGTM! Thanks!

abhinav (approved):
LGTM! Thanks!

PR #1282, opened 2023-05-08
Swap expected and got fields

JacobOaks:
This is just a small nit from #1281 - I think the expected and got fields of the test case structs are flipped.


rabbbit (approved):
do we actually have a recommendation somewhere on ordering the fields within the tests struct?

PR #1286, opened 2023-05-13
feat: user defined log levels

### LINKED ISSUE #1277: Allow specifying user defined log level names

As a user of `zap` I'd like to define my own custom log level, let's call it `TraceLevel`. It's destined to be more verbose than [Debug](https://github.com/uber-go/zap/blob/4451b47c5e4c4263828b042d6fb12d6d14c6ca2e/zapcore/level.go#L37).

Because currently the log levels are checked against a known, concrete set of values https://github.com/uber-go/zap/blob/4451b47c5e4c4263828b042d6fb12d6d14c6ca2e/zapcore/level.go#L134-L156 when pretty printing the log level name I cannot define a name on my level and see it in the logs.

All I can get is [`fmt.Sprintf("LEVEL(%d)", l)`](https://github.com/uber-go/zap/blob/4451b47c5e4c4263828b042d6fb12d6d14c6ca2e/zapcore/level.go#L154)

Define an interface that user defined log levels can be checked against and if they implement it then use it to get the name of the log level.

If that fails then resort to `fmt.Sprintf("LEVEL(%d)", l)`.

**Describe alternatives you've considered**

None exist to the best of my knowledge.

No.

SoulPancake:
This is a starter to the approach for - resolves #1277

### DIFF
--- a/zapcore/level.go
+++ b/zapcore/level.go
@@ -24,12 +24,43 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"sync"
 )
 
 var errUnmarshalNilLevel = errors.New("can't unmarshal a nil *Level")
 
 // A Level is a logging priority. Higher levels are more important.
 type Level int8
+
+type CustomLevel interface {
+	String() string
+	CapitalString() string
+	Enabled(lvl Level) bool
+}
+
+// Type to hold custom levels
+type CustomLevels struct {
+	mu     sync.RWMutex
+	levels map[Level]CustomLevel
+}
+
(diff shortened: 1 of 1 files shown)


sywhang (changes requested):
hey @SoulPancake , sorry I missed the original issue you were working on but we do not want to add this feature to Zap as this is a backwards incompatible change. See my comment https://github.com/uber-go/zap/issues/1277#issuecomment-1548366070.

SoulPancake:
Understood, Thanks @sywhang  Closing this PR

PR #1297, opened 2023-07-03
feat(field): add zap.Dict to Field constructor list

hhk7734:
When I want to log a sub dictionary, it's hard to find that I can just log `zap.Object("k", ObjectMarshalerFunc(func(inner ObjectEncoder) error { ... }))`, and it seems unintuitive.

So I added `zap.Dict`.

## Example

```go
zap.Dict("1",
	zap.String("1-1", "b"),
	zap.Bool("1-2", true),
	zap.Dict("1-3",
		zap.String("1-3-1", "d"),
	),
),
```

```json
{"1": {"1-1": "b", "1-2": true, "1-3": {"1-3-1": "d"}}}
```

```json
{
  "1": {
    "1-1": "b",
    "1-2": true,
    "1-3": {
      "1-3-1": "d"
    }
  }
}
```


mway:
LGTM - thanks @hhk7734 and @abhinav!

PR #1300, opened 2023-07-20
Replacing the path to the "clock" library

BorzdeG:
Replacing the path to the "[clock](https://github.com/benbjohnson/clock)" library, since the author of the original library stopped development

### DIFF
--- a/benchmarks/go.mod
+++ b/benchmarks/go.mod
@@ -16,7 +16,6 @@ require (
 )
 
 require (
-	github.com/benbjohnson/clock v1.3.0 // indirect
 	github.com/go-logfmt/logfmt v0.5.1 // indirect
 	github.com/go-stack/stack v1.8.1 // indirect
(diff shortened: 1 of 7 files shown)


rabbbit:
I don't expect we'll migrate to a clock library that was created 2 days ago. Maybe in some time, but not immediately.

(If other maintainers disagree, please re-open)

BorzdeG:
It's not created 2 days ago - it's forked from the original repository. In it, for today, only errors from the linter have been fixed and no more improvements have been made

sywhang:
@BorzdeG thanks for this PR.

I don't think there is any pressing need to migrate to the fork (at least just yet), especially since we don't know which fork of this will be the most actively and well-maintained library, unless this particular fork is where the original author/maintainers have pointed to this, but I don't see that. Please correct me if I'm wrong. 

We can revisit replacing this dependency when there is a more pressing need to do so.

PR #1302, opened 2023-07-25
Improve Any stack usage

cdvr1993:
By doing this unsafe approach we are able to save one call to morestack.

This is another approach for #1301.

The benchmarks look way better than the other one:

```
goos: linux
goarch: amd64
pkg: go.uber.org/zap
cpu: AMD EPYC 7B13
BenchmarkAny/normal             15682330                76.51 ns/op
BenchmarkAny/normal-2           16135849                73.94 ns/op
BenchmarkAny/normal-4           16358181                72.74 ns/op
BenchmarkAny/normal-8           16008602                74.66 ns/op
BenchmarkAny/normal_with_logger                  8867184               129.7 ns/op
BenchmarkAny/normal_with_logger-2               10477144               119.9 ns/op
BenchmarkAny/normal_with_logger-4                9692742               121.4 ns/op
BenchmarkAny/normal_with_logger-8                9572548               127.4 ns/op
BenchmarkAny/normal_new_goroutine                1224696               913.0 ns/op
BenchmarkAny/normal_new_goroutine-2              2121248               591.4 ns/op
BenchmarkAny/normal_new_goroutine-4              2760704               394.2 ns/op
BenchmarkAny/normal_new_goroutine-8              5203736               304.2 ns/op
PASS
ok      go.uber.org/zap 19.748s
```

Stack usage is now:

```
field.go:422          0xe6a93                 4881ec40010000                  SUBQ $0x140, SP
```

320 bytes.

### DIFF
--- a/field.go
+++ b/field.go
@@ -23,7 +23,9 @@ package zap
 import (
 	"fmt"
 	"math"
+	"reflect"
 	"time"
+	"unsafe"
 
 	"go.uber.org/zap/zapcore"
 )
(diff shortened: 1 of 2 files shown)


rabbbit:
Could you post the full benchmarks? :)

sywhang:
I believe the tradeoff here is not worth. While this approach's improvements on new goroutine stack allocation scenario is certainly impressive, the more frequent case is when we already have a stable goroutine whose stack size isn't growing anymore and we're logging happily. (the normal_with_logger case) 

Let's discuss options in #1301.

PR #1303, opened 2023-07-27
Optimize zap.Any to allocate less memory on stack

rabbbit:
This is an alternative to #1301 and #1302. It's not as fast as these two options, but it still gives us half the stack reduction without the `unsafe` usage.

This halves the stack usage of `Any`:
```
 field.go:420          0xd16c3                 4881ecf8120000          SUBQ $0x12f8, SP   // 4856
```
to
```
  field.go:420          0xefb80                 4881ec40060000                  SUBQ $0x640, SP   // 1600
```

Interestingly it seems that on both arm64 and amd64 the new code, with the closure, is faster than the plain old switch.
We do see a ~5-10ns delay on `Any` creation if it's used without `logger`, but that's minimal and not realistic.

Bunch of credit for this goes to @cdvr1993, we started independently, I was about to give up but the conversations pushed me forward. In the end he ended up going into a more advanced land where I dare not to enter.

Longer version:

We have identified an issue where zap.Any can cause performance degradation due to stack size.

This is apparently cased by the compiler assigning 4.8kb (a zap.Field per arm of the switch statement) for zap.Any on stack. This can result in an unnecessary runtime.newstack+runtime.copystack. A github issue against Go language is pending.

This can be particularly bad if `zap.Any` was to be used in a new goroutine, since the default goroutine sizes can be as low as 2kb (it can vary depending on average stack size - see golang/go#18138).

This is an alternative to #1301, @cdvr and me were talking about this, and he inspired this idea with the closure.

By using a function and a closure we're able to reduce the size and remove the degradation.
At least on my laptop, this change result in a new performance gain, as all benchmarks show reduced time.

Results can be compared against #1311 

```
❯  go test -bench BenchmarkAny -benchmem -cpu 1

goos: darwin
goarch: arm64
pkg: go.uber.org/zap
BenchmarkAny/string/field-only/typed    158039978                7.462 ns/op           0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      58012468                20.14 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            2925488               416.3 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              2830414               431.0 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed         1000000              1171 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any           1239298               983.8 ns/op           128 B/op          3 allocs/op
[...]

PR #1305, opened 2023-07-28
Optimize zap.Any to use less memory on stack (inline version)

rabbbit:
This is an alternative to:
- #1301 and #1302, and #1304 - a series of PRs that are faster than this one. However, they rely on unsafe.
- #1303 - my own PR that uses closures to reduce the stack size by 60%.

This PR reduces the stack size from:
```
 field.go:420          0xd16c3                 4881ecf8120000          SUBQ $0x12f8, SP   // 4856
```
to
```
  field.go:420          0xcb603                 4881ecb8000000          SUBQ $0xb8, SP // 184
```
so by ~96%. More crucially, `zap.Any` is now as fast as correctly typed methods, like `zap.String`, etc.

The downside is the (slight) increase in the code maintenance - we inline and rely on the compiler correctly re-using small variable sizes. While this is not pretty, it feels safe - the changes were purely mechanical. Future changes and extensions should be easy to review.

Additionally, the new code is (slightly) faster in all cases since we remove 1-2 function calls from all paths. The "in new goroutine" path is most affected, as shown in the benchmarks below.

This was largely inspired by conversations with @cdvr1993. We started looking at this in parallel, but I would have given up if it wasn't for our conversations.
This particular version was inspired by an earlier version of #1304 - where I realized that @cdvr1993 is doing a similar dispatching mechanism that Zap already does via `zapcore` - a possible optimization.

Longer version:

We have identified an issue where zap.Any can cause performance degradation due to stack size.

This is apparently caused by the compiler assigning 4.8kb (a zap.Field per arm of the switch statement) for zap.Any on stack. This can result in an unnecessary runtime.newstack+runtime.copystack. A GitHub issue against Go language is pending.

This can be particularly bad if `zap.Any` was to be used in a new goroutine since the default goroutine sizes can be as low as 2kb (it can vary depending on average stack size - see golang/go#18138).

*Most crucially, `zap.Any` is now as fast as a direct dispatch like `zap.String`.*

Results can be compared with #1311 

```

goos: darwin
goarch: arm64
pkg: go.uber.org/zap
BenchmarkAny/string/field-only/typed    152969097                7.460 ns/op           0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      90461920                13.21 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            3073038               389.1 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              3030273               393.0 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed         1000000              1141 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any           1000000              1143 ns/op             128 B/op          3 allocs/op
BenchmarkAny/stringer/field-only/typed  159827430                7.613 ns/op           0 B/op          0 allocs/op
BenchmarkAny/stringer/field-only/any    52378398                22.89 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/log/typed          3283455               372.7 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log/any            3125164               378.3 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log-go/typed       1000000              1108 ns/op             112 B/op          3 allocs/op
BenchmarkAny/stringer/log-go/any         1000000              1145 ns/op             128 B/op          3 allocs/op
PASS
ok      go.uber.org/zap 19.275s

goos: darwin
goarch: arm64
pkg: go.uber.org/zap
BenchmarkAny/string/field-only/typed    159988640                7.383 ns/op           0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      88060196                13.90 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            2805666               430.7 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              2714698               438.8 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed         1000000              1163 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any           1000000              1178 ns/op             128 B/op          3 allocs/op
BenchmarkAny/stringer/field-only/typed  156450748                7.673 ns/op           0 B/op          0 allocs/op
BenchmarkAny/stringer/field-only/any    51093813                24.31 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/log/typed          2974875               414.9 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log/any            2853394               434.3 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log-go/typed       1000000              1177 ns/op             112 B/op          3 allocs/op
BenchmarkAny/stringer/log-go/any          993225              1188 ns/op             128 B/op          3 allocs/op
PASS
ok      go.uber.org/zap 19.440s
goos: linux
goarch: amd64
pkg: go.uber.org/zap
cpu: AMD EPYC 7B13
BenchmarkAny/string/field-only/typed    47207966                25.48 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      48939066                24.15 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            1651234               739.9 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              1624704               753.6 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed          526701              2288 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any            505273              2331 ns/op             128 B/op          3 allocs/op
BenchmarkAny/stringer/field-only/typed  48647834                24.69 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/field-only/any    30350282                38.55 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/log/typed          1801276               673.0 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log/any            1584859               733.6 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log-go/typed        523908              2307 ns/op             112 B/op          3 allocs/op
BenchmarkAny/stringer/log-go/any          507704              2418 ns/op             128 B/op          3 allocs/op
PASS
ok      go.uber.org/zap 19.012s
```

### DIFF
--- a/field.go
+++ b/field.go
@@ -418,132 +418,311 @@ func Inline(val zapcore.ObjectMarshaler) Field {
 // them. To minimize surprises, []byte values are treated as binary blobs, byte
 // values are treated as uint8, and runes are always treated as integers.
 func Any(key string, value interface{}) Field {
+	// To work around go compiler assigning unreasonably large space on stack
+	// (4kb, one `Field` per arm of the switch statement) which can trigger
+	// performance degradation if `Any` is used in a brand new goroutine.
+	var (
+		t     zapcore.FieldType
+		i     int64
+		s     string
+		iface any
+	)
 	switch val := value.(type) {
 	case zapcore.ObjectMarshaler:
-		return Object(key, val)
+		t = zapcore.ObjectMarshalerType
+		iface = val
 	case zapcore.ArrayMarshaler:
-		return Array(key, val)
+		t = zapcore.ArrayMarshalerType
+		iface = val
 	case bool:
-		return Bool(key, val)
+		var ival int64
+		if val {
+			ival = 1
+		}
+		t = zapcore.BoolType
+		i = ival
 	case *bool:
-		return Boolp(key, val)
+		if val == nil {
+			t = zapcore.ReflectType
+			break
+		}
+		var ival int64
+		if *val {
+			ival = 1
+		}
+		t = zapcore.BoolType
+		i = ival
 	case []bool:
-		return Bools(key, val)
+		t = zapcore.ArrayMarshalerType
+		iface = bools(val)
 	case complex128:
-		return Complex128(key, val)
+		t = zapcore.Complex128Type
+		iface = val
 	case *complex128:
-		return Complex128p(key, val)
+		if val == nil {
+			t = zapcore.ReflectType
+			break
+		}
+		t = zapcore.Complex128Type
+		iface = *val
 	case []complex128:
-		return Complex128s(key, val)
+		t = zapcore.ArrayMarshalerType
+		iface = complex128s(val)
 	case complex64:
-		return Complex64(key, val)
+		t = zapcore.Complex64Type
+		iface = val
 	case *complex64:
-		return Complex64p(key, val)
+		if val == nil {
+			t = zapcore.ReflectType
+			break
+		}
+		t = zapcore.Complex64Type
+		iface = *val
 	case []complex64:
-		return Complex64s(key, val)
+		t = zapcore.ArrayMarshalerType
+		iface = complex64s(val)
 	case float64:
-		return Float64(key, val)
+		t = zapcore.Float64Type
+		i = int64(math.Float64bits(val))
 	case *float64:
-		return Float64p(key, val)
+		if val == nil {
+			t = zapcore.ReflectType
+			break
+		}
+		t = zapcore.Float64Type
+		i = int64(math.Float64bits(*val))
 	case []float64:
-		return Float64s(key, val)
+		t = zapcore.ArrayMarshalerType
+		iface = float64s(val)
 	case float32:
-		return Float32(key, val)
+		t = zapcore.Float32Type
+		i = int64(math.Float32bits(val))
 	case *float32:
-		return Float32p(key, val)
+		if val == nil {
+			t = zapcore.ReflectType
+			break
+		}
+		t = zapcore.Float32Type
+		i = int64(math.Float32bits(*val))
 	case []float32:
-		return Float32s(key, val)
+		t = zapcore.ArrayMarshalerType
+		iface = float32s(val)
 	case int:
-		return Int(key, val)
+		t = zapcore.Int64Type
+		i = int64(val)
 	case *int:
-		return Intp(key, val)
+		if val == nil {
+			t = zapcore.ReflectType
+			break
+		}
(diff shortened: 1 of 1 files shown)


field.go
```diff
@@ -418,132 +418,311 @@ func Inline(val zapcore.ObjectMarshaler) Field {
...
+		if val {
+			ival = 1
+		}
+		t = zapcore.BoolType
+		i = ival
 	case *bool:
-		return Boolp(key, val)
+		if val == nil {
+			t = zapcore.ReflectType
+			break
```
rabbbit, line 447: I also considered pulling all the *p types out of the switch into a separate switch, like

```
	switch val := value.(type) {
	case *int, *bool, *int64, ...:
		if val == nil {
			t = zapcore.ReflectType
			return Field ...
		}
```
but that doesn't work, each of the types would need it's own branch, like

```
	switch val := value.(type) {
	case *int:
		if val == nil {
			t = zapcore.ReflectType
			return Field ...
	case *int64:
		if val == nil {
			t = zapcore.ReflectType
			return Field ...
		}
```
so each of the `*` cases would need to be duplicated in both switches.

Slight benefit in reduction of `if/else` nil pointer checks, but increase in the total function length. Not sure.

cdvr1993:
Even though this doesn't use unsafe, I feel it is still harder to maintain than #1303. Here you are dropping all the typed constructors. Then if in the future anything changes on how we created Fields, we would need to update in 2 places: constructors and Any().

abhinav:
> Even though this doesn't use unsafe, I feel it is still harder to maintain than #1303. Here you are dropping all the typed constructors. Then if in the future anything changes on how we created Fields, we would need to update in 2 places: constructors and Any().

Agreed! I don't think we should merge this as-is.
If I have time this weekend, I may explore auto-generating the unrolled version by having it inspect the current implementation.

rabbbit:
I pushed a slightly smaller version of this to https://github.com/uber-go/zap/compare/pawel/any-int5 - it reduced the LOC changed around a half. It does seem slightly easier to parse than ^.

rabbbit:
Closed in favor of #1310.

PR #1307, opened 2023-07-29
Reduce Any stack usage with map + list of lookups

prashantv:
Alternative to the other Any stack optimizations,  building on top of the tests in https://github.com/uber-go/zap/pull/1303.

This is a different approach to the Any function using a prebuilt a map by type to determine which function to call. The map only works for concrete types, so we also need a list for interfaces, ordered in precedence order (since multiple interfaces may match a single type).

I'm using the output of `go build -gcflags -S` to determine the stack frame size, which allows us to compare the approaches.
```
# master
go.uber.org/zap.Any STEXT size=16672 args=0x20 locals=0x1308 funcid=0x0 align=0x0
        0x0000 00000 (/Users/prashant/dev/3p/zap/field.go:420)  TEXT    go.uber.org/zap.Any(SB), ABIInternal, $4880-32
```
In the above, locals is 0x1308 = 4872, which is 8 bytes less than the frame size on the next line ($4880).

Using this, we can compare the different approaches directly,
```
PR #1301:
go.uber.org/zap.Any STEXT size=5072 args=0x20 locals=0x3e8 funcid=0x0 align=0x0
        0x0000 00000 (/Users/prashant/dev/3p/zap/field.go:421)  TEXT    go.uber.org/zap.Any(SB), ABIInternal, $1008-32

PR #1302:
go.uber.org/zap.Any STEXT size=4048 args=0x20 locals=0x148 funcid=0x0 align=0x0
        0x0000 00000 (/Users/prashant/dev/3p/zap/field.go:422)  TEXT    go.uber.org/zap.Any(SB), ABIInternal, $336-32

PR #1303
go.uber.org/zap.Any STEXT size=6320 args=0x20 locals=0x648 funcid=0x0 align=0x0
        0x0000 00000 (/Users/prashant/dev/3p/zap/field.go:420)  TEXT    go.uber.org/zap.Any(SB), ABIInternal, $1616-32

PR #1304:
go.uber.org/zap.Any STEXT size=3664 args=0x20 locals=0x128 funcid=0x0 align=0x0
        0x0000 00000 (/Users/prashant/dev/3p/zap/field.go:428)  TEXT    go.uber.org/zap.Any(SB), ABIInternal, $304-32

PR #1305:
go.uber.org/zap.Any STEXT size=7120 args=0x20 locals=0xb8 funcid=0x0 align=0x0
        0x0000 00000 (/Users/prashant/dev/3p/zap/field.go:420)  TEXT    go.uber.org/zap.Any(SB), ABIInternal, $192-32

Branch pawel/any-int5
go.uber.org/zap.Any STEXT size=7216 args=0x20 locals=0xc8 funcid=0x0 align=0x0
        0x0000 00000 (/Users/prashant/dev/3p/zap/field.go:420)  TEXT    go.uber.org/zap.Any(SB), ABIInternal, $208-32

This PR:
go.uber.org/zap.Any STEXT size=448 args=0x20 locals=0xf8 funcid=0x0 align=0x0
        0x0000 00000 (/Users/prashant/dev/3p/zap/field.go:506)  TEXT    go.uber.org/zap.Any(SB), ABIInternal, $256-32
```

Benchmarks for this version show `Any` and `String` perform similarly:
```
BenchmarkAny/str-in-go            782828              1422 ns/op              89 B/op          2 allocs/op
BenchmarkAny/any-in-go            723226              1446 ns/op              88 B/op          2 allocs/op
BenchmarkAny/str-in-go-with-stack                 887086              1436 ns/op              88 B/op          2 allocs/op
BenchmarkAny/any-in-go-with-stack                 757072              1455 ns/op              88 B/op          2 allocs/op
```

### DIFF
--- a/field.go
+++ b/field.go
@@ -421,26 +421,11 @@ func typedToAnyFunc[T any](f func(string, T) Field) anyFuncType {
 
 // anyTypeLookup is a map from concrete types to the field constructors.
 var anyTypeLookup = buildAnyTypeLookup()
-var anyInterfaceLookup = []struct {
-	match func(v any) bool
-	fn    func(k string, v any) Field
-}{
-	{
-		match: func(v any) bool { _, ok := v.(zapcore.ObjectMarshaler); return ok },
-		fn:    func(k string, v any) Field { return Object(k, v.(zapcore.ObjectMarshaler)) },
-	},
-	{
-		match: func(v any) bool { _, ok := v.(zapcore.ArrayMarshaler); return ok },
-		fn:    func(k string, v any) Field { return Array(k, v.(zapcore.ArrayMarshaler)) },
-	},
-	{
-		match: func(v any) bool { _, ok := v.(error); return ok },
-		fn:    func(k string, v any) Field { return NamedError(k, v.(error)) },
-	},
-	{
-		match: func(v any) bool { _, ok := v.(fmt.Stringer); return ok },
-		fn:    func(k string, v any) Field { return Stringer(k, v.(fmt.Stringer)) },
-	},
+var anyInterfaceLookup = buildAnyInterfaceLookup()
+
+type interfaceMatcher struct {
+	match func(any) bool
+	field func(string, any) Field
 }
 
 func addTypeLookup[T any](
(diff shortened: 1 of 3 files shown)


rabbbit:
An actual question, not a hidden message - I really don't know.

How do these locals+stack sizes work wtr to nested functions? I imagine they wouldn't show up as stack size, possibly not as locals too? Is there an easy way to measure the whole thing?

What does the `size` mean here? You're clearly winning in this one :)

prashantv:
> How do these locals+stack sizes work wtr to nested functons? I imagine they wouldn't show up as stack size, possibly not as locals too? Is there an easy way to measure the whole thing?

That's right, these only measure the cost of the specific function, not the nested calls as far as I know.

> What does the `size` mean here? You're clearly winning in this one :)

Size is the amount of code directly in the function. With the switch, all the cases are in code, there's a lot of tests + branches to get to the right case. With this approach, we've encoded the cases into data (hashmap), and the lookup logic is also through a separate call, so it makes the size smaller, but it doesn't really mean a huge amount (less cache impact, but we're also calling other functions which could counteract that).

I think the benchmarks you added are the best way to measure the impact, wdyt about landing those as a separate PR since it's useful to have the same benchmarks to compare the different approaches? We may want to add a benchmark that does Any with many types to measure the impact of the different branches as well.

rabbbit:
> I think the benchmarks you added are the best way to measure the impact, wdyt about landing those as a separate PR since it's useful to have the same benchmarks to compare the different approaches? We may want to add a benchmark that does Any with many types to measure the impact of the different branches as well.

SGTM, can put it up. I opted for `string` for now since more complicated things were allocating more so the impact wasn't as obvious. Maybe we're pairing too negative picture then.

prashantv:
Closing since this was primarily intended to improve the stack usage, which #1310 brings in a more minimal way.

This may be worth coming back to for maintainability and possibly other performance reasons, but that should be considered separately.

PR #1308, opened 2023-07-30
Optimize zap.Any to use less memory on stack (intermediate props verion)

rabbbit:
This is an alternative to:
- #1301 and #1302 and #1304 - a series of PRs that are faster than this one. However, they rely on unsafe.
- #1303 - my own PR that uses closures, to reduce the stack size by 60%.
- #1305 - my own PR that inline bunch of loops
- https://github.com/uber-go/zap/compare/pawel/any-int5 that does the same as above, but is slightly easier to parse
- #1307 - a reflect.TypeOf lookup version

THIS PR IS INCOMPLETE - it shows a possible approach, but I wanted to get reviewers thoughts before typing everything in.

I originally thought we can use a `type props strucy` intermediary struct to store the data, but that hits the same problem: every `props` would get it's own slot on the stack. This avoids this by returning the raw data.

Pros:
- the implementation is shared between `Any` and strongly typed Fields
- no reflect or unsafe
- reduced the stack significantly - we should be able to get to the same ~180 bytes as ~1305.
- no peft penalty for strongly typed versions, at least on ARM64 it's compiled away.

Cons:
- the code gets a bit harder to maintain. It's significantly better than #1305 I would say though.

### DIFF
--- a/field.go
+++ b/field.go
@@ -37,6 +37,20 @@ var (
 	_maxTimeInt64 = time.Unix(0, math.MaxInt64)
 )
 
+// toField converts props to a Field.
+// This is syntastic sugar that allows implementation sharing between
+// strongly typed methods and zap.Any. We expect `toField` to be inlined
+// and have no cost.
+func toField(key string, t zapcore.FieldType, i int64, s string, iface any) Field {
+	return Field{
+		Key:       key,
+		Type:      t,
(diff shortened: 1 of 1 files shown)


rabbbit:
Closed in favor of #1310.

PR #1310, opened 2023-07-30
zap.Any: Reduce stack size with generics

abhinav:
Yet another attempt at reducing the stack size of zap.Any,
borrowing from #1301, #1303, #1304, #1305, #1307, and #1308.

This approach defines a generic data type for field constructors
of a specific type. This is similar to the lookup map in #1307,
minus the map lookup, the interface match, or reflection.

    type anyFieldC[T any] func(string, T) Field

The generic data type provides a non-generic method
matching the interface:

    interface{ Any(string, any) Field }

**Stack size**:
The stack size of zap.Any following this change is 0xc0 (192 bytes).

    % go build -gcflags -S 2>&1 | grep ^go.uber.org/zap.Any
    go.uber.org/zap.Any STEXT size=5861 args=0x20 locals=0xc0 funcid=0x0 align=0x0

This is just 8 bytes more than #1305,
which is the smallest stack size of all other attempts.

**Allocations**:
Everything appears to get inlined with no heap escapes:

    % go build -gcflags -m 2>&1 |
      grep field.go |
      perl -n -e 'next unless m{^./field.go:(\d+)}; print if ($1 >= 413)' |
      grep 'escapes'
    [no output]

(Line 413 declares anyFieldC)

Besides that, the output of `-m` for the relevant section of code
consists of almost entirely:

    ./field.go:415:6: can inline anyFieldC[go.shape.bool].Any
    ./field.go:415:6: can inline anyFieldC[go.shape.[]bool].Any
    ./field.go:415:6: can inline anyFieldC[go.shape.complex128].Any
    [...]
    ./field.go:415:6: inlining call to anyFieldC[go.shape.complex128].Any
    ./field.go:415:6: inlining call to anyFieldC[go.shape.[]bool].Any
    ./field.go:415:6: inlining call to anyFieldC[go.shape.bool].Any

Followed by:

    ./field.go:428:10: leaking param: key
    ./field.go:428:22: leaking param: value

**Maintainability**:
Unlike some of the other approaches, this variant is more maintainable.
The `zap.Any` function looks roughly the same.
Adding new branches there is obvious, and requires no duplication.

**Performance**:
This is a net improvement against master on BenchmarkAny's
log-go checks that log inside a new goroutine.

```
name                           old time/op    new time/op    delta
Any/string/field-only/typed      25.2ns ± 1%    25.6ns ± 2%     ~     (p=0.460 n=5+5)
Any/string/field-only/any        56.9ns ± 3%    79.4ns ± 0%  +39.55%  (p=0.008 n=5+5)
Any/string/log/typed             1.47µs ± 0%    1.49µs ± 4%   +1.58%  (p=0.016 n=4+5)
Any/string/log/any               1.53µs ± 2%    1.55µs ± 1%   +1.37%  (p=0.016 n=5+5)
Any/string/log-go/typed          5.97µs ± 6%    5.99µs ± 1%     ~     (p=0.151 n=5+5)
Any/string/log-go/any            10.9µs ± 0%     6.2µs ± 0%  -43.32%  (p=0.008 n=5+5)
Any/stringer/field-only/typed    25.3ns ± 1%    25.5ns ± 1%   +1.09%  (p=0.008 n=5+5)
Any/stringer/field-only/any      85.5ns ± 1%   124.5ns ± 0%  +45.66%  (p=0.008 n=5+5)
Any/stringer/log/typed           1.43µs ± 1%    1.42µs ± 2%     ~     (p=0.175 n=4+5)
Any/stringer/log/any             1.50µs ± 1%    1.56µs ± 6%   +4.20%  (p=0.008 n=5+5)
Any/stringer/log-go/typed        5.94µs ± 0%    5.92µs ± 0%   -0.40%  (p=0.032 n=5+5)
Any/stringer/log-go/any          11.1µs ± 2%     6.3µs ± 0%  -42.93%  (p=0.008 n=5+5)

name                           old alloc/op   new alloc/op   delta
Any/string/field-only/typed       0.00B          0.00B          ~     (all equal)
Any/string/field-only/any         0.00B          0.00B          ~     (all equal)
Any/string/log/typed              64.0B ± 0%     64.0B ± 0%     ~     (all equal)
Any/string/log/any                64.0B ± 0%     64.0B ± 0%     ~     (all equal)
Any/string/log-go/typed            112B ± 0%      112B ± 0%     ~     (all equal)
Any/string/log-go/any              128B ± 0%      128B ± 0%     ~     (all equal)
Any/stringer/field-only/typed     0.00B          0.00B          ~     (all equal)
Any/stringer/field-only/any       0.00B          0.00B          ~     (all equal)
Any/stringer/log/typed            64.0B ± 0%     64.0B ± 0%     ~     (all equal)
Any/stringer/log/any              64.0B ± 0%     64.0B ± 0%     ~     (all equal)
Any/stringer/log-go/typed          112B ± 0%      112B ± 0%     ~     (all equal)
Any/stringer/log-go/any            128B ± 0%      128B ± 0%     ~     (all equal)

name                           old allocs/op  new allocs/op  delta
Any/string/field-only/typed        0.00           0.00          ~     (all equal)
Any/string/field-only/any          0.00           0.00          ~     (all equal)
Any/string/log/typed               1.00 ± 0%      1.00 ± 0%     ~     (all equal)
Any/string/log/any                 1.00 ± 0%      1.00 ± 0%     ~     (all equal)
Any/string/log-go/typed            3.00 ± 0%      3.00 ± 0%     ~     (all equal)
Any/string/log-go/any              3.00 ± 0%      3.00 ± 0%     ~     (all equal)
Any/stringer/field-only/typed      0.00           0.00          ~     (all equal)
Any/stringer/field-only/any        0.00           0.00          ~     (all equal)
Any/stringer/log/typed             1.00 ± 0%      1.00 ± 0%     ~     (all equal)
Any/stringer/log/any               1.00 ± 0%      1.00 ± 0%     ~     (all equal)
Any/stringer/log-go/typed          3.00 ± 0%      3.00 ± 0%     ~     (all equal)
Any/stringer/log-go/any            3.00 ± 0%      3.00 ± 0%     ~     (all equal)
```

It causes a regression in "field-only"
which calls the field constructor and discards the result
without using it in a logger.
I believe this is acceptable because that's not a real use case;
we expect the result to be used with a logger.

### DIFF
--- a/field.go
+++ b/field.go
@@ -410,6 +410,43 @@ func Inline(val zapcore.ObjectMarshaler) Field {
 	}
 }
 
+// We discovered an issue where zap.Any can cause a performance degradation
+// when used in new goroutines.
+//
+// This happens because the compiler assigns 4.8kb (one zap.Field per arm of
+// switch statement) of stack space for zap.Any when it takes the form:
+//
+//	switch v := v.(type) {
+//	case string:
+//		return String(key, v)
+//	case int:
+//		return Int(key, v)
+//		// ...
+//	default:
+//		return Reflect(key, v)
+//	}
+//
+// To avoid this, we use the type switch to assign a value to a single local variable
+// and then call a function on it.
+// The local variable is just a function reference so it doesn't allocate
+// when converted to an interface{}.
+//
+// A fair bit of experimentation went into this.
+// See also:
+//
+// - https://github.com/uber-go/zap/pull/1301
+// - https://github.com/uber-go/zap/pull/1303
+// - https://github.com/uber-go/zap/pull/1304
+// - https://github.com/uber-go/zap/pull/1305
+// - https://github.com/uber-go/zap/pull/1308
+type anyFieldC[T any] func(string, T) Field
+
+func (f anyFieldC[T]) Any(key string, val any) Field {
+	v, _ := val.(T)
+	// val is guaranteed to be a T, except when it's nil.
+	return f(key, v)
+}
+
 // Any takes a key and an arbitrary value and chooses the best way to represent
 // them as a field, falling back to a reflection-based approach only if
 // necessary.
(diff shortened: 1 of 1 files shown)


field.go
```diff
@@ -410,6 +410,14 @@ func Inline(val zapcore.ObjectMarshaler) Field {
 	}
 }
 
+type anyFieldC[T any] func(string, T) Field
+
+func (f anyFieldC[T]) Any(key string, val any) Field {
```
rabbbit, line 415: Should this be private? Does it matter if it's public?

field.go
```diff
@@ -418,132 +426,136 @@ func Inline(val zapcore.ObjectMarshaler) Field {
 // them. To minimize surprises, []byte values are treated as binary blobs, byte
 // values are treated as uint8, and runes are always treated as integers.
 func Any(key string, value interface{}) Field {
-	switch val := value.(type) {
+	var c interface{ Any(string, any) Field }
```
rabbbit, line 429: TIL

rabbbit (commented):
Very nice. I did wonder if generics could help here, but this is clearly beyond me.

For the fun of it, I'll try running this later on arm and amd to compare against previous numbers. I'll also try running the current head against gotip just to verify that this didn't change and that go1.22 doesn't change anything.

After Prashant suggestion,I also planned pulling the benchmarks into a separate PR and adding some other cases. Stay tuned.

field.go
abhinav: It doesn't matter if it's public or private here.
An advantage of `Any` is that it's not going to conflict with `any`, the built-in type.
I also try to document the "public" interface of internal types by making them public like this.
(So someone using anyFieldC knows they can use Any, but they can't use someInternalStateManagement.)

rabbbit:
I ran the new benchmarks:

```

goos: darwin
goarch: arm64
pkg: go.uber.org/zap
BenchmarkAny/string/field-only/typed    157439328                7.418 ns/op           0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      57899338                20.86 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            2960004               409.2 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              2793056               436.6 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed          967309              1228 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any           1000000              1179 ns/op             128 B/op          3 allocs/op
BenchmarkAny/stringer/field-only/typed  160436690                7.584 ns/op           0 B/op          0 allocs/op
BenchmarkAny/stringer/field-only/any    36087421                33.24 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/log/typed          3102452               386.9 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log/any            2901717               429.1 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log-go/typed       1000000              1162 ns/op             112 B/op          3 allocs/op
BenchmarkAny/stringer/log-go/any         1002906              1153 ns/op             128 B/op          3 allocs/op
PASS
ok      go.uber.org/zap 20.298s

goos: darwin
goarch: arm64
pkg: go.uber.org/zap
BenchmarkAny/string/field-only/typed    156370705                7.367 ns/op           0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      57934047                21.47 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            2716717               442.1 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              2668359               465.0 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed         1000000              1223 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any            964839              1223 ns/op             128 B/op          3 allocs/op
BenchmarkAny/stringer/field-only/typed  157792915                7.782 ns/op           0 B/op          0 allocs/op
BenchmarkAny/stringer/field-only/any    35110689                34.05 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/log/typed          3024481               410.7 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log/any            2727921               445.9 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log-go/typed       1000000              1138 ns/op             112 B/op          3 allocs/op
BenchmarkAny/stringer/log-go/any         1000000              1216 ns/op             128 B/op          3 allocs/op
PASS
ok      go.uber.org/zap 19.557s
pawel@go(pawel.devpod-us-va) ~/zap
goos: linux
goarch: amd64
pkg: go.uber.org/zap
cpu: AMD EPYC 7B13
BenchmarkAny/string/field-only/typed    47630637                25.28 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      25443076                47.87 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            1578100               738.2 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              1540040               784.3 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed          484922              2332 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any            489837              2414 ns/op             128 B/op          3 allocs/op
BenchmarkAny/stringer/field-only/typed  48447813                24.75 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/field-only/any    17729155                68.85 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/log/typed          1693166               681.6 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log/any            1609806               788.9 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log-go/typed        522900              2318 ns/op             112 B/op          3 allocs/op
BenchmarkAny/stringer/log-go/any          502071              2356 ns/op             128 B/op          3 allocs/op
PASS
ok      go.uber.org/zap 19.148s
```

Which (1) seems comparable with all the other PRs (2) shows improvement vs #1311 (apart from the field-only `Any` as you pointed out yourself)

rabbbit (approved):
Approved, but I don't think I can rebase this for you, can I?

abhinav:
Rebased!

field.go
```diff
@@ -410,6 +410,14 @@ func Inline(val zapcore.ObjectMarshaler) Field {
 	}
 }
 
+type anyFieldC[T any] func(string, T) Field
```
prashantv, line 413: can you add a comment about why we're doing all of this

ideally we'd have an upstream issue linked which describes the compiler issue

abhinav: Added more details. I don't think there's an upstream compiler issue yet. I believe there are plans to write one.

rabbbit: Issue now exists, adding a link in https://github.com/uber-go/zap/pull/1387

rabbbit:
@abhinav do we want to revert this with 

> The compiler in Go 1.23 can now overlap the stack frame slots of local variables accessed in disjoint regions of a function, which reduces stack usage for Go applications.

I kinda like your code, though. After the initial 5 minutes of wth, I appreciate the elegance of it. 
(the 5 minutes of wth happens every time I look at it too)

abhinav:
I think it makes sense to keep it as-is until Go 1.24 is released, since Go 1.22 is still supported by Zap.
Once 1.23 is the minimum supported version, we can revert this if we want to.
(I don't have strong feelings on that. The code is not overly complex so the 110 bytes may be worth it.)

PR #1311, opened 2023-07-30
zap.Any add benchmarks

rabbbit:
This is a prefactor for #1301, #1302, #1304, #1305, #1307, #1308 and #1310.

We're writing various approaches to reduce the stock size and it's painful to keep copy-pasting the tests between PRs. This was suggested in @prashantv in #1307.

The tests are mostly based on tests in #1303, but made "more generic", as #1307 we might want to test across more than just a single type. It does make the tests a bit harder to setup. Some of the setup is inconvenient (we duplicate the value in both `typed` and `any` version of the tests) but hopefully okay to understand. A fully non-duplicated alternative would likely require something like #1310 itself.

For #1307 in particular a test against interface type would likely be needed, so adding it here too.

The tests compare two code paths, with the same arguments, one using a strongly typed method and second using `zap.Any`. We have:
- a simple "create field" case for a baseline
- a "create and log" case for a realistic case (we typically log the fields)
- a "create and log in a goroutine" case for the pathological case we're trying to solve for.
- -a "create and long in goroutine in a pre-warmed system" that does the above- we decided it's not worth the complication.

The test results are:
```
goos: darwin
goarch: arm64
pkg: go.uber.org/zap
BenchmarkAny/string/field-only/typed    161981473                7.374 ns/op           0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      82343354                14.67 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            2965648               416.1 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              2920292               418.8 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed         1000000              1158 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any            553144              2152 ns/op             128 B/op          3 allocs/op
BenchmarkAny/stringer/field-only/typed  160509367                7.548 ns/op           0 B/op          0 allocs/op
BenchmarkAny/stringer/field-only/any    51330402                23.45 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/log/typed          3221404               377.0 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log/any            2726443               393.5 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log-go/typed       1000000              1129 ns/op             112 B/op          3 allocs/op
BenchmarkAny/stringer/log-go/any          558602              2147 ns/op             128 B/op          3 allocs/op
PASS
ok      go.uber.org/zap 19.426s
```

On gotip:
```

goos: darwin
goarch: arm64
pkg: go.uber.org/zap
BenchmarkAny/string/field-only/typed    155084869                7.603 ns/op           0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      82740788                14.55 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            2800495               411.6 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              2896258               411.7 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed         1000000              1155 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any            551599              2168 ns/op             128 B/op          3 allocs/op
BenchmarkAny/stringer/field-only/typed  159505488                7.578 ns/op           0 B/op          0 allocs/op
BenchmarkAny/stringer/field-only/any    51406354                23.78 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/log/typed          3011210               388.6 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log/any            3010370               395.1 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log-go/typed       1000000              1161 ns/op             112 B/op          3 allocs/op
BenchmarkAny/stringer/log-go/any          553860              2161 ns/op             128 B/op          3 allocs/op
PASS
ok      go.uber.org/zap 19.391s
```

on amd64 (similar, 2x worse stack growth impact)
```
goos: linux
goarch: amd64
pkg: go.uber.org/zap
cpu: AMD EPYC 7B13
BenchmarkAny/string/field-only/typed    47534053                25.23 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/field-only/any      36913526                32.57 ns/op            0 B/op          0 allocs/op
BenchmarkAny/string/log/typed            1693508               725.7 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log/any              1576172               765.3 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string/log-go/typed          516832              2343 ns/op             112 B/op          3 allocs/op
BenchmarkAny/string/log-go/any            243692              4404 ns/op             128 B/op          3 allocs/op
BenchmarkAny/stringer/field-only/typed  48735537                24.73 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/field-only/any    26115684                47.24 ns/op            0 B/op          0 allocs/op
BenchmarkAny/stringer/log/typed          1761630               677.5 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log/any            1646913               705.4 ns/op            64 B/op          1 allocs/op
BenchmarkAny/stringer/log-go/typed        534187              2275 ns/op             112 B/op          3 allocs/op
BenchmarkAny/stringer/log-go/any          273787              4348 ns/op             128 B/op          3 allocs/op
PASS
ok      go.uber.org/zap 18.890s
```

### DIFF
--- a/logger_bench_test.go
+++ b/logger_bench_test.go
@@ -22,6 +22,8 @@ package zap
 
 import (
 	"errors"
+	"runtime"
+	"sync"
 	"testing"
 	"time"
 
@@ -236,5 +238,132 @@ func Benchmark100Fields(b *testing.B) {
 			second[i] = Int("foo", i+batchSize)
 		}
 		logger.With(first...).Info("Child loggers with lots of context.", second...)
+	}
+}
+
+func dummy(wg *sync.WaitGroup, s string, i int) string {
+	if i == 0 {
+		wg.Wait()
+		return "1" + s
+	}
+	return dummy(wg, s, i-1)
+}
+
+func stackGrower(n int) *sync.WaitGroup {
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+
+	go dummy(&wg, "hi", n)
+	return &wg
+}
+
+func BenchmarkAny(b *testing.B) {
+	var (
+		logger = New(
+			zapcore.NewCore(
+				zapcore.NewJSONEncoder(NewProductionConfig().EncoderConfig),
+				&ztest.Discarder{},
+				DebugLevel,
+			),
+		)
+		key = "some-long-string-longer-than-16"
+	)
+
+	tests := []struct {
+		name   string
+		typed  func() Field
+		anyArg any
+	}{
+		{
+			name:   "string",
+			typed:  func() Field { return String(key, "yet-another-long-string") },
+			anyArg: "yet-another-long-string",
+		},
+		{
+			name:   "stringer",
+			typed:  func() Field { return Stringer(key, InfoLevel) },
+			anyArg: InfoLevel,
+		},
+	}
+
+	for _, tt := range tests {
+		b.Run(tt.name+"-typ-no-logger", func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				f := tt.typed()
+				runtime.KeepAlive(f)
+			}
+		})
+		b.Run(tt.name+"-any-no-logger", func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				f := Any(key, tt.anyArg)
+				runtime.KeepAlive(f)
+			}
+		})
+		b.Run(tt.name+"-typ-logger", func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				logger.Info("", tt.typed())
+			}
+		})
+		b.Run(tt.name+"-any-logger", func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				logger.Info("", Any(key, tt.anyArg))
+			}
+		})
+		b.Run(tt.name+"-typ-logger-go", func(b *testing.B) {
+			wg := sync.WaitGroup{}
+			wg.Add(b.N)
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				go func() {
+					logger.Info("", tt.typed())
+					wg.Done()
+				}()
+			}
+			wg.Wait()
+		})
+		b.Run(tt.name+"-any-logger-go", func(b *testing.B) {
+			wg := sync.WaitGroup{}
+			wg.Add(b.N)
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				go func() {
+					logger.Info("", Any(key, tt.anyArg))
+					wg.Done()
+				}()
+			}
+			wg.Wait()
+		})
+		// The stack growing below simulates production setup where some other
+		// goroutines exist and affect the starting goroutine stack size up.
+		// Otherwise, for tests with 2+ goroutines, the cost of starting the goroutine
+		// dominates and the cost of `any` stack overallocation is not visible.
+		b.Run(tt.name+"-typ-logger-go-stack", func(b *testing.B) {
+			wg := sync.WaitGroup{}
+			wg.Add(b.N)
+			defer stackGrower(2000).Done()
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				go func() {
+					logger.Info("", tt.typed())
+					wg.Done()
+				}()
+			}
+			wg.Wait()
+			b.StopTimer()
+		})
+		b.Run(tt.name+"-any-logger-go-stack", func(b *testing.B) {
+			wg := sync.WaitGroup{}
+			wg.Add(b.N)
+			defer stackGrower(2000).Done()
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				go func() {
+					logger.Info("", Any(key, tt.anyArg))
+					wg.Done()
+				}()
+			}
+			wg.Wait()
+			b.StopTimer()
+		})
 	}
 }
--- a/logger_bench_test.go
+++ b/logger_bench_test.go
@@ -249,18 +249,6 @@ func dummy(wg *sync.WaitGroup, s string, i int) string {
 	return dummy(wg, s, i-1)
 }
 
-// increaseAvgStack starts a background goroutine with a variable
-// stack size. The goal is to move the average stack size higher,
-// since https://go-review.googlesource.com/c/go/+/345889 this affects
-// goroutine starting stack size.
-func increaseAvgStack(n int) *sync.WaitGroup {
-	wg := sync.WaitGroup{}
-	wg.Add(1)
-
-	go dummy(&wg, "hi", n)
-	return &wg
-}
-
 func BenchmarkAny(b *testing.B) {
 	key := "some-long-string-longer-than-16"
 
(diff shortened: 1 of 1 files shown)


abhinav (approved):
LGTM

logger_bench_test.go
```diff
@@ -238,3 +240,130 @@ func Benchmark100Fields(b *testing.B) {
...
+			wg := sync.WaitGroup{}
+			wg.Add(b.N)
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				go func() {
+					logger.Info("", tt.typed())
+					wg.Done()
+				}()
+			}
+			wg.Wait()
```
abhinav, line 322: Actually, is this right? It'll spawn `b.N` goroutines and count that as part of a run.
I think maybe you want a fixed number of goroutines spawned inside each b.N loop, and wait for them to finish also in the loop?

```
const Count = 1000
for i := 0; i < b.N; i++ {
    wg.Add(Concurrency)
    for j := 0; j < Count; j++ {
      go func() { ... }()
    }
    wg.Wait()
}
```

logger_bench_test.go
```diff
@@ -238,3 +240,130 @@ func Benchmark100Fields(b *testing.B) {
...
+		})
+		b.Run(tt.name+"-typ-logger", func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				logger.Info("", tt.typed())
+			}
+		})
+		b.Run(tt.name+"-any-logger", func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				logger.Info("", Any(key, tt.anyArg))
+			}
```
abhinav, line 310: For these two cases, we might want to use withBenchedLogger.

https://github.com/abhinav/zap/blob/1058d37e5320ddd1a6da01e76247fa64dd8ad26d/logger_bench_test.go#L53

That's used in existing logging benchmarks:

https://github.com/abhinav/zap/blob/1058d37e5320ddd1a6da01e76247fa64dd8ad26d/logger_bench_test.go#L155-L159

Although it'll use `b.RunParallel` -- I don't know if that's okay?

logger_bench_test.go
rabbbit: So the actual issue, as (badly) reported by a user was: "if I start a new goroutine and `zap.Any` in it, it's slow" - this is where this came from, with an exact translation.

The goal was also to have sequential progress:
- how long does it take to create a `Field`
- how long does it take to log a `Field`
- how long does it take to log a `Field` in a new goroutine - including the setup and teardown cost.

Spawning o 1000 goroutines would at least break the sequentially above, I think?

It seemed that as long as we're comparing the same cost for typed version and `Any`, the benchmarks are representative.

logger_bench_test.go
rabbbit: Yeah, I don't know if it matters.

Looking at https://pkg.go.dev/testing#B.RunParallel, it means we'll start a GOMAXPROCS goroutines, and then distribute the actual tests (which will start their own goroutines) there.

Does it matter? Does it not matter? I can test/try.

rabbbit: Oh, just meant these two cases, rather than all loggers.

```
BenchmarkAny/string-typ-logger                   2959306               406.9 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string-typ-logger-12               15433677                80.36 ns/op           64 B/op          1 allocs/op
BenchmarkAny/string-any-logger                   2845083               413.0 ns/op            64 B/op          1 allocs/op
BenchmarkAny/string-any-logger-12               15772636                79.69 ns/op           64 B/op          1 allocs/op
```

the difference for 12- gomacprocs is huge.

I guess it makes sense to include it like this.

logger_bench_test.go
```diff
@@ -238,3 +240,130 @@ func Benchmark100Fields(b *testing.B) {
...
+		},
+		{
+			name:   "stringer",
+			typed:  func() Field { return Stringer(key, InfoLevel) },
+			anyArg: InfoLevel,
+		},
+	}
+
+	for _, tt := range tests {
+		b.Run(tt.name+"-typ-no-logger", func(b *testing.B) {
```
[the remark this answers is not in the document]
rabbbit: Nice indeed. The nesting is unpleasant, but more bearable than I had thought.

logger_bench_test.go
rabbbit: Added to all tests.

logger_bench_test.go
```diff
@@ -238,3 +240,130 @@ func Benchmark100Fields(b *testing.B) {
...
+					logger.Info("", Any(key, tt.anyArg))
+					wg.Done()
+				}()
+			}
+			wg.Wait()
+		})
+		// The stack growing below simulates production setup where some other
+		// goroutines exist and affect the starting goroutine stack size up.
+		// Otherwise, for tests with 2+ goroutines, the cost of starting the goroutine
+		// dominates and the cost of `any` stack overallocation is not visible.
```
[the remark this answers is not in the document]
rabbbit: So I'm hesitating about the value of this.

In my earlier experiments (#1303) the value of this was clear - the cost of "typed" version was half of the `any` if the average stack was higher. Especially on high GOMAXPROCS tests.

As the benchmarking setup grew, this became less pronounced:
```
BenchmarkAny/string/log-go/typed-12              2151031               550.7 ns/op            99 B/op          2 allocs/op
BenchmarkAny/string/log-go/any-12                1931274               650.3 ns/op           115 B/op          2 allocs/op
BenchmarkAny/string/log-go-stack/typed-12        2042646               520.8 ns/op            96 B/op          2 allocs/op
BenchmarkAny/string/log-go-stack/any-12          2068048               622.4 ns/op           112 B/op          2 allocs/op
```
to an extend that I'm wondering about skipping this altogether.

I'll tinker with this a bit more, but maybe we should remove it. and rely on GOMACPROCS=2 as indication of our changes working instead.

logger_bench_test.go
```diff
@@ -238,3 +240,132 @@ func Benchmark100Fields(b *testing.B) {
...
+					wg := sync.WaitGroup{}
+					wg.Add(b.N)
+					b.ResetTimer()
+					withBenchedLogger(b, func(log *Logger) {
+						go func() {
+							log.Info("", Any(key, tt.anyArg))
+							wg.Done()
+						}()
+					})
+					wg.Wait()
```
abhinav, line 335: Are you sure this works? (Related to the prior comment about `b.N` goroutines.)
Each iteration (whether with RunParallel or with a plain for loop), will last only as long as spawning the goroutine takes. I'm not sure this will do what we want? 
I know we're trying to measure the cost inside a new goroutine, but this doesn't seem right?
Maybe we need to spawn and wait for the goroutine in each iteration of the loop?

```go
var wg sync.WaitGroup
wg.Add(1)
go func() {
    ...
    wg.Done()
}()
wg.Wait()
```

This will measure the cost of the whole operation (including spawning the goroutine) per iteration.

logger_bench_test.go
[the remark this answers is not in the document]
rabbbit: Actually, it seems to work.

```
BenchmarkAny/string/log-go-stack/typed           1826714               659.1 ns/op           112 B/op          3 allocs/op
BenchmarkAny/string/log-go-stack/typed-12        8111128               143.7 ns/op           112 B/op          3 allocs/op
BenchmarkAny/string/log-go-stack/any             1606474               687.5 ns/op           128 B/op          3 allocs/op
BenchmarkAny/string/log-go-stack/any-12          3320598               373.5 ns/op           128 B/op          3 allocs/op
```

`any-12` is twice slower than `typed-12`, while `any` is relatively as slow `typed` (single maxprocs test)

logger_bench_test.go
rabbbit: Hm.

It was testing something reasonable, because the changes we were making to the stack were clearly having an effect. I agree that this is better though, thanks.

abhinav (approved):
LGTM! Thanks for all the fixes.

Please get @prashantv's stamp on the stack growing end of things.

logger_bench_test.go
```diff
@@ -238,3 +240,132 @@ func Benchmark100Fields(b *testing.B) {
...
+// stack size. The goal is to move the average stack size higher,
+// since https://go-review.googlesource.com/c/go/+/345889 this affects
+// goroutine starting stack size.
+func increaseAvgStack(n int) *sync.WaitGroup {
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+
+	go dummy(&wg, "hi", n)
+	return &wg
+}
```
prashantv, line 262: We can use a simple `var data[8000]byte` to grow the stack size. 
```go
func growAvgStack() {
	var wg sync.WaitGroup
	block := make(chan struct{})
	for i := 0; i < 1000; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()

			// Effectively the minimum size for the starting goroutine size.
			var data [8000]byte
			<-block
			runtime.KeepAlive(data)
		}()
	}

	close(block)
	wg.Wait()
}
```

Verified that it works by reading + printing the starting stack size of `metrics.Sample{Name: "/gc/stack/starting-size:bytes"}`

https://go.dev/play/p/3MFl3S7Zz-A

prashantv (approved):

PR #1317, opened 2023-08-08
[draft/wip] Adding (initial) fortio.org/log entry

ldemailly:
Ran `make updatereadme` and `go test -v -bench="." -benchtime=3s -benchmem` on my M1 max laptop

### DIFF
--- a/benchmarks/go.mod
+++ b/benchmarks/go.mod
@@ -5,6 +5,7 @@ go 1.19
 replace go.uber.org/zap => ../
 
 require (
+	fortio.org/log v1.10.0
 	github.com/apex/log v1.9.0
 	github.com/go-kit/log v0.2.1
 	github.com/rs/zerolog v1.28.0
@@ -13,6 +14,11 @@ require (
 	go.uber.org/zap v1.23.0
 	golang.org/x/exp v0.0.0-20230310171629-522b1b587ee0
 	gopkg.in/inconshreveable/log15.v2 v2.0.0-20221122034931-555555054819
+)
+
+replace (
+// 	fortio.org/log => ../../../fortio.org/log
+ 	go.ubder.org/zap => ../
 )
 
 require (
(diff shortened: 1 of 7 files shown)


benchmarks/go.mod
```diff
@@ -15,6 +16,9 @@ require (
 	gopkg.in/inconshreveable/log15.v2 v2.0.0-20221122034931-555555054819
 )
 
+//	fortio.org/log => ../../../fortio.org/log
+replace go.ubder.org/zap => ../
```
abhinav, line 20: This entry already exists at the top of the file.

benchmarks/zap_test.go
```diff
@@ -34,20 +34,21 @@ import (
 var (
 	errExample = errors.New("fail")
 
-	_messages   = fakeMessages(1000)
-	_tenInts    = []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 0}
+	_messages = fakeMessages(1000)
+	// Go optimizes away allocations for numbers below 256
+	_tenInts    = []int{1001, 1002, 1003, 1004, 1005, 1006, 1007, 1008, 1009, 1010}
```
abhinav, line 39: Fair point. If you're still experimenting with this PR, you can probably get this specific change merged as a separate PR.

abhinav (commented):
I'm guessing this is still WIP.
I'm supportive of adding another entry to the list when you're ready.

benchmarks/go.mod
ldemailly: opsa! I was looking at the go.uber.org/zap v1.23.0 in require and wanted to check I was running the latest, but didn't see the existing replace, thx

benchmarks/zap_test.go
ldemailly: thanks btw, how / on which machine do you run `make updatereadme` as the absolute (and seemingly even some times the relative numbers) depends on the hw/arch/os it's being ran on ?

abhinav: I think original runs were on Intel Macbook Pros, and since then on Apple Silicon Macbook Pros.
Absolute numbers are less important than the relative performance, so as long as the numbers are consistent and the benchmarks were all run on the same machine, it should be good to check-in.

PR #1319, opened 2023-08-14
Add WithLazy Logger method

jquirke:
As per internal Uber discussion


logger.go
```diff
@@ -178,11 +180,57 @@ func (log *Logger) With(fields ...Field) *Logger {
...
+/* deferred with core is a zap core that
+ * defers the evaluation of the core's with
+ * until the first log is written.
+ *
+ * This saves significant memory allocation
+ * and CPU with complex cores
+ */
+
+type deferredWithCore struct {
+	zapcore.Core
```
jquirke, line 203: put the sync.Once object adjacent to Core to increase the likelihood it ends up on the same cache line

logger.go
```diff
@@ -178,11 +180,57 @@ func (log *Logger) With(fields ...Field) *Logger {
...
+	if log.deferWith {
+		log.core = &deferredWithCore{Core: log.core, fields: fields}
+		return log
+	}
 	l := log.clone()
 	l.core = l.core.With(fields)
 	return l
 }
 
+// todo move this elsewhree
```
jquirke, line 192: this should be in core package, with NewDeferredWithCore method

logger.go
```diff
@@ -178,11 +180,57 @@ func (log *Logger) With(fields ...Field) *Logger {
...
+
+/* deferred with core is a zap core that
+ * defers the evaluation of the core's with
+ * until the first log is written.
+ *
+ * This saves significant memory allocation
```
jquirke, line 198: document with benchmarks

logger.go
```diff
@@ -178,11 +180,57 @@ func (log *Logger) With(fields ...Field) *Logger {
...
+	sync.Once
+}
+
+func (d *deferredWithCore) initOnce() {
+	d.Once.Do(func() {
+		d.Core = d.Core.With(d.fields)
+	})
+}
+
+func (d *deferredWithCore) With(fields []zapcore.Field) zapcore.Core {
```
jquirke, line 214: explain why this is needed (in case people directly access the core)

logger.go
```diff
@@ -178,11 +180,57 @@ func (log *Logger) With(fields ...Field) *Logger {
 	if len(fields) == 0 {
 		return log
 	}
+	if log.deferWith {
+		log.core = &deferredWithCore{Core: log.core, fields: fields}
```
jquirke, line 184: this will cause cross talk
```suggestion
                 l := *log
		l.core = &deferredWithCore{Core: log.core, fields: fields}
		return &l
```

logger.go
```diff
@@ -178,6 +179,11 @@ func (log *Logger) With(fields ...Field) *Logger {
 	if len(fields) == 0 {
 		return log
 	}
+	if log.deferWith {
+		l := *log
```
rabbbit, line 183:
```suggestion
		l := log.clone()
```

zapcore/deferred_with.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+func (d *deferredWithCore) initOnce() {
+	d.Once.Do(func() {
+		d.Core = d.Core.With(d.fields)
+	})
+}
+
+func (d *deferredWithCore) With(fields []Field) Core {
```
rabbbit, line 51: from internal discussion

I still wonder if this couldn't do something like
```
func (d *deferredWithCore) With(fields []Field) Core {
	return &deferredWithCore{
		Core:   core,
		fields: d.fields+fields, // paraphrasing, likely as a new copy
	}
```

it should be cheaper to both create, but also use later on?

rabbbit (commented):
My internal comment was meant as "we should report this before making internal changes", so you went further than I thought we would here - thanks!

It does look surprisingly clean, I thought it'd be more painful.

zapcore/deferred_with.go
jquirke: Yeah I like it, its even more performant on the Benchmark5WithsDeferredUsed, I just need to think through it a bit more carefully, is it correct in all cases?

```
goos: darwin
goarch: arm64
pkg: go.uber.org/zap
Benchmark5WithsUsed-10               	  496836	      2543 ns/op
Benchmark5WithsNotUsed-10            	  570384	      2251 ns/op
Benchmark5WithsDeferredUsed-10       	  770275	      1456 ns/op
Benchmark5WithsDeferredNotUsed-10    	 2621470	       464.9 ns/op
```

jquirke: Seems to be, added another test TestLoggerChainedWith

abhinav (commented):
The With state tracking isn't right. 

---

Separately from that, you can probably drop the synchronization and the overhead of the atomic sync.Once if you do something like @mway's context core.

As a reminder, the contract for writing with Cores is:

```
if ce := core.Check(e, ce); ce != nil { ce.Write(fields) }
```

Where `core.Check` is responsible for adding the core to the CheckedEntry with CheckedEntry.AddCore, and `ce.Write` eventually calls `Core.Write` on all added cores.

@mway's context core takes advantage of the fact that `Core.Write` will never be called directly, only after being added to the CheckedEntry with AddCore. It does so by splitting the core into two: the bit that just tracks `With`s and everything else, and the writer bit that is used with Check and Write.

So the way that works out is something like this:

```go
type myCore struct{ /* some state */ }

func (*myCore) With(fields []Field) Core { return &myCore{...} }

func (c *myCore) Check(e Entry, ce *CheckedEntry) *CheckedEntry {
  ce.AddCore(&myCoreWriter{c: c.wrappedCore, /* stuff */})
  // This is the important bit.
  // The core added to the CheckedEntry
  // is a completely different object that has the relevant state
  // computed.
  // You can even pool relevant things here,
  // and release them in myCoreWriter.Write.
}

type myCoreWriter struct{ /* state */ }

func (*myCoreWriter) Write(fields []Field) {
  // ...
}
```

The Write method on myCore panics because that's invalid usage that breaks the contract. The non-Write methods on myCoreWriter delegate or panic as necessary.

abhinav:
Also, API-wise, I'm happy to be convinced otherwise, but my gut says this shouldn't be an Option.
I'll welcome input from others here as well, but I don't think this is a global logger-level choice.
I would guess this is a per-`With` call choice?
Because whether those fields can be deferred or not, what their lifetimes are, that's known separately at each call site.
Would a `logger.WithDeferred` (or something similarly named) that initializes the deferred core be more appropriate?

rabbbit:
> Because whether those fields can be deferred or not, what their lifetimes are, that's known separately at each call site.

Yeah, this was a bit unclear. And you can implicitly end up with some loggers behaving differently than others.

On the other hand, having it per-logger means you can write `logger.With(String("a", "b").With(String("c", "d")` and the optimization still holds. Having explicit `WithDeferred` on each call site will likely be hard to remember, and then unused.

We were wondering - what's the exact reason for not deferring always? Would that be a very strongly impossible option?

abhinav:
> We were wondering - what's the exact reason for not deferring always? Would that be a very strongly impossible option?

The contract right of `With` right now is that the value inside Field is evaluated "right now" and then it's yours to do whatever with -- modify state, etc. So if you have a value in the `Interface` field that is modified after the `With`, it'll behave differently.

For example, here's a convoluted test that passes today:

```go
func Example_modifyAfterWith() {
	logger := zap.NewExample()
	defer logger.Sync()

	var x int
	arr := zapcore.ArrayMarshalerFunc(func(enc zapcore.ArrayEncoder) error {
		enc.AppendInt(x)
		return nil
	})

	logger = logger.With(zap.Array("a", arr))
	x = 1
	logger = logger.With(zap.Array("b", arr))
	x = 2
	logger.Info("hello", zap.Array("c", arr))

	// Output:
	// {"level":"info","msg":"hello","a":[0],"b":[1],"c":[2]}
}
```

If you always defer Withs (by adding the option to the logger), you'll get:

```
{"level":"info","msg":"hello","a":[2],"b":[2],"c":[2]}
```

zapcore/deferred_with.go
```diff
@@ -0,0 +1,82 @@
...
+	allFields := make ([]Field, 0, len(fields) + len(d.fields))
+	allFields = append(allFields, d.fields...)
+	allFields = append(allFields, fields...)
+	d.fields = allFields
+	return d
```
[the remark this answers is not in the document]
jquirke: interestingly its not breaking, which surprised me (see the cross talk test)

jquirke:
> 

Do you think then that perhaps we express in documentation and tests that this difference in behaviour will result and the consumer can then decide whether it is the right approach for them?

jquirke:
> On the other hand, having it per-logger means you can write logger.With(String("a", "b").With(String("c", "d") and the optimization still holds. Having explicit WithDeferred on each call site will likely be hard to remember, and then unused

I mean the inspiration for this was for 99% cases, this greatly reduced heap allocations across the application, I tested this simply by doing

```
fx.Decorate(func(l *zap.Logger) *zap.Logger {
   return l.WithOptions(zap.DeferWith())
}),
```

At the same time however, a global change to the top level logger may have unexpected behaviour as you mention with the way that Withs are evaluated. However, the most common case of this is an internal zapfx.Context(...) use case at Uber which is very expensive to pre-evaluate

jquirke:
> Separately from that, you can probably drop the synchronization and the overhead of the atomic sync.Once if you do something like @mway's context core.
> 
> 
> 
> As a reminder, the contract for writing with Cores is:
> 
> 
> 
> ```
> 
> if ce := core.Check(e, ce); ce != nil { ce.Write(fields) }
> 
> ```
> 
> 
> 
> Where `core.Check` is responsible for adding the core to the CheckedEntry with CheckedEntry.AddCore, and `ce.Write` eventually calls `Core.Write` on all added cores.
> 
> 
> 
> @mway's context core takes advantage of the fact that `Core.Write` will never be called directly, only after being added to the CheckedEntry with AddCore. It does so by splitting the core into two: the bit that just tracks `With`s and everything else, and the writer bit that is used with Check and Write.
> 

While I appreciate the cleverness of this solution, it seems very difficult to understand for maintainers. 

We discussed internally, once.Do() is relatively low over head in the non contended initialization case (it's a read acquire operation).

In the contended case where say, the deferred with logger is passed to a bunch of go routines that do perform operations on the logger I can see that would result in stalled execution though.

abhinav:
> Do you think then that perhaps we express in documentation and tests that this difference in behaviour will result and the consumer can then decide whether it is the right approach for them?

Yeah, the difference in behavior will need to be expressed in documentation.
I was answering @rabbbit's question of why we don't do this by default.

The other point I was bringing up was about whether this is a logger-level configuration or a `With` callsite-level switch (with a different method). My initial impression was that it made sense as a per-callsite change, but there's also a valid argument to make it logger-level. I'm open to either now. I'd love to hear from other folks as well.

However, if it's to be a logger-level option, there needs to be a way to disable it after you've enabled it for a logger. To accommodate that, the `DeferWith` option should take a boolean argument to turn the switch off.

```go
func DeferWith(enable bool) Option
```

This will allow:

```go
logger = logger.WithOptions(zap.DeferWith(true))
// ...
foo(logger.WithOptions(zap.DeferWith(false)))
```

As an example of where we failed to do this in the past: [AddCaller](https://pkg.go.dev/go.uber.org/zap#AddCaller) was added without a boolean switch, so we ended up adding [WithCaller](https://pkg.go.dev/go.uber.org/zap#WithCaller) to be able to turn that back.

logger_test.go
```diff
@@ -124,20 +124,65 @@ func TestLoggerInitialFields(t *testing.T) {
...
+		},
+		{
+			"deferred logger",
+			[]Option{Fields(Int("foo", 42)), DeferWith(true)},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			withLogger(t, DebugLevel, tt.opts, func(logger *Logger, logs *observer.ObservedLogs) {
+				// Child loggers should have copy-on-write semantics, so two children
+				// shouldn't stomp on each other's fields or affect the parent's fields.
```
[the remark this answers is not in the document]
jquirke: TestLoggerChainedWith covers that, but I'll merge that into this test.

zapcore/deferred_with.go
jquirke: I haven't given up.
I wish to write the tests that show the difference in behaviour first with the sync.Once core, and then port in the "mway" style pattern in zapfx context_core.go once I have confidence in the tests. Just been busy with day to day tasks :)

jquirke: I updated the tests - TestLoggerDeferredWithCapture

They are still using the "sync.Once" new core pattern, please check if the behaviour looks correct. Then I will try the mway pattern.

jquirke: Using your pattern, there is a subtle difference:

```
--- FAIL: TestLoggerDeferredWithCapture (0.00s)
    --- FAIL: TestLoggerDeferredWithCapture/deferred_logger_evaluates_at_time_of_logging,_and_stops_deferring_on_children (0.00s)
        logger_test.go:221:
            	Error Trace:	/Users/qjeremy/zap/logger_test.go:221
            	Error:      	Not equal:
            	            	expected: zap.res{A:[]int{2}, B:[]int{2}, C:[]int{2}, D:[]int{3}, E:[]int{4}}
            	            	actual  : zap.res{A:[]int{4}, B:[]int{4}, C:[]int{2}, D:[]int{4}, E:[]int{4}}

            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,6 +2,6 @@
            	            	  A: ([]int) (len=1) {
            	            	-  (int) 2
            	            	+  (int) 4
            	            	  },
            	            	  B: ([]int) (len=1) {
            	            	-  (int) 2
            	            	+  (int) 4
            	            	  },
            	            	@@ -11,3 +11,3 @@
            	            	  D: ([]int) (len=1) {
            	            	-  (int) 3
            	            	+  (int) 4
            	            	  },
            	Test:       	TestLoggerDeferredWithCapture/deferred_logger_evaluates_at_time_of_logging,_and_stops_deferring_on_children
```

logger_test.go
```diff
@@ -124,20 +125,104 @@ func TestLoggerInitialFields(t *testing.T) {
...
+				enc.AppendInt(x)
+				return nil
+			})
+
+			logger = logger.With(Array("a", arr))
+			x = 1
+			logger = logger.With(Array("b", arr))
+			x = 2
+			logger.Info("hello", Array("c", arr))
+			x = 3
```
jquirke, line 206: @rabbbit @abhinav do you agree this is the correct behaviour?

rabbbit: I think so, but the example seems painful. I don't know `zap` enough to figure out an easier one (@abhinav?). It does seem like "a" and "b" are always equal - could you simplify and remove if they're really unnecessarily duplicated?

And generally, I'd personally split the PR into two or even four:
- one that modifies `TestLoggerWith` - this seems non-controversial, perhaps adding benchmarks too.
- one that adds `TestLoggerDeferredCapture` to formalize the current behavior as a test
- one that adds the `WithDefer` behavior.

Apologies for being a pain.

prashantv:
> Also, API-wise, I'm happy to be convinced otherwise, but my gut says this shouldn't be an Option. I'll welcome input from others here as well, but I don't think this is a global logger-level choice. I would guess this is a per-`With` call choice? Because whether those fields can be deferred or not, what their lifetimes are, that's known separately at each call site. Would a `logger.WithDeferred` (or something similarly named) that initializes the deferred core be more appropriate?

+1, I don't think this should be a logger-level option either -- a library using a `*zap.Logger` may assume that the `With` call only uses the passed-in reference for the duration of the call, but this isn't clear in the typesystem or to callers, so passing a logger that defers `With` will incorrectly hold the reference for longer than expected. The most obvious problem is for mutable objects like `Binary`.

We may want to be more explicit about when fields passed-in to `With` will be used, but I think the current expectation is that they're used at the time of the call.

jquirke:
> > Also, API-wise, I'm happy to be convinced otherwise, but my gut says this shouldn't be an Option. I'll welcome input from others here as well, but I don't think this is a global logger-level choice. I would guess this is a per-`With` call choice? Because whether those fields can be deferred or not, what their lifetimes are, that's known separately at each call site. Would a `logger.WithDeferred` (or something similarly named) that initializes the deferred core be more appropriate?
> 
> +1, I don't think this should be a logger-level option either -- a library using a `*zap.Logger` may assume that the `With` call only uses the passed-in reference for the duration of the call, but this isn't clear in the typesystem or to callers, so passing a logger that defers `With` will incorrectly hold the reference for longer than expected. The most obvious problem is for mutable objects like `Binary`.
> 
> We may want to be more explicit about when fields passed-in to `With` will be used, but I think the current expectation is that they're used at the time of the call.

logger_test.go
jquirke: 1a/ https://github.com/uber-go/zap/pull/1325 (benchmarks)
2/ https://github.com/uber-go/zap/pull/1326 (deferred capture - cleaner)

rabbbit:
> > Also, API-wise, I'm happy to be convinced otherwise, but my gut says this shouldn't be an Option. I'll welcome input from others here as well, but I don't think this is a global logger-level choice. I would guess this is a per-`With` call choice? Because whether those fields can be deferred or not, what their lifetimes are, that's known separately at each call site. Would a `logger.WithDeferred` (or something similarly named) that initializes the deferred core be more appropriate?
> 
> +1, I don't think this should be a logger-level option either -- a library using a `*zap.Logger` may assume that the `With` call only uses the passed-in reference for the duration of the call, but this isn't clear in the typesystem or to callers, so passing a logger that defers `With` will incorrectly hold the reference for longer than expected. The most obvious problem is for mutable objects like `Binary`.
> 
> We may want to be more explicit about when fields passed-in to `With` will be used, but I think the current expectation is that they're used at the time of the call.

I see the breaking change, but the `DeferWith` is significantly less ergonomic. You have to remember to use it throughout the new code, but also, the old code will not improve - I see today 550 cases of `logger.With.().With.()` in our monorepo, I had hoped they'd magically become less expensive. (it's 550 cases of two+ With being on the same line). Arguably this could be a linter check though.

The proposed `fx` based linter override at that application startup would make things much "easier to use". A service owner could opt-in themselves and "own the change" fully.

>  a library using a *zap.Logger may assume that the With call only uses the passed-in reference for the duration of the call

Theoretically, a library could disable deferring at the entry point ... but yeah, this would be a breaking change.

tldr; sadness.

jquirke:
> tldr; sadness.

Nevertheless, I implemented it in exp/ package for future reference.

abhinav:
I don't want this PR to grow inactive because it does solve a legitimate problem I've seen before, so in the interest of actually getting this feature in, I'd like to discuss what we need here.

My opinion is: If we're already open to naming this WithLazy (which is a name I like more than WithDefer), it can be in zap.Logger proper instead of a wrapper. @jquirke if you have a preference for experimenting with it in exp first, we can do the wrapper you're suggesting instead.

@rabbbit, your opinion that this is called `With` -- does that change with the knowledge that this isn't trying to address `logger.With(..).With(..)`? Also note that if it's `logger.With(..).With(..).Info(..)`, this will not speed those up; in fact it'll slow them. 

@prashantv, @mway, @sywhang Do you have any thoughts on this?

mway:
Some thoughts:

1. This is a nice feature, looking forward to having this capability.
2. Agree with the general take that this shouldn't be an `Option`, and that this is a "per-sublogger-field-evaluation" level concern rather than a logger-level concern. For me, this became a lot easier to rationalize when thinking of it as "a detail of sublogger creation" vs "a property of the resulting logger".
3. I think `WithLazy` is preferable to the other formulations. While the net effect (from an entry encoding perspective) is the same, it's semantically different than `With` from the caller's perspective.
4. I don't think `exp/` and/or a wrapper are necessary here unless (1) we're really unsure about the API or (2) the viability of the feature in general. Put another way: IMO this feels "natural", and doesn't particularly seem like something we'll want to walk back later.
5. Regarding transparently optimizing `l.With().With()` cases: I think this is an orthogonal concern. We have other options for those (like linting as mentioned above, or potentially `gopatch`ing). More generally, spurious/ineffectual `With()` calls are discouraged, and so aren't something we should feel especially compelled to optimize for.
6. Regarding breaking changes: there have been a lot of features/changes in the past that were not implemented (or were implemented differently) to avoid breaking changes. We're very change-averse for ad hoc API changes, and IMO that general ethos should continue. Relatedly, we should start curating a list of any "desirable API changes" to incorporate as a group of changes (features, ergonomics, or otherwise) to be released together if/when we cut v2, rather than releasing them individually.

jquirke:
> Some thoughts:
> 
> 1. This is a nice feature, looking forward to having this capability.
> 2. Agree with the general take that this shouldn't be an `Option`, and that this is a "per-sublogger field-evaluation-time" level concern rather than a logger-level concern. For me, this became a lot easier to rationalize when thinking of it as "a detail of sublogger creation" vs "a property of the resulting logger".
> 3. I think `WithLazy` is preferable to the other formulations. While the net effect (from an entry encoding perspective) is the same, it's semantically different than `With` from the caller's perspective.
> 4. I don't think `exp/` and/or a wrapper are necessary here unless (1) we're really unsure about the API or (2) the viability of the feature in general. Put another way: IMO this feels "natural", and doesn't particularly seem like something we'll want to walk back later.
> 5. Regarding transparently optimizing `l.With().With()` cases: I think this is an orthogonal concern. We have other options for those (like linting as mentioned above, or potentially `gopatch`ing). More generally, spurious/ineffectual `With()` calls are discouraged (and can be considered misuse), and so aren't something we should feel especially compelled to optimize for.
> 6. Regarding breaking changes: there have been a lot of features/changes in the past that were not implemented (or were implemented differently) to avoid breaking changes. We're very change-averse for ad hoc API changes, and IMO that general ethos should continue. Relatedly, we should start curating a list of any "desirable API changes" to incorporate as a group of changes (features, ergonomics, or otherwise) to be released together if/when we cut v2, rather than releasing them individually.

Taken both comments by @abhinav and @mway - I've already tested this on a non prod branch and seen significant improvement in GC time by replacing With with WithLazy(..). I created the /exp/ branch for future reference, but I agree, after this discussion this seems a worthwhile breaking API change to make. 

Therefore, I've updated the commit to move it out of the /exp/ package.

jquirke:
I've been thinking, if we are going to introduce breaking changes (new functions) then perhaps WithLazy is not the right answer 

Perhaps we generalize this as such 

Instead of logger.WithLazy(fields...)

We instead create some method With Opts, and the usage therefore is something like 

logger.WithOpts(Lazy, Fields(fields))

Thoughts?

jquirke:
Bump, whats the verdict on this. I've provided and implemented a menu of choices and ther

1/ Logger level option: Not a breaking change, but we all agreed its wrong. 
2/ /exp/ Wrapper for immediate use
3/ Breaking change (adding zap/Logger.WithLazy)
4/ With level option e.g. logger.WithOption(LazyFields(....))). Not breaking, but needs to be supported forever.
5/ Internal fork of our own OSS project, but this seems absurd

sywhang:
Sorry I'm late to the game here. Looked through the different options @jquirke put up, and IMO the best option here is https://github.com/uber-go/zap/commit/150586869edd5ae9b62ab57237c6d4f1df0788bd (adding WithLazy method to Logger). As I chatted offline with Jeremy, this level of change is something we've done in the past and is compatible enough for us to accept. 

As for putting in exp/ i don't see much benefit added from doing so, since the API seems pretty solid. For other things we put in exp/ it was either because it depended on experimental packages itself (i.e. slog) or only for a temporary solution to bake things in a bit. 

Thoughts? @abhinav @mway @prashantv

jquirke:
> Sorry I'm late to the game here. Looked through the different options @jquirke put up, and IMO the best option here is [1505868](https://github.com/uber-go/zap/commit/150586869edd5ae9b62ab57237c6d4f1df0788bd) (adding WithLazy method to Logger). As I chatted offline with Jeremy, this level of change is something we've done in the past and is compatible enough for us to accept.

Cool. Switched to the WithLazy implementation without exp (2)

abhinav:
> [..] IMO the best option here is [1505868](https://github.com/uber-go/zap/commit/150586869edd5ae9b62ab57237c6d4f1df0788bd) (adding WithLazy method to Logger). [..]
> 
> Thoughts? @abhinav @mway @prashantv

Agreed.
I haven't had a chance to review the change again since the original couple iterations, but yeah, I'm on-board with a WithLazy method.

logger_test.go
```diff
@@ -124,58 +124,121 @@ func TestLoggerInitialFields(t *testing.T) {
...
-			{Context: []Field{Int("foo", 42), String("three", "four")}},
-			{Context: []Field{Int("foo", 42), String("five", "six"), String("seven", "eight")}},
-			{Context: []Field{Int("foo", 42)}},
-		}, logs.AllUntimed(), "Unexpected cross-talk between child loggers.")
-	})
+	for _, tt := range []struct {
+		name           string
+		initialFields  []Field
+		withMethodExpr func(*Logger, ...Field) *Logger
+	}{
```
abhinav: nit: We avoid `for _, tt := range []struct { ... }`; instead we place the table in a separate variable. (I don't know if the entire codebase matches that.) It makes for slightly better readability.

abhinav (commented):
I'm a bit busy this week, so I can't give it a full review. I've pushed a couple minor tweaks.

I don't know if the tests already cover this but I want to make sure at least the following are handled, specifically with regards to evaluation of fields.

- `WithLazy.With`
- `WithLazy.WithLazy`

jquirke:
> * `WithLazy.With`
> * `WithLazy.WithLazy`

Added test for With.WithLazy, WithLazy.With, WithLazy.WithLazy
Also sorry I force pushed, overwriting your doco cleanup. If you have the commit locally do you mind putting it back, apologies.

abhinav:
> Also sorry I force pushed, overwriting your doco cleanup. If you have the commit locally do you mind putting it back, apologies.

Not a problem! Pushed it back.

sywhang (approved):

sywhang:
The table tests hurts the readability of the tests a little bit, but we can address these later. 

Thanks!

PR #1318, opened 2023-08-09
Update to work with released log/slog

shreyner:

PR #1320, opened 2023-08-14
Update to work with released log/slog

jcmuller:
This is based on #1318


exp/go.mod
```diff
@@ -1,11 +1,10 @@
 module go.uber.org/zap/exp
 
-go 1.19
+go 1.21
```
abhinav, line 3: Increasing the minimum required Go version isn't necessary if we're using `//go:build` directives to support exp/'s slog package for older versions.

exp/zapslog/slog_old.go
```diff
@@ -0,0 +1,190 @@
+//go:build go1 && !go1.21
```
abhinav, line 1: Typically, for Go-version-specific files, we've named them `foo_pre_go$version` and `foo_go$version`.
For example, we have https://github.com/uber-go/zap/blob/fcfd368624a286486e62d46deba7e42c20164616/array_go118.go.
(There's no pre-Go version for that.)

So this should probably be called slog_pre_go121.go.

exp/zapslog/slog.go
```diff
@@ -1,3 +1,5 @@
+//go:build go1.21
```
abhinav, line 1: And this should be called slog_go121.go.

abhinav (commented):
Thanks for this change! I've left some suggestions.
Would you mind also updating the zapslog/doc.go (or creating one if it doesn't exist)
to indicate that it'll use golang.org/x/exp/slog for versions of Go before 1.21, and log/slog for versions after.

abhinav:
PR updated:

- Merged master in so tests run against 1.20 and 1.21
- Made the renames I suggested above
- Lowered the `go` directive to Go 1.20
- Updated doc.go as suggested above

abhinav (commented):
A couple comments probably out of scope for this PR,
but should be addressed before we consider upgrading this to the main Zap package.

jcmuller:
Thanks, @abhinav. Just added two commits fixing test and lint running.

abhinav (approved):

abhinav:
@sywhang @mway please merge when you have a chance. I can't do anything about the coverage error. I expect we should fix that when we move this to the non-exp Zap module.

mway:
> @sywhang @mway please merge when you have a chance. I can't do anything about the coverage error. I expect we should fix that when we move this to the non-exp Zap module.

Ack, no worries.

PR #1325, opened 2023-08-18
Clarify argument capturing behaviour of With

jquirke:
Make it clear that the contract with the With and logging methods is such that any arguments that could change are captured at the time of the With and logging method, and that any other behaviour introduced in future is intentionally contrary to this contract.

There are proposed (https://github.com/uber-go/zap/pull/1319) future opt-in, changes that will explicitly relax this contract.


abhinav (approved):
Great! I tweaked the test a bit: assert.JSONEq plus an encoder that doesn't include time so that we don't have to json.Unmarshal.

Also, the test case was a bit off: `b: [1]` will not be present in the second entry
because `Array("b", ..)` was only used with the `logger.Info`, not `logger.With`.

rabbbit:
I'm late, but thanks for doing this!

PR #1326, opened 2023-08-18
Add benchmarks for chained Withs

jquirke:
Add benchmarks to show the performance of chaining child loggers via 'With' and whether the logger is used or not used.

This is intended to establish a baseline for a proposed optimisation in PR https://github.com/uber-go/zap/pull/1319.

### DIFF
--- a/logger_bench_test.go
+++ b/logger_bench_test.go
@@ -23,6 +23,7 @@ package zap
 import (
 	"errors"
 	"runtime"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


logger_bench_test.go
```diff
@@ -198,6 +198,26 @@ func BenchmarkAddCallerAndStacktrace(b *testing.B) {
...
+	withBenchedLogger(b, func(log *Logger) {
+		make5ChainedChildLoggers(log).Info("used")
+	})
+}
+
+func Benchmark5WithsNotUsed(b *testing.B) {
+	withBenchedLogger(b, func(log *Logger) {
+		l := make5ChainedChildLoggers(log)
+		runtime.KeepAlive(l)
+	})
```
abhinav, line 211: Instead of hard-coding the 5, how about something like this?

```go
func benchmarkWithUsed(b *testing.B, N int, use bool) {
  keys := make([]string, N)
  values := make([]string, N)
  for i := 0; i < N; i++ {
    keys[i] = "k" + strconv.Itoa(i)
    values[i] = "v" + strconv.Itoa(i)
  }

  b.ResetTimers()

  withBenchedLogger(b, func(log *Logger) {
    for i := 0; i < N; i++ {
      log = log.With(String(keys[i], values[i]))
    }
    if use {
      log.Info("used")
    } else {
      runtime.KeepAlive(log)
    }
  })
}
```

abhinav (commented):
In favor of this change. Just leaving a minor suggestion for if/when we want to expand this benchmark.

abhinav (approved):

rabbbit:
I'm late, but thanks for doing this!

PR #1338, opened 2023-08-23
zapslog: Delete x/exp-based implementation

### LINKED ISSUE #1337: zapslog: Delete exp-based implementation
We should delete the zapslog implementation based on golang.org/x/exp since log/slog is in the standard library now. It's not worth the effort to maintain two parallel implementations just for pre-1.20 support of an experimental module.

abhinav:
Deletes the implementation of slog.Handler based on
golang.org/x/exp/slog.
Maintaining two implementations doesn't serve much value here.

Remaining files have been renamed to better match their contents:

- handler.go holds the slog.Handler
- handler_test.go tests that
- example_test.go contains example tests

Files that remain are still tagged with `go1.21`
so they're only visible to Go 1.21 or newer.

This way, `go test` in this directory with Go 1.20 will not fail:

```
% GOTOOLCHAIN=go1.20 go test -v
?       go.uber.org/zap/exp/zapslog     [no test files]
```

Resolves #1337


mway (approved):
LGTM

Issue #1329, opened 2023-08-19
zapslog.Handler: Add stack traces


zekth:
First i'm really happy to see the fast work on making zap compliant on go 1.21 with `slog` !

**Is your feature request related to a problem? Please describe.**

This describes just the miss of `stack trace` serialization via the slog handler.

**Describe the solution you'd like**

Taking this example:
```go
package main

import (
	"errors"
	"log/slog"
	"os"

	"go.uber.org/zap"
	"go.uber.org/zap/exp/zapslog"
	"go.uber.org/zap/zapcore"
)

func main() {
	config := zapcore.EncoderConfig{
		MessageKey:    "msg",
		LevelKey:      "level",
		EncodeLevel:   zapcore.CapitalLevelEncoder,
		TimeKey:       "ts",
		EncodeTime:    zapcore.ISO8601TimeEncoder,
		CallerKey:     "caller",
		EncodeCaller:  zapcore.ShortCallerEncoder,
		StacktraceKey: "stack_trace",
	}
	core := zapcore.NewCore(
		zapcore.NewJSONEncoder(config),
		zapcore.Lock(os.Stdout),
		zap.DebugLevel,
	)
	zapInstance := zap.New(core,
		zap.AddCaller(),
		zap.AddStacktrace(zapcore.ErrorLevel),
	)
	slogInstance := slog.New(zapslog.NewHandler(zapInstance.Core(), &zapslog.HandlerOptions{AddSource: true}))
	err := errors.New("failure")
	zapInstance.Error("zap", zap.Error(err))
	slogInstance.Error("slog", slog.Any("error", err))
}

```
It outputs:
```console
{"level":"ERROR","ts":"2023-08-19T12:10:36.636+0200","caller":"go/main.go:35","msg":"zap","error":"failure","stack_trace":"main.main\n\t/Users/vincent/kong/go/main.go:35\nruntime.main\n\t/usr/local/Cellar/go/1.21.0/libexec/src/runtime/proc.go:267"}
{"level":"ERROR","ts":"2023-08-19T12:10:36.636+0200","caller":"go/main.go:36","msg":"slog","error":"failure"}
```
I'd hope to find a solution to be able to have the `stack_trace` also via the `slog` handler.

**Describe alternatives you've considered**

[...]

PR #1339, opened 2023-08-23
feat(zapslog): implement stack trace handling

zekth:
fix: https://github.com/uber-go/zap/issues/1329

As discussed in the issue, replacing the `HandlersOptions` with functional options, i also renamed `addSource` to `addCaller` to properly match the `zap` semantic.

### DIFF
--- a/exp/zapslog/handler.go
+++ b/exp/zapslog/handler.go
@@ -120,9 +120,8 @@ func (h *Handler) Enabled(ctx context.Context, level slog.Level) bool {
 
 // Handle handles the Record.
 func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
-	zapLevel := convertSlogLevel(record.Level)
 	ent := zapcore.Entry{
-		Level:      zapLevel,
+		Level:      convertSlogLevel(record.Level),
 		Time:       record.Time,
 		Message:    record.Message,
 		LoggerName: h.name,
--- a/exp/zapslog/handler.go
+++ b/exp/zapslog/handler.go
@@ -37,7 +37,7 @@ type Handler struct {
 	core       zapcore.Core
 	name       string // logger name
 	addCaller  bool
-	addStack   zapcore.LevelEnabler
+	addStackAt slog.Level
 	callerSkip int
 }
 
@@ -45,8 +45,8 @@ type Handler struct {
 // with options.
 func NewHandler(core zapcore.Core, opts ...Option) *Handler {
 	h := &Handler{
-		core:     core,
-		addStack: zapcore.ErrorLevel,
+		core:       core,
+		addStackAt: slog.LevelError,
 	}
 	for _, v := range opts {
 		v.apply(h)
@@ -145,7 +145,7 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
 		}
 	}
 
-	if h.addStack.Enabled(zapLevel) {
+	if record.Level >= h.addStackAt {
 		// Skipping 3:
 		// zapslog/handler log/slog.(*Logger).log
 		// slog/logger log/slog.(*Logger).log
--- a/exp/zapslog/handler.go
+++ b/exp/zapslog/handler.go
@@ -33,36 +33,24 @@ import (
 
 // Handler implements the slog.Handler by writing to a zap Core.
 type Handler struct {
-	core      zapcore.Core
-	name      string // logger name
-	addSource bool
-}
-
-// HandlerOptions are options for a Zap-based [slog.Handler].
-type HandlerOptions struct {
-	// LoggerName is used for log entries received from slog.
-	//
-	// Defaults to empty.
-	LoggerName string
-
-	// AddSource configures the handler to annotate each message with the filename,
-	// line number, and function name.
-	// AddSource is false by default to skip the cost of computing
-	// this information.
-	AddSource bool
+	core       zapcore.Core
+	name       string // logger name
+	addCaller  bool
+	addStack   zapcore.LevelEnabler
+	callerSkip int
 }
 
 // NewHandler builds a [Handler] that writes to the supplied [zapcore.Core]
-// with the default options.
-func NewHandler(core zapcore.Core, opts *HandlerOptions) *Handler {
-	if opts == nil {
-		opts = &HandlerOptions{}
+// with options.
+func NewHandler(core zapcore.Core, opts ...Option) *Handler {
+	h := &Handler{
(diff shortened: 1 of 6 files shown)


stacktrace.go
```diff
@@ -125,6 +125,12 @@ func (st *stacktrace) Next() (_ runtime.Frame, more bool) {
 	return st.frames.Next()
 }
 
+// TakeStacktrace returns the stack trace,
+// and skips the provided number of frames.
+func TakeStacktrace(skip int) string {
```
zekth, line 130: Should we `deprecate` this and add a comment that it's just a temporary export?

exp/zapslog/slog_pre_go121.go
```diff
@@ -131,20 +119,19 @@ func (h *Handler) Enabled(ctx context.Context, level slog.Level) bool {
...
 // Handle handles the Record.
 func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
+	zapLevel := convertSlogLevel(record.Level)
 	ent := zapcore.Entry{
-		Level:      convertSlogLevel(record.Level),
+		Level:      zapLevel,
 		Time:       record.Time,
 		Message:    record.Message,
 		LoggerName: h.name,
-		// TODO: do we need to set the following fields?
```
zekth, line 139: `Stack` was not required for the implementation

stacktrace.go
abhinav: I don't like the idea of exporting this functionality from the main Zap package even as a temporary unsupported thing.
Instead, I've moved the relevant bits into an internal package in https://github.com/uber-go/zap/pull/1341.

That makes it easy to use them here.

abhinav (commented):
Thanks for doing this, and thanks for waiting for the review. It's a bit busy these days.

I'll try to get https://github.com/uber-go/zap/pull/1341 landed soon to capture the stack traces without exporting them from the main Zap package.

And yeah, exp/ will need a `replace` directive. We'll have to be careful about updating exp/ to the latest Zap release when we tag it.

abhinav:
I've pushed some minor fixups to your branch. Please be sure not to lose them if you `git push -f`.

stacktrace.go
zekth: Thanks for this i agree on this approach too

exp/zapslog/slog_go121.go
```diff
@@ -157,6 +144,10 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
 		}
 	}
 
+	if h.addStack.Enabled(zapLevel) {
+		ce.Stack = zap.TakeStacktrace(4 + h.callerSkip)
```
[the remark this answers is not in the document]
zekth: Sure I'll add revelant comments

zekth:
@abhinav I rebased and applied all the comments

zekth:
@abhinav should i add tests for the options or we add them in another PR? I'm happy to write a follow up one if needed

exp/zapslog/handler.go
```diff
@@ -157,6 +145,14 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
 		}
 	}
 
+	if h.addStack.Enabled(zapLevel) {
+		// Skipping 3:
+		// zapslog/handler log/slog.(*Logger).log
+		// slog/logger log/slog.(*Logger).log
+		// slog/logger log/slog.(*Logger).<level>
+		ce.Stack = stacktrace.Take(3 + h.callerSkip)
```
abhinav, line 153: Thanks for adding this. Much more obvious what's going on now.

exp/zapslog/options.go
```diff
@@ -0,0 +1,71 @@
...
+	return optionFunc(func(handler *Handler) {
+		handler.addCaller = enabled
+	})
+}
+
+// AddCallerSkip increases the number of callers skipped by caller annotation
+// (as enabled by the AddCaller option). When building wrappers around the
+// Logger and SugaredLogger, supplying this Option prevents zap from always
+// reporting the wrapper code as the caller.
+func AddCallerSkip(skip int) Option {
```
abhinav, line 59: Minor outdated comment here. This is no Logger/SugaredLogger-based.

Plus this should probably be named WithCallerSkip to match WithCaller above.

exp/zapslog/options.go
```diff
@@ -0,0 +1,71 @@
...
+	return optionFunc(func(log *Handler) {
+		log.callerSkip += skip
+	})
+}
+
+// AddStacktrace configures the Logger to record a stack trace for all messages at
+// or above a given level.
+func AddStacktrace(lvl zapcore.LevelEnabler) Option {
```
abhinav, line 67: Just thinking out loud (would love your opinion on this, @zekth):
I wonder if this should be a slog Level instead. Or a new zapslog.LevelEnabler interface based on slog.Level.

```go
type LevelEnabler interface{ Enabled(context.Context, Level) bool }
```

The reasoning is this: the result of zapslog.NewHandler is a slog Handler. It talks in the language of slog.

abhinav (commented):
Hey @zekth, apologies; I haven't forgotten about this, but I've had (and continue to have) a busy few days. I've taken a pass; this mostly looks good to go with minor comments. I'll address a few myself in a few minutes.

Tagging other Zap maintainers as well. CC @sywhang @mway @prashantv 
(Context: This PR turns HandlerOptions into functional options per prior discussion, and adds support for stack traces. Stack traces are captured for Error and above -- same as Zap's production logger.)

RE: tests: There's no need to backfill tests for existing options at this time--tests for the new option you're adding for stack traces should be sufficient. We do intend to fill in the rest of the tests before moving this to the main Zap package, though, but that's out of scope for this PR.

abhinav:
On LevelEnabler and whether this should use a Zap level or slog level.
The more I think about it, the slog level makes sense.
The slog level has more information; conversion from slog to Zap level is a lossy operation.
I've renamed the option to AddStacktraceAt.
That way, if we want to add a LevelEnabler in the future, we can still use an `AddStacktrace(LevelEnabler)` option.
(The proposed LevelEnabler's Enabled signature matches slog.Handler.Enabled, so such a user-provided function
would have all the important information about the record to determine whether it should capture a stacktrace.)

abhinav (approved):
LGTM!

@mway @sywhang would appreciate a second look.
FYI, the coverage drop is because some of the existing functionality is untested; we will want to fix that before considering this stable anyway.

sywhang (approved):
LGTM.

PR #1340, opened 2023-08-23
Add LazyLogger experimental wrapper

jquirke:
This is a preview of a breaking API change proposed in https://github.com/uber-go/zap/pull/1319. It is included as an /exp/ wrapper for those interested in trying it out in their particular use cases as it does introduce a performance optimisation.

One common performance problem we see is that Loggers and their children may be composed  with one or more With() invocations, but the logger/s is/are not used in most invocations of the function.

For example, an author might initialise a logger with some function-scoped contexttual information at the start of a function:

```
l := logger.With(zap.String("tripID", trip.ID))
```

but never use 'l' except in seldom taken branches. However, in the majority cases the expensive with operation has been performed which can result in the wasted effort (in terms of alloc, CPU) of cloning (e.g of the JSON encoder).

This commit adds a new method on Logger, WithLazy, which defers all of this expense until the time of first use. This creates considerable performance improvement  for the cases where a logger is not regularly used.

This does require an explicit opt in, because the author must be aware there is the possibility of objects living longer than the scope of the method wherein WithLazy is called.

```
 % go test -bench=Benchmark5With
goos: darwin
goarch: arm64
pkg: go.uber.org/zap/exp/lazylogger
Benchmark5WithsUsed-10           	  527385	      2390 ns/op
Benchmark5WithsNotUsed-10        	  563468	      2246 ns/op
Benchmark5WithLazysUsed-10       	  475064	      2642 ns/op
Benchmark5WithLazysNotUsed-10    	 2255786	       527.4 ns/op
PASS
```

Furthermore, one core microservice in Uber has seen a reduction of 10% heap allocation by simply using this option.

### DIFF
--- a/exp/go.mod
+++ b/exp/go.mod
@@ -9,9 +9,11 @@ require (
 )
 
 require (
+	github.com/benbjohnson/clock v1.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	go.uber.org/atomic v1.10.0 // indirect
 	go.uber.org/multierr v1.10.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace go.uber.org/zap v1.24.0 => ../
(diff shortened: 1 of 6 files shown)


jquirke:
Won't do: will go ahead with #1319

PR #1341, opened 2023-08-25
chore: Move stacktrace capturing into internal/

abhinav:
Moves the functionality to capture and format stack traces
into an internal stacktrace package
and exports the relevant bits out for the logger and field.go to use.

This will be used by zapslog.Handler to capture stack traces
so it needs to be in a shared location.

Refs #1329, #1339

### DIFF
--- a/field.go
+++ b/field.go
@@ -25,6 +25,7 @@ import (
 	"math"
 	"time"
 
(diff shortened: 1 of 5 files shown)


mway (approved):
LGTM

PR #1342, opened 2023-08-28
zap: always return a new child logger from `*Logger.With`

liuzengh:
even if fields parameters is empty to make `*Logger.With`'s behavior consistent with `*Logger.WithOptions(Fields())`'s behavior.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -177,9 +177,6 @@ func (log *Logger) WithOptions(opts ...Option) *Logger {
 // to the child don't affect the parent, and vice versa. Any fields that
 // require evaluation (such as Objects) are evaluated upon invocation of With.
 func (log *Logger) With(fields ...Field) *Logger {
-	if len(fields) == 0 {
-		return log
-	}
 	l := log.clone()
 	l.core = l.core.With(fields)
 	return l
--- a/logger_test.go
+++ b/logger_test.go
@@ -124,21 +124,31 @@ func TestLoggerInitialFields(t *testing.T) {
 }
 
 func TestLoggerWith(t *testing.T) {
-	fieldOpts := opts(Fields(Int("foo", 42)))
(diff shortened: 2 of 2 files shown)


liuzengh:
This refactor PR #201  “altered the external behavior of `With` method.  
```golang
// https://github.com/uber-go/zap/pull/201/files#diff-ff87b7c4777a35588053a509583d66c9f404ccbea9e1c71d2a5f224d7ad1323eL75
func (log *logger) With(fields ...Field) Logger {
	clone := &logger{
		Meta: log.Meta.Clone(),
	}
	addFields(clone.Encoder, fields)
	return clone
}
```

And,  I would be happy to see the behavior of  `*Logger.With` and `*Logger.WithOptions(Fields())` is same.

rabbbit:
The PR you linked was from 2016 so likely the current state is not upsetting a lot of people.

Is this PR solving an actual issue you encountered? If so, could you elaborate on why you need the logger to be a new object?
Partially related, we were discussing this in #1340.

liuzengh:
I guess almost no one would use `With` with an empty 'fields',  and they don't care whether the two Loggers before and after are the same. So people haven't been upset about current state.

1. If people `With` an non-empty fields', it is not necessary to check whether the 'fields' is empty in the `With` method.  This PR remove the  empty 'fields' check logic, it looks good.

2. If people `With` with an empty 'fields'. Let's discuss two scenarios. In scenario 1, he actually doesn't want to deal with empty fields. His code implementation is flawed. Instead of using zap package to check for empty fields within the `With` method, he should handle the empty field checks within his business logic code.  In scenario 2, he intentionally uses empty fields in the `With` method to  get a new logger,  and performs some operations on it, such as setting the log level for the new logger( implementing the `SetLevel` method on the wrapped zap.Logger).

Maybe,  I should reword the CL "zap: always return a new child logger from *Logger.With"  to  "zap: Remove the  empty 'fields' check logic in *Logger.With" after I see https://github.com/uber-go/zap/pull/1340.

Issue #1334, opened 2023-08-23
zapslog.Handler: Test with slogtest, fix all failures


abhinav:
zapslog.Handler is not yet tested with slogtest on master.
We should add a test for that and address any problems that occur.

#1335 attempts to add such a test. Pending failures:

- [x] https://github.com/uber-go/zap/pull/1344
- [x] Handler should ignore a zero Record.Time
    - #1369
- [ ] Handler should inline Attrs of a group with an empty key
    - #1263
- [ ] Handler should not output groups for an empty Record

SoulPancake:
Can I help with this or is it done? @abhinav

abhinav:
Hey, @SoulPancake. Thanks for your interest!
This isn't done yet. I've created #1335 with the test, but I haven't had time to fix the failures (save #1344).
Looks like @mway made an attempt to fix one of the failures in (#1263) but hasn't had time to follow up.
If you have interest in helping with this, please do go ahead.

Thanks!

justinhwang:
Hey @abhinav, as far as "Handler should ignore a zero Record.Time" - isn't this up to each individual `Core` to determine? We could update the `ioCore` w/ `jsonEncoder` (i.e. have a zero check in `jsonEncoder.EncodeEntry`) to be compliant which would take care of the error from #1335 but `zapslog` wouldn't always be compliant no?

abhinav:
@justinhwang Yeah, that's a fair point. I think it makes sense that the official encoders that come with Zap support that behavior, and we should possibly consider documenting that encoder implementations are highly encouraged to skip writing the time for a log entry if it's a zero value.

arukiidou:
@abhinav 
- These GHA logs have expired, I made each test error a separate Issue
  - #1401
  - #1402

PR #1344, opened 2023-09-01
zapslog: Handler should ignore an empty Attr

abhinav:
Per the slog.Handler contract,
handlers should not log attributes that are equal to the zero value.
This is equivalent to Zap's `zap.Skip()`.

Discovered by #1335
Refs #1334


sywhang (approved):
👍

PR #1349, opened 2023-09-08
Replace benbjohnson/clock with custom MockClock

### LINKED ISSUE #1331: Use of github.com/benbjohnson/clock which is no longer maintained
**Describe the bug**
This project uses github.com/benbjohnson/clock which is no longer maintained

**To Reproduce**
N/A

**Expected behavior**
Use an alternative package (for example, github.com/jonboulle/clockwork) or bring this functionality into this repo

**Additional context**
While it seems to be a test-only dependency, this is still a potential issue and is being flagged by some SBOM checking tools

abhinav:
This drops the dependency on benbjohnson/clock as it's no longer maintained.
It replaces it with a hand-rolled mock clock implementation.

The core of the functionality of MockClock is provided by the following:

- waiter: a function waiting to be executed
- runAt: schedules a function to be executed when time progresses
- Add: moves time forward, running all functions in range

The rest of the necessary functionality is built on top of these.

This is a pretty simple implementation.
We schedule work, but nothing happens until Add is called.
There are no goroutines running additional work in the background.

Resolves #1331

### DIFF
--- a/internal/ztest/clock.go
+++ b/internal/ztest/clock.go
@@ -107,7 +107,12 @@ func (c *MockClock) Add(d time.Duration) {
 	// This ensures that any waiters that are resolved
 	// are resolved at the time they were expecting.
 
-	for w, ok := c.waiters.PopLTE(newTime); ok; w, ok = c.waiters.PopLTE(newTime) {
+	for {
+		w, ok := c.waiters.PopLTE(newTime)
+		if !ok {
+			break
+		}
+
 		// The waiter is within range.
 		// Travel to the time of the waiter and resolve it.
 		c.now = w.until
--- a/internal/ztest/clock.go
+++ b/internal/ztest/clock.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Uber Technologies, Inc.
+// Copyright (c) 2023 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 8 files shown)


internal/ztest/clock.go
```diff
@@ -21,30 +21,167 @@
...
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.now
 }
 
 // NewTicker returns a time.Ticker that ticks at the specified frequency.
+//
+// As with [time.NewTicker],
+// the ticker will drop ticks if the receiver is slow,
+// and the channel is never closed.
```
prashantv, line 64: is it worth mentioning that `Stop` won't work?

internal/ztest/clock.go
```diff
@@ -21,30 +21,167 @@
...
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	newTime := c.now.Add(d)
+	// newTime won't be recorded until the end of this method.
+	// This ensures that any waiters that are resolved
+	// are resolved at the time they were expecting.
+
+	for w, ok := c.waiters.PopLTE(newTime); ok; w, ok = c.waiters.PopLTE(newTime) {
```
prashantv, line 107: nit: rather than duplicating the call to `PopLTE` and making sure they're the same, it may be cleaner to move that logic into the body,
```
for {
  w, ok := c.waiters.PopLTE(newTime)
  if !ok {
    break
  }

}
```

internal/ztest/clock.go
```diff
@@ -21,30 +21,167 @@
...
+		// Travel to the time of the waiter and resolve it.
+		c.now = w.until
+
+		// The waiter may schedule more work
+		// so we must release the lock.
+		c.mu.Unlock()
+		w.fn()
+		// Sleeping here is necessary to let the side effects of waiters
+		// take effect before we continue.
+		time.Sleep(1 * time.Millisecond)
```
prashantv, line 118: just to understand this better, this is to make sure any asynchronous side-effects of waiters are run right? I'm guessing the main case is after writing to a ticker channel, we want to make sure the other side reads from it.

I guess if this sleep is too low, the ticker consumer may not have finished processing, and will miss ticks. If it's too high, we're slowing down tests. May be worth calling out that this is a best-effort.

internal/ztest/clock.go
```diff
@@ -21,30 +21,167 @@
...
+	until time.Time
+	fn    func()
+}
+
+// waiters is a thread-safe collection of waiters
+// with the next waiter to be resolved at the front.
+//
+// Use the methods on this type to manipulate the collection.
+// Do not modify the slice directly.
+type waiters struct{ heap waiterHeap }
```
prashantv, line 143: for a generic mock clock, I think a heap makes sense, though for the limited use-cases in zap, I wonder if we should do something simpler -- just use a slice sorted by time. Push is just append + resort, and pop is remove the first element

internal/ztest/clock.go
```diff
@@ -21,30 +21,167 @@
...
+}
+
+func (h *waiterHeap) Push(x interface{}) {
+	*h = append(*h, x.(waiter))
+}
+
+func (h *waiterHeap) Pop() interface{} {
+	old := *h
+	n := len(old)
+	x := old[n-1]
```
prashantv, line 184: since we have a pointer in `waiter` (the function), good practice to also zero it out to avoid memory leaks, `old[n-1] = waiter{}`

internal/ztest/clock.go
abhinav: Yeah, I'll mention Stop is a no-op.
Also note that it'll become a no-op in std in the future too ([ref](https://github.com/golang/go/issues/61542)).

internal/ztest/clock.go
abhinav: Yeah. I found that using `runtime.Gosched()` was not sufficient; worked for the unit tests in this package, but caused flakiness elsewhere.

A millisecond sleep for other computations to run was the most reliable way. This is also what benbjohnson/clock does:  https://github.com/benbjohnson/clock/blob/fa525c8ac4e629b5385b7b6546d69e0e7de0560e/clock.go#L417

I'll mention that side effects of operations waiting on the mock clock will resolve at a best effort basis.

internal/ztest/clock.go
abhinav: Heh, that's what I did in my first implementation of this, and realized I didn't like sorting the slice each `Add` call, and switched to a heap. I can switch back to that.

internal/ztest/clock.go
prashantv: It's a little different -- with that change, `Stop` is not required for cleanup. However, `Stop` still still have a functional impact -- it stops sending ticks to the channel. With this mock implementation, `Stop` doesn't actually stop anything, so we're still writing tickets to the channel periodically

prashantv (approved):

internal/ztest/clock.go
abhinav: Oh, fair. Yeah. Unfortunately, there's no way to intercept that Stop with the exported APIs. benbjohnson/clock gets around it with its own Ticker type, but I figured that's worth avoiding.

PR #1350, opened 2023-09-08
Improve string encoding by following json approach

cdvr1993:
Recently we found an application were using zap.Reflect was faster than using zapcore.ObjectMarshaler. After profiling we found that string encoding is really expensive. I replicated what encoding/json does and was able to get greater performance than Reflect.

I had to modify the benchmark to have a greater number of strings.

### Benchmark results

```
goos: linux
goarch: amd64
pkg: go.uber.org/zap/zapcore
cpu: AMD EPYC 7B13
               │ /tmp/old.txt │            /tmp/new.txt             │
               │    sec/op    │   sec/op     vs base                │
ZapJSON-8         89.10µ ± 1%   33.38µ ± 3%  -62.54% (p=0.000 n=10)
StandardJSON-8    40.74µ ± 1%   42.46µ ± 1%   +4.22% (p=0.000 n=10)
geomean           60.25µ        37.65µ       -37.52%
```

### DIFF
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -43,7 +43,7 @@ func (b *Buffer) AppendByte(v byte) {
 }
 
 // AppendByteV writes a single byte to the Buffer.
-func (b *Buffer) AppendByteV(v ...byte) {
+func (b *Buffer) AppendBytes(v []byte) {
 	b.bs = append(b.bs, v...)
 }
 
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -42,7 +42,7 @@ func (b *Buffer) AppendByte(v byte) {
 	b.bs = append(b.bs, v)
 }
 
-// AppendByteV writes a single byte to the Buffer.
+// AppendBytes writes a single byte to the Buffer.
 func (b *Buffer) AppendBytes(v []byte) {
 	b.bs = append(b.bs, v...)
 }
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -42,6 +42,11 @@ func (b *Buffer) AppendByte(v byte) {
 	b.bs = append(b.bs, v)
 }
 
+// AppendByteV writes a single byte to the Buffer.
+func (b *Buffer) AppendByteV(v ...byte) {
+	b.bs = append(b.bs, v...)
+}
+
 // AppendString writes a string to the Buffer.
 func (b *Buffer) AppendString(s string) {
 	b.bs = append(b.bs, s...)
--- a/zapcore/json_encoder.go
+++ b/zapcore/json_encoder.go
@@ -563,18 +563,20 @@ func safeAppendStringLike[S []byte | string](
 			continue
 		}
 
-		appendTo(buf, s[last:i])
-
 		r, size := decodeRune(s[i:])
 		if r == utf8.RuneError && size == 1 {
+			// Invalid UTF-8 sequence.
+			// Replace it with the Unicode replacement character.
+			appendTo(buf, s[last:i])
 			buf.AppendString(`\ufffd`)
 			i++
 			last = i
 			continue
 		}
-		appendTo(buf, s[i:i+size])
+
+		// No special handling required.
+		// Skip over this rune and continue.
 		i += size
-		last = i
 	}
 
 	// add remaining
--- a/zapcore/json_encoder.go
+++ b/zapcore/json_encoder.go
@@ -23,8 +23,10 @@ package zapcore
 import (
 	"encoding/base64"
 	"math"
+	"reflect"
 	"time"
 	"unicode/utf8"
+	"unsafe"
 
 	"go.uber.org/zap/buffer"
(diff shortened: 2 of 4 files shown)


buffer/buffer.go
```diff
@@ -42,6 +42,11 @@ func (b *Buffer) AppendByte(v byte) {
 	b.bs = append(b.bs, v)
 }
 
+// AppendByteV writes a single byte to the Buffer.
+func (b *Buffer) AppendByteV(v ...byte) {
+	b.bs = append(b.bs, v...)
+}
```
abhinav, line 48: This should probably be named `AppendBytes` and take a `[]byte`, not a vararg. (Also, the docstring is inaccurate.)

(FYI, there's also Buffer.Write which does the same, while satisfying io.Writer, but there's no problem with also having this method. However, if we have both, maybe Write should call AppendBytes.)

zapcore/json_encoder.go
```diff
@@ -486,67 +488,65 @@ func (enc *jsonEncoder) appendFloat(val float64, bitSize int) {
...
-			continue
-		}
-		enc.buf.AppendString(s[i : i+size])
-		i += size
-	}
+	enc.safeAddByteString(*(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
+		Data: (*reflect.StringHeader)(unsafe.Pointer(&s)).Data,
+		Len:  len(s),
+		Cap:  len(s),
+	})))
```
abhinav, line 495: Putting aside that this is decidedly Not Safe (in a function called safeAddString), starting Go 1.20, the above isn't the best way for an unsafe string to byte slice conversion.

It's better to now do: `unsafe.Slice(unsafe.StringData(s), len(s))`.

https://go.dev/play/p/mGnV97K5tu_w

zapcore/json_encoder.go
```diff
@@ -486,67 +488,65 @@ func (enc *jsonEncoder) appendFloat(val float64, bitSize int) {
...
 
 // safeAddByteString is no-alloc equivalent of safeAddString(string(s)) for s []byte.
 func (enc *jsonEncoder) safeAddByteString(s []byte) {
+	start := 0
 	for i := 0; i < len(s); {
-		if enc.tryAddRuneSelf(s[i]) {
+		if s[i] < utf8.RuneSelf {
+			if s[i] >= 0x20 && s[i] != '\\' && s[i] != '"' {
+				i++
+				continue
```
abhinav, line 505: Very nice! This right here is the performance win.
Scanning until the next position escaping is needed instead of appending one byte at a time.

I'm in favor of such a change, but I would prefer if we could do this without the unsafe.

abhinav (commented):
Very nice!
I would prefer to do this without unsafe, and I have some ideas on how we could do that. Let me try something out locally.

abhinav:
@cdvr1993 I just pushed a change that drops the unsafe/reflect in favor of generics.
It needs function references for DecodeUTF8 and AppendString vs AppendBytes,
but otherwise it's almost as good as the unsafe version -- only 4% slower than your version.

```
name     old time/op  new time/op  delta
ZapJSON   779µs ± 0%   813µs ± 1%  +4.34%  (p=0.000 n=9+10)
```

I think that may be acceptable since this is still a pretty massive net improvement. WDYT?

cdvr1993:
LGTM, on my computer the delta is like 1-2%, so probably you have some noise there.

abhinav:
@cdvr1993 I realized that the same idea (skip over characters that don't need special handling) could be used for the multi-byte runes as well. That yields a small improvement too.

I've pushed that and a small readability fix on top.

jquirke (commented):
tangential: why did we never escape \f as well?

abhinav:
> tangential: why did we never escape \f as well?

I don't know what the original reasoning for that choice is, but it's definitely handled:

https://github.com/uber-go/zap/blob/82c728b02339ba8ce9beb07b5d4ef5904a9c7a5f/zapcore/json_encoder_impl_test.go#L73-L76

We could change that and still be okay.

abhinav (approved):
CC @prashantv @rabbbit @sywhang

abhinav:
Thanks, @cdvr1993!

sywhang (approved):
Retrospective LGTM! 

I measured this on a pretty beefy machine but here's the results:
```
name        old time/op  new time/op  delta
ZapJSON-96  57.1µs ±25%  27.8µs ± 2%  -51.34%  (p=0.000 n=10+8)
```

buffer/buffer.go
```diff
@@ -42,6 +42,11 @@ func (b *Buffer) AppendByte(v byte) {
 	b.bs = append(b.bs, v)
 }
 
+// AppendBytes writes a single byte to the Buffer.
```
makkes, line 45: Just came here from the 1.26.0 release notes. Great job! This comment, though, doesn't seem to be right in that the method writes **all** bytes to the buffer.

abhinav: Oops, yes, you're right. I missed this. Thanks!

PR #1362, opened 2023-09-18
buffer: Fix AppendBytes doc string

abhinav:
Bad copy-paste; AppendBytes appends all the given bytes,
not just one byte.

Thanks to @makkes for catching this.


mway (approved):
Thanks!

PR #1363, opened 2023-09-22
test: migrate slog benchmark from golang.org/x/exp/slog to log/slog

chenyanchen:
Bechmark test `log/slog`.


chenyanchen:
Did you have uniform benchmark environment? If you do, please update benchmark result in README.

abhinav (commented):
Hey, @chenyanchen. Thanks for this PR. This makes sense, and .

However, I think this could benefit from a small change:
What you've done in `b.Run("slog"` is great.
I think we need a second `b.Run("slog/LogAttrs"` that is the same, except it uses [Logger.LogAttrs](https://pkg.go.dev/log/slog#Logger.LogAttrs) and the old `fakeSlogFields()` list.
This will provide a comparison of both, slog with `any`, and slog with `slog.Attr`.

Would you be willing to make that change?
If not, one of us will try to get to it as soon as we have some time to do it.

chenyanchen:
It's very thoughtful of you.

The `slog` `Logger.LogAttrs` benchmark been added.

chenyanchen:
> Did you have uniform benchmark environment? If you do, please update benchmark result in README.

If you do not ask for specific hardware, I can do benchmarks on my `M1` Mac.

abhinav:
> > Did you have uniform benchmark environment? If you do, please update benchmark result in README.
> 
> If you do not ask for specific hardware, I can do benchmarks on my `M1` Mac.

Hey, we do not have a uniform benchmark environment.
The benchmarks are updated with whatever machine someone was on at the time they ran the benchmarks.
It hasn't been a problem because the absolute numbers are less important than the relative numbers, which tend to follow the same trend regardless of where we're running the benchmarks.

chenyanchen:
Benchmars has been updated. The results looks as expected.

abhinav (approved):
It's weird that there's no difference between slog with any and slog with LogAttrs, but that's something to investigate later.

Thanks!

PR #1364, opened 2023-09-26
grow /exp/zapfield

mikhalytch:
adds following generic Field constructors:
- Uint
- Uint64
- Uint32
- Uint16
- Uint8
- Int
- Int64
- Int32
- Int16
- Int8
- Float64
- Float32
- String
- Strings


abhinav (commented):
Hey, @mikhalytch, thanks for this PR, and apologies for the delay.

I'd like to better understand the motivation behind this proposal.
What problem are we trying to solve with these type-parameterized field constructors?
Is this another version of the problem from #1275?

PR #1369, opened 2023-10-04
zapcore: Ignore nil Entry.Time in encoders

justinhwang:
Per the comment on `Encoder.EncodeEntry`, any fields that are empty including fields on the `Entry` type should be omitted. Omit the `Time` field when we have empty time.
This also aligns with slog.Handler contract.

Refs #1334
Discovered by #1335

### DIFF
--- a/exp/zapslog/handler_test.go
+++ b/exp/zapslog/handler_test.go
@@ -23,17 +23,13 @@
 package zapslog
 
 import (
-	"bytes"
-	"encoding/json"
 	"log/slog"
 	"testing"
-	"testing/slogtest"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zapcore"
-	"go.uber.org/zap/zaptest"
 	"go.uber.org/zap/zaptest/observer"
 )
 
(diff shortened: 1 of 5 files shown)


exp/zapslog/handler_test.go
```diff
@@ -189,3 +193,42 @@ func TestAttrKinds(t *testing.T) {
 		},
 		entry.ContextMap())
 }
+
+func TestSlogtest(t *testing.T) {
+	t.Skip() // TODO: skip for now until test failures are fixed (#1334)
```
abhinav, line 198: I would prefer to keep this test off master until we actually have it passing.
`t.Skip`s don't really get follow ups that often—I'd prefer we track the pending work in issues as we're already doing that.

Would you mind deleting this change from this PR?

zapcore/json_encoder_test.go
```diff
@@ -109,6 +109,20 @@ func TestJSONEncodeEntry(t *testing.T) {
 				}),
 			},
 		},
+		{
+			desc: "zero_time_omitted",
+			expected: `{
```
abhinav, line 114: Nice. Would you mind adding a similar check to the console_encoder_test.go as well?

abhinav (commented):
Thanks for doing this!
The PR largely LGTM. Just two changes, please:

- Delete the skipped test. We'd prefer to track that pending work in a GH issue.
- Add a similar check to console_encoder_test as the JSON case.

abhinav (approved):
I realized that the console encoder didn't have an existing test for something like this. I added one and a test for this.

LGTM!

mway (approved):
LGTM, thanks @justinhwang and @abhinav!

Issue #514, opened 2017-10-12
stacktrace collection should be overridable


zllak:
# Stacktrace collection
While trying to implement an encoder that is compatible with GCP Stackdriver logging and also for GCP Stackdriver Error Reporting, I suspected that it's not working properly because the generated stacktrace is not compatible with what Error Reporting is expected.

The documentation says (https://cloud.google.com/error-reporting/docs/formatting-error-messages):
```Go: Must be the return value of runtime.Stack()```

The code in go sources (`cloud.google.com/go/errors/errors.go`) shows that `cloud.google.com/go/errors.(*Client).logInternal` indeed calls `runtime.Stack`.

Unfortunately, the stack is grabbed by `trackStacktrace` in `stracktrace.go`, and there is apparently no other way around to provide the stracktrace.

Unless I'm missing something, we cannot provide the way the stacktrace is grabbed by zap

akshayjshah:
There's no way to customize this now, but I don't mind adding an option for this. We could make a `zap.Option` that takes a `func() string` and overrides the current stacktrace-snapping functionality.

However, this wouldn't affect the output of the `zap.Stack` field if it's used explicitly - the user would have to use some alternate field for stackdriver compatibility.

djui:
(I don't want to hijack this issue but would rather like to add a similar question and ask for a recommendation) Besides adding an option for the stacktrace-snapping functionality, is it currently already possible to modify the stack trace output? Would it work by implementing a `zapcore.Encoder`?

We would like to trim off some stacks that are not adding much to the signal and are rather noise.

I know trimming stacks is a inherently difficult task, but I would at least like to go give our heuristic a try.

evanj:
FWIW: It is possible to wrap the existing JSON encoder to hack stack traces. I've done this to format stacks so they get picked up by Google Cloud's Stackdriver Error Report (where it needs to look like a "real" panic). It would be much nicer to have a customization hook similar to `EncoderConfig.EncodeTime`, but this is possible today. See: https://github.com/evanj/gcplogs/blob/master/gcpzap/encoder.go#L48

FlorianLoch:
Any progress regarding adding support for the mentioned `zap.Option` allowing a custom function to provide the stack trace? That would be very handy to have ;)

sywhang:
@FlorianLoch nope, the issue is still a valid ask and we welcome PRs!

PR #1371, opened 2023-10-09
Allow configurable stacktrace encoding

arwineap:
This PR adds a `StacktraceEncoder` type to the `EncoderConfig` struct

Users can override the `EncodeStacktrace` field to configure how stacktraces are outputed This PR aims to resolve https://github.com/uber-go/zap/issues/514 by mirroring the behavior of the `EncodeCaller` field

The `EncodeStacktrace` field has been inserted as a required field, and has been added to `NewDevelopmentConfig` and `NewProductionConfig` as sane defaults however, as a required field it can cause a panic if a user is manually building their config without these helpers.


zapcore/encoder.go
```diff
@@ -270,6 +270,18 @@ func (e *DurationEncoder) UnmarshalText(text []byte) error {
...
 
+// A StacktraceEncoder serializes a stacktrace
+//
+// This function must make exactly one call
+// to a PrimitiveArrayEncoder's Append* method.
+type StacktraceEncoder func(string, PrimitiveArrayEncoder)
+
+// FullStacktraceEncoder passes down the full stacktrace as a string to the enc
+func FullStacktraceEncoder(stacktrace string, enc PrimitiveArrayEncoder) {
+	// TODO: consider using a byte-oriented API to save an allocation.
```
abhinav, line 281: FYI: This TODO will be unfixable once this change is released and these APIs are made public. We can discuss this more in the PR if you'd like, but I think this AppendString won't cause an alloc. (We've already paid for the stacktrace alloc elsewhere. )

zapcore/encoder.go
```diff
@@ -343,10 +355,11 @@ type EncoderConfig struct {
...
 	// seconds since epoch, while others may prefer ISO8601 strings.
-	EncodeLevel    LevelEncoder    `json:"levelEncoder" yaml:"levelEncoder"`
-	EncodeTime     TimeEncoder     `json:"timeEncoder" yaml:"timeEncoder"`
-	EncodeDuration DurationEncoder `json:"durationEncoder" yaml:"durationEncoder"`
-	EncodeCaller   CallerEncoder   `json:"callerEncoder" yaml:"callerEncoder"`
+	EncodeLevel      LevelEncoder      `json:"levelEncoder" yaml:"levelEncoder"`
+	EncodeTime       TimeEncoder       `json:"timeEncoder" yaml:"timeEncoder"`
+	EncodeDuration   DurationEncoder   `json:"durationEncoder" yaml:"durationEncoder"`
+	EncodeCaller     CallerEncoder     `json:"callerEncoder" yaml:"callerEncoder"`
+	EncodeStacktrace StacktraceEncoder `json:"stacktraceEncoder" yaml:"stacktraceEncoder"`
```
abhinav, line 362: If we're adding JSON/YAML tags to this, we'll probably want to implement UnmarshalText on the StacktraceEncoder type so that this is actually configurable.
e.g.

https://github.com/uber-go/zap/blob/87577d85d58b6d92d0158967df29303d04d30e36/zapcore/encoder.go#L295-L303

abhinav (commented):
Thanks for the PR, @arwineap! This change looks largely in the right direction.
However:

> The EncodeStacktrace field has been inserted as a required field [..] it can cause a panic if a user is manually building their config without these helpers

That's a breaking change and we can't merge that. The new field must be optional.
The console and JSON encoders already have handling for optional overrides like this one; this should be another such optional field. For example:

https://github.com/uber-go/zap/blob/87577d85d58b6d92d0158967df29303d04d30e36/zapcore/json_encoder.go#L381-L387

zapcore/encoder.go
arwineap: Thank you, I had misunderstood what this was doing and I appreciate learning the new pattern

arwineap:
Thank you for the clarification, I'm unsure how I missed the default catch for the nil cases. This has been added.

One other concern I had was with the console encoder; I'm not sure I can avoid calling `getSliceEncoder()` a second time because of the split in output between structured fields and unstructured panic data. This felt ineffecient, but the code seems cleaner than the alternative

zapcore/console_encoder.go
```diff
@@ -119,10 +120,23 @@ func (c consoleEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer,
...
+		if stacktraceEncoder == nil {
+			stacktraceEncoder = FullStacktraceEncoder
+		}
+		stacktraceEncoder(ent.Stack, arr)
+		for i := range arr.elems {
+			if i > 0 {
+				line.AppendString(c.ConsoleSeparator)
+			}
+			fmt.Fprint(line, arr.elems[i])
+		}
```
abhinav, line 138: The contract for stacktraceEncoder is that it makes at most one `Append*` call.
So we don't need the full `arr.elems` handling here. It can just be:
```suggestion
		if len(arr.elems) > 0 {
			fmt.Fprint(line, arr.elems[0])
		} else {
			// User-provided encoder was a no-op.
			// Fall back to full encoder.
			line.AppendString(ent.Stack)
		}
```

abhinav (commented):
This is great, thanks. I've left some minor comments.

I have a couple open questions that I maybe need input from other maintainers on.

- Should the encoder be primitives-based like the other encoder overrides or a full one capable of posting objects -- e.g. if someone parses the stack trace?
- Should the fallback behavior be to encode the full stack trace, or should we take that as an indication that they don't want to encode the stack trace? 

CC @prashantv @sywhang @mway

abhinav:
Related: We'll want to make a decision on https://github.com/uber-go/zap/issues/1373 before we release this because choices there affect the design for this.

zapcore/json_encoder.go
```diff
@@ -420,7 +420,21 @@ func (enc *jsonEncoder) EncodeEntry(ent Entry, fields []Field) (*buffer.Buffer,
...
+		stacktraceEncoder := final.EncodeStacktrace
+		if stacktraceEncoder == nil {
+			stacktraceEncoder = FullStacktraceEncoder
+		}
+
+		stacktraceEncoder(ent.Stack, final)
+		if cur == final.buf.Len() {
+			// User-supplied EncodeStacktrace was a no-op. Fall back to strings to
+			// keep output JSON valid.
+			final.AppendString("")
```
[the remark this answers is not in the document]
arwineap: Good catch, this was actually intentional.

The idea being that a user can disable stacktrace key / values in json output by configuring `StacktraceKey` to be blank

However, if the user created the custom encoder which emptied out the stack trace, it seems like we should respect their key / value, and append their `StacktraceKey` with an empty string

The issue is that at this part of the code the `StacktraceKey` has already been added to the json output, so we need a string value to make the json valid, so an empty string makes sense. Perfectly reasonable to only add the Key if a value exists, just need some rearranging

It's a matter of design, and it's more important that it's consistent across all of the options, so I will defer to the zap team here :+1:

devdrops:
@abhinav @arwineap hi! Any plans about this PR? Any responses are really appreciated :slightly_smiling_face:

arwineap:
I will work to rebase this with modern changes; but depending on any decision in https://github.com/uber-go/zap/issues/1373 we may try to send the entire stack object into the function instead of the string

win5do:
We need this API, the stacktrace contains newlines and tabs, which are very difficult to read in json.

PR #1372, opened 2023-10-09
Make it possible to extract fields from the context in the slog adapter

recht:
Since slog has context logging built in it would be nice to be able to utilize that with a zap handler, for example to log tracing context.

### DIFF
--- a/exp/zapslog/handler.go
+++ b/exp/zapslog/handler.go
@@ -32,13 +32,17 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+// ContextExtractor can be used to extract fields from a context.
+type ContextExtractor func(ctx context.Context) []zapcore.Field
+
 // Handler implements the slog.Handler by writing to a zap Core.
 type Handler struct {
-	core       zapcore.Core
-	name       string // logger name
-	addCaller  bool
-	addStackAt slog.Level
-	callerSkip int
+	core             zapcore.Core
+	name             string // logger name
+	addCaller        bool
+	addStackAt       slog.Level
+	callerSkip       int
+	contextExtractor ContextExtractor
 }
 
 // NewHandler builds a [Handler] that writes to the supplied [zapcore.Core]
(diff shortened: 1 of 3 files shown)


r-hang (commented):
This makes sense to me! I think the name WithFieldsFromContext is a bit more descriptive than WithContextExtractor.

recht:
> This makes sense to me! I think the name WithFieldsFromContext is a bit more descriptive than WithContextExtractor.

I can change it to that, but do you then have any suggestion for the field/type name?

r-hang:
My apologies, I've thought about it some more and looked at more code and propose

WithContextFieldExtractors(extractors ...ContextFieldExtractor), where a ContextFieldExtractor takes a context.Context and returns a single zapcore.Field. 

I think this will make composing and sharing logic between multiple extractors a bit easier than if the option constructor took in a single extractor function that returned multiple fields.

recht:
Sorry for the late reply - we could do that instead, but it will lead to some amount of duplicated work. My specific use case is to add tracing info to logs, and for that I have something like this:
```
zapslog.WithContextExtractor(func(ctx context.Context) []zapcore.Field {
		span := trace.SpanFromContext(ctx)
		sctx := span.SpanContext()
		if sctx.IsValid() {
			return []zapcore.Field{
				zap.Stringer("trace_id", sctx.TraceID()),
				zap.Stringer("span_id", sctx.SpanID()),
				zap.Int8("trace_flags", int8(sctx.TraceFlags())),
			}
		}
		return nil
	})
```
Splitting that into 3 extractors instead then requires me to get the span context 3 times, check if it's valid, and then emit single fields.

How about instead allowing multiple extractors, but allow them to still return a slice of fields?

r-hang (approved):
Makes sense to me.

Providing options for multiple extractors that can provide multiple fields seems adaptable to both cases where someone might want to re-use context extractors or provide a single one for all of their fields.

ixugo:
Is there any more progress?

swiftxy:
Is there any more progress?

We need to add `trace_id` globally throughout the project, which requires changing:
```go
slog.Info("something")
```
to:
```
slog.Info(ctx, "something")
```

This approach forces context passing in every log call and doesn't provide a clean way to share logger with trace_id in context.

ixugo:
> Is there any more progress?还有什么进展吗？
> 
> We need to add `trace_id` globally throughout the project, which requires changing:我们需要在整个项目全局添加 `trace_id` ，这需要更改：
> 
> ```go
> slog.Info("something")
> ```
> 
> to:  到：
> 
> ```
> slog.Info(ctx, "something")
> ```
> 
> This approach forces context passing in every log call and doesn't provide a clean way to share logger with trace_id in context.这种方法强制在每次日志调用中传递上下文，并且不提供在上下文中与 trace_id 共享记录器的干净方法。

This is how I solved it

https://github.com/ixugo/goddd/blob/master/pkg/logger/slog.go

PR #1378, opened 2023-10-31
Extend WithLazy() method to SugaredLogger

### LINKED ISSUE #1377: Extend `WithLazy` method to `SugaredLogger`
Hi, `zap` maintainers!

The recently added `WithLazy` method is amazing (PR: https://github.com/uber-go/zap/pull/1319). This capability is not yet available for `SugaredLogger`

The straightforward idea is to expose a similar signature from `SugaredLogger` using `WithLazy` method :

```go
func (s *SugaredLogger) WithLazy(args ...interface{}) *SugaredLogger {
	return &SugaredLogger{base: s.base.WithLazy(s.sweetenFields(args)...)}
}
```

There is still overhead from the `sweetenFields` call, but it's nonetheless much less than using `With()`.

Nope

If you approve, I'm ready to work on that because it's somewhat blocking us

defval:
Hey there! Resolves #1377

Notes:
- I reused all tests from the `With()` method to work with both `With()` and `WithLazy()`
- To highlight the differences between `With()` and `WithLazy()`, I added a test similar to [this one](https://github.com/uber-go/zap/pull/1319/files#diff-55a2db70576eb079880167200221ea3787a01fa4bbc958503769eba24084d842R179). I kept only the test cases that clearly show the distinctions

Looking forward to your thoughts!

### DIFF
--- a/sugar_test.go
+++ b/sugar_test.go
@@ -187,7 +187,7 @@ func TestSugarWithCaptures(t *testing.T) {
 			},
 		},
 		{
-			name:        "lazy with captures arguments at time of With or Logging",
+			name:        "lazy with captures arguments at time of Logging",
 			withMethods: []withAny{(*SugaredLogger).WithLazy},
 			wantJSON: []string{
 				`{
--- a/sugar_test.go
+++ b/sugar_test.go
@@ -22,6 +22,8 @@ package zap
 
 import (
 	"errors"
+	"fmt"
+	"strconv"
 	"testing"
 
 	"go.uber.org/zap/internal/exit"
@@ -53,6 +55,9 @@ func TestSugarWith(t *testing.T) {
 			Context: []Field{Error(err)},
 		}
 	}
+
+	type withAny func(*SugaredLogger, ...interface{}) *SugaredLogger
+	withMethods := []withAny{(*SugaredLogger).With, (*SugaredLogger).WithLazy}
 
 	tests := []struct {
 		desc     string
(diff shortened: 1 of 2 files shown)


sugar_test.go
```diff
@@ -141,16 +146,94 @@ func TestSugarWith(t *testing.T) {
...
+				`{
+					"m": "world 0",
+					"a0": [1],
+					"c0": [2]
+				}`,
+			},
+		},
+	}
+
+	for _, tt := range tests {
```
r-hang, line 207: https://github.com/uber-go/zap/pull/1319/files also has a few more scenarios for the capture tests. While you are testing some aspect of chained With and WithLazy, I think it makes sense to also test interleaved calls of With then WithLazy and vis-versa as well.

r-hang (commented):
Sorry for the delay, the implementation looks good. I think we should at least add tests for interleaved With and WithLazy calls, what do you think?

sugar_test.go
```diff
@@ -141,16 +146,94 @@ func TestSugarWith(t *testing.T) {
...
+				}`,
+				`{
+					"m": "world 0",
+					"a0": [0],
+					"c0": [2]
+				}`,
+			},
+		},
+		{
+			name:        "lazy with captures arguments at time of With or Logging",
```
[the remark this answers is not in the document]
defval: Yeah, I think so, fixed

sugar_test.go
defval: I was actually thinking about including those interleaved `With` and `WithLazy` tests at first. But then I thought, since `lazyWithCore` already covers what we need, maybe it’s better not to repeat the same stuff? I just focused on showing how `With` and `WithLazy` are different, especially with how `WithLazy` uses that extra core.

I'm totally fine with adding more tests if it really helps. What do you think?

r-hang: I agree that there would be testing overlap between the two loggers. 

I think it would be nice for zap to set up testing in a way that would be easy to verify sugared and non sugared behavior through the same process where they should be similar. However, that doesn't need to be part of this change.

r-hang (approved):
thank you for the contribution!

PR #1384, opened 2023-11-08
Update goleak to 1.3.0

michel-laterman:
Update goleak to 1.3.0 to remove `x/lint` and `x/tools` as well as their dependencies


r-hang:
Change concept looks good to me but the lint job is still failing. I assume you ran go mod tidy?

michel-laterman:
I was running `make tidy-lint` but didn't catch that it stops proccessing after hitting an error in one file. They should now all be updated

sywhang (approved):
LGTM, thank you for the contribution.

Issue #1394, opened 2023-12-11
Allow copying `fields` when using `WrapCore`


adrianlungu:
**Is your feature request related to a problem? Please describe.**

https://go.dev/play/p/vlsJxJfxhUL

Any fields added via the config, or via `With` to the logger before calling `zap.WrapCore` are not seen at all by the observer via the context. 

**Describe the solution you'd like**

After building a logger with fields using `.Build()`, allow using `zap.WrapCore` to be able to split, or replace the core of a logger, but keep the fields that were added via the configuration, i.e.

This could be done either via exposing `Context` on `core`, or internally by copying the context from one core to the other, although allowing the developer to choose the scenario that fits him best would be better. 

**Describe alternatives you've considered**

None so far that wouldn't involve heavy refactoring. 

**Is this a breaking change?**

It should not be a breaking change. 

**Additional context**
N/A/

PR #1395, opened 2023-12-13
Allow retrieval of previously added fields from the Core

adrianlungu:
As per the feature requested opened [here](https://github.com/uber-go/zap/issues/1394):

- extended the `Core` interface with `Fields()` which can be used to retrieve fields that have been previously added to the core
- updated the `ioCore.With` to store the fields in an internal private array
- updated the `ioCore.Clone` to clone the fields as well
- added unit tests to cover these cases
- updated the types implementing `Core` to also have `Fields()`

### DIFF
--- a/zapcore/core.go
+++ b/zapcore/core.go
@@ -42,6 +42,8 @@ type Core interface {
 	Write(Entry, []Field) error
 	// Sync flushes buffered logs (if any).
 	Sync() error
+	// Fields returns any added structured context from the Core.
+	Fields() []Field
 }
 
 type nopCore struct{}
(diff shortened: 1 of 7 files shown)


JacobOaks:
Hey @adrianlungu, thanks for the PR.

As it stands currently, this is a breaking change. Adding a method to an exported interface will cause all custom implementations to no longer satisfy the interface.

I'm also curious about your use case. Usually I see `observer` used in-place of a normal logger, passed into some functionality where you want to test what messages get logged. I don't usually see it used in combination with another logger. If you want to test what fields get added to a logger, is there any way you can pass an `observer` before the fields get added?

Maybe you can create a separate implementation of `Core` that wraps an underlying core but keeps track of the fields you add (similar to what this PR is having `ioCore` do), and use that in your code instead of baking it into `ioCore` by default.

adrianlungu:
Hello @JacobOaks , thank you for the review and insight! 

I have a logger built on top of zap which has some mandatory fields that should be present, populated from the environment, but the logger can also be customized with additional fields via context or options. 

While trying to write a unit test to assert that the required fields are there, and then that the optional fields are there, I stumbled into the `Context` fields which does not return the fields if they are initialized via the `InitialFields` property, or if they are added to the core before I get a chance to inject the `Observable core`. 

Using a config `Build()` and wrapping the core in there also doesn't help because the `WrapCore` func would be called after the fields would be added to the core so they would not be seen by the observer core.

Issue #1390, opened 2023-11-28
Vulnerability scanning shows unsanitized input, path traversal issue in zap


candita:
**Describe the bug**
Snyk code vulnerability scanner was run on vendored uber-go code and found an issue:

> Error: SNYK_CODE_WARNING ([CWE-23](https://cwe.mitre.org/data/definitions/23.html)):
> vendor/go.uber.org/zap/sink.go:139:9: error[go/PT]: Unsanitized input from the request URL flows into os.OpenFile, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to open arbitrary files.
> 
This is an old line number reference, but the issue is still visible in the repo head, at https://github.com/uber-go/zap/blob/master/sink.go#L158C33-L158C33.

**To Reproduce**
Run any scanner that picks up Path Traversal or CWE vulnerabilities.  CWE-23 will appear and reference https://github.com/uber-go/zap/blob/master/sink.go#L158C33-L158C33.

**Expected behavior**
No vulnerabilities seen.

**Additional context**
CWE-23 is in the top 25 Common Weaknesses.  Zap "... uses external input to construct a pathname that should be within a restricted directory, but it does not properly neutralize sequences such as ".." that can resolve to a location that is outside of that directory.  This allows attackers to traverse the file system to access files or directories that are outside of the restricted directory."

Thanks for your attention.

abhinav:
Thanks for the report, @candita!

I think this is a false positive.
The URL there is used to distinguish between different output sources in configuration.
That code path will be hit only if you use the URL `file://...` when configuring your logger.

This is where that sink gets registered:

https://github.com/uber-go/zap/blob/5acd569b6a5264d4c7433cbb278a8336d491715c/sink.go#L70

(schemeFile is "file" here.)

So when someone specified `zap.Open("file://path/to/file")`, that'll hit this line:

https://github.com/uber-go/zap/blob/5acd569b6a5264d4c7433cbb278a8336d491715c/writer.go#L71

Which will dispatch on the `file://` to open enter newSinkFromURL, which will call the vulnerable line linked above.

The input to the line vulnerable line will not come from an HTTP external request.
If someone did `zap.Open("http://../foo")`, it would dispatch to a sink registered for the HTTP protocol, not the file sink.
So the only way to hit this with a request input is if someone specifically did: `zap.Open("file://" + req.URL.Path)`.

Is there something we can do to help address this false positive?
[...]

PR #1397, opened 2023-12-18
zap.Open: Invalidate relative paths and paths with ".." segments

r-hang:
Currently, zap.Open doesn't prevent someone from explicitly doing something like zap.Open("file://../../../secret").

zap.Open already documents that paths passed to it must be absolute. Add validation to error if zap.Open is called with a relative paths that could write files outside of intended file directory hierarchy.

This change addresses https://cwe.mitre.org/data/definitions/23.html

ref https://github.com/uber-go/zap/issues/1390

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -23,6 +23,9 @@ package zap
 import (
 	"fmt"
 	"io"
+	"net/url"
+	"path/filepath"
+	"strings"
 
 	"go.uber.org/zap/zapcore"
 
@@ -48,6 +51,10 @@ import (
 // os.Stdout and os.Stderr. When specified without a scheme, relative file
 // paths also work.
 func Open(paths ...string) (zapcore.WriteSyncer, func(), error) {
+	if err := checkOpenPaths(paths); err != nil {
+		return nil, nil, err
+	}
+
 	writers, closeAll, err := open(paths)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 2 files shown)


writer.go
```diff
@@ -84,6 +91,23 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
...
+	var openErr error
+	for _, path := range paths {
+		if !strings.HasPrefix(path, "file:") {
+			continue
+		}
+		u, err := url.Parse(path)
+		if err != nil {
+			return err
+		}
+		if !(filepath.IsAbs(u.Path) && !strings.Contains(u.Path, "..")) {
```
abhinav, line 104: if filepath.IsAbs returns true, we're good. It's okay if an absolute path has `/../` because that'll just modify the absolute path. `/foo/bar/../baz` is just `/foo/baz`.

writer.go
```diff
@@ -84,6 +91,23 @@ func open(paths []string) ([]zapcore.WriteSyncer, func(), error) {
 	return writers, closeAll, nil
 }
 
+func checkOpenPaths(paths []string) error {
+	var openErr error
+	for _, path := range paths {
+		if !strings.HasPrefix(path, "file:") {
+			continue
+		}
```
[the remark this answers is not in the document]
r-hang: I just experimented and yes this location allows for the reuse of a lot more code.

writer.go
r-hang: @abhinav, I think that means we don't need this feature then?

`sr.openFile`, given a rawUrl with a file scheme, always consumes an absolute path because `url.Parse` return object's Path field only contains an absolute path under this constraints. `filepath.IsAbs` will always return true here.

Experimentation:
https://go.dev/play/p/Xcjj0PhmgXo

If that make sense I can close the issue?

abhinav: @r-hang can we confirm with a test that it's not possible to provide a relative path or one with '..' to that API?

r-hang: https://github.com/uber-go/zap/pull/1398 as a successor since I'm not able to reopen this one.

PR #1398, opened 2023-12-19
zap.Open: Invalidate relative path roots

r-hang:
Add validation to ensure file schema paths passed to zap.Open are absolute since this is already documented.

Tests are also added to demonstrate that double dot segements within file schema URIs passed to zap.Open remain within the specified file hierarchy.

This change addresses https://cwe.mitre.org/data/definitions/23.html

ref https://github.com/uber-go/zap/issues/1390

This PR succeeds https://github.com/uber-go/zap/pull/1397

### DIFF
--- a/sink.go
+++ b/sink.go
@@ -103,6 +103,11 @@ func (sr *sinkRegistry) newSink(rawURL string) (Sink, error) {
 	if err != nil {
 		return nil, fmt.Errorf("can't parse %q as a URL: %v", rawURL, err)
 	}
+
+	if u.Scheme == schemeFile && !filepath.IsAbs(u.Path) {
+		return nil, fmt.Errorf("file URI %q attempts a relative path", rawURL)
+	}
+
 	if u.Scheme == "" {
 		u.Scheme = schemeFile
 	}
--- a/sink.go
+++ b/sink.go
@@ -104,12 +104,9 @@ func (sr *sinkRegistry) newSink(rawURL string) (Sink, error) {
 		return nil, fmt.Errorf("can't parse %q as a URL: %v", rawURL, err)
 	}
 
-	if u.Scheme == schemeFile && !filepath.IsAbs(u.Path) {
-		return nil, fmt.Errorf("file URI %q attempts a relative path", rawURL)
-	}
-
+	// No scheme specified. Assume absolute or relative file path.
 	if u.Scheme == "" {
-		u.Scheme = schemeFile
+		return sr.newFileSinkFromPath(rawURL)
 	}
 
 	sr.mu.Lock()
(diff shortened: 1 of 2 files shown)


abhinav:
Okay, so I had a chance to go over the issue again and the url.Parse behavior you mentioned.
While it's correct that .Path from `url.Parse` will never give us a relative path, #1390 specifically wants to stop us from allowing `..` in that path.

I see that you added the check to `newSink` instead of `newFileSinkFromURL` because we still want `zap.Open` to support relative paths when `file://` is not specified.
This should've been explained in a comment there, ideally, but I think it might be better if we just diverge those two code paths: `newSink` should call `newFileSinkFromURL` if a scheme was specified, and `newFileSinkFromPath` directly if not. That keeps this simple.

I've pushed a commit that does that but:

- I've removed some invalid `file:foo/bar` tests -- that's not a valid URL. 
- I didn't understand the purpose of the TestOpenRelativeValidated test, so I've skipped it for now. I'll let you fix/remove it based on what you were trying to do there.

r-hang:
@abhinav by TestOpenRelativeValidated, I think that's TestOpenDotSegmentsSanitized because as that is the one skipped. 

This test is no longer required if we are updating the validation to reject all double dot segments from file schema paths. I've deleted that test wholesale. 

My original intention was to use that test to demonstrate that zap.Open would still function in a safe way even when provided with double dot segments in the path.

abhinav (approved):

sink.go
```diff
@@ -145,6 +147,10 @@ func (sr *sinkRegistry) newFileSinkFromURL(u *url.URL) (Sink, error) {
 		return nil, fmt.Errorf("file URLs must leave host empty or use localhost: got %v", u)
 	}
 
+	if strings.Contains(u.Path, "..") {
```
prashantv, line 150: A filename can contain `..`, so I don't think this check is right
```
$ vim foo..bar

$ cat foo..bar
hello
```

should we instead look for an absolute path before calling `newFileSinkFromPath`?

prashantv (changes requested):

sink.go
abhinav: newFileSinkFromPath must support relative paths. Paths without the file:// form are allowed to be relative. This is documented and tested already. 

Paths that take the file:// form will use net/url, and the path will always be absolute (because u.Path will always start with / for them).

The ask was to specifically reject path traversals with the URLs. We could check for /../ instead.

TBH, I'm not convinced that there's a vulnerability here, but I'll believe that a security expert could use some combination of symlinks and .. to get arbitrary file access.

prashantv: The path traversal security issue may be a false positive, since we have the hostname check above(`file://foo/bar` parses to host=`foo`, path=`/bar`).

Stepping back, what is the security vulnerability exactly? If the user has control over the paths, they can specify relative paths like `../foo` which is a valid supported path -- does it matter if it's via URL or a relative path?

abhinav: As I said, I'm not convinced that there's a vulnerability here.
But you're right, given that the user has full control of this input, this is even less a question of sanitization.
If we want to appease the linter, we can add a `filepath.Clean` there, but the actual check may be unnecessary.

abhinav: Commented in https://github.com/uber-go/zap/issues/1390#issuecomment-1867932885

Issue #1333, opened 2023-08-23
zapslog stabilization tracking issue


abhinav:
This issue tracks everything we need to do before zapslog can be moved to the main go.uber.org/zap module.

- [ ] Add stack traces #1329
- [ ] Test with slogtest, fix all failures
- [ ] Ensure adequate test coverage
- [ ] Delete exp-based implementation
- [ ] Add a zapslog.Core that writes Zap logs to a slog.Handler

arukiidou:
I would like to contribute  zapslog GA.

> zapslog: Add a zapslog.Core that writes Zap logs to a slog.Handler
> We may want to rename zapslog.Options to HandlerOptions when we do this.

Can you tell us more about this?

abhinav:
> I would like to contribute zapslog GA.
> 
> > zapslog: Add a zapslog.Core that writes Zap logs to a slog.Handler
> > We may want to rename zapslog.Options to HandlerOptions when we do this.
> 
> Can you tell us more about this?

Hello! Yes!

Okay, so right now, zapslog takes a Zap core, and returns a slog handler. `func NewHandler(zapcore.Core) slog.Handler`.
That's for applications that are already using Zap, and want to connect to a library that's using slog, to have the library feed their slog logs into their Zap handler.

The proposal was for the reverse: your application is using slog, and you want to use a library that uses Zap, so you want Zap to feed into your slog as the backend. Basically, `func NewCore(slog.Handler) zapcore.Core`.

Thinking about it out loud, though I don't know if this is actually necessary. @mway @prashantv @sywhang @r-hang What do y'all think?

On the second point: if this was added, then we'd want to rename the `Option` type to `HandlerOption` because we'd need another Option type (CoreOption) for the `NewCore` function.

One thing we could do is defer implementing NewCore until the need presents itself, but rename Option to HandlerOption anyway so that if/when that happens, we have the "option" name free.

arukiidou:
- Is this what you mean?: #1411
If so, it seems to me that we can submit this as a separate item now.

That way we could move zapslog to the root moduke without having to change anything, and easier to migrate from exp for users..

abhinav:
Yeah, I think following the rename, we can defer adding `NewCore` to when we need it. 
That PR looks good! Thanks!

arukiidou:
Started working on custom mapping.

- #1412
  - #1413

arukiidou:
- Added #1413, please take a look?

mblaschke:
[...]

PR #1408, opened 2024-02-02
zapslog: fix all with slogtest, support inline group, ignore empty group.

### LINKED ISSUE #1334: zapslog.Handler: Test with slogtest, fix all failures

### LINKED ISSUE #1401: zapslog: Handler should not output groups for an empty Record
- part of #1334

**Describe the bug**

`slogtest` cause errors:

```
unexpected key "H": a Handler should not output groups for an empty Record (zap/exp/zapslog/slogtest.go:165)
```

actual logs:

```
zap\exp\zapslog\logger.go:130: 2024-01-28T01:08:54.927+0900	INFO	msg	{"a": "b", "G": {"c": "d", "H": {}}}
```

**To Reproduce**
run slogtest
```go
f: func(l *slog.Logger) {
	l.With("a", "b").WithGroup("G").With("c", "d").WithGroup("H").Info("msg")
},
```

**Expected behavior**

so that:

```
zap\exp\zapslog\logger.go:130: 2024-01-28T01:08:54.927+0900	INFO	msg	{"a": "b", "G": {"c": "d"}}
```

**Additional context**
none.

### LINKED ISSUE #1402: zapslog: Handler should inline the Attrs of a group with an empty key
- part of #1334
  - may fixed by #1263

**Describe the bug**

`slogtest` cause errors:

```
missing key "c": a Handler should inline the Attrs of a group with an empty key (zap/exp/zapslog/slogtest.go:126)
```

actual logs:

```
zap\exp\zapslog\logger.go:130: 2024-01-28T01:08:54.902+0900	INFO	msg	{"a": "b", "": {"c": "d"}, "e": "f"}
```
**To Reproduce**
run `testing/slogtest`

```go
f: func(l *slog.Logger) {
	l.Info("msg", "a", "b", slog.Group("", slog.String("c", "d")), "e", "f")
},
```

**Expected behavior**

so that:

```
exp\zapslog\logger.go:130: 2024-01-28T01:08:54.902+0900	INFO	msg	{"a": "b", "c": "d", "e": "f"}
```

**Additional context**
none.

arukiidou:
Support all slog features, add `testing/slogtest`, and fixes all slogtest failure.

- part of #1333
- fixes #1334
  - fixes #1402
  - fixes #1401
  - closes #1263

### DIFF
--- a/exp/zapslog/handler.go
+++ b/exp/zapslog/handler.go
@@ -88,6 +88,10 @@ func convertAttrToField(attr slog.Attr) zapcore.Field {
 	case slog.KindUint64:
 		return zap.Uint64(attr.Key, attr.Value.Uint64())
 	case slog.KindGroup:
+		if attr.Key == "" {
+			// Inlines recursively.
+			return zap.Inline(groupObject(attr.Value.Group()))
+		}
 		return zap.Object(attr.Key, groupObject(attr.Value.Group()))
 	case slog.KindLogValuer:
 		return convertAttrToField(slog.Attr{
--- a/exp/zapslog/handler.go
+++ b/exp/zapslog/handler.go
@@ -26,6 +26,7 @@ import (
 	"context"
 	"log/slog"
 	"runtime"
+	"slices"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/internal/stacktrace"
@@ -39,6 +40,7 @@ type Handler struct {
 	addCaller  bool
 	addStackAt slog.Level
 	callerSkip int
+	holdGroup  string // holds latest group.
 }
 
 // NewHandler builds a [Handler] that writes to the supplied [zapcore.Core]
@@ -161,11 +163,17 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
 		ce.Stack = stacktrace.Take(3 + h.callerSkip)
 	}
 
-	fields := make([]zapcore.Field, 0, record.NumAttrs())
+	fields := make([]zapcore.Field, 0, record.NumAttrs()+1)
 	record.Attrs(func(attr slog.Attr) bool {
-		fields = append(fields, convertAttrToField(attr))
+		f := convertAttrToField(attr)
+		if h.holdGroup != "" && f != zap.Skip() {
+			fields = slices.Insert(fields, 0, zap.Namespace(h.holdGroup))
+			h.holdGroup = ""
+		}
+		fields = append(fields, f)
 		return true
 	})
+
 	ce.Write(fields...)
 	return nil
 }
@@ -175,7 +183,12 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
 func (h *Handler) WithAttrs(attrs []slog.Attr) slog.Handler {
 	fields := make([]zapcore.Field, 0, len(attrs))
 	for _, attr := range attrs {
-		fields = append(fields, convertAttrToField(attr))
+		f := convertAttrToField(attr)
+		if h.holdGroup != "" && f != zap.Skip() {
+			fields = slices.Insert(fields, 0, zap.Namespace(h.holdGroup))
+			h.holdGroup = ""
+		}
+		fields = append(fields, f)
 	}
 	return h.withFields(fields...)
 }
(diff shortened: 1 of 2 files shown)


arukiidou:
@abhinav @tchung1118
Please take a look at this one? Thanks!

exp/zapslog/handler.go
```diff
@@ -157,29 +162,45 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
...
-	fields := make([]zapcore.Field, 0, record.NumAttrs())
+	fields := make([]zapcore.Field, 0, record.NumAttrs()+1)
 	record.Attrs(func(attr slog.Attr) bool {
-		fields = append(fields, convertAttrToField(attr))
+		f := convertAttrToField(attr)
+		if h.holdGroup != "" && f != zap.Skip() {
+			fields = append(fields, zap.Namespace(h.holdGroup))
+			h.holdGroup = ""
+		}
+		fields = append(fields, f)
```
abhinav, line 172: I don't think this bit is right. We cannot modify `h`'s internal state in a `Handle`.
If we have two concurrent Handle calls, this will include the group in only one of those calls.
I think what we want to do here is record a boolean in the closure outside the Attrs loop, and flip it after the first `zap.Namespace`.

Something like this:

```
var addedNamespace bool
record.Attrs(... {
  if ... && !addedNamespace {
    fields = append(...)
    addedNamespace = true
  }
})
```

This way two concurrent calls will not fight each other on this.

exp/zapslog/handler.go
```diff
@@ -157,29 +162,45 @@ func (h *Handler) Handle(ctx context.Context, record slog.Record) error {
...
 // both the receiver's attributes and the arguments.
 func (h *Handler) WithAttrs(attrs []slog.Attr) slog.Handler {
-	fields := make([]zapcore.Field, 0, len(attrs))
+	fields := make([]zapcore.Field, 0, len(attrs)+1)
 	for _, attr := range attrs {
-		fields = append(fields, convertAttrToField(attr))
+		f := convertAttrToField(attr)
+		if h.holdGroup != "" && f != zap.Skip() {
+			fields = append(fields, zap.Namespace(h.holdGroup))
+			h.holdGroup = ""
```
abhinav, line 188: Same here. Whether or not the namespace was added should be considered a concern local to this WithAttrs call.

After that, in the child handler constructed and returned below—that's where we unset the group because it's already been "used" here.

abhinav (commented):
Thanks for this, @arukiidou! Really appreciate the work you put into matching the slog.Handler requirements. This is looking pretty good at a high-level, but there's one issue with the state management: we can't change the handler state like that.
Once a Handler is constructed, consider all its fields immutable. Only new handlers (like the ones returned by WithAttrs) can have different values for those fields—that's the only time we can mutate them. After construction, these handlers will be re-used for concurrent log calls, so they can't modify their own state. I've left suggestions on how I think we could fix the issue.

arukiidou:
Must adding `WithGroup` in `WithGroup` tests.

arukiidou:
Today, we begin working on this again.

exp/zapslog/handler.go
arukiidou: fixed.

exp/zapslog/handler.go
arukiidou: fixed.

abhinav (approved):
Thanks for the quick fix, @arukiidou!

I made a few minor fixups and changed the `holdGroup` to a slice because the string can't handle cases like `WithGroup("foo").WithGroup("bar").Info("msg")`.

The change looks good to me now. I'll ping a couple of the other maintainers to try to get their approval as well.

Again, thanks for the contribution!

abhinav:
Got an offline LGTM from @mway.
Merging!

abhinav:
Thanks, @arukiidou!

PR #1409, opened 2024-02-03
ci: Test with Go 1.22

arukiidou:
With the release of Go 1.22,
CI should run against Go 1.21 and 1.22

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,9 +16,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ["1.20.x", "1.21.x", "1.22.0-rc.2"]
+        go: ["1.21.x", "1.22.x"]
         include:
-        - go: 1.21.x
+        - go: 1.22.x
 
     steps:
     - name: Checkout code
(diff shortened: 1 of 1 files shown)


BorzdeG:
Maybe add something like ">=1.22.0-rc.1"? Then any latest version will be tested

abhinav:
Thanks for the PR, @arukiidou!
I don't expect that RC version will be merged, but I'm glad we can confirm that Zap still works as-is against 1.22.
Once 1.22 is released, we can upgrade this PR to test against 1.22 proper and merge that.

BorzdeG:
I use such a matrix to check the work in the beta version, but not blocking the work of CI

```yaml
    strategy:
      matrix:
        go: ["1.20.x", "1.21.x"]
        continueOnError: [false]
        include:
          - go: ">=1.22.0-rc.1"
            continueOnError: true
```

arukiidou:
passed in Go1.22

arukiidou:
Should these also be Go1.21 bumped?

https://github.com/uber-go/zap/blob/7db06bc9b095571d3dc3d4eebdfbe4dd9bd20405/go.mod#L3

https://github.com/uber-go/zap/blob/7db06bc9b095571d3dc3d4eebdfbe4dd9bd20405/exp/go.mod#L3

sywhang (approved):
Thanks!

> Should these also be Go1.21 bumped?
I don't think these changes should occur in this PR - we can bump it to Go 1.20 to be safe, since starting from Go 1.21 the go module version is the minimum required version for the project to work as expected which isn't the case. That can be a separate discussion, though.

arukiidou:
Thank you!
Okay, now this pr is completed.

JacobOaks (approved):

PR #1411, opened 2024-02-06
zapslog: rename Option to HandlerOption

arukiidou:
- relates #1333

Rename `Option` to `HandlerOption` to avoid future conflicts with [zap.Option](https://pkg.go.dev/go.uber.org/zap#Option).
There is no change in functionality.

### DIFF
--- a/exp/zapslog/options.go
+++ b/exp/zapslog/options.go
@@ -24,29 +24,29 @@ package zapslog
 
 import "log/slog"
 
-// An Option configures a slog Handler.
-type Option interface {
+// An HandlerOption configures a slog Handler.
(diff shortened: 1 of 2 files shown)


exp/zapslog/options.go
```diff
@@ -24,29 +24,29 @@ package zapslog
 
 import "log/slog"
 
-// An Option configures a slog Handler.
-type Option interface {
+// An HandlerOption configures a slog Handler.
```
abhinav, line 27:
```suggestion
// A HandlerOption configures a slog Handler.
```

abhinav (approved):
Thanks!

PR #1414, opened 2024-02-07
zapcore: Improve errors unit tests

justinhwang:
Updates and adds some unit tests for the zapcore/errors.go file.

### DIFF
--- a/zapcore/error.go
+++ b/zapcore/error.go
@@ -27,8 +27,12 @@ import (
 	"go.uber.org/zap/internal/pool"
 )
 
-// Encodes the given error into fields of an object. A field with the given
-// name is added for the error message.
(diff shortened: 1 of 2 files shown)


zapcore/error.go
```diff
@@ -37,6 +41,16 @@ import (
 // causer (from github.com/pkg/errors), a ${key}Causes field is added with an
 // array of objects containing the errors this error was comprised of.
 //
+// In the case of a ObjectMarshaler,
+//
+//	{
+//	  "error": {
```
justinhwang, line 47: Technically this is a breaking change as we would have objects here instead of strings but I would argue that this would be the intended behavior when someone implements `ObjectMarshaler` on their error type.

tchung1118 (approved):

Issue #1423, opened 2024-03-27
ci broken: golangci-lint changes breaks lint.


arukiidou:
**Describe the bug**

ci breaks from golangci-lint changes.

```
run golangci-lint
  Running [/home/runner/golangci-lint-1.57.1-linux-amd64/golangci-lint run --out-format=github-actions --version] in [] ...
  Error: unknown flag: --version
  Failed executing command with error: unknown flag: --version
  
  Error: golangci-lint exit with code 3
  Ran golangci-lint in 80ms
```

**To Reproduce**

Just run CI, No need to change anything,
- example: arukiidou/zap/pull/6

**Expected behavior**

**Additional context**

Removing flag will resolve it.

- test passed:  arukiidou/zap/pull/7

`golangci-lint run` no longer have `--version` flags.

```
Usage:
  golangci-lint run [flags]

Flags:
  -c, --config PATH                    Read config from file path PATH
      --no-config                      Don't read config file
  -D, --disable strings                Disable specific linter
      --disable-all                    Disable all linters
  -E, --enable strings                 Enable specific linter
      --enable-all                     Enable all linters
      --fast                           Enable only fast linters from enabled linters set (first run won't be fast)
  -p, --presets strings                Enable presets (bugs|comment|complexity|error|format|import|metalinter|module|performance|sql|style|test|unused) of linters. Run 'golangci-lint help linters' to see them. This option implies option --disable-all
      --enable-only strings            Override linters configuration section to only run the specific linter(s)
  -j, --concurrency int                Number of CPUs to use (Default: number of logical CPUs) (default 12)
      --modules-download-mode string   Modules download mode. If not empty, passed as -mod=<mode> to go tools
      --issues-exit-code int           Exit code when issues were found (default 1)
      --go string                      Targeted Go version
      --build-tags strings             Build tags
      --timeout duration               Timeout for total work (default 1m0s)
      --tests                          Analyze tests (*_test.go) (default true)
      --allow-parallel-runners         Allow multiple parallel golangci-lint instances running. If false (default) - golangci-lint acquires file lock on start.
      --allow-serial-runners           Allow multiple golangci-lint instances running, but serialize them around a lock. If false (default) - golangci-lint exits with an error if it fails to acquire file lock on start.
      --out-format string              Formats of output: colored-line-number|line-number|json|tab|checkstyle|code-climate|html|junit-xml|github-actions|teamcity (default "colored-line-number")
      --print-issued-lines             Print lines of code with issue (default true)
[...]

PR #1424, opened 2024-03-27
remove golangci-lint --version flags

arukiidou:
fixes: #1423 
remove golangci-lint --version flags, ci fixed.

- #1423

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -61,8 +61,7 @@ jobs:
     - uses: golangci/golangci-lint-action@v4
       name: Install golangci-lint
       with:
-        version: latest
-        args: --version  # make lint will run the linter
+        version: latest  # make lint will run the linter
 
     - run: make lint
       name: Lint


.github/workflows/go.yml
```diff
@@ -61,8 +61,7 @@ jobs:
     - uses: golangci/golangci-lint-action@v4
       name: Install golangci-lint
       with:
-        version: latest
-        args: --version  # make lint will run the linter
+        version: latest  # make lint will run the linter
```
abhinav, line 64: This will run the linter twice (`make lint` does it too).
We can keep the old hack in place but with a different argument:

```suggestion
        version: latest
        # Hack: Use the official action to download, but not run.
        # make lint below will handle actually running the linter.
        args: --help
```

abhinav (commented):
Thanks for catching and fixing. I've got a minor change suggestion.

abhinav (approved):

abhinav:
@r-hang @JacobOaks It looks like codecov token isn't set.

https://github.com/uber-go/zap/actions/runs/8454727528/job/23160466933?pr=1424#step:6:48

```
error - 2024-03-27 15:53:08,413 -- Upload failed: {"detail":"Tokenless has reached GitHub rate limit. Please upload using a token: https://docs.codecov.com/docs/adding-the-codecov-token. Expected available in 1886 seconds."}
```

JacobOaks (approved):
Thank you!

PR #1425, opened 2024-03-27
Feature/upgrade go mod

arukiidou:
I upgraded go version to 1.21 in go.mod.
copy of #1181

- see: #1181
- it behind of #1424

### DIFF
--- a/exp/go.mod
+++ b/exp/go.mod
@@ -1,6 +1,6 @@
 module go.uber.org/zap/exp
 
-go 1.19
+go 1.21
 
 require (
 	github.com/stretchr/testify v1.8.1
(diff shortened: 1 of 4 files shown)


arukiidou:
- passed test: https://github.com/arukiidou/zap/pull/5

JacobOaks:
Hey @arukiidou - my understanding is that the go line in a library's go.mod file now sets a requirement for all consumers of the library, so I actually think we may want to be more conservative when it comes to updating it so as to not require all consumers to upgrade, unless there's a very good reason. But I'm open to other thoughts!

arukiidou:
Yes, it will need to upgrade most of go.mods to pass the test.

### Why upgrades?
- zapslog is approaching move to core. but, zapslog requires go 1.21.
- [According to the readme](https://github.com/uber-go/zap?tab=readme-ov-file#installation), zap may only support go 1.21 or later.

- You make a good point , could postpone this for several months.

JacobOaks:
Hey @arukiidou -

* For zaplog, is not upgrading zap blocking usage (and eventual merging) of zapslog? The advantage of upgrading here is that we can get rid of some of the build tags, but my understanding is that it can still be used as long as the _consumer_ of zap is using >= 1.21.
* It's true that Zap only _officially supports_ the latest two versions, but upgrading go.mod makes it a _hard requirement_, rather than a suggestion.

If it's ok with you, I think we should probably wait on this.

arukiidou:
Ok, I understand that this is still early. 

Please let me know when the time comes.

PR #1429, opened 2024-04-11
zapcore: add warning as Level

florianl:
Some packages set their value for WarnLevel to "warning". For easier integration add "warning" as accepted WarnLevel.

Example for a package that is using "warning":
https://github.com/sirupsen/logrus/blob/dd1b4c2e81afc5c255f216a722b012ed26be57df/logrus.go#L68

### DIFF
--- a/zapcore/level.go
+++ b/zapcore/level.go
@@ -183,7 +183,7 @@ func (l *Level) unmarshalText(text []byte) bool {
 		*l = DebugLevel
 	case "info", "INFO", "": // make the zero value useful
 		*l = InfoLevel
-	case "warn", "WARN":
+	case "warn", "warning", "WARN":
 		*l = WarnLevel
 	case "error", "ERROR":
 		*l = ErrorLevel
--- a/zapcore/level_test.go
+++ b/zapcore/level_test.go
@@ -84,7 +84,6 @@ func TestLevelText(t *testing.T) {
 		err := unmarshaled.UnmarshalText(input)
 		assert.NoError(t, err, `Unexpected error unmarshaling text %q to level.`, string(input))
 		assert.Equal(t, WarnLevel, unmarshaled, `Text %q unmarshaled to an unexpected level.`, string(input))
-
 	})
 }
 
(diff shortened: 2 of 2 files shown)


florianl:
Friendly ping for feedback to @abhinav or @JacobOaks .
Thanks in advance :pray:

r-hang:
Hey @florianl, I think this change incurs some technical debt for a small amount of flexibility. The current level [MarshalText](https://github.com/uber-go/zap/blob/v1.27.0/zapcore/level.go#L160) logic is simplified because there is either a capitalized or lower case representation of the each level, where the level is represented in a consistent tense.

From my perspective, the special check case in level_test.go makes the level test code a bit more difficult to main.

florianl:
Hey @r-hang thanks for the feedback!
I agree with you, that current level [MarshalText](https://github.com/uber-go/zap/blob/v1.27.0/zapcore/level.go#L160) logic is simplified and should stay that way. 

There are cases where this packages is expected to handle input, that is generated from other packages, that are using the same levels, but slightly different level text - `warning` instead of `warn`. Currently this package is not able to pick up these slightly differences. In such cases where this package is expected to handle input with level `warning`, it will ignore this level, and will instead use `level=info` - which is not correct.
The intention of this PR is, that this package is able to handle input that originates from other sources and transparently uses the correct level.

florianl:
@r-hang what do you think about having a dedicated test for unmarshaling `warning` rather mixing it with the existing `TestLevelText`? This should decrease the maintenance level in the test you have mentioned.

florianl:
@abhinav or @JacobOaks  or @r-hang with https://github.com/uber-go/zap/pull/1429/commits/126ea63be89efd6961a7678da70689e64fabe35f I have moved the test to guarantee compatability to a dedicated sub test. Please let me know, if this works for you.

florianl:
friendly ping to @abhinav or @JacobOaks or @r-hang - do you have some feedback?

r-hang (approved):

r-hang:
Makes sense to me @florianl! Thanks for helping keep our tests more maintainable and sorry for the delay.

sywhang (approved):

PR #1444, opened 2024-06-06
recognize capitalized "warning" as a level

sywhang:
All Zap level string configurations are recognized when capitalized. For example, "INFO" gets parsed as InfoLevel.

We recently added "warning" as a config string for WarnLevel, which but did not add "WARNING".

This simply adds that for sake of consistency.


zapcore/level.go
```diff
@@ -183,7 +183,7 @@ func (l *Level) unmarshalText(text []byte) bool {
 		*l = DebugLevel
 	case "info", "INFO", "": // make the zero value useful
 		*l = InfoLevel
-	case "warn", "warning", "WARN":
+	case "warn", "warning", "WARN", "WARNING":
```
abhinav, line 186: is this actually needed? unmarshalText is called with `bytes.ToLower` at some point.

sywhang: yea i just saw that we call it two times, once with the text itself and another time with ToLower()... 

In this case I think we should just remove all the capitalized cases in unmarshalText since they're redundant.

abhinav (approved):

PR #1452, opened 2024-07-20
zaptest: add ability to filter observer logs by logger name

mauri870:
This is useful for assertions on logs in a particular named path, currently this is only possible through calls to Filter but it is more verbose.

Added a FilterLoggerName function that plays nicely with the other filters:

	zapLogs.FilterLoggerName("my.logger").FilterMessage("hello")

### DIFF
--- a/zaptest/observer/observer.go
+++ b/zaptest/observer/observer.go
@@ -91,7 +91,7 @@ func (o *ObservedLogs) FilterMessage(msg string) *ObservedLogs {
 	})
 }
 
-// FilterLoggerName filters entries to those that have the specified logger name.
+// FilterLoggerName filters entries to those logged through logger with the specified logger name.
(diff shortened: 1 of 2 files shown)


zaptest/observer/observer.go
```diff
@@ -91,6 +91,13 @@ func (o *ObservedLogs) FilterMessage(msg string) *ObservedLogs {
 	})
 }
 
+// FilterLoggerName filters entries to those that have the specified logger name.
```
sywhang, line 94:
```suggestion
// FilterLoggerName filters entries to those logged through logger with the specified logger name.
```

sywhang (commented):
This looks good to me. Thank you for the contribution!

sywhang (approved):

Issue #1458, opened 2024-08-14
Expose `DictObject`


MayCXC:
I would like to expose the ObjectMarshaler here: https://github.com/uber-go/zap/blob/v1.27.0/field.go#L425 like this:

```
func DictObject(val ...Field) zapcore.ObjectMarshaler {
    return dictObject(val)
}
```

so that field functions can be used to construct arbitrary object marshalers, to use with other functions like `Objects`:

```
zap.Objects("tuple", []zapcore.ObjectMarshaler{zap.DictObject(zap.String("foo", "bar"), zap.Int("F00", 3840)), zap.DictObject(zap.String("more", "fields"))})
```

and this would be especially nice if the signature of `Objects` were changed to use a vararg:

```
func Objects[T zapcore.ObjectMarshaler](key string, values ...T) Field
```

abhinav:
>  and this would be especially nice if the signature of `Objects` were changed to use a vararg:
>
> ```
> func Objects[T zapcore.ObjectMarshaler](key string, values ...T) Field
> ```

That would be a breaking change. We can't do that.

Also, note that even if we did, Objects wouldn't necessarily behave the way we want it to.
With the T type parameter, it's expected that all `values` are the same type.
It would be invalid to do this:

```
Objects("foo",
  DictObject(...),
  &MyObject{...},   // given func (*MyObject) MarshalLogObject(...)
)
```

This functionality is more conveniently possible with a separate function that takes a non-type-parameterized version:

```
func(name string, objs ...zapcore.ObjectMarshaler) zap.Field
```

I can't think of a good name for that function that stands apart from the other such helpers.

MayCXC:
> > and this would be especially nice if the signature of `Objects` were changed to use a vararg:
> > ```
> > func Objects[T zapcore.ObjectMarshaler](key string, values ...T) Field
> > ```
> 
> That would be a breaking change. We can't do that.
> 
> Also, note that even if we did, Objects wouldn't necessarily behave the way we want it to. With the T type parameter, it's expected that all `values` are the same type. It would be invalid to do this:
> 
> ```
> Objects("foo",
>   DictObject(...),
>   &MyObject{...},   // given func (*MyObject) MarshalLogObject(...)
> )
> ```
> 
> This functionality is more conveniently possible with a separate function that takes a non-type-parameterized version:
> 
> ```
> func(name string, objs ...zapcore.ObjectMarshaler) zap.Field
> ```
> 
[...]

PR #1460, opened 2024-08-15
Add `func DictObject`

MayCXC:
Allows fields to be used to construct `ObjectMarshaler`s that can then be used with other functions like `func Objects`. Link to issue: https://github.com/uber-go/zap/issues/1458

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -388,6 +388,60 @@ func ExampleObjects() {
 	// {"level":"debug","msg":"opening connections","addrs":[{"ip":"123.45.67.89","port":4040},{"ip":"127.0.0.1","port":4041},{"ip":"192.168.0.1","port":4042}]}
 }
 
+func ExampleDictObject() {
+	logger := zap.NewExample()
+	defer logger.Sync()
+
+	// Use DictObject to create zapcore.ObjectMarshaler implementations from Field arrays,
+	// then use the Object and Objects field constructors to turn them back into a Field.
+
+	logger.Debug("worker received job",
+		zap.Object("w1",
+			zap.DictObject(
+				zap.Int("id", 402000),
+				zap.String("description", "compress image data"),
+				zap.Int("priority", 3),
+			),
+		))
+
+	d1, err := time.ParseDuration("68ms")
+	if(err != nil) {
+		panic(err)
+	}
+	d2, err := time.ParseDuration("79ms")
+	if(err != nil) {
+		panic(err)
+	}
+	d3, err := time.ParseDuration("57ms")
+	if(err != nil) {
+		panic(err)
+	}
+
+	logger.Info("worker status checks",
+		zap.Objects("job batch enqueued",
+			[]zapcore.ObjectMarshaler{
+				zap.DictObject(
+					zap.String("worker", "w1"),
+					zap.Int("load", 419),
+					zap.Duration("latency", d1),
+				),
+				zap.DictObject(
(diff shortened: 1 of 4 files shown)


abhinav:
Thanks for the PR, @MayCXC.
This change makes sense to me.
Would you mind adding some documentation and tests for this?

MayCXC:
> Thanks for the PR, @MayCXC. This change makes sense to me. Would you mind adding some documentation and tests for this?

you got it, I added doc comments & tests.

example_test.go
```diff
@@ -388,6 +388,60 @@ func ExampleObjects() {
...
+		panic(err)
+	}
+	d2, err := time.ParseDuration("79ms")
+	if(err != nil) {
+		panic(err)
+	}
+	d3, err := time.ParseDuration("57ms")
+	if(err != nil) {
+		panic(err)
+	}
```
abhinav, line 418: can we just use time.Duration values directly here?

```suggestion
	d1 := 68 * time.Millisecond
	d2 := 79 * time.Millisecond
	d3 := 57 * time.Millisecond
```

field.go
```diff
@@ -431,6 +431,12 @@ func (d dictObject) MarshalLogObject(enc zapcore.ObjectEncoder) error {
 	return nil
 }
 
+// DictObject exposes the [zapcore.ObjectMarshaler] for an array of [Field] to
+// use with functions like [Object] and [Objects].
```
abhinav, line 435: I would reword this a bit. Object/Objects aren't the only uses of this.

```suggestion
// DictObject constructs a [zapcore.ObjectMarshaler] with the given list of fields.
// The resulting object may be used as input to [Object], [Objects],
// or anywhere else that an object is expected.
```

zapcore/field_test.go
```diff
@@ -320,6 +320,16 @@ func TestEquals(t *testing.T) {
 			b:    zap.Dict("k", zap.String("a", "d")),
 			want: false,
 		},
+		{
+			a:    zap.Object("k", zap.DictObject(zap.String("a", "b"))),
+			b:    zap.Object("k", zap.DictObject(zap.String("a", "b"))),
+			want: true,
+		},
+		{
+			a:    zap.Object("k", zap.DictObject(zap.String("a", "b"))),
```
abhinav, line 329: The example test is good, and this equality check makes sense.
Would you also mind adding a test similar to [TestDict](https://github.com/uber-go/zap/blob/3f27eb9c787ea9b2ddfc1f707ef291c9070655fe/field_test.go#L294) that asserts the shape of the resulting JSON-ey object.

example_test.go
MayCXC: yep, this looks much better

field.go
MayCXC: I reworded it a bit based on this feedback

zapcore/field_test.go
MayCXC: sounds good, I put one here: https://github.com/uber-go/zap/pull/1460/files#diff-218ab55776e43f87dc44a4b65eebeff94670cebeb5dcf80ca3143b5577b50e7dR318

abhinav (approved):
LGTM!
Will wait for an Uber maintainer to also stamp before merging.

CC @sywhang @r-hang @JacobOaks

sywhang (approved):
this looks good to me. Thanks for the PR @MayCXC !

Issue #1168, opened 2022-08-29
provides an option to turn off errorVerbose output


OldSmokeGun:
Please provide a way where we can turn off the errorVerbose output! 

When I print pkg/errors, it prints errorVerbose, which is quite confusing, since StackTrace is provided, why should it output the stack trace of error

The output of errorVerbose is rather messy and not easy to read, which should not be forced on the user!

PR #1487, opened 2025-01-30
Add an option to disable errorVerbose

### LINKED ISSUE #650: Expose the ability to disable errorVerbose via logger config
I use Zap for basically every service I write, it's great! There's one issue I have though. Since I use pkg/errors, each error satisfies the fmt.Formatter interface and thus, every error log comes with a stack trace. This is great in production when that information is vital, but during development, (i.e. with `NewDevelopmentConfig`) they still show up; which, given the less vital scenario, is a bit annoying.

I notice there's `DisableStacktrace` which removes the stack trace that originates from the log call itself but no way to disable `errorVerbose`.

In development, I usually don't need to know the full stack trace, only the top-most element (or, the top-most `errors.New` call). And if I do, in a perfect world, I could flip a env var to enable it!

makotonakai:
This PR adds `disableVerbose` option to `Logger` type.

This option allows users to disable the output of error verbose when they use pkg/errors, which enables simpler logging and provides clearer understanding of the current situation in their applications, especially in their production environments.

Closes #650 #1168

### DIFF
--- a/zapcore/field.go
+++ b/zapcore/field.go
@@ -102,11 +102,12 @@ const (
 // context. Most fields are lazily marshaled, so it's inexpensive to add fields
 // to disabled debug-level log statements.
 type Field struct {
-	Key       string
-	Type      FieldType
-	Integer   int64
-	String    string
-	Interface interface{}
+	Key            string
+	Type           FieldType
+	Integer        int64
+	String         string
+	Interface      interface{}
+	DisableVerbose bool
 }
 
 // AddTo exports a field through the ObjectEncoder interface. It's primarily
@@ -173,7 +174,7 @@ func (f Field) AddTo(enc ObjectEncoder) {
 	case StringerType:
 		err = encodeStringer(f.Key, f.Interface, enc)
 	case ErrorType:
-		err = encodeError(f.Key, f.Interface.(error), enc)
+		err = encodeError(f.Key, f.Interface.(error), enc, f.DisableVerbose)
 	case SkipType:
 		break
 	default:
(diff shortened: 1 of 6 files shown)


zapcore/field.go
```diff
@@ -102,11 +102,12 @@ const (
...
-	Type      FieldType
-	Integer   int64
-	String    string
-	Interface interface{}
+	Key            string
+	Type           FieldType
+	Integer        int64
+	String         string
+	Interface      interface{}
+	DisableErrorVerbose bool
```
prashantv, line 110: We avoid adding new struct fields to the `Field` type for performance reasons, as the struct is passed as a value.

We could add a new `FieldType` to indicate an error that should only have the message included, though that could have backwards compatibility concerns with some encoders.

makotonakai: @prashantv 

I appreciate your feedback. I will create a new type with error field and DisableErrorVerbose field.

Amphaal:
Also eager to see this merged ! Many thanks in advance for your consideration :)

prashantv:
Here's a simple test that shows that this approach does not work in common cases:
```go
func TestDisableErrorVerbose(t *testing.T) {
	err := errTooManyUsers(4) // custom error with %+v formatting.

	logger := zap.NewExample(zap.DisableErrorVerbose())
	logger.Error("test error log", zap.Error(err))
	logger.Info("test info log", zap.Error(err))
	if ce := logger.Check(zap.InfoLevel, "test check"); ce != nil {
		ce.Write(zap.Error(err))
	}
}
```

This outputs:
```
{"level":"error","msg":"test error log","test error log":"4 too many users"}
{"level":"info","msg":"test info log","error":"4 too many users","errorVerbose":"4 too many users verbose"}
{"level":"info","msg":"test check","error":"4 too many users","errorVerbose":"4 too many users verbose"}
```

Though the `Error` method is handled, other methods are not, and the path for `logger.Check` can't be handled with this approach.

This approach also has some performance implications as it iterates over every field and checks the type, though it only impacts the `DisableErrorVerbose` case.

Before a code review, I'd recommend discussing the design on the issue to make sure it will solve the problem fully (and ideally in a performant way).

Issue #1500, opened 2025-05-27
zap.Object panics with nil or a nil interface


alshopov:
**Describe the bug**
zap.Object("test", nil) panics during runtime

**To Reproduce**
either: 
```
zap.Object("test", nil)
```
or
```
var m zapcore.ObjectMarshaler
logger.Info("object", zap.Object("test", m))
```

**Expected behavior**

The logging framework should not panic. Compare this with the tests in [TestFieldAddingError](https://github.com/uber-go/zap/blob/9aa9ce1300672f39f708913a259393d316fb2a8d/zapcore/field_test.go#L105): Special effort is made to prevent common cases where there are panics and zap successfully logs what happens

**Additional context**
As the bare minimum - this behavior should be pointed out. `zap.Object` gets suggested by IDEs too often and it is easy to use it instead of `zap.Any`

Note:
A possible improvement is coming.

alshopov:
This issue has been fixed via #1501

PR #1501, opened 2025-05-27
Prevent zap.Object from panicing on nils

alshopov:
It is possible to call zap.Object with nil or nil interface This will cause panic in zap bringing down the app using it

The standard way of handling this is via nilField - compare all the constructors on pointers. As interfaces are similar to pointers - lets handle this case the same way

Add tests in zapcore for equality on zap.Object(nil) Show ReflectType does not panic on nils
Add test that zap.Object(nil) works

Update benchmarks and use such a field to show performance and allocation are not affected excessively

Refs #1500

### DIFF
--- a/benchmarks/zap_test.go
+++ b/benchmarks/zap_test.go
@@ -134,6 +134,7 @@ func fakeFields() []zap.Field {
 		zap.Times("times", _tenTimes),
 		zap.Object("user1", _oneUser),
 		zap.Object("user2", _oneUser),
+		zap.Object("user3", nil),
 		zap.Array("users", _tenUsers),
 		zap.Error(errExample),
 	}
--- a/field.go
+++ b/field.go
@@ -398,6 +398,9 @@ func Durationp(key string, val *time.Duration) Field {
 // struct-like user-defined types to the logging context. The struct's
 // MarshalLogObject method is called lazily.
 func Object(key string, val zapcore.ObjectMarshaler) Field {
+	if val == nil {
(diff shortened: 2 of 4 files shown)


JacobOaks (approved):

alshopov:
@JacobOaks , @tchung1118 - rebase done, all lint checks run and green

prashantv:
> What I solved is the case of a missing value in the interface.

To make sure we're using the same terminology, I'm using it as described [here](https://go.dev/tour/methods/11):
> interface values can be thought of as a tuple of a value and a concrete type

The check added here only works when the **type** is nil (and when that's the case, the value is also nil).

It does not help if the type is non-nil, but the value is nil, since it will try to make a method call.

> What you point out is a `MarshalLogObject` method with a bug. It does nil dereference without check and the code rightfully panics.

Even if `MarshalLogOjbect` is on a value type (where it can't do a nil dereference), you can still cause a panic by passing a `nil` value:
https://go.dev/play/p/YLIJxL91nrq

```go
type user struct{}

func (u user) MarshalLogObject(enc zapcore.ObjectEncoder) error {
	enc.AddString("foo", "bar")
	return nil
}

func main() {
	logger := zap.NewExample()
	defer logger.Sync()

	var u *user
	logger.Info("proper coded nil object works", zap.Object("nilUser", u))
}
```

The solution used for stringer will also help here, but I wanted to call out that the solution here helps in limited cases, and there's other `nil` value cases for which #1500 is not solved.

Issue #1504, opened 2025-06-03
Badly coded MarshalLogObject can panic zap


alshopov:
**Describe the bug**
A panic in implementer of zapcore.ObjectMarshaler (i.e. badly coded `MarshalLogObject(ObjectEncoder) error`) causes zap to panic which causes the app using it to panic

**To Reproduce**
Check this playgound:  https://go.dev/play/p/HqAKQ6nGxgy 

Here is the important part:
```
type user struct{}

func (u user) MarshalLogObject(enc zapcore.ObjectEncoder) error {
	panic("Catch me if you can")
}
```

**Expected behavior**
zap should not panic but make a good effort to signal this problem

**Additional context**
Check similar cases that are guarded against panics in (encodeStringer](https://github.com/uber-go/zap/blob/master/zapcore/field.go#L214)
Also see discussion in this thread: https://github.com/uber-go/zap/pull/1501#issuecomment-2933659745

------

This problem also shows in zap.Object, zap.Inline, zap.Array, zap.Dict (if one of the fields has the problem) - basically any field for which we call a method on the logged. This is similar to the workaround for Stringer. Note that there are similar cases for zap.Error when not all cases are covered - for example a panic with an object  that causes a panic itself
when being converted to a string

As a minimum we need some coverage for all these cases or refactoring in a way that shows that we reuse enough of the implementation.

PR #1505, opened 2025-07-06
Safer behaviour of (mis)logging badly coded marshallers

alshopov:
ArrayMarshalerType, ObjectMarshalerType, InlineMarshalerType all call
user provided methods on user provided values.
They may fail/panic and have to be guarded.

Use guard similar to the one for stringer with the following caveats:
- ObjectMarshalerType in case of nil-ness will be logged as string rather than map.
- ArrayMarshalerType - we have to insert a synthetic empty marshaller to log an empty slice.
   Otherwise the default shape of map shines through
- InlineMarshalerType - in case of any error it will be de-inlined to show exact cause of error.
  Otherwise error will be attached to the wrapping object which will make debugging and fixing harder

While we introduce enough coverage, the test harness is not using currently subtests. 
This patch does not address this to keep changes to a minimum and give time to discuss if this is necessary.
Thus there are 3 tests that were changed - the error cases, the success cases and inline case (separate).


alshopov:
@JacobOaks, @tchung1118: as discussed on #1501, I am proposing a fix for #1504.
The commit messages should outline decisions made in implementing this. Have a look.

zapcore/field.go
```diff
@@ -231,3 +232,70 @@ func encodeStringer(key string, stringer interface{}, enc ObjectEncoder) (retErr
...
+	// the MarshalLogObject() method, similar to https://golang.org/src/fmt/print.go#L540
+	defer func() {
+		if err := recover(); err != nil {
+			// If it's a nil pointer, just say "<nil>". The likeliest causes are a
+			// MarshalLogObject that fails to guard against nil or a nil pointer for a
+			// value receiver, and in either case, "<nil>" is a nice result.
+			if v := reflect.ValueOf(objectMarshaller); v.Kind() == reflect.Ptr && v.IsNil() {
+				enc.AddString(key, "<nil>")
+				return
+			}
```
tchung1118, line 247: It doesn't seem necessary to log nil here separately because the error returned from here will already indicate there's a nil panic without this. Let's not do this because reflection is expensive.

tchung1118 (commented):
on behalf of @sywhang from group code review

Issue #1516, opened 2025-10-27
Improve memory consumption of the log sampler


jm-franc:
Hi!

The [zapcore/sampler.go](https://github.com/uber-go/zap/blob/07077a697f639389cc998ff91b8885feb25f520d/zapcore/sampler.go) allocates ~0.5MB per instance. 

That's quite a bit considering that some binaries instantiate quite a few. For example, [OpenTelemetry collectors monitoring Kubernetes cluters](https://opentelemetry.io/docs/platforms/kubernetes/collector/) can instantiate tens of thousands of those.

Here I suggest an update that divides the memory consumption by 7 and improves performance, with very little drawback.

### Details

The sampler allocates `numberOfLogLevels * 4096` instances of the `counter` structure (128 bits). There are [7 different log levels](https://github.com/uber-go/zap/blob/master/zapcore/level.go#L34-L55). Once an log message is emitted, it's hashed to one of those 7*4096 buckets.

In practice, all 7 log levels are not used uniformly (far from it), and some (e.g. DPanic, Panic, Fatal) re barely used at all.

I thus propose to only instantiate 4096 instances of the `counter` structure (dividing the memory footprint in 7) and include the log level as an input to the hash function. Since the 7 log levels are use very unevenly, this only marginally impact the risks of collision while improving cache locality.

jm-franc:
Indeed hopefully _completely unused_ log levels hopefully make it to virtual memory. The number of entries of _not completely unused_ log levels is highly unbalanced though (info >> warning >> error): I have measured significant memory savings in large Kubernetes clusters using OpenTelemetry collectors thanks to this PR.

I didn't touch the number of buckets as I thought it's difficult to know which value fits every use case, but this PR mechanically leads to a bit less buckets (since they are now shared for every log level).

Considering this, do you think this change is a net improvement?

PR #1517, opened 2025-10-28
Reduce memory footprint.

jm-franc:
Divides the memory footprint of a sampler instance by 7 (from ~0.5MB to 65KB), at the expense of a small increase in over-sampling chances.

https://github.com/uber-go/zap/issues/1516

### DIFF
--- a/zapcore/sampler.go
+++ b/zapcore/sampler.go
@@ -26,8 +26,7 @@ import (
 )
 
 const (
-	_numLevels        = _maxLevel - _minLevel + 1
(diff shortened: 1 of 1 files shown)


prashantv:
[Sharing my comment here too](https://github.com/uber-go/zap/issues/1516#issuecomment-3540198218)

> I think this is a good idea, though I'd prefer if it was opt-in, maybe as a setting on the SamplingConfig.

My worry is that a version upgrade changing the sampling behaviour significantly could be surprising.

PR #1546, opened 2026-04-23
zapgrpc: fix V() to use verbosity, not severity

### LINKED ISSUE #1544: zapgrpc: verbosity level check is incorrect
grpclog has two concepts of log level: severity level, and verbosity level.
Severity level is analogous to the zap logger level, verbosity is supplemental.

The [LoggerV2.V](https://pkg.go.dev/google.golang.org/grpc@v1.80.0/grpclog/internal#LoggerV2) is documented as follows:

```go
	// V reports whether verbosity level l is at least the requested verbose level.
	V(l int) bool
```

The zapgrpc implementation confuses verbosity level for severity level: https://github.com/uber-go/zap/blob/0ab0d5aae5986395e2ca497385d977ccd7cdfc5e/zapgrpc/zapgrpc.go#L234-L237

alliasgher:
## What this changes

`grpclog.LoggerV2.V` is documented as:

> V reports whether verbosity level l is at least the requested verbose level.

`zapgrpc.Logger.V` was instead consulting a mapped zap severity (`grpcLvlInfo → InfoLevel`, `grpcLvlWarn → WarnLevel`, …) via `levelEnabler.Enabled`. Callers gating verbose traces on `V(l)` either saw them always on or always off depending on zap's core level, independent of the verbosity they requested.

This PR:

- Adds a `verbosity` field on `Logger`.
- Adds a `WithVerbosity(int)` option to configure it at construction time (default 0, matching grpclog's default).
- Rewrites `V(level int) bool` to `return level <= l.verbosity`.
- Rewrites `TestLoggerV` against the new semantics (table-driven across verbosity 0/1/3) and adds `TestLoggerVDefaultVerbosity` confirming default behaviour is independent of zap's severity.

## Behaviour note

This is a behaviour change. Callers that relied on the previous (severity-based) behaviour will need to pass `WithVerbosity` at construction time to opt back into chatty V-logging. Happy to switch to an opt-in flag (e.g. `WithStrictVerbosity()`) if the maintainers prefer to preserve the existing default, but the issue description and grpclog docs both suggest the current behaviour is incorrect.

Closes #1544

### DIFF
--- a/zapgrpc/zapgrpc.go
+++ b/zapgrpc/zapgrpc.go
@@ -72,6 +72,16 @@ func WithDebug() Option {
 	})
 }
 
+// WithVerbosity configures the Logger's verbosity level. V(l) returns true
+// whenever l is less than or equal to the configured verbosity, matching
+// grpclog.LoggerV2's semantics. The default is 0, so V(l) returns true only
+// for V(0). Pass a higher value to allow callers to emit more verbose logs.
+func WithVerbosity(verbosity int) Option {
+	return optionFunc(func(logger *Logger) {
+		logger.verbosity = verbosity
+	})
+}
+
 // withWarn redirects the fatal level to the warn level, which makes testing
 // easier. This is intentionally unexported.
 func withWarn() Option {
@@ -140,6 +150,7 @@ type Logger struct {
 	levelEnabler zapcore.LevelEnabler
 	print        *printer
 	fatal        *printer
+	verbosity    int
 	// printToDebug bool
 	// fatalToWarn  bool
 }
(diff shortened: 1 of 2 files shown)


sywhang:
thanks for the PR - unfortunately the opt-out behavior change is not something we would want. Could you modify this so that it's an opt-in flag instead?

alliasgher:
Thanks @sywhang, that is fair. Made it opt-in in 4ad8f73.

`V` now keeps the existing severity-based behaviour by default, so nothing changes for current users, and only switches to grpclog's documented verbosity semantics when `WithVerbosity` is passed:

```go
func (l *Logger) V(level int) bool {
	if l.useVerbosity {
		return level <= l.verbosity
	}
	return l.levelEnabler.Enabled(_grpcToZapLevel[level])
}
```

`WithVerbosity` sets the flag as well as the level, so the opt-in is explicit rather than inferred from a zero value.

Tests cover both paths: `TestLoggerVDefaultKeepsSeverityBehaviour` pins the default to the core's level (`V(grpcLvlInfo)` false at `ErrorLevel`, true at `DebugLevel`), and `TestLoggerVOptInIsIndependentOfCoreLevel` shows the opt-in path answers from the requested verbosity regardless of the core level. The original `TestLoggerV` table is unchanged and still exercises the verbosity semantics.

The docs on both `V` and `WithVerbosity` now spell out which behaviour you get and why the correct one is opt-in.

PR #1554, opened 2026-06-13
fix: encode NaN/Inf floats as null in JSON encoder

### LINKED ISSUE #690: float NaN are encoded as string "NaN"
float NaN, +Inf, -Inf are encoded as corresponding string, which is inconsistent with json.Marshal in golang stdlib. json.Marshal will return error if float value can not be encoded in json. 
For the following code
```golang
type C struct {
    Field float64 `json:"field"`
}

func (c C) MarshalLogObject(enc zapcore.ObjectEncoder) error {
    enc.AddFloat64("field", c.Field)
    return nil
}

c := C{
    Field: math.NaN(),
}
logger.Error("reflection", zap.Reflect("c", c))
logger.Error("marshall", zap.Object("c", c))
```
The two logging statement will output different values
```
{"msg":"reflection","cError":"json: unsupported value: NaN"}
{"msg":"marshall","c":{"field":"NaN"}}
```
Output string for float value is undesired, since they are different types.

MD-Mushfiqur123:
## Problem

The JSON encoder currently encodes NaN, +Inf, and -Inf as quoted strings (`"NaN"`, `"+Inf"`, `"-Inf"`), producing **invalid JSON**. This is inconsistent with Go's `encoding/json`, which returns an `UnsupportedValueError` for these values.

This means:
1. `zap.Reflect()` with a struct containing NaN returns an error, but `enc.AddFloat64()` silently produces a string — inconsistent behavior within zap itself.
2. The output is not valid JSON.
3. Consumers parsing zap's JSON output cannot distinguish between a legitimate string field and a mis-encoded float.

## Fix

Encode NaN, +Inf, and -Inf as `null` in the JSON encoder. `null` is valid JSON and the most common representation for non-finite numbers in JSON serializers.

## Changes
- `zapcore/json_encoder.go`: Change `appendFloat()` to emit `null` instead of `"NaN"`/`"+Inf"`/`"-Inf"`
- `zapcore/json_encoder_impl_test.go`: Update test expectations for NaN/Inf float64 and float32 cases

Fixes #690


justinhwang:
Closing as duplicate of #1553

PR #1557, opened 2026-06-26
Add context-aware logging methods (Context option + *Ctx methods)

### LINKED ISSUE #1018: Support context's values logging
We want to use the values such as CorrelationId, RequestPath, etc. of each incoming request in web service projects, when we trigger log logEvent anywhere in the application. Because we need these values to observe in what case the log event is thrown. For this, we do not want to use zap.WithField in the middleware and create a new logger to use values such as CorrelationId, RequestPath etc. and move them within the application with the context. This both corrects the cost of creating a new instance and we will need to do reflection in the places we will use for the logger object we carry in the context.

FournyP:
## Summary

Adds context-aware logging to zap: a new `Context` option plus `*Ctx` logging methods that extract fields from a `context.Context` at the log site, without having to build a new `With(...)` logger at every middleware boundary.

This revives the long-inactive #1019 (open since 2021, now conflicting with `master`) and addresses #1018. It is re-implemented against current `master`, fixes a latent bug in the original, broadens the surface to the methods added since 2021, and adds benchmarks.

```go
logger, _ := zap.NewProduction(zap.Context(func(ctx context.Context) []zap.Field {
    if id, ok := ctx.Value(correlationIDKey).(string); ok {
        return []zap.Field{zap.String("correlation-id", id)}
    }
    return nil
}))

logger.InfoCtx(ctx, "request handled", zap.Int("status", 200))
// {"level":"info","msg":"request handled","correlation-id":"...","status":200}
```

## What's included

- **`options.go`** — `Context(func(context.Context) []Field) Option` and a `contextFunc` field on `Logger`.
- **`Logger`** — `LogCtx`, `DebugCtx`, `InfoCtx`, `WarnCtx`, `ErrorCtx`, `DPanicCtx`, `PanicCtx`, `FatalCtx`.
- **`SugaredLogger`** — full symmetry with the existing surface across all four styles and every level: print (`InfoCtx`, …), `w` (`InfowCtx`, …), `f` (`InfofCtx`, …), `ln` (`InfolnCtx`, …), plus `Log*Ctx` for each style.
- Each `*Ctx` method is placed directly next to its non-context sibling.
- Tests (`context_test.go`) and benchmarks (`context_bench_test.go`).

Context fields are prepended to any fields passed at the log site. The change is purely additive — no existing signatures change.

## Difference from #1019

- Re-based on current `master` (no conflicts; covers `Log`/`Logw`/`Logf`/`Logln` which didn't exist in 2021).
- **Bug fix:** #1019's `generateFields` returned `nil` whenever no `Context` option was set or `ctx` was nil, which silently dropped the explicit log-site fields. Here the log-site fields are always preserved; context fields are prepended only when both a `Context` option and a non-nil `ctx` are present. A regression test covers this.
- Extends the same pattern to `SugaredLogger` and all log levels/styles.

## Benchmarks

`go test -bench BenchmarkLoggerContext -benchmem` (Apple M3 Pro, arm64). Three configs per style: existing baseline, `*Ctx` with no `Context` option (nil-check fast path), and `*Ctx` extracting one field.

```
Logger  Info/Baseline            48.3 ns/op    64 B/op   1 allocs/op
Logger  InfoCtx/NoOption         49.7 ns/op    64 B/op   1 allocs/op
Logger  InfoCtx/WithOption      111.1 ns/op   257 B/op   3 allocs/op
Sugar   Print/Baseline           28.1 ns/op     0 B/op   0 allocs/op
Sugar   PrintCtx/NoOption        32.1 ns/op     0 B/op   0 allocs/op
Sugar   PrintCtx/WithOption     107.6 ns/op   209 B/op   4 allocs/op
Sugar   Infow/Baseline           80.9 ns/op   128 B/op   1 allocs/op
Sugar   InfowCtx/WithOption     156.7 ns/op   386 B/op   5 allocs/op
```

When no `Context` option is configured, the overhead over the non-context method is negligible and allocation-free. The extra cost in the `WithOption` case is the user's `contextFunc` producing fields plus the prepend — inherent to the feature.

## Testing

- `make lint` — clean
- `go test -race ./...` — passing
- New tests cover every method family, the prepend ordering, the panic/fatal paths, and the field-preservation regression.

Closes #1018.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+Enhancements:
+* [#1557][]: Add a `Context` option and context-aware `*Ctx` logging methods (`InfoCtx`, `ErrorCtx`, etc. on `Logger`; `InfowCtx`, `InfofCtx`, etc. on `SugaredLogger`) that extract fields from a `context.Context`.
+
+[#1557]: https://github.com/uber-go/zap/pull/1557
+
 ## 1.28.0 (27 Apr 2026)
 Enhancements:
 * [#1534][]: Add `zapcore.CheckPreWriteHook` and `CheckedEntry.Before` method for transforming entries before they are written to any Cores.
--- a/context_bench_test.go
+++ b/context_bench_test.go
@@ -1,0 +1,143 @@
+// Copyright (c) 2026 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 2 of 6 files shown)


jalaziz (commented):
Really excited about this as it can really simplify some things.

It would also be great if the context could be available to hooks and cores too.

This may be entirely out of scope for this PR, but when I last looked into this it _seemed_ that context could be added to `Entry` without breaking the existing interface. The new `*Ctx` methods could pass in the actual `context` and the original non-ctx methods could pass in `context.TODO()`.

FournyP:
Hi @JacobOaks, @sywhang

Can you consider reviewing this MR? I would love to stick with zap in my projects but the lack of context awareness is a problem.

sywhang:
thanks for the PR! Just letting you know that I have a large backlog of reviews that I'll try to get to this week, but this is on top of my review stack.

PR #1565, opened 2026-07-27
BUGFIX: Suppress "failed to get caller" diagnostic for std-log bridges

inteon:
The standard-library bridges (NewStdLog, NewStdLogAt, RedirectStdLog) bake in a fixed AddCallerSkip that only resolves a caller for messages routed through log.Output -> loggerWriter.Write. A direct write to the returned io.Writer (e.g. via log.Writer()) arrives on a shallower stack where no caller can be resolved, causing check to spam the error output with "failed to get caller".

Add an internal suppressCallerFailure option and set it on the loggers backing these bridges so the expected caller-resolution failure stays silent.


inteon:
@sywhang do you have time to leave a review?

PR #1567, opened 2026-08-05
fix: set content type for AtomicLevel HTTP responses

fudongyingluck:
`AtomicLevel.ServeHTTP` returns JSON responses, but does not explicitly set`Content-Type`. As a result, `net/http` detects these responses as`text/plain; charset=utf-8`.

This change:

- sets `Content-Type: application/json` for normal and business error responses;
- sets `Content-Type: text/plain; charset=utf-8` for the fallback internal error;
- adds tests covering both response types.

Response bodies and status codes are unchanged.

Thank you for your time. Any feedback is welcome.


sywhang (approved):
lgtm. Thank you for the contribution.

PR #1569, opened 2026-08-20
Fix Field.Equals panic on uncomparable Stringer and Inline values

nileshpatil6:
## The bug

`Field.Equals` ends in a struct comparison:

```go
switch f.Type {
case BinaryType, ByteStringType:
	return bytes.Equal(f.Interface.([]byte), other.Interface.([]byte))
case ArrayMarshalerType, ObjectMarshalerType, ErrorType, ReflectType:
	return reflect.DeepEqual(f.Interface, other.Interface)
default:
	return f == other
}
```

`StringerType` and `InlineMarshalerType` both land in `default`, and both store an arbitrary user value in `Field.Interface`. If that value is backed by a slice, map or func, `f == other` panics.

Every other `Interface`-carrying type is already routed to `reflect.DeepEqual`. `InlineMarshalerType` was introduced in #912 (2021) four years after the switch was written in #444 (2017), and was never added to the list. `StringerType` predates `Equals` and was missed from the start.

## It is reachable from the public API

`zaptest/observer`'s `FilterField` is the documented way to assert on logged fields, and it calls `Equals` on every context field:

```go
core, logs := observer.New(zapcore.DebugLevel)
log := zap.New(core)
log.Info("hello", zap.Stringer("tags", tags{"a", "b"}))

logs.FilterField(zap.Stringer("tags", tags{"a", "b"}))
```

where `tags` is `[]string` with a `String()` method. That panics:

```
panic: runtime error: comparing uncomparable type observer_test.tags
	type:.eq.go.uber.org/zap/zapcore.Field
	go.uber.org/zap/zapcore.Field.Equals            zapcore/field.go:204
	(*ObservedLogs).FilterField.func1               zaptest/observer/observer.go:112
	(*ObservedLogs).Filter                          zaptest/observer/observer.go:140
```

`zap.Inline` on a map- or slice-backed `ObjectMarshaler` panics the same way.

## The fix

Add `InlineMarshalerType` and `StringerType` to the existing `reflect.DeepEqual` case, so all six `Interface`-carrying types are handled consistently.

## Tests

Four cases added to the existing `TestEquals` table using a slice-backed type that implements both `fmt.Stringer` and `ObjectMarshaler`, covering equal and unequal values for `zap.Stringer` and `zap.Inline`.

Without the fix `TestEquals` panics with `comparing uncomparable type zapcore_test.userNames`; with it, it passes. `go build ./...` and `go test ./...` are clean apart from pre-existing Windows-only failures in `TestConfig`, `TestOpen`, `TestOpenOtherErrors` and `TestIOCore`, which fail identically on an unmodified checkout here.


fallenmi (approved):
Reviewed exact head 814a4582a7fe7f1a9e793558826fd4bb0f11f8af with Codex.

APPROVE — this closes the panic for both Interface-carrying field types without changing the specialized byte-slice path. On exact base bbd4ecbd8760678fdbac94c93d9aca8069c83b2f, an independent four-case Stringer/Inline oracle panics on the first slice-backed value. On this head, all four equal/unequal cases pass 20/20; the existing TestEquals table passes 20/20; and the full go test ./... suite is green. git diff --check is also clean.

PR #1572, opened 2026-08-25
feat: add CheckedEntry.Free to return entries to the pool without writing it

justinhwang:
**What**
Adds (*CheckedEntry).Free(), which returns a CheckedEntry to zap's internal pool without writing it.

**Why**
A Core that wants to filter on field values (e.g. drop zap.Error(err) logs for known-benign errors) must override Check so that it-not the core it wraps-gets added to the CheckedEntry, since fields only arrive in Write. To decide whether the downstream chain (crucially, the sampler) actually wants the entry, such a core can probe it:

```go
func (c *core) Check(ent zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
    probe := c.Core.Check(ent, nil) // consults the inner chain, incl. the sampler
    if probe == nil {
        return ce // inner chain declined
    }
    probe.Free() // <- previously impossible
    return ce.AddCore(ent, c)
}
```

- Before this change there was no way to hand that probe back to the pool: putCheckedEntry is unexported and only reachable via Write. Dropping the probe on the floor collapses the sync.Pool hit rate, allocating a fresh CheckedEntry plus its 4-element []Core backing array on every logged line.

- Reuses the existing dirty best-effort misuse detection, so double-free / use-after-Free is guarded the same way double-Write already is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+Enhancements:
+* [#1572][]: Add `CheckedEntry.Free` method to return a checked entry to the pool without writing it.
+
+[#1572]: https://github.com/uber-go/zap/pull/1572
+
 ## 1.28.0 (27 Apr 2026)
 Enhancements:
 * [#1534][]: Add `zapcore.CheckPreWriteHook` and `CheckedEntry.Before` method for transforming entries before they are written to any Cores.
--- a/zapcore/entry.go
+++ b/zapcore/entry.go
@@ -292,6 +292,31 @@ func (ce *CheckedEntry) Write(fields ...Field) {
 	putCheckedEntry(ce)
 }
 
+// Free returns the CheckedEntry to the internal pool without writing it.
+//
+// Free is intended for Core implementations that obtain a CheckedEntry by
+// probing a wrapped Core — calling its Check method with a nil CheckedEntry —
+// and then decide not to log the entry. Returning the unused CheckedEntry to
+// the pool with Free keeps the pool effective; dropping it on the floor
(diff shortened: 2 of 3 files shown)


JacobOaks (commented):
Hey @justinhwang thanks for the contribution!

I think the perf justification for adding this could make sense, but I'm not sure how good of an idea it would be to expose an entrypoint into zap's internal pooling mechanisms like this so I wanted to dig a little more into your use case and try to understand better how you're using Zap.

You mentioned sampling as an example. If we were using a sampler core, is there a reason you'd structure that to be wrapped underneath other cores rather than an outer core? If it were an outer core, it could avoid calling `.Check` on inner cores entirely when making its sampling decision, rather than having this situation where inner cores are deciding to add to an entry only to have an outer core try to back away from it.

You also mention removing certain errors from logs. This again seems like something that could be an outer wrapper instead to avoid the same "reverse back-away" problem. Or, we could also use standard library errors methods `errors.Is` or `errors.As` to avoid the invocation of the logger entirely if we already have some set of known errors we want to avoid logging. Any reason this wouldn't work well for you?

Maybe @moisesvega or @sywhang would also have thoughts here.

justinhwang:
So we're seeing an increase of usage of request hedging--see https://research.google/pubs/the-tail-at-scale/, when an initial request takes over X ms, we send off a second request. The slower of the two gets canceled. What this means for my service is that we're seeing an uptick in context canceled / deadline exceeded. Thus far, we've handled this by:
- Ignoring those errors at individual callsites
- Ignoring it within logging frameworks (like CFF's instrumentation)

This is not very scalable and we miss a lot of logs and just have to put up more changes to catch them. Ideally we could put in some sort of sampler, when `zap.Error(canceled || deadlineExceeded)` at a service-level instead of at individual callsites. This means calling `Check` as we need to inspect the fields before we can decide to sample. 

Will defer to you guys @JacobOaks @moisesvega on what the best way to achieve that is

PR #1588, opened 2026-09-09
feat(zapcore): add AppendNumber for full-precision epoch time/duration

huww98:
## Summary

`EpochTimeEncoder`, `EpochMillisTimeEncoder`, and `SecondsDurationEncoder` convert through `float64`, which silently loses low nanosecond digits for any value past `2^53` ns — i.e. for essentially every real-world timestamp (anything after ~104 days past the Unix epoch). This PR adds `buffer.AppendNumber(n int64, scale int)`, which formats a signed integer scaled by `10^-scale` as an exact decimal computed from the integer, and wires the three encoders to use it. The output stays a JSON number, so existing float parsers keep working; a capable parser now reads all the significant digits.

## The problem

```go
nanos := t.UnixNano()           // 1735689600123456789
sec := float64(nanos) / float64(time.Second)
//   = 1735689600.1234567  ← last digits corrupted by float64 rounding
```

`float64` can only represent integers exactly up to `2^53 ≈ 9.0e15` ns (~104 days of epoch time). Beyond that, the low bits of `nanos` are rounded off before the division, so the sub-second digits emitted by these encoders are wrong. This was reported in #275 back in 2017; at the time it was resolved by *adding* `RFC3339NanoTimeEncoder` (string) and `EpochNanosTimeEncoder` (int64 nanos) as full-precision alternatives. This change fixes the float64 encoders directly.

## Design

- **New primitive:** `buffer.AppendNumber(n int64, scale int)` writes `n × 10^-scale` as an exact decimal. It writes the magnitude via `strconv.AppendInt`, then inserts the decimal point in place (`copy` is overlap-safe per the `builtin.copy` contract), and strips trailing fractional zeros. Zero allocations on the common path.
- **Non-breaking optional interface:** `appendNumberEncoder` (mirroring the existing `appendTimeEncoder`/`AppendTimeLayout` pattern from #786). `jsonEncoder` implements it; the memory/test encoder does not, so it uses the `AppendFloat64` fallback. External `PrimitiveArrayEncoder` implementations are unaffected — they get the float64 fallback, identical to today.
- **Minimal-form output:** trailing fractional zeros are stripped, matching `strconv`'s `'f', -1` float formatting. Consequently, values that `float64` already represented exactly keep their bytes (`"0"`, `"1008720000"`, `"1008720000000"`); only previously-*lossy* timestamps change (they gain their missing significant digits). So this is a precision fix, not a format change for clean values.

## Behavior

| input (nanos) | scale | old (float64) | new (AppendNumber) |
|---|---|---|---|
| `0` | 9 | `0` | `0` |
| `1008720000000000000` | 9 | `1008720000` | `1008720000` |
| `1735689600123456789` | 9 | `1735689600.1234567` | `1735689600.123456789` |
| `1735689600123456789` | 6 | `1735689600123.4568` | `1735689600123.456789` |

The first two rows are unchanged; the last two recover the lost digits.

### In context — `EpochMillisTimeEncoder` via `EncodeEntry`, same timestamp

The "old" column was produced by checking out `zapcore/encoder.go` from the parent commit (pure float64) while keeping this branch's benchmark file.

```
goos: darwin
goarch: arm64
pkg: go.uber.org/zap/zapcore
cpu: Apple M2 Pro
                                       │   old.txt    │               new.txt               │
                                       │    sec/op    │   sec/op     vs base                │
EpochMillisTimeEncoder/float64-12         116.1n ± 1%   116.6n ± 2%        ~ (p=0.128 n=10)
EpochMillisTimeEncoder/AppendNumber-12   116.65n ± 1%   99.44n ± 1%  -14.75% (p=0.000 n=10)
geomean                                   116.4n        107.7n        -7.47%
```

### Isolated primitive — `buffer.AppendNumber` vs `AppendFloat`

`AppendFloat` (the float64 path, unchanged by this PR) vs `AppendNumber` on the same input values. All paths are 0 B/op and 0 allocs/op. `AppendNumber` is faster in every case.

```
BenchmarkAppendNumber/subOne/scale-12           96812282                11.97 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/subOne/float-12           44059466                27.05 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/negativeSubOne/scale-12           99220358                11.96 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/negativeSubOne/float-12           43696346                27.59 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/fullMillis/scale-12               58202042                20.10 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/fullMillis/float-12               35877031                33.14 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/fullSeconds/scale-12              58790746                19.84 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/fullSeconds/float-12              33487358                35.29 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/trailingZeros/scale-12            59279254                19.88 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/trailingZeros/float-12            40252302                30.27 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/negativeScale/scale-12            184328799                6.508 ns/op           0 B/op          0 allocs/op
BenchmarkAppendNumber/negativeScale/float-12            52012316                23.40 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/largeScale/scale-12               74782275                15.92 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/largeScale/float-12               19906879                61.94 ns/op            0 B/op          0 allocs/op
BenchmarkAppendNumber/zero/scale-12                     213517701                5.328 ns/op           0 B/op          0 allocs/op
BenchmarkAppendNumber/zero/float-12                     154237894                7.672 ns/op           0 B/op          0 allocs/op
BenchmarkAppendNumber/integer/scale-12                  236471412                5.122 ns/op           0 B/op          0 allocs/op
BenchmarkAppendNumber/integer/float-12                  53292375                22.20 ns/op            0 B/op          0 allocs/op
```

The encoder-level delta (-14.75%) is smaller than the primitive one (fullMillis, -39.35%) because time encoding is a fraction of a full `EncodeEntry`; the isolated number shows the raw speedup of the int-format path (`strconv.AppendInt` plus an in-place decimal-point insert beats `strconv.AppendFloat`'s shortest-representation float-to-decimal conversion).

## Compatibility

- No changes to public interfaces (`PrimitiveArrayEncoder`/`ArrayEncoder`/`ObjectEncoder` gain no methods).
- External `PrimitiveArrayEncoder` implementations get the float64 fallback — byte-identical to today.
- `jsonEncoder` output changes only for timestamps that `float64` was already corrupting; values that were exact are unchanged.
- All existing tests pass unmodified.


huww98:
Added license header.

PR #1589, opened 2026-09-10
fix: validate Level before opening sinks in Config.Build

kevin9327:
A `zap.Config` that is missing `level` still creates the log file and leaks its file descriptor.

```go
cfg := zap.Config{
	// Level unset -- e.g. a YAML config where the "level" key was
	// misspelled, so it unmarshals into the zero AtomicLevel.
	Encoding:         "json",
	EncoderConfig:    zap.NewProductionEncoderConfig(),
	OutputPaths:      []string{"/var/log/app.log"},
	ErrorOutputPaths: []string{"stderr"},
}

logger, err := cfg.Build()
// err = "missing Level", logger = nil
// ...and /var/log/app.log now exists, held open by a descriptor
// that nothing can reach.
```

`Build` returns no closer, so the descriptor stays open for the life of the
process. A service that retries `Build` on config reload leaks one per
attempt; on Windows the file also cannot be deleted or rotated.

## Cause

`Build` opens the sinks before it finishes validating:

```go
enc, err := cfg.buildEncoder()      // validates EncoderConfig
if err != nil {
	return nil, err
}

sink, errSink, err := cfg.openSinks()   // side effect: opens/creates files
if err != nil {
	return nil, err
}

if cfg.Level == (AtomicLevel{}) {       // validation, after the side effect
	return nil, errors.New("missing Level")
}
```

`buildEncoder` is pure and already runs first. The `Level` check is the only
validation that happens after `openSinks`, and `openSinks` only cleans up
after itself when it fails partway (`closeOut()` on the second `Open`); it has
no way to know a later check will reject the config.

The existing `"missing level"` case in `TestConfigWithMissingAttributes` passes
`Config{Encoding: "json"}` with no `OutputPaths`, and `Open()` with no paths
returns a `io.Discard` syncer — so nothing is opened and the leak is invisible.

## Fix

Move the `Level` check above `openSinks`. It stays below `buildEncoder`, so the
precedence between the two validation errors is unchanged.

## Tests

`TestConfigWithMissingLevelDoesNotOpenSinks` builds an invalid `Config` pointed
at a temp file and asserts the file was never created.

Before, on unmodified master:

```
=== RUN   TestConfigWithMissingLevelDoesNotOpenSinks
    config_test.go:160:
        	Error:      	Expected error with "file does not exist" in chain but got nil.
        	Messages:   	Build shouldn't open output paths for an invalid Config.
    testing.go:1617: TempDir RemoveAll cleanup: unlinkat ...\001\test.log: The process cannot access the file because it is being used by another process.
--- FAIL: TestConfigWithMissingLevelDoesNotOpenSinks (1.73s)
```

The second line is the leaked descriptor: `t.TempDir()` cleanup cannot remove
the file because it is still open.

After:

```
=== RUN   TestConfigWithMissingLevelDoesNotOpenSinks
--- PASS: TestConfigWithMissingLevelDoesNotOpenSinks (0.01s)
ok  	go.uber.org/zap	0.364s
```

`TestConfigWithMissingAttributes` (both the `missing level` and
`missing encoder time in encoder config` cases) and `TestConfigWithInvalidPaths`
pass before and after, confirming the error precedence is unchanged.

## Verification

```
go vet ./...     # clean
gofmt -l         # clean
go test ./...
```

`go test ./...` produces the same set of failures before and after this change:
`TestConfig`, `TestConfigWithSamplingHook`, `TestOpen`, `TestOpenOtherErrors`,
`TestStacktraceFiltersVendorZap` and `zapcore.TestIOCore`. All are pre-existing
and Windows-only on this machine — `t.TempDir()` cleanup cannot delete a file
zap still has open, and the stacktrace test needs symlink privileges. They fail
identically on an unmodified checkout.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kevin9327:
recheck

justinhwang (approved):

PR #1591, opened 2026-09-10
fix(zapslog): return the receiver from WithGroup("")

kevin9327:
Wrapping a `zapslog.Handler` with a group name that turns out to be empty nests
every subsequent attribute under an empty key.

```go
var h slog.Handler = zapslog.NewHandler(core)
h = h.WithGroup(name)   // name == "", e.g. an unset config value

slog.New(h).With("a", "b").Info("msg", "c", "d")
```

```json
{"":{"a":"b","c":"d"}}
```

expected

```json
{"a":"b","c":"d"}
```

## Cause

The `slog.Handler` interface requires:

> If the name is empty, WithGroup returns the receiver.

`zapslog.Handler.WithGroup` appends the name unconditionally:

```go
func (h *Handler) WithGroup(group string) slog.Handler {
	newGroups := make([]string, len(h.groups)+1)
	copy(newGroups, h.groups)
	newGroups[len(h.groups)] = group
	...
}
```

so `""` is carried in `h.groups` and emitted by `appendGroups` as
`zap.Namespace("")`.

`slog.Logger.WithGroup` drops empty names before they reach the handler, so
code that goes through `*slog.Logger` never hits this. It shows up when a
`Handler` is composed directly — which slog middleware and handler wrappers do,
and which the interface contract explicitly allows.

`testing/slogtest` does not cover this case (its group cases are `empty-group`,
`inline-group`, `WithGroup`, `empty-group-record` and
`nested-empty-group-record`), which is why the existing `slogtest.TestHandler`
run passes.

## Fix

Return the receiver when the name is empty, matching the contract and the
existing empty-key handling in `convertAttrToField`, which already inlines
`slog.KindGroup` attributes whose key is `""`.

## Tests

A new `empty name` subtest in `TestWithGroup` asserts both halves of the
contract: that the receiver is returned, and that attributes are not nested.

Before, on unmodified master:

```
=== RUN   TestWithGroup/empty_name
    handler_test.go:238:
        	Error:      	Not same:
        	            	expected: &zapslog.Handler{..., groups:[]string(nil)} (*zapslog.Handler)(0x...8c0)
        	            	actual  : &zapslog.Handler{..., groups:[]string{""}} (*zapslog.Handler)(0x...910)
        	Messages:   	Expected WithGroup to return the receiver
    handler_test.go:244:
        	Error:      	Not equal:
        	            	expected: map[string]interface {}{"a":"b", "c":"d"}
        	            	actual  : map[string]interface {}{"":map[string]interface {}{"a":"b", "c":"d"}}
        	Messages:   	Unexpected context
--- FAIL: TestWithGroup/empty_name (0.00s)
```

After:

```
--- PASS: TestWithGroup (0.00s)
    --- PASS: TestWithGroup/nested (0.00s)
    --- PASS: TestWithGroup/nested_empty (0.00s)
    --- PASS: TestWithGroup/empty_group (0.00s)
    --- PASS: TestWithGroup/empty_name (0.00s)
    --- PASS: TestWithGroup/skipped_field (0.00s)
    --- PASS: TestWithGroup/reuse (0.00s)
```

The five sibling subtests that pin the current grouping behaviour pass before
and after, as does `TestSlogtest`, which runs `slogtest.TestHandler`.

## Verification

```
cd exp
go vet ./...       # clean
gofmt -l zapslog   # clean
go test ./...
```

```
ok  	go.uber.org/zap/exp/zapfield	1.232s
ok  	go.uber.org/zap/exp/zapslog	1.816s
```

The `exp` module is fully green before and after this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


justinhwang (approved):

justinhwang:
Thanks!

